PR #17, opened 2020-02-13
Move all the packaging recipes to build-aux/

bochecha:
We have multiple packaging recipes:

* `packaging/linux/fedora/` contains some information about the Fedora packaging;
* `packaging/linux/gentoo/` has information about the Gentoo packaging;
* `packaging/windows/` has the recipes to build the app on Windows;
* `snap/` has the recipes to build a Snap;

I'm about to add Flatpak recipes as well, and was wondering where to put them in a way that doesn't add clutter to the top dir.

The obvious solution would be to put it under `packaging/`, but then the Snap recipes stant out on their own.

The Fedora/Gentoo information could be anywhere, they are just text files. Similarly the Windows recipes don't seem to have any restrictions on where they must be. I could add the Flatpak recipes anywhere, there also isn't any restriction.

However Snapcraft is the strictest, and only allows two locations for its build recipes:

* a `snap/` directory at the top of the sources;
* a `build-aux/snap/` directory, since Snapcraft 3.3;

That seems like a reasonable solution to put all the other packaging recipes, which is what this pull request does.

I'll add the Flatpak recipes in a subsequent pull request.

cc @ppd1990


ppd1990:
Snaps build and lgtm in general

scorpio810:
Thanks Mathieu. ;-)

PR #18, opened 2020-02-14
build: Add a Flatpak manifest

bochecha:
This allows building the master branch of QElectroTech with a simple
command:

$ flatpak-builder _build build-aux/flatpak/org.qelectrotech.QElectroTech.json

---

For those who don't know Flatpak much, here are a few additional instructions.

1. you need to install Flatpak, Flatpak Builder and configure the Flathub remote: https://flatpak.org/setup/
2. you will need to install the KDE platform and sdk to build;
3. to make rebuilds faster you want to use the `--ccache` and `--force-clean` options;
4. once the app is built, you need to export it to a local repo, configure that local repo, then install from that repo; this can be automated with the `--install` and `--user` options;

All in all, the following two commands should do the right thing:

```
$ flatpak install flathub org.kde.Platform//5.14 org.kde.Sdk//5.14
$ flatpak-builder --force-clean --ccache --user --install _build build-aux/flatpak/org.qelectrotech.QElectroTech.json
```

If this is merged, we can look at having automatic builds in your CI, published to a development repo, etc…

And then I can help you send the next release to Flathub if you're interested. :slightly_smiling_face:

### DIFF
--- a/build-aux/flatpak/org.qelectrotech.QElectroTech.json
+++ b/build-aux/flatpak/org.qelectrotech.QElectroTech.json
@@ -1,0 +1,37 @@
+{
+  "id": "org.qelectrotech.QElectroTech",
+  "runtime": "org.kde.Platform",
+  "runtime-version": "5.14",
+  "sdk": "org.kde.Sdk",
+  "command": "qelectrotech",
+  "rename-desktop-file": "qelectrotech.desktop",
+  "rename-appdata-file": "qelectrotech",
+  "rename-icon": "qelectrotech",
+  "copy-icon": true,
+  "finish-args": [
+    "--socket=wayland",
+    "--socket=x11",
+    "--device=dri",
+    "--share=ipc",
+    "--filesystem=host"
+  ],
+  "modules": [
+    {
+      "name": "qelectrotech",
+      "buildsystem": "qmake",
+      "post-install": [
+        "mv /app/share/mime/packages/qelectrotech.xml /app/share/mime/packages/org.qelectrotech.QElectroTech.xml"
+      ],
+      "sources": [
+        {
+          "type": "dir",
+          "path": "../.."
+        },
+        {
+          "type": "patch",
+          "path": "patches/0001-build-Fix-the-installation-paths.patch"
+        }
+      ]
+    }
+  ]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 *.snap
+.flatpak-builder
--- a/build-aux/flatpak/patches/0001-build-Fix-the-installation-paths.patch
+++ b/build-aux/flatpak/patches/0001-build-Fix-the-installation-paths.patch
@@ -1,0 +1,44 @@
+From 579ee22f1d2bef560ec90d324a0e476b81faf495 Mon Sep 17 00:00:00 2001
+From: Mathieu Bridon <mathieu@hashbang.fr>
+Date: Tue, 7 Jan 2020 14:39:08 +0100
+Subject: [PATCH] build: Fix the installation paths
+
+---
+ qelectrotech.pro | 14 +++++++-------
+ 1 file changed, 7 insertions(+), 7 deletions(-)
+
+diff --git a/qelectrotech.pro b/qelectrotech.pro
+index 07e51fb..2a86b91 100644
+--- a/qelectrotech.pro
++++ b/qelectrotech.pro
+@@ -5,20 +5,20 @@
+ # Chemins utilises pour la compilation et l'installation de QET
+ unix {
+ 	# Chemins UNIX
+-	COMPIL_PREFIX              = '/usr/local/'
+-	INSTALL_PREFIX             = '/usr/local/'
++	COMPIL_PREFIX              = '/app/'
++	INSTALL_PREFIX             = '/app/'
+ 	QET_BINARY_PATH            = 'bin/'
+ 	QET_COMMON_COLLECTION_PATH = 'share/qelectrotech/elements/'
+ 	QET_COMMON_TBT_PATH        = 'share/qelectrotech/titleblocks/'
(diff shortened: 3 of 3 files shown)


build-aux/flatpak/org.qelectrotech.QElectroTech.json
```diff
@@ -0,0 +1,37 @@
+{
+  "id": "org.qelectrotech.QElectroTech",
```
bochecha: This might require some discussion.

Flatpak enforces a long-standing, not entirely written convention in the desktop space, which comes all the way from DBus, of having an "application id" formed like a reverse-DNS.

QElectroTech is maintained over at https://qelectrotech.org, so the obvious application id is `org.qelectrotech.QElectroTech`. We can change it to anything you like, but I'd **strongly** recommend to prefix it by `org.qelectrotech.`.

By convention, this application id corresponds to the name the application takes on DBus (if any), to the name of the desktop file, the appdata file, the icons, …

Flatpak actually enforces this convention, so the application id is important.

build-aux/flatpak/org.qelectrotech.QElectroTech.json
```diff
@@ -0,0 +1,37 @@
...
+  "runtime": "org.kde.Platform",
+  "runtime-version": "5.14",
+  "sdk": "org.kde.Sdk",
+  "command": "qelectrotech",
+  "rename-desktop-file": "qelectrotech.desktop",
+  "rename-appdata-file": "qelectrotech",
+  "rename-icon": "qelectrotech",
```
bochecha: This renames the desktop file, the appdata file, and the main launcher icons.

This is required because Flatpak enforces the aforementioned naming convention, and will only export out of the sandbox the files named like the application id.

Those files need to be exported outside of the sandbox, or else the desktop launcher (e.g an application menu) won't be able to see the app, and users will be very confused how to run it. :stuck_out_tongue: 

Similarly the appdata must be renamed for software centers like GNOME Software or KDE Discover to see it and offer users to install the app.

Renaming all of those directly in the source tree would be better and simpler, but it is not required to make a Flatpak build. I'm happy to send the changes in a future merge request if they are desired.

build-aux/flatpak/org.qelectrotech.QElectroTech.json
```diff
@@ -0,0 +1,37 @@
...
+    "--device=dri",
+    "--share=ipc",
+    "--filesystem=host"
+  ],
+  "modules": [
+    {
+      "name": "qelectrotech",
+      "buildsystem": "qmake",
+      "post-install": [
+        "mv /app/share/mime/packages/qelectrotech.xml /app/share/mime/packages/org.qelectrotech.QElectroTech.xml"
```
bochecha: This renames the mime type handling file, for the same reason as the icon, desktop file, etc…

This is indeed required for Flatpak to export this file outside the sandbox, so that the desktop can see it and do the right thing.

build-aux/flatpak/org.qelectrotech.QElectroTech.json
ppd1990: We should probably put the renaming etc. of resource files into the qmake project iself. Similar to this: 

https://github.com/solvespace/solvespace/blob/master/res/CMakeLists.txt#L129

build-aux/flatpak/org.qelectrotech.QElectroTech.json
ppd1990: Strongly agree. The AppStream metadata file should also be named accordingly.

build-aux/flatpak/org.qelectrotech.QElectroTech.json
bochecha: I'd just plain remove the source files. But like I said that's not necessary for this merge request, so I'd rather do that separately.

My point here was to explain to the maintainers who might not know about the convention **what** this does and **why** it exists, not to start a conversation on **how** to do it in the future. (I've done it for multiple apps in the past)

build-aux/flatpak/org.qelectrotech.QElectroTech.json
bochecha: There is no appstream file for an app, the appstream file is at the repo level and concatenates all the individual app appdata/metainfo files.

If you meant the appdata file, I did say so in my comment already, and this is taken care of with the `rename-appdata` option which I detailed in another comment.

scorpio810:
Merged, thanks Mathieu.

scorpio810:
```
flatpak install flathub org.kde.Platform//5.14 org.kde.Sdk//5.14     
Looking for matches…


        ID                                           Branch           Op           Remote            Download
 1. [✗] org.freedesktop.Platform.openh264            19.08            i            flathub                    593,6 kB / 593,4 kB
 2. [✓] org.gtk.Gtk3theme.Breeze                     3.22             i            flathub                    109,5 kB / 156,7 kB
 3. [✗] org.kde.Platform                             5.14             i            flathub           1.0 kB / 322,5 MB
 4. [ ] org.kde.Sdk.Locale                           5.14             i            flathub                  < 340,6 MB (partial)
 5. [ ] org.kde.Sdk                                  5.14             i            flathub                  < 766,0 MB

Warning: org.freedesktop.Platform.openh264 not installed
Error: Error deploying: Failed to read commit ad2084647a51518d03cde7fb83b7e8e979fa7ad0dad16919e8d1cbee5f7af1a6: No such metadata object 3e6db835526778ba64504a7866caa620a0a881f70c6d02d98b477a169b1db53b.dirtree
error: Failed to install org.kde.Platform: Error deploying: Failed to read commit ad2084647a51518d03cde7fb83b7e8e979fa7ad0dad16919e8d1cbee5f7af1a6: No such metadata object 3e6db835526778ba64504a7866caa620a0a881f70c6d02d98b477a169b1db53b.dirtree
```

scorpio810:
https://download.tuxfamily.org/qet/builds/Flatpak/

PR #22, opened 2020-02-17
flatpak: Properly rename the appdata file

bochecha:
Adding the full file name was required, just like for renaming the
desktop files, unlike for renaming the icons.


scorpio810:
Merged, thanks. ;-)

PR #23, opened 2020-02-23
Traduction and add symbols

sdeffaux:


scorpio810:
https://git.tuxfamily.org/qet/qet.git/commit/?id=6535a93b66bf09bd475685e909c44930b3d1f4a8

PR #34, opened 2020-04-16
Modernize snap

ppd:
Thorough modernization of the snap. 

Main benefits:

## Usage of kde-neon extension
- Up-to-date Qt and KDE libs. We depend directly on upstream's distribution
- Fewer stage-packages = fewer rebuilds required for security updates
- Snaps are ~83% smaller which leads to faster downloads and faster startup

## Building with kde-frameworks-5-core18-sdk
No need to pollute snapcraft.yaml with tons of build dependencies. We use the kde-neon's corresponding sdk snap instead

## Misc
- Small optimizations for startup performance
- User config and elements are stored in SNAP_USER_COMMON instead of SNAP_USER_DATA which is more correct for data that should not be backed up (mostly the collection) on each update
- Set PYTHONPATH explicitly to make sure qet_tb_generator actually works

It's probably a good idea to test this a little before committing. I'll attach a snap with these changes included.

### DIFF
--- a/build-aux/snap/local/launchers/qelectrotech-launch
+++ b/build-aux/snap/local/launchers/qelectrotech-launch
@@ -1,5 +1,15 @@
-#!/bin/bash
+#!/bin/sh
 
+# migrate .qet directory from SNAP_USER_DATA to SNAP_USER_COMMON
+from="$SNAP_USER_DATA/.qet"
+to="$SNAP_USER_COMMON/.qet"
+if [ ! -d "$to" ] && [ -d "$from" ]; then
+  echo "Migrating user data from $from to $to"
+  mkdir "$to"
+  cp -av "$from/." "$to"
+fi
+
+# link DXFtoQET so that QET finds it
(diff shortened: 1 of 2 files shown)


ppd:
A compiled snap
[qelectrotech_0.8-dev-206-g1296fb1f2_amd64.snap.zip](https://github.com/qelectrotech/qelectrotech-source-mirror/files/4489143/qelectrotech_0.8-dev-206-g1296fb1f2_amd64.snap.zip)


```shell
wget https://github.com/qelectrotech/qelectrotech-source-mirror/files/4489143/qelectrotech_0.8-dev-206-g1296fb1f2_amd64.snap.zip
unzip qelectrotech_0.8-dev-206-g1296fb1f2_amd64.snap.zip
snap install --dangerous qelectrotech_0.8-dev-206-g1296fb1f2_amd64.snap
```

scorpio810:
Merged, thanks Max.

PR #43, opened 2020-05-14
Recent update of norwegian translation

JoelAs-au2x:
Thanks to Laurent who could insert my first try. I hope to become better after a while...


scorpio810:
Thanks Joel, merged.

PR #45, opened 2020-06-08
Multi edit element editor

Murmele:
Edit multiple elements in the element editor in one step

### DIFF
--- a/sources/editor/arceditor.cpp
+++ b/sources/editor/arceditor.cpp
@@ -128,6 +128,10 @@ CustomElementPart *ArcEditor::currentPart() const {
 	return(part);
 }
 
+QList<CustomElementPart*> ArcEditor::currentParts() const {
+    return style_->currentParts();
+}
+
 /**
  * @brief ArcEditor::updateArcS
  * Update the start angle of the arc according to  the edited value.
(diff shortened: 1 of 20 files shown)


scorpio810:
wget https://github.com/qelectrotech/qelectrotech-source-mirror/pull/45.patch

```
~/qet_git$ git am 45.patch 
Application de  not needed, because it is the same than a few lines below
Application de  WIP
Application de  multiterminal change
Application de  make partTextEditor capable to modify multiple parts at the same time
Application de  forgotten to commit texteditor
Application de  solves #7
Application de  - static_cast because it is a condition that it is PartTerminal
Application de  multipart edit
Application de  static_cast is enough
Application de  add multiedit feature also for dynamic text
Application de  implement currentParts() for every elementItemEditor so the editor can be extended to multi part edit in a later step
Application de  do not set parts again when they are set already
Application de  not needed, because it is the same than a few lines below
error: le patch a échoué : sources/editor/qetelementeditor.cpp:629
error: sources/editor/qetelementeditor.cpp : le patch ne s'applique pas
le patch a échoué à 0013 not needed, because it is the same than a few lines below
```

scorpio810:
Can't merge.

PR #46, opened 2020-06-08
when model is not yet loaded, but the element editor should be opened…

Murmele:
…, a segmentation fault occurs. Do not set current index when m_model is not yet loaded


scorpio810:
This patch has no conflit and can be merged

scorpio810:
Merged, thanks

PR #47, opened 2020-06-08
UUID

Murmele:
Use uuid instead of a dynamic created id


scorpio810:
```
git am 47.patch
Application de  when model is not yet loaded, but the element editor should be opened, a segmentation fault occurs. Do not set current index when m_model is not yet loaded
error: le patch a échoué : sources/ElementsCollection/elementscollectionwidget.cpp:123
error: sources/ElementsCollection/elementscollectionwidget.cpp : le patch ne s'applique pas
le patch a échoué à 0001 when model is not yet loaded, but the element editor should be opened, a segmentation fault occurs. Do not set current index when m_model is not yet loaded
```

scorpio810:
Can't merged

PR #51, opened 2020-07-08
Add file .gitmessage.txt

De-Backer:
Git will use ths file as the default initial message when you commit.

for use run this instruction:
git config --local commit.template .gitmessage.txt

https://qelectrotech.org/forum/misc.php?action=pun_attachment&item=1452&preview

In essence, the intention is to use the commits for the change-log for that we need to use tools like a template.
Of course this is just a start, and i would like to get extra info for the definitive file.

So please make your comment.


scorpio810:
It 's could be nice to merge it and add this information in contrib.MD, what do you think?

PR #60, opened 2020-08-03
Elementeditor: remove redundant calls to StyleEditor::setPart

damianCaceres:
During bug fix 0197 it is discovered that part editors make redundant calls to style editor.
Thanks


scorpio810:
Merged, thanks

PR #67, opened 2020-08-15
Fix dxf export for copied conductors

c-born:
When exporting as DXF a drawing with conductors that had been copied, all the copies appeared in the same location as the original, so visually they disappeared.
This minor modification fixes that problem.

There are still (at least) a couple of problems with DXF export:
1. The frame/border is offset incorrectly
2. Text that has been auto-sized to fit in a cell is exported at original size and overflows

### DIFF
--- a/sources/exportdialog.cpp
+++ b/sources/exportdialog.cpp
@@ -578,12 +578,14 @@ void ExportDialog::generateDxf(Diagram *diagram, int width, int height, bool kee
 	}
 
 	//Draw conductors
-	foreach(Conductor *cond, list_conductors) {
-		foreach(ConductorSegment *segment, cond -> segmentsList()) {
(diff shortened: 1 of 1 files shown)


scorpio810:
Merged, thanks

scorpio810:
> [14:17] <cborn> I've made small bug-fix to DXF export and added it as a pull-request to the Github repo, let me know if that isn't the right place to do it
> [14:19] <cborn> Problem was if you made copies which included a conductor, the segment list remained the same in the copies, just the x and y position in the conductor is changed, the dxf code exportdialog.cpp didn't take account of that
> [14:20] <cborn> There are still a couple of other problems, in the border and titlebar text, needing work\ 
> [14:31] <cborn> Merged already, that was FAST! I'm used to projects where they will discuss merges for days...
Thank you so much.
Merged, and I build  all packages. ;-)

PR #69, opened 2020-08-16
Fix dxf Export for Border

c-born:
This fixes the border rectangle and row/column headers when exporting to dxf, taking into account the diagram margins.
It doesn't address the string sizing issue (auto-size strings to fit in title block cells), there seem to be some problems with the dxf text alignment which will require more analysis, possible a dxf version issue.


scorpio810:
Merged, thanks.
But I deleted "Correct dxf export for copied drivers" in your PR because it already applied.
The Github repository is  amirror and is synchronized with the upstream GIT remote master every hour.

c-born:
Hi Laurent,

Yes, I wasn't sure if I should be doing a PR against the Github mirror or
upstream, and as I know how to do it from my fork on Github I went with
that!
Please let me know if there is a better/correct way.

I've spent most of the day working with the DXF spec trying to sort a few
details out with regard to text alignment and fitting, but it appears other
programs (including AutoCAD) don't honor the spec anyway, so some of the
things the code is adding to the DXFfile are just ignored. Lots of
comments about this on the AutoCAD forum archives once you start looking.

I've just made another dxf  fix to properly align text in the titlebox
cells, however the required offsets were obtained heuristically (x is the
Diagram::margin but y is slightly less) and I wasn't going to commit the
change until I have a chance to work through the cascaded scaling logic
that precedes writing the text.

I also want to get the auto-size text working to fit long text in the
cells, and possibly take a look at saving image cells as well, but given
tomorrow is Monday I may not have a chance to get back on it for a bit.
However if anybody is looking for it, let me know and I can commit/push
what I have.

Regards,

*David Varley*
C-Born Software Systems

*t:* +61 3 5229 0144
*e:* David.Varley@cborn.com
*w:* www.cborn.com

scorpio810:
Hi David,
maybe Ronny post can help you 
https://qelectrotech.org/forum/viewtopic.php?id=1465

Best Regards,
Laurent

c-born:
Hi Laurent,

Ronny is right in that it would be a good idea to add different regions
such as title blocks and borders as separate layers, but this is a separate
issue.

The immediate problem was that whoever coded the current dxy output
correctly followed the dxy specification from AutoCAD, however other
programs have not themselves implemented it correctly! A good way to test
this is with one of the free programs available here from ODA
<https://www.opendesign.com/guestfiles>

The current Qelectrotech export code uses  dxf  TEXT group codes 72 and 73
for text alignment and fitting. The specification says if these codes are
zero then the first alignment point will be used, but if either is non-zero
then the first will be ignored and only the second alignment point used.
This is very easy to test with a simple file and text editor, and in fact
the first alignment point is always used and the second ignored!

That means alignment/fitting has to be handled internally by our code. I'll
take a look into that, and also see about coding image cells.

Regards,
David

c-born:
Hi Laurent,

I've made the changes so that the auto-sized title-block text cell export
to DXF format works now.
Somebody made a lot of format-only changes to the code overnight which
might make the PR a bit confusing, but it should sort itself out by
fast-forwarding over the common stuff.

Regards,
David

PR #76, opened 2020-10-09
use pkgconfig for sqlite3

remicollet:
To detected early, and use propre build/link flags

Without library
```
$ qmake-qt5 *pro
Project ERROR: sqlite3 development package not found
```

With, build passes and
```
$ ldd qelectrotech
	linux-vdso.so.1 (0x00007ffed41e9000)
	libsqlite3.so.0 => /lib64/libsqlite3.so.0 (0x00007f39d0e3d000)
...
```


scorpio810:
Merged, thanks Remi.

PR #88, opened 2020-10-31
Update qet_en.ts

Mies-bvd89:
The first letter of the word needs to be in upper case to align with the rest of the menu.


scorpio810:
Merged, thanks

PR #93, opened 2020-11-17
German language files updated

Bisku:
Translation of new items added.

Note:
It would be nice if the translation of "Label" (QObject-Label) could be specified differently in the nomenclature than in the component properties. The German translation "Betriebsmittelkennzeichen" is a fixed technical term, but it is too long for the nomenclature table. Therefor this translation uses a abbreviation ("BMK") now. 

In order to improve user-friendliness, especially for beginners, two different strings for translation would be need. One string for the object properties and one for the nomenclature table. Maybe this can be implemented in future?


scorpio810:
Merged, thanks

scorpio810:
Please add this information in the bugtracker, thanks.

Bisku:
Done.

PR #94, opened 2020-11-17
changed and added some translations in qet_directory - files and elements

plc-user:
changed and added some translations in qet_directory - files and elements


scorpio810:
Merged, thanks

PR #95, opened 2020-11-19
modified / added german translations and added some Elements

plc-user:
The german translations to qet_directory - files (EN 60617) are now taken from the german version of the Standard.


scorpio810:
Merged, thanks

PR #98, opened 2020-11-30
Init of Cmake

De-Backer:
what a mess!
should we look to use static libs anyway?

can we involve Joshua what happens best, 
two choices 
we keep the project structure and therefore a CMakeLists.txt in each folder, 
or we postpone this until we renew the project structure

this is not finished yet because all files must be in the CMakeLists.txt (the use of vars is not recommended by Cmake => so no * .h or * .cpp)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,7 @@
+cmake_minimum_required(VERSION 3.5)
+
+project(qelectrotech LANGUAGES CXX)
+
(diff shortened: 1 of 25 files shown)


De-Backer:
NOTE: do not put this branch in the QET repo!!
I will continue to adapt this comit
so that there are no unnecessary files.

De-Backer:
this is already in the git repo by now

PR #106, opened 2020-12-21
re-organized directory, added manufacturer- and designation-info and added some new elements

plc-user:
... as discussed in Elements-forum.

The resize of front-graphics isn't done yet, because this discussion isn't finished.
Will resize all front-views and create pull-request, when the decision is done!


scorpio810:
Merged, thanks

PR #114, opened 2021-01-23
some german translations in "qet_directory"

plc-user:
added some german translations for manufacturers


scorpio810:
Merged, thanks

PR #116, opened 2021-01-29
Updated German translation qet_de.ts

Bisku:
Translation for 'Mongol' and 'Modifier les informations de plusieurs éléments' added.


scorpio810:
Merged, thanks

PR #119, opened 2021-01-31
added MDT-elements (by a.funke) and some cleanup of elements

plc-user:
some hundred elements contain this fragment "&#xd;" and I removed that from some elements. 
Is this of any use? 
I noticed that it sometimes prevents QET from using that element, so I removed it.

Additionally I added some elements from the german manufacturer MDT created by a.funke as explained in the german forum-topic: https://qelectrotech.org/forum/viewtopic.php?id=1919


scorpio810:
Merged, thanks

PR #141, opened 2021-02-21
SaveAsPDF uses project's filename instead of project's title now

Bisku:
When creating a PDF file QElectroTech will use the filename of the project file (if exists) instead of the project's title now.


Bisku:
Found a typo in the same file that prevents PDF files from shown in the dialog. I'll fix that and create a new pull request with both adjustments.

PR #142, opened 2021-02-22
Modifications to SaveFile-Dialog for "Save As PDF"

Bisku:
Fixed a typo that prevented existing PDF files from being displayed in SaveFileDialog for PDFs.

The way the file name for the PDF is generated has changed. If the project has already been saved, the PDF has the same file name (with .pdf of course); If not, the file name is generated from the project title (= same behavior as Save as - dialog for a .qet project file).


scorpio810:
Merged, thanks

PR #148, opened 2021-03-09
Fixed grammer issues in German language file

Bisku:


scorpio810:
```
git am 148.patch 
Application de  Fixed grammer issues in German language file
error: lang/qet_de.qm : ne correspond pas à l'index
error: lang/qet_de.ts : ne correspond pas à l'index
le patch a échoué à 0001 Fixed grammer issues in German language file
```

PR #150, opened 2021-03-13
centered some elements on DIN-rail (allows rotating without changing Y-position)

plc-user:
as the subject says.
And one new element


scorpio810:
Merged, thanks

PR #171, opened 2021-06-04
Changed scaling of (some) front-views to 100mm -> 200px

plc-user:
... and some additional elements with the same scaling.


scorpio810:
Merged, thanks

PR #172, opened 2021-08-10
BugFix, Labelling, Beautification of existing elements and some new elements

plc-user:
the subject says it: 
additionally to WAGO-Elements I modified some existing elements where I think some modifications would improve them.


scorpio810:
Merged, thanks

PR #177, opened 2021-08-28
fixed fonts / added schematics for MID-Counters

plc-user:
- reduced font-size for parts in arduino-folder 
- schematic-symbols in wago-folder


scorpio810:
Merged, thanks

PR #182, opened 2021-09-13
Add possibility to rotate items in element editor using the spacebar

al1ads:
I had to override the setRotation method of PartRectangle, PartLine and PartPolygon so that they don't rotate around the origin of the canvas but around their own origin. Also override PartTerminal so that setRotation triggers a setOrientation.


scorpio810:
Merged, thanks

PR #190, opened 2021-09-23
Build fix for non-C++17 compatible systems

Carbenium:
std::variant/std::visit was only introduced with C++17. Remove its usage.
We don't even need it in these cases since QColor has an implicit constructor accepting Qt::GlobalColor.

Follow-up for b69c7b10277257fcd2b4b993e303049f41061a6f


scorpio810:
Thanks.
You could provide an howto to compile QElectroTech with MSVC and KDE 5 Framework?
Your QElectroTech MSVC binary is it more faster than my MXE cross-compile Windows package?

PR #198, opened 2021-11-07
upgraded elements and renamed company

plc-user:
as the subject says


scorpio810:
Merged, thanks 

Not useful to add multiple translation by lang in elmt if is the same name. ;-)

```
        <name lang="en">750-642</name>
￼        <name lang="de">750-642</name>		￼        <name lang="fr">750-642</name>
￼        <name lang="cs">750-642</name>		￼        <name lang="cs">750-642</name>
￼        <name lang="de">750-642</name>
```

PR #205, opened 2022-01-13
snap: Remove framework snap prompt

ppd:
Now that https://github.com/snapcore/snapcraft/pull/3596 has been
released in snapcraft 6.0.1, drop the prompt that tells users to
disconnect the old framework snap.

Also drop --enable-experimental-extensions from CI  because kde-neon
has been declared stable for core20.


scorpio810:
Merged, thanks

PR #207, opened 2022-01-18
projectview: Re-enable appending qet suffix if not running as flatpak

### LINKED ISSUE #206: File dialog should enforce suffix
@scorpio810 This commit 
https://github.com/qelectrotech/qelectrotech-source-mirror/commit/b121dad5625512658b1511059e264161affa6341
disables appending the correct suffix. 

People have a hard time finding those files again because they get saved without the .qet suffix. 

For some reason, the QFileDialog::getSaveFileName() does not display a default name when using the xdg-desktop-portal backend. Could be a Qt bug.

ppd:
It was removed in b121dad for all platforms.
Only disable appending the suffix for the flatpak platform
by testing for the FLATPAK_ID environment variable.

Fixes #206

### DIFF
--- a/sources/projectview.cpp
+++ b/sources/projectview.cpp
@@ -339,8 +339,8 @@ QString ProjectView::askUserForFilePath(bool assign) {
 	// if no filepath is provided, return an empty string
 	if (filepath.isEmpty()) return(filepath);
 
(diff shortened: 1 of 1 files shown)


scorpio810:
Merged, thanks

PR #208, opened 2022-01-26
projectview: Only append .qet if not snap or flatpak

ppd:
We need to disable the suffix for snaps also because older xdg-desktop-portal's use the document portal indiscriminately, even in the user's home folder.


scorpio810:
Merged, thanks

PR #216, opened 2022-03-20
rework and new elements (Hager)

plc-user:
reworked some Hager front-views to scale 100mm -> 200px and removed terminals in this directory:
elements/10_electric/98_graphics/99_assembly_plan/01_thumbnails_mounting_plate/hager/
also added some graphics for common parts that are used in Germany


scorpio810:
Merged, thanks

PR #219, opened 2022-06-06
Changed many front-views and some new elements

plc-user:
- changed many front-views from link-type "simple" to "thumbnail"
- added some parts from WAGO-Series 2003
- added some parts from manufacturer Weidmüller
-


scorpio810:
Merged, thanks

PR #222, opened 2022-06-22
Add 'Other' option for slave device contact type

TadeasPilar:
Master - slave elements are super useful for splitting one part into multiple logical elements. This makes schematic much more readable.
My issue is, QET always assumes the slave is a switch. And draws appropriate switch symbol next to the slave link(see pic.). I would like to use the master-slave elements for other parts as well. My specific use-case has power meters with digital outputs. It makes sense to separate the low voltage digital part into second sheet.

There is demand for this among the community. I am not the first one to stumble upon this issue.
[link1](https://qelectrotech.org/forum/viewtopic.php?id=1093)
[link2](https://qelectrotech.org/forum/viewtopic.php?id=1223)

This PR adds `Other` option into the slave switch state selection. If you select this option, and your `Cross reference`->`XRef slave position` is set to `View contacts`, QET will only draw the XRef without contact type. This change doesn't make sense for `cross` style of contacts. In that case, no link will be shown.

image

Before:
image

After:
image


scorpio810:
Merged, thanks

scorpio810:
Added fix see : https://qelectrotech.org/forum/viewtopic.php?pid=16477#p16477

PR #223, opened 2022-07-05
Cleanups related to MIME types

pinotree:
- drop the autogenerated XML files from the XML file with XDG MIME type definitions
- drop the old desktop MIME types, for KDE 3.x and older
- refresh the flatpak patch accordingly


scorpio810:
Hello Pino, thanks for your expertise.
Merged, thanks

PR #226, opened 2022-07-21
Remove useless titles

pkess:
With these changes i removed the "Form" and "Dialog" window titles from widgets where a title is never shown and so no translation is needed.

### DIFF
--- a/lang/qet_ar.ts
+++ b/lang/qet_ar.ts
@@ -491,94 +491,93 @@
 <context>
     <name>AutoNumberingManagementW</name>
     <message>
-        <location filename="../sources/autoNum/ui/autonumberingmanagementw.ui" line="26"/>
         <source>Form</source>
-        <translation>شكل</translation>
+        <translation type="vanished">شكل</translation>
     </message>
     <message>
-        <location filename="../sources/autoNum/ui/autonumberingmanagementw.ui" line="149"/>
+        <location filename="../sources/autoNum/ui/autonumberingmanagementw.ui" line="146"/>
         <source>Project Status:</source>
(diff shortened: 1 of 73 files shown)


pkess:
Hello,
the basic change here is to the ui files. In those files i removed the definition for the window title where no title would be displayed. In those windows there was only a generic text like "Form" or "Dialog". After this i ran lupdate to get the ts files updated. I was using this command, which is basically the same:
```bash
for i in $(ls lang/*.ts);
do
       	lupdate -ts $i -pro qelectrotech.pro -no-obsolete;
done
```
You can see the different steps by checking the single commits.

There is one point which is hard to check here: If anyone wrote a translation for a window where there was only "Form" defined but the translation had a useful text this might be removed with this.

pkess:
One more question: Do you want to get the qm files generated inside a merge request? From my point of view those qm files are binary files and when you generate them in each merge request this will raise much of merge conflicts. So my basic suggestion would be to remove qm files and generate them as a build step instead.

scorpio810:
Qm files was pushed by translators when update translation.
FYI, Form" and "Dialog" was added to dictionnary in qtlinguist and need only one translation.

pkess:
Sorry, but could you explain why you want to keep this text inside the ui files and the translations? This title does not really cover any real purpose of the attached widgets or dialogs and will never be displayed i think. So why confusing people that want to contribute translations with those texts?

scorpio810:
I don't want your changes to break the UI.

pkess:
Wo if i  split this in smaller pr and check the usage of the ui elements you would agree to change this?

scorpio810:
Yes of course.

PR #225, opened 2022-07-20
Add german translations

pkess:
Hello,

i added and changed some translations for the terminal editor to german.

PR #227, opened 2022-07-23
Fix german translations

pkess:
New try for https://github.com/qelectrotech/qelectrotech-source-mirror/pull/225


scorpio810:
Merged, thanks

PR #229, opened 2022-09-30
fix corrupted elements that could not be loaded

plc-user:
when playing with a new version of my ElementScaler I found some Elements with corrupted XML


scorpio810:
Merged, thanks

PR #240, opened 2022-12-15
add and rework elements

plc-user:
- added Telecontrol-Controller (WAGO)
- simplified graphics and added terminal-names for circuit-breakers
  (among others: labeling generalised for elements/10_electric/10_allpole/200_fuses_protective_gears/11_circuit_breakers/eaton_40a.elmt because I didn't find any manufacturer-specific parts in element. Next step would be the renaming of the element to a more common name.)


scorpio810:
Merged, thanks

PR #241, opened 2023-01-06
avoid WARNING: mime_xml.path is not defined

remicollet:
```
Info: creating stash file /dev/shm/BUILD/qet-0.9/.qmake.stash
WARNING: mime_xml.path is not defined: install target not created

WARNING: mime_desktop.path is not defined: install target not created

WARNING: mime_xml.path is not defined: install target not created

WARNING: mime_desktop.path is not defined: install target not created


```


scorpio810:
Merged, thanks

PR #245, opened 2023-01-13
some German translations

plc-user:
In German we don't say "Zeichnungskopf" for "title block".
In technical drawings we say "Schriftfeld".
Changed many different occurrences of that word and added some other translations.

### DIFF
--- a/README
+++ b/README
@@ -8,7 +8,7 @@ QET utilise le format XML pour ses éléments et ses schémas et inclut un édit
 
 [de]
 QElectroTech ist eine Qt5 Software, um Schaltpläne zu erstellen.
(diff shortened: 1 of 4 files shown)


scorpio810:
Merged, thanks plc-user
I pushed in manual misc/qelectrotech.xml

```
erreur : misc/qelectrotech.xml : le patch ne s'applique pas
l'application de la rustine a échoué à 0001 some German translations
astuce: Utilisez 'git am --show-current-patch=diff' pour visualiser le patch en échec

```

plc-user:
I don't know what I did differently when editing that file compared to the others...

Je ne vois pas ce que j'ai fait de différent en éditant ce fichier...
Merci de l'avoir quand même repris, Laurent !

PR #249, opened 2023-01-23
Only display git revision when available

remicollet:
1/ avoid tons of warnings during the build such as (git not installed)

`/bin/sh: ligne 1: git : commande introuvable`

or (when build from sources archive)

`/bin/sh: ligne 1: fatal: not a git repository (or any of the parent directories): .git`

2/ display git revision when available


scorpio810:
Merged, merci Remi.

PR #251, opened 2023-03-12
some German translations and Element-Corrections

plc-user:
... for elements and directories in a subject area "foreign" to me.


scorpio810:
Merged, thanks

PR #252, opened 2023-03-12
German translations and corrected directory-name

plc-user:
The directory-name contained two underscores ...


scorpio810:
Merged, thanks

PR #255, opened 2023-05-21
Fixed some italian translations

cbacchianif1:
I've fixed some translations that were missing or wrong. I couldn't make a test build though


scorpio810:
Merged, thanks

PR #256, opened 2023-06-11
corrections, new terminals and new sub-dir "discontinued"

plc-user:
- corrected front-view of 750-8215
- added new "distribution-terminals" 2006-803x and 2206-803x
- added directory "discontinued" and sorted some elements to it (would like to keep those elements here because they are still in "heavy use" by end-users although they aren't sold as new anymore)


scorpio810:
Merged, thanks

PR #257, opened 2023-06-13
fixed terminal-positions

plc-user:
There are so many element files that have not been created very carefully: Terminal positions are not on values (n*10) and sometimes not even on integers...
Corrected some of them - many element-files are still left unchanged...


scorpio810:
Merged, thanks

PR #258, opened 2023-06-18
scaling and fixing of front-views

plc-user:
I would like to continue scaling of those many elements in the sub-dirs of
`elements/10_electric/98_graphics/99_assembly_plan/01_thumbnails_mounting_plate/`
which are not to scale.
In this context I would then also remove the terminals from these elements ... may I?


scorpio810:
Merged, thanks.
Why not, but sometime is usefull to have terminal in these element...
https://download.tuxfamily.org/qet/scorpio/2022-7-23_Vibrafloor.pdf=page6
https://download.tuxfamily.org/qet/scorpio/2022-7-23_Vibrafloor.pdf=page16
https://download.tuxfamily.org/qet/scorpio/2022-7-23_Vibrafloor.pdf=page17

scorpio810:
Hallo PLC-user,

so ok

PR #268, opened 2023-10-13
Too many parts

scorpio810:


scorpio810:
I tried with 1000 primitives seem's to work quickly.

PR #273, opened 2023-12-10
Fix Cmake build

De-Backer:
When creating new files.
Add the new files to: "cmake/qet_compilation_vars.cmake".

So that Cmake stays up to date.

thank you.


scorpio810:
Merged, thank you

PR #280, opened 2024-03-14
Fix sizes

plc-user:
- fixes size of SVG-export in diagram-editor
 - fixes font-size of "input"-elements in element-editor


scorpio810:
Thanks you, plc-user. ;-)

PR #287, opened 2024-04-10
Revert "some minor changes"

scorpio810:
Reverts qelectrotech/qelectrotech-source-mirror#286

Strange but my script use 

`cat sources/qetversion.cpp | grep "return QVersionNumber{ 0, "| head -n 1| cut -c32-40| sed -e 's/,/./g' -e 's/ //g'`
return:
`.0};`
Instead of `0.100.0`

packaging name is now qelectrotech-.0};.r8073...

### DIFF
--- a/sources/autoNum/ui/selectautonumw.cpp
+++ b/sources/autoNum/ui/selectautonumw.cpp
@@ -226,7 +226,7 @@ void SelectAutonumW::on_buttonBox_clicked(QAbstractButton *button)
 							   ));
 				break;
 			}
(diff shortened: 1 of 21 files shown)


plc-user:
Sorry, Laurent, din't know you pull the version-number with a script!

Line 26 in qetversion.cpp changed, too:
It began with eight blanks that are now replaced by two &lt;tab&gt;

scorpio810:
Thanks plc-user for your  awk tip.

PR #289, opened 2024-04-24
correct more indentations / whitespace

plc-user:
I have left out some files/lines from the "TerminalStrip" section as these are under development.
The copyright notice should be completely safe...

Code compiles without additional errors or warnings!


scorpio810:
Thanks

PR #292, opened 2024-05-06
hide SVG background checkbox in print preferences

plc-user:
it is not necessary here and only confuses the user ...


scorpio810:
I don't saw this option in print preferences, see https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/sources/print/projectprintwindow.cpp

plc-user:
That part of export-preferences is used twice. See comment in source-code from here:
[Comment](https://github.com/qelectrotech/qelectrotech-source-mirror/blob/9b4bbfbf0d018fb6d91bb0476c63f7bce3f1abf6/sources/exportpropertieswidget.cpp#L98)

QET_Print-Settings

scorpio810:
Arhg, I'd forgotten about this menu, good find, thanks and sorry for the noise...
Laurent

PR #297, opened 2024-06-16
add terminal-names to connection in qet-file

plc-user:
It may be a good idea to add the terminal-names to the connections in QET-file. 
This could make it easier for a programmer of an external tool or plugin to develop something...


scorpio810:
Thanks plc-user ;-)

PR #300, opened 2024-07-14
move code into else-clause to avoid possible crashes

plc-user:
In my opinion, we should move the code Laurent inserted for the tags "element1_linked" and "element2_linked" to the else branch of the IF query for the UUID of the terminal to avoid possible crashes? Users may still have elements in their projects and/or collections that do not have a UUID.


scorpio810:
Thank you, I think you're right

PR #303, opened 2024-08-20
QET_ElementScaler: fix error for Qt 5.9 and added mirroring

plc-user:
Laurent found a compile-error with Ubuntu 18.04 (!) which should be fixed with this PR.

But mainly:
In this PR there is an improvement for QET_ElementScaler - Import:
I added another input for mirroring of the imported element!

Bildschirmfoto_2024-08-20_11-43-02

In this PR I added language-files for German and English but the other ones have to be rebuild...


scorpio810:
It seems to be fixed, thanks.
https://launchpad.net/~scorpio/+archive/ubuntu/qelectrotech-dev/+packages

PR #308, opened 2024-10-06
use MessageBox to inform user about additional info when importing scaled element

plc-user:
There is no special need for this PR as the message-box will "only" be shown, if single parts of the imported elmt-file had to be dropped...


scorpio810:
Thanks plc-user

PR #310, opened 2024-10-08
About QET: improvements in usability

plc-user:
I was irritated that I first had to select the most recent LOG file and that the most recent entry in the file was not displayed...


scorpio810:
Good job plc-user, thanks.

PR #312, opened 2024-11-16
Force light mode in collections like projects

Arusekk:
When in dark mode, symbols shown in collections menu are very difficult to view.
Forcing bright palette (like projects tree already does) solves the issue for me.

This is probably not the best way to code it (duplicated code), this palette should ideally be created in one place and reused. Feel free to adapt my suggestions in any way you like.


scorpio810:
Try this css, extract it and put in home QET like : $HOME/.qet/style.css
In settings -> configure QElectreotech -> in appearence  field  uncheck case use system color:
[style.zip](https://github.com/user-attachments/files/17789803/style.zip)

scorpio810:
Since you can use your patch and  also can choice of user style.css, I've decided to merge it, thank you.

PR #320, opened 2024-12-31
Fix: Only scroll diagram-view, when moved text leaves visible area

plc-user:
As described in this Forum-Thread:
[https://qelectrotech.org/forum/viewtopic.php?id=2864](url)


scorpio810:
Tanks, plc-user

PR #326, opened 2025-01-15
Update qet_cs.ts

pafri:
I've updated the translation of the Czech localization file. Please commit it.

Greetings.


scorpio810:
Thanks.

PR #331, opened 2025-01-27
machine_info: fix element-count and make static text a bit shorter

plc-user:
The custom elements were counted twice instead of adding company-elements...


scorpio810:
Hello plc-user and thank you for proofreading the code and correcting my mistatake miscalculation, uhu, where has my brain gone...

PR #338, opened 2025-02-01
minor changes

plc-user:
- remove dead code (local variables that were never used)
- typos
- white space
- comments

### DIFF
--- a/lang/qet_de.ts
+++ b/lang/qet_de.ts
@@ -3009,7 +3009,7 @@ Alle Bauteile und Unterordner von diesem Ordner werden ebenso gelöscht.</transl
     <message>
         <location filename="../sources/exportpropertieswidget.cpp" line="214"/>
         <source>SVG: fond transparent</source>
-        <translation>SVG-transparenter Hintergrund</translation>
+        <translation>SVG: transparenter Hintergrund</translation>
     </message>
 </context>
 <context>
--- a/sources/editor/graphicspart/partdynamictextfield.cpp
+++ b/sources/editor/graphicspart/partdynamictextfield.cpp
@@ -177,9 +177,7 @@ void PartDynamicTextField::fromXml(const QDomElement &dom_elmt) {
 		setFont(font_);
 	}
 	else if (dom_elmt.hasAttribute("font_size")) {
-#if TODO_LIST
-#pragma message("@TODO plc-user: remove again, when elements are converted!")
-#endif
+		// plc-user: for convenience - to support font-size from old 'input' ?")
 		setFont(QETApp::dynamicTextsItemFont(dom_elmt.attribute("font_size", QString::number(9)).toInt()));
 	}
 	else {
--- a/sources/editor/graphicspart/partpolygon.cpp
+++ b/sources/editor/graphicspart/partpolygon.cpp
@@ -116,7 +116,7 @@ void PartPolygon::fromXml(const QDomElement &qde)
 
 /**
 	@brief PartPolygon::toXml
-	Export this polygin in xml
+	Export this polygon in xml
 	@param xml_document : Xml document to use for create the xml element
 	@return an xml element that describe this polygon
 */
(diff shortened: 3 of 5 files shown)


scorpio810:
Hallo plc-user,

about "remove dead code (local variables that were never used)" this function create "rx" and "ry" anglr fot rectangle corner in element editor, ok so I comment these parts and I can draw a rectangle with rounded corners
This code was also user for qetshapeitem: https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/sources/qetgraphicsitem/qetshapeitem.cpp#L907

plc-user:
Hmm ... Let me explain how I see it:
We have a function here in which local variables “x” and “y” are declared. 
Values are assigned to these variables, but these variables are no longer used in the further course of this function.
How should the values from the variables then get into the corner rounding?

If I use my executable file, I can use it to create rectangles with rounding!:
`<definition hotspot_x="35" type="element" hotspot_y="35" link_type="simple" height="70" width="70" version="0.100.0">
    <uuid uuid="{7def9ac3-14f5-4482-8ee9-e464f2180d3b}"/>
    <names>
        <name lang="de">abgerundetes Rechteck</name>
    </names>
    <elementInformations/>
    <informations></informations>
    <description>
        <rect style="line-style:normal;line-weight:normal;filling:none;color:black" y="-30" x="-30" height="60" ry="5" rx="10" antialias="false" width="60"/>
    </description>
</definition>
`
Bildschirmfoto_2025-02-01_abgerundetes_Rechteck

scorpio810:
Thanks,
m_rect.normalized() was used further in the code, so the function was called twice.

plc-user:
Ok, let's take a closer look at the code part, maybe I missed something?

The contents of the local variables “x” and “y” are determined by assignment (“=”) from “m_xRadius” and “m_yRadius” respectively, but are not used any further, so they can be omitted.

The result of the call “m_rect.normalized()” is assigned to the locally created variable “rect”.

The variable “rect” can be omitted because nothing else is done with it.

What (perhaps) cannot be omitted is the call of “m_rect.normalized()”: This will (maybe) do something with the rectangle “m_rect”, which I haven't seen yet, but which might be necessary.

The Qt documentation about [QrectF::normalized()](https://doc.qt.io/qt-6/qrectf.html#normalized) says:
`Returns a normalized rectangle; i.e., a rectangle that has a non-negative width and height.`

In this context, this means that nothing is done with the “m_rect”, but that data from “m_rect” is assigned to the local variable “rect”, which is not used in the further course.   

From my point of view, part of the code can therefore be removed or remain removed without having to fear any effects on the rectangle “m_rect”!

plc-user:
This code-fragment seems to be the same:
[sources/qetgraphicsitem/qetshapeitem.cpp#L909](https://github.com/qelectrotech/qelectrotech-source-mirror/blob/de0fa1fa209d57a78a362bd81d2450726a245337/sources/qetgraphicsitem/qetshapeitem.cpp#L909)
The code from line 909 to 918 can be removed from my point of view.

scorpio810:
what's strange is the "to XML" and not from XML, so it's not a backward compatibility code.

scorpio810:
Function was made here https://github.com/qelectrotech/qelectrotech-source-mirror/blob/de0fa1fa209d57a78a362bd81d2450726a245337/sources/qetgraphicsitem/qetshapeitem.cpp#L714

scorpio810:
> This code-fragment seems to be the same: [sources/qetgraphicsitem/qetshapeitem.cpp#L909](https://github.com/qelectrotech/qelectrotech-source-mirror/blob/de0fa1fa209d57a78a362bd81d2450726a245337/sources/qetgraphicsitem/qetshapeitem.cpp#L909) The code from line 909 to 918 can also be removed from my point of view. At least I haven't seen any effects since I commented it out in my local executable of QET.

I 'm think you can make a PR to delete these dead code?

plc-user:
> Function was made here
> 
> https://github.com/qelectrotech/qelectrotech-source-mirror/blob/de0fa1fa209d57a78a362bd81d2450726a245337/sources/qetgraphicsitem/qetshapeitem.cpp#L714

it seems the code for rectangle was introduced in sources/qetgraphicsitem/qetshapeitem.cpp in this [commit](https://github.com/qelectrotech/qelectrotech-source-mirror/commit/eeb49d5a1859dec378271b28bd569260c66279bd#diff-a449b0d029dcce84f47803722b7ed94465f5896c2b59bfadf5178b5af8fdeba1) by blacksun on Jul 14, 2018

PR is prepared and on the way...!

PR #341, opened 2025-02-02
more precise Log-Text for search of "qet_tb_generator"

plc-user:
to avoid misunderstandings:
as long as plugin is searched, "not found: search-location" is logged
as soon as the plugin is found, the search ends and the place of plugin is written to log


scorpio810:
Thanks plc-user,
see why : https://qelectrotech.org/forum/viewtopic.php?pid=20955#p20955

PR #344, opened 2025-02-05
some clean-up for element-file and in code

plc-user:
- for some time now and for whatever reason, element-editor
  sometimes adds element-information without content –> do not
  save info-lines without any content
- sort element-information alphabetically by name in element-file
- use trimmed strings for element-information to remove leading
  and trailing whitespace
- fix typo
- NamesList: move while-loop into "else"-clause, to be more precise 
- added English text to comments


scorpio810:
Hallo plc-user,

maybe see this article
https://stackoverflow.com/questions/27378143/qt-5-produce-random-attribute-order-in-xml

PR #345, opened 2025-02-06
export: set maximum width / height according specifications of export-type

plc-user:
see also discussion in QET-Forum from here:
https://qelectrotech.org/forum/viewtopic.php?pid=21000#p21000


scorpio810:
Thanks. ;)

PR #350, opened 2025-02-09
some cleaning for element-file

plc-user:
limit decimal places for floats
and typos and whitespace -corrections


scorpio810:
After some quick checks, I saw no problems, thanks plc-user. ;-)

PR #358, opened 2025-02-16
some more rotation, mirror and flip

plc-user:
The “text” elements look so harmless, but are tricky to mirror and flip!
I'll have to sleep on it for another night or two...


scorpio810:
Great, thanks

scorpio810:
Please keep your nights quiet , we have the time . ;-)

plc-user:
Don't worry: There will be fewer PRs from me in the near future. So I will get more sleep again!  ;-)

PR #362, opened 2025-02-18
Fix typo widht to width

pkess:
While searching for something different i saw that there is a settings key not written in a consistent way. So in some places it was already spelled `dynamic_text_width` but within evaluation it was written `dynamic_text_widht` which is only a typo i think.

I fixed this and found another place where it was spelled wrong as well

I did not do testing on this cause i do not really know what is meant here.


scorpio810:
Merged, thanks

scorpio810:
> I did not do testing on this cause i do not really know what is meant here.
Just user's config for dynamic text saved in settings, I think users could re save it without issue.

PR #369, opened 2025-02-25
Added slovak translation to org.qelectrotech.qelectrotech.desktop

prescott66:
Added slovak translation for app name and comment.


scorpio810:
Thanks

PR #376, opened 2025-03-02
simplify and use the same code for Qt5 & Qt6

plc-user:
When reading and comparing [Qt5-docs](https://qthub.com/static/doc/qt5/qtwidgets/qstyleoptiongraphicsitem.html#levelOfDetailFromTransform) and [Qt6-docs](https://doc.qt.io/qt-6/qstyleoptiongraphicsitem.html#levelOfDetailFromTransform), I read, that there are no differences in the functions!
So it is not necessary to have the differentiation between the Qt-Versions.
Code compiles without errors or warnings for Qt5 and Qt6 and there are no visible differences in elements.
But check yourself, before merging...


scorpio810:
I haven't had any problems from the Ubuntu PPA CI (builder's) for focal, Jammy and of course Noble. So it's ok . ;)

PR #379, opened 2025-03-05
unify some more code for Qt5 & Qt6 (and more)

plc-user:
- set default-value for QFont::Weight to prevent warning at runtime
- remove unused NamesList from qetproject.cpp / qetproject.h
- unify NamesList "imported elements" to be used with Qt5 & Qt6
  we still need to decide whether to use list of QChars or written text in utf-8; kept both variants for now
- corrected comment
- use same code for Qt5 & Qt6 for XRef-list
  Until now the code was just added for Qt6, but it can be used by Qt5 as well.

### DIFF
--- a/sources/ElementsCollection/xmlelementcollection.cpp
+++ b/sources/ElementsCollection/xmlelementcollection.cpp
@@ -46,125 +46,64 @@ XmlElementCollection::XmlElementCollection(QETProject *project) :
 	collection.appendChild(import);
 
 	NamesList names;
-#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0) // ### Qt 6: remove
-
-	const QChar russian_data[24] = {
-		0x0418, 0x043C, 0x043F, 0x043E, 0x0440, 0x0442, 0x0438, 0x0440,
-		0x043E, 0x0432, 0x0430, 0x043D, 0x043D, 0x044B, 0x0435, 0x0020,
-		0x044D, 0x043B, 0x0435, 0x043C, 0x0435, 0x043D, 0x0442, 0x044B};
+	const QChar chinese_data[4] ={
+		QChar(0x5BFC), QChar(0x5165), QChar(0x5143), QChar(0x4EF6)};
 	const QChar greek_data[18] = {
-		0x0395,
-		0x03b9,
-		0x03c3,
-		0x03b7,
-		0x03b3,
-		0x03bc,
-		0x03ad,
-		0x03bd,
-		0x03b1,
-		0x0020,
-		0x03c3,
-		0x03c4,
-		0x03bf,
-		0x03b9,
-		0x03c7,
-		0x03b5,
-		0x03af,
-		0x03b1};
-	const QChar turkish_data[12] = {
-		0x0130,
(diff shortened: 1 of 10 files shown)


scorpio810:
Something has changed, it seems to me that in the display  diagrams, 
the lines of conductors, elements, etc, are thinner since this PR, on my UHD screen it tires the sight I have to zoom a little to see the diagrams...

plc-user:
The setting of QPen for being "Cosmetic" has changed, but this change will be used as standard for Qt6!

As you can read in [Qt5-docs](https://doc.qt.io/qt-5/qstyleoptiongraphicsitem-obsolete.html), "levelOfDetail" was marked as obsolete already in Qt5!

Cite:
_(...) is provided to keep old source code working.
We strongly advise against using it in new code.

Use QStyleOptionGraphicsItem::levelOfDetailFromTransform()
together with QPainter::worldTransform() instead._
End Cite.

Maybe we need to adjust the value of "1.0" to set the QPen as cosmetic?
`t.setCosmetic(options && options -> levelOfDetailFromTransform(painter->worldTransform()) < 1.0);`

Can you try to use other values to be more "eye-friendly", Laurent?

plc-user:
Maybe start with conductors:
when changing the value from 1.0 to 2.0 [here](https://github.com/qelectrotech/qelectrotech-source-mirror/blob/40a79959931a4148d43c60e69e78570210620a00/sources/qetgraphicsitem/conductor.cpp#L523C62-L523C67) conductors look thicker!

```
		//Use a cosmetique line, below a certain zoom
	if (options && options->levelOfDetailFromTransform(painter->worldTransform()) < 2.0)
```

scorpio810:
here with a value < 0.5 it displays much better and the lines are blacker and more visible than before.

Can you try it yourself, plc-user?



```
--- sources/qetgraphicsitem/conductor.cpp
+++ sources/qetgraphicsitem/conductor.cpp
@@ -520,7 +520,7 @@ void Conductor::paint(QPainter *painter, const QStyleOptionGraphicsItem *options
 	final_conductor_pen.setJoinStyle(Qt::SvgMiterJoin); // better rendering with dot
 
 		//Use a cosmetique line, below a certain zoom
-	if (options && options->levelOfDetailFromTransform(painter->worldTransform()) < 1.0)
+	if (options && options->levelOfDetailFromTransform(painter->worldTransform()) < 0.5)
 	{
 		final_conductor_pen.setCosmetic(true);
 	}
--- sources/qetgraphicsitem/element.cpp
+++ sources/qetgraphicsitem/element.cpp
@@ -225,7 +225,7 @@ void Element::paint(
 	QBrush brush;
 	painter->setPen(pen);
 	painter->setBrush(brush);
-	if (options && options->levelOfDetailFromTransform(painter->worldTransform()) < 1.0)
+	if (options && options->levelOfDetailFromTransform(painter->worldTransform()) < 0.5)
 	{
 		painter->drawPicture(0, 0, m_low_zoom_picture);
 	} else {
```

scorpio810:
Ok and thanks so you can merge it this change?

scorpio810:
Thanks!

PR #381, opened 2025-03-06
implement QtConcurrent::run - call for "writeBackup" (Qt6)

plc-user:
We need to add a QT_VERSION_CHECK(6, 0, 0) also to the 
function in qet.cpp/qet.h itself, because the call with 
QtConcurrent::run needs a "const" for first parameter, 
as I read from docs and compile-errors...
Seems to work here: stalefile(s) are created!
But as always: Please double-check! ;)

### DIFF
--- a/sources/qet.cpp
+++ b/sources/qet.cpp
@@ -754,7 +754,14 @@ QActionGroup *QET::depthActionGroup(QObject *parent)
 	return action_group;
 }
 
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0) // ### Qt 6: remove
 bool QET::writeToFile(QDomDocument &xml_doc, QFile *file, QString *error_message)
+#else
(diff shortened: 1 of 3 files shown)


elevatormind:
This fixes it for me. But is the QT_VERSION_CHECK needed? The point of the qt6-cmake branch is to migrate away from qt5.
Added this change in parallel on my repo's qt6-cmake branch as I needed tis fix!

plc-user:
You're right, of course: If we think of Qt6 alone, we don't need the version check!
Must have gotten mixed up with all the changes and additions in the last days...

We can let Laurent decide: 
This PR can go to the master branch and if we remove the version check (what you already did for your repo, @elevatormind?), then that reduced version can go to the qt6-cmake - branch.

plc-user:
Close this Pull-Request, because the code is not needed in master-branch for Qt5 and @elevatormind uses this new call in his qt6-cmake - branch!

PR #387, opened 2025-03-15
only calculate grid-point-size, when min != max

plc-user:
and some English comments

### DIFF
--- a/sources/diagram.cpp
+++ b/sources/diagram.cpp
@@ -186,7 +186,7 @@ void Diagram::drawBackground(QPainter *p, const QRectF &r) {
 	p -> drawRect(r);
 
 	if (draw_grid_) {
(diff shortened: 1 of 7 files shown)


scorpio810:
Hallo plc-user,
I put you to admin since long time ago, no need I think to make an PR you can push it directly, I don't health for control each PR !

plc-user:
Salut Laurent !
Having a second reader here is always a bit reassuring not to have uploaded some rubbish, and I find the discussion about changes beneficial too! 
Remember our discussion about dead code? 
That gives an opportunity to think more carefully about the changes uploaded, again.
I don't want to push you too hard in this state of health, so I'll try to think even more carefully about the code before releasing it!
Thank you for your confidence!

PR #389, opened 2025-04-01
Added updated automatic doxygen build on push +  theme to make it fit with docs page

Int-Circuit:
In effort to modernize the  QET doc and source code, i have implemented a GitHub action to automatically create and deploy the doxygen doc on push. This allows it to be more frequently follow upstream and be up-to-date.

Looks have also been update to follow the docs new theme.

You can try the deployment here:
https://int-circuit.github.io/qelectrotech-source-mirror/

This MR is for the time being marked as draft to get community opinion and perhaps get knowledge on improvements to the doxygen, since the [used theme](https://jothepro.github.io/doxygen-awesome-css/index.html) allows for customization. Feel free to comment


scorpio810:
```
time doxygen Doxyfile
type lookup cache used 40528/65536 hits=255647 misses=41264
symbol lookup cache used 50914/65536 hits=317159 misses=50914
finished...

real    1m24,466s
user    7m39,602s
sys     5m2,236s
```

It 's use 2.5% CPU here Threadripper 3960X

scorpio810:
Yes we will be interested but if you can clean MR history?
Git patch was very big https://github.com/qelectrotech/qelectrotech-source-mirror/pull/389.patch

scorpio810:
So we can't remove QElectroTech.qch, it's only used for working with qtcreator!

Int-Circuit:
Yes it is. So the setup part is the following:
- Go to the user setting page -> developper settings.
- Create a finegrated personal access token (PAT) with contents: write and pull-requests: write scopes. (Required to have the action run on push). I recommend limiting the token to the source-mirror repo only.
- Add that token to the repo's secrets. (Repo settings -> secrets and variables -> actions -> repo secrets
- **(OPTIONAL STEP)** if you name your secret other than the default in this MR (Here MR_TOKEN), you will need to change this in the action code:
```
        token: ${{ secrets.MR_TOKEN }}
```
to whatever you use before merging, otherwise the action won't work.
- merge the mr. After the action executes, you should see an open mr with the labels qch and cicd that is trying to merge the updated qch file. To test it, import it into qt creator you should see the docs (might need to reopen the program if you see no changes, had this problem and a restart fixed it).
And voila, doxygen is built and the qch file stays up to date.

PS: regarding the mr action, it won't open a new one if there is one already open, it will just update its content

> How the action behaves:
  If there are changes (i.e. a diff exists with the checked-out base branch), the changes will be pushed to a new branch and a pull request created.
 If there are no changes (i.e. no diff exists with the checked-out base branch), no pull request will be created and the action exits silently.
  If a pull request already exists it will be updated if necessary. Local changes in the Actions workspace, or changes on the base branch, can cause an update. If no update is required the action exits silently.
    If a pull request exists and new changes on the base branch make the pull request unnecessary (i.e. there is no longer a diff between the pull request branch and the base), the pull request is automatically closed. Additionally, if delete-branch is set to true the branch will be deleted.

https://github.com/peter-evans/create-pull-request/tree/main?tab=readme-ov-file#action-behaviour

Please look at the readme for more details about the action used

Int-Circuit:
Screenshot_20260105-130343_1

Int-Circuit:
@scorpio810 Did you add the secret? The action fails because it cannot find the secret
```

Invalid workflow file: .github/workflows/auto-doxygen.yml#L1
(Line: 31, Col: 16): Unrecognized named-value: 'MR_TOKEN'. Located at position 1 within expression: MR_TOKEN
```

PR #404, opened 2025-09-08
Fixing translation file list in CMake

arummler:
No rule to make target ... qelectrotech-source-mirror/lang/qet_be.ts

Compilation fails.

It had been renamed and two other languages missing in the CMakeList.txt


De-Backer:
**While you are right about Cmake**, it is indeed recommended to switch from QMake to CMake.
You will have to repair CMake continuously, if anyone create a file, rename it and delete it.

Just look at this log:
https://github.com/qelectrotech/qelectrotech-source-mirror/commits/master/?author=De-Backer

I wish you lots of strength and perseverance.

out of curiosity:
Do you have KF5 libraries working in CMake yet?

PR #412, opened 2025-12-28
Delete outdated QET docs

Int-Circuit:
Delete outdated odt and pdf docs as these are superseded by the updated sphinx docs.


scorpio810:
docs/Documentation.webloc
Was used in macOS scripts::

https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/misc/MacQetDeploy.sh#L245

PR #422, opened 2026-02-05
Add Korean language support in xml element collection

Kyle-Code-CA:
Fixed as requested: added Korean translation using QChar unicode only, without changing formatting or logic. 고맙습니다.


scorpio810:
I still need to package the Appimages, Flatpak, and macOS Intel and Apple Silicon bundles.

PR #449, opened 2026-04-09
Automatic Terminal Numbering Tool

Kellermorph:
This Pull Request introduces a new feature for the automatic numbering of terminals across an entire project. This tool significantly speeds up the workflow when dealing with large terminal blocks, replacing manual labeling or external scripts.
Integrated GUI: A new dialog (TerminalNumberingDialog) allows users to choose between X-axis or Y-axis priority for sorting.

Smart Sorting: Terminals are sorted by their prefix, then by folio index, and finally by their physical coordinates on the diagram (with a 1px tolerance for alignment).

Alphanumeric Support: Users can choose to include or exclude alphanumeric suffixes.

Preservation of Special Terminals: Standard terminal labels like "N", "PE", or "L1" are preserved to maintain electrical correctness, while still being counted in the sequence to ensure correct numbering for subsequent terminals.

Full Undo/Redo Support: The numbering process is encapsulated in a QUndoCommand macro, allowing users to revert all changes with a single "Undo" action.

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -88,6 +88,7 @@ set(QET_RES_FILES
   ${QET_DIR}/sources/ui/dynamicelementtextitemeditor.ui
   ${QET_DIR}/sources/ui/elementinfopartwidget.ui
   ${QET_DIR}/sources/ui/elementinfowidget.ui
+  ${QET_DIR}/sources/ui/terminalnumberingdialog.ui
   ${QET_DIR}/sources/ui/formulaassistantdialog.ui
   ${QET_DIR}/sources/ui/imagepropertieswidget.ui
   ${QET_DIR}/sources/ui/importelementdialog.ui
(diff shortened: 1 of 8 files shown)


scorpio810:
Maybe on the forum.

Alfmat01:
Hi!
This is a great feature! And it works really well. Thanks @Kellermorph !
As a suggestion:
Would it be possible to select only the terminal blocks to which you want to apply automatic numbering?
Currently, all terminals on a terminal block are automatically numbered, unless you mark each terminal as “exclude from automatic numbering.”
I think that by adding the selection option I suggested above, the tool would be more flexible for users.

Kellermorph:
Hi Alfmat01,
I'm glad to hear that the feature is helpful.
Yes, the exclusion option is intended for individual terminals that should definitely retain their own labels.
I’d have to see if and how it would be possible to read the terminal blocks before numbering them and then select separately which ones should be affected. I’d like to focus on a few other features first. Then I might have time to look into it. Or why don’t you grab Gemini and try it yourself?
Tip: You’ll need to use Pro mode for this.

PR #452, opened 2026-04-19
Fix losing Focus on moving diagram position with keyboard

ChuckNr11:
Hi Laurent,
I hope this solves the problem.


scorpio810:
Hi  Achim, that's fix it, thanks you.

PR #455, opened 2026-04-29
Fix Thumbnail in Makrotree

Kellermorph:


scorpio810:
Hallo Thomas that didn't work here, I pushed your PR without testing it. :(

Kellermorph:
Hi Laurent
What isnt working? 
It worked by me.
Maybe i deleted to much by deleting the geman notes? I will take a look as soon as i am home

scorpio810:
Pixmaps in makro tree isn't refreshed. And now we haven't any pixmaps in makro tree..

Kellermorph:
Yes, previously the wrong images were being displayed in the macro tree. It was showing some random components. That's why I've now intentionally removed it entirely. My thinking was, better nothing than something faulty.
I wanted to tackle another bug first, but I can get to it after that. How should the preview in the macro tree look?

scorpio810:
I’m fine with that; better no pixmap than a faulty one.

PR #464, opened 2026-05-21
New element: Line definition

Kellermorph:
New element for wire definition.
Advantages: It doesn't create an open end like a sheet reference, and I can filter it out of the wiring diagram in the subsequent pull request to prevent incorrect connections from being displayed, unlike with the sheet reference.


scorpio810:
You could make an little video for explain?

Kellermorph:
yeah i could do it but it's the same like a sheet reference. create it, place the dynamic text, save and place on the sheet. 
Should I do it anyway?

scorpio810:
It's related to Achim 's cable management?

Kellermorph:
No, it's a standalone app.
It's more about what's described in this forum post.

https://qelectrotech.org/forum/viewtopic.php?id=3071

And since conductor properties are no longer displayed via the layer reference, I can filter them out of the wiring diagram and have them displayed correctly without the open ends

PR #477, opened 2026-06-01
Fix: Dynamic element text shifting/jumping when duplicating diagrams

Kellermorph:
When duplicating a diagram, the content is exported and imported via XML. During this process, the layout calculation of Qt triggers shifts in dynamic texts.
The safeguards introduced in PR #399 (correctTextPos() and restoreText()) to freeze and restore text alignments were not applied here because they were declared as protected inside the Diagram class, making them inaccessible to the ElementsPanelWidget.

### DIFF
--- a/sources/diagram.h
+++ b/sources/diagram.h
@@ -142,10 +142,11 @@ class Diagram : public QGraphicsScene
 		void wheelEvent (QGraphicsSceneWheelEvent *event) override;
 		void keyPressEvent (QKeyEvent *event) override;
 		void keyReleaseEvent (QKeyEvent *) override;
-		void correctTextPos(Element* elmt);
-		void restoreText(Element* elmt);
+
 	
 	public:
+		void correctTextPos(Element* elmt);
+		void restoreText(Element* elmt);
 		QUuid uuid();
 		void setEventInterface (DiagramEventInterface *event_interface);
 		void clearEventInterface();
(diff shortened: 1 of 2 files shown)


scorpio810:
e/5.15.17/QtCore -I/usr/include/x86_64-linux-gnu/qt5/QtCore -Idebug -I. -I/usr/lib/x86_64-linux-gnu/qt5/mkspecs/linux-g++ -o debug/qetmessagebox.o ../sources/qetmessagebox.cpp
../sources/elementspanelwidget.cpp:28:10: fatal error: element.h: Aucun fichier ou dossier de ce nom
   28 | #include "element.h"
      |          ^~~~~~~~~~~
compilation terminated.
make: *** [Makefile.Debug:5244: debug/elementspanelwidget.o] Error 1
make: *** Attente des tâches non terminées....

Kellermorph:
Hi Laurent.
Sorry for the late reply. I just got back from vacation.
It's strange that there's an error. When I compiled it on my PC, it worked fine. Anyway, have you fixed the error yet, or do I need to update it?

scorpio810:
Hi Thomas,  I fixed this see https://github.com/qelectrotech/qelectrotech-source-mirror/commit/1572c23d51ba5c6fa2d46adefaaf18db52fdcc66

Kellermorph:
Thank you

PR #479, opened 2026-06-04
fix possible crashes in crossrefitem

ChuckNr11:
fix access to QList with potentially out-of-bounds index


scorpio810:
Thanks, Achim, I think will try  your new patch these next days.

Issue #309, opened 2024-10-07
CLI  export option


BotiKepi:
Hi,
I'm not an every day user of QElectrotech, but I do appreciate it (thanks for that !).
Yet, I'm used to version control my files (with git), even files such as .qet .

Having the ability to export pdf (or such graphic output) through CLI let the ability to graphically enlight differences when making a git diff,  before committing for instance (the what-have-I-done-since-last-commit syndrom ;-) ).
This can be done using git attributes, which let us automate the process : exporting 2 versions of graphic rendering -- pdf, png, svg ...-- and diffing the graphics (in various ways).

I use that process with KiCad  (PCB design) : it's not perfect, but quite convenient.

Hence, having a CLI export option (like for instance, --export-pdf)  would be appreciated .

This is somewhat related to #162 , yet the need here is smaller : no need for a scripting language, "just" restricted actions such as export, through options (and without launching GUI).

Yeah, sorry for the "just" word, I don't know how big effort this would take, but I guess this is not a small piece of cake.

ispyisail:
This is now available in master — the headless CLI added in #483 and extended in #489 covers what's requested here:

- **Graphic export for git diffs:** `qelectrotech --export-pdf <project.qet> <out.pdf>` (also `--export-png` / `--export-svg`), which is exactly the KiCad-style 'render two revisions and diff them' workflow described.
- **Connections as a list** (the follow-up ask): `--export-cables` and `--export-wires` (CSV), plus `--export-nets` (JSON, connected-terminal groups) and `--export-links` (cross-references) from #489.

So the CLI export need here looks fully met. @scorpio810 — this could probably be closed (the broader scripting/CI vision lives on in #162). Thanks @BotiKepi and the others for the original request.

ispyisail:
Closing per the above — CLI export (--export-pdf/png/svg/cables/wires/nets/links) fully covers what was asked here.

Issue #162, opened 2021-03-25
Feature Request: CLI for scripts


jkrenzer:
Hi!

I just wanted to drop an idea:

Adding a command-line-interface for functions like opening files, saving, import and export operations would be really helpful. This could also be achieved by exposing all functionality internally to a language-bindung (like python) so a CLI could be developed independently from the main-gui program which would then be used by the CLI as backend. In latter case integrating a scripting language natively would also be within grasp.

The ability to run without GUI on a shell would be needed either way.

Just two example use-case:

1. Scripting locally
    - Use a script to automate getting all derived files (CSV, PDFs etc.) exported after you have done a new revision of a file
    - Use a unified script so all projects done will get the same layout of derived files
2. Using CI
    - Keep your *.qet files in a git, hosted on github, gitlab or a similar webservice with CI backend
    - Use github/gitlab CI actions to build and publish plan PDFs, BOM, list-of-cables, CSV-exports etc. automatically after each push or merge to the repository

Cheers,

Jörn

scorpio810:
Hi Jörn,

you are welcome.
Can you take care of it?

jkrenzer:
Sorry, this somewhat drowned in my inbox. Currently I am quite flooded with work, so I cannot afford to spend time on this. I might take a look later on if this issue is still open then. Still I think it is good to keep this feature-request around to keep track of this for the time being.

ispyisail:
Hi Laurent,

Now that the CLI export/inspection tools are in (#483, #489, #490), I've been thinking about the bigger idea in this issue — exposing QET's internals as a **Python binding** so scripts could drive it as a library, rather than only through discrete CLI flags.

One thing the CLI work showed me: QET's core already runs fine headless (`QETProject`, `Diagram`, `Element`, etc. all work under `QT_QPA_PLATFORM=offscreen`), so binding those existing classes (e.g. via pybind11) looks feasible without major surgery to the core — it'd run offscreen like the CLI does.

The use cases are the same ones in this issue, plus the automation/AI angle we touched on — agents and pipelines can do a lot more with a real API than with flags.

Before I look into it seriously, though, I wanted to gauge your appetite: **would you want Python bindings upstream at all?** It's a real addition — a new build dependency and a binding surface that becomes part of what the project maintains long-term — so it's very much your call, not something I'd want to build speculatively and drop on you. Totally fine if it's not a direction you want; I'm happy to keep the CLI tools as the scripting surface either way.

[...]

PR #483, opened 2026-06-10
Add headless command-line export (PDF / PNG / SVG / cable-list / wire-list)

ispyisail:
## Headless command-line export

Adds a headless (no-GUI) command-line export so projects can be rendered and
exported in batch / CI pipelines without opening the editor. This is a
long-standing request (qelectrotech.org bugtracker #171, GitHub #309; the
list export addresses @pkess's request in #162 to "export all connections as
a list").

### Usage

```
qelectrotech --export-pdf    <project.qet> <output.pdf>
qelectrotech --export-png    <project.qet> <output_dir>
qelectrotech --export-svg    <project.qet> <output_dir>
qelectrotech --export-cables <project.qet> <output.csv>
qelectrotech --export-wires  <project.qet> <output.csv>
```

- **pdf** — one multi-page document, one diagram per page.
- **png / svg** — one file per diagram, named `<NN>_<title>.<ext>`.
- **cables** — wiring list (one row per conductor) as CSV.
- **wires** — list of distinct wire numbers as CSV.

### How it works

- The export request is detected in `main()` **before** `SingleApplication`
  is created, so a batch run does not attach to / focus a running GUI
  instance. A plain `QApplication` is used (offscreen-capable).
- Rendering reuses `Diagram::render()` over
  `BorderTitleBlock::borderAndTitleBlockRect()` — the same path the GUI export
  uses — so output matches the on-screen drawing.
- The image/PDF code lives in a new `CLIExport` namespace
  (`sources/cli_export.{h,cpp}`); `main.cpp` only gains a small dispatch hook.

### List exports

The CSV list exports reuse the existing exporters rather than duplicating
logic:

- `--export-cables` → `WiringListExport`. To make it usable headlessly,
  `WiringListExport::toCsv()` (which mixed CSV generation with a file dialog
  and file write) is refactored: a new const method `toCsvString()` builds and
  returns the CSV, and `toCsv()` now calls it before doing the dialog + write.
  **No behavioural change to the existing GUI export.**
- `--export-wires` → `ConductorNumExport::wiresNum()` (already a pure string
  producer).

### Testing

- PDF/PNG/SVG verified against native example projects (e.g.
  `Habitat-Schemas_developpes.qet`) — output opens and matches the GUI render.
- `--export-cables` / `--export-wires` verified end-to-end: both exit 0 and
  produce the same CSV the GUI exporters produce.

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -107,6 +107,8 @@ set(QET_RES_FILES
   ${QET_DIR}/sources/ui/configpage/generalconfigurationpage.ui
   )
 set(QET_SRC_FILES
+  ${QET_DIR}/sources/cli_export.cpp
+  ${QET_DIR}/sources/cli_export.h
   ${QET_DIR}/sources/borderproperties.cpp
   ${QET_DIR}/sources/borderproperties.h
   ${QET_DIR}/sources/bordertitleblock.cpp
--- a/sources/cli_export.cpp
+++ b/sources/cli_export.cpp
@@ -1,0 +1,211 @@
+/*
+	Copyright 2006-2025 The QElectroTech Team
+	This file is part of QElectroTech.
+
+	QElectroTech is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 2 of the License, or
+	(at your option) any later version.
+
+	QElectroTech is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with QElectroTech.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#include "cli_export.h"
+
+#include "bordertitleblock.h"
+#include "diagram.h"
+#include "qetproject.h"
+
+#include <QDir>
+#include <QFileInfo>
+#include <QPainter>
+#include <QPdfWriter>
+#include <QSvgGenerator>
+#include <QTextStream>
+
+namespace {
+
+QTextStream out(stdout);
+QTextStream err(stderr);
+
+/// All export option flags, mapped to a short format name.
+const QHash<QString, QString> &exportFlags()
+{
+	static const QHash<QString, QString> flags {
+		{"--export-pdf", "pdf"},
+		{"--export-png", "png"},
+		{"--export-svg", "svg"},
+	};
+	return flags;
+}
+
+/// Pixel rect of a diagram's border + title block (the printable page area).
+QRect diagramRect(Diagram *diagram)
+{
+	QRectF r = diagram->border_and_titleblock.borderAndTitleBlockRect();
+	r.adjust(0, 0, 1, 1); // include the 1px border line
+	return r.toAlignedRect();
+}
+
+/// A filesystem-safe per-diagram file stem: "01_Title".
+QString diagramStem(Diagram *diagram, int index)
+{
+	QString title = diagram->title();
+	title.replace(QRegularExpression("[^\\w \\-]"), "_");
+	title = title.simplified();
+	if (title.isEmpty())
+		title = "diagram";
+	return QStringLiteral("%1_%2")
+		.arg(index, 2, 10, QChar('0'))
+		.arg(title);
+}
+
+/// Render @p diagram into @p painter, fitting @p target to the page rect.
+void renderDiagram(Diagram *diagram, QPainter &painter, const QRectF &target)
+{
+	const QRect source = diagramRect(diagram);
+	const bool was_drawing_grid = false; // export without the editor grid
+	Q_UNUSED(was_drawing_grid)
+	diagram->render(&painter, target, source, Qt::KeepAspectRatio);
+}
+
+int exportPdf(QETProject &project, const QString &output)
+{
+	const QList<Diagram *> diagrams = project.diagrams();
+	if (diagrams.isEmpty()) {
+		err << "No diagrams to export.\n";
+		return 1;
+	}
+
+	QPdfWriter writer(output);
+	writer.setCreator("QElectroTech");
+	writer.setResolution(96);
+
+	QPainter painter;
+	bool first = true;
+	for (Diagram *diagram : diagrams) {
+		const QRect r = diagramRect(diagram);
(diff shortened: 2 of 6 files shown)


scorpio810:
Missing   in qelectrotech.pro
${QET_DIR}/sources/cli_export.cpp
${QET_DIR}/sources/cli_export.h

scorpio810:
```
nu/libQt5Xml.so /usr/lib/x86_64-linux-gnu/libQt5Sql.so /usr/lib/x86_64-linux-gnu/libQt5Concurrent.so /usr/lib/x86_64-linux-gnu/qt5/mkspecs/modules/../../../../../lib/x86_64-linux-gnu/libKF5CoreAddons.so /usr/lib/x86_64-linux-gnu/libQt5Core.so -lGL   
/usr/bin/x86_64-linux-gnu-ld.bfd : debug/cli_export.o : dans la fonction « CLIExport::isExportRequest(QStringList const&) » :
/home/laurent/Qet-svn/git/qet/build/../sources/cli_export.cpp:199 : définitions multiples de « CLIExport::isExportRequest(QStringList const&) »; debug/cli_export.o:/home/laurent/Qet-svn/git/qet/build/../sources/cli_export.cpp:199 : défini pour la première fois ici
/usr/bin/x86_64-linux-gnu-ld.bfd : debug/cli_export.o : dans la fonction « CLIExport::run(QStringList const&) » :
/home/laurent/Qet-svn/git/qet/build/../sources/cli_export.cpp:207 : définitions multiples de « CLIExport::run(QStringList const&) »; debug/cli_export.o:/home/laurent/Qet-svn/git/qet/build/../sources/cli_export.cpp:207 : défini pour la première fois ici
collect2: error: ld returned 1 exit status
make: *** [Makefile.Debug:1707: qelectrotech] Error 1
```

ispyisail:
Thanks Laurent, appreciate you catching and fixing that. I'd missed that incremental qmake builds don't re-run the $$files() globs until the .pro itself changes — good to know for next time. 🙏

scorpio810:
@ispyisail , it is my fault :  
sources/*.cpp → flat glob, covers the root of sources/ only → no explicit entry needed
sources/SubModule/*.cpp → each subdirectory has its own dedicated $$files(...) in the list

So any .cpp sitting directly in sources/ is already covered — no need to add it manually.

scorpio810:
Hi ispyisail, thanks you for your work, but please disable the grid on CLI PDF export, I saw also we cant use PDF linking with XREFS, folio reports.

ispyisail:
Thanks Laurent!

- **Grid:** fixed in #485. `renderDiagram()` had a leftover no-op stub (`was_drawing_grid` was set but `Q_UNUSED`'d), so the grid was still being painted. It now toggles `setDisplayGrid(false)` around the render and restores it after — and since PDF/PNG/SVG all share that path, all three formats are covered.
- **PDF linking with XREFs / folio reports:** good catch. That one's a bigger piece — the export draws through `QPainter`/`QPdfWriter`, which doesn't expose clickable link annotations for drawn content, so it'd need a different approach (overlaying link rectangles at the XREF/folio targets). I'll look at it as a separate enhancement rather than hold up the grid fix.

scorpio810:
https://github.com/qelectrotech/qelectrotech-source-mirror/commit/cd76b6a1d63d21d038dee2db3cfc41c79f080262

ispyisail:
Quick correction to my last message — I dug into `ProjectPrintWindow` and I see QET already does exactly this for the GUI PDF export (`QPdfEngine::drawHyperlink()` over the `CrossRefItem`/folio rects, plus the URI→GoTo/FitR post-processing). So it's not new ground — the CLI just renders via `QPdfWriter` without injecting the links. The right fix is to reuse that logic for the CLI rather than reinvent it; I'll look at factoring it so both share the same path. Thanks for flagging.

ispyisail:
Following up on PDF cross-reference links for CLI parity — I went through your GUI implementation (cd76b6a) and most of it looks reusable:

- `pdfConvertUriToGoTo()` is already a static free function, so the CLI can use it as-is.
- The link injection in `printDiagram()` only needs the paint engine to be a `QPdfEngine`, and `QPdfWriter` (what the CLI export uses) wraps one too — so `drawHyperlink()` is reachable from the CLI path.
- The `gui-private` build setup you added already covers `cli_export.cpp`.

The one non-trivial part is the scene→page geometry: the injection is matched to `QGraphicsScene::render()` under the QPrinter page layout, whereas the CLI renders via `QPdfWriter` (96 dpi, per-page point size, no margins), so that mapping has to be re-derived for the CLI.

My proposed approach: factor the injection + post-processing into a small shared helper (e.g. `pdf_links.{cpp,h}`) that both `ProjectPrintWindow` and the CLI `exportPdf` call, so there's no duplicated logic. But since that refactors your print window, I'd rather check with you first:

- Are you happy with extracting a shared helper out of `ProjectPrintWindow`, or would you prefer the CLI keep its own copy / a different split?
- Any preference on sequencing — before or after the other CLI verification/export tools I still have pending?

Happy to do it whichever way you'd like.

scorpio810:
1. Shared helper yes, but keep it minimal

Extracting pdf_links.{cpp,h} is the right approach. pdfConvertUriToGoTo() and the link-injection logic belong there. I'd keep ProjectPrintWindow as a pure caller — no logic moved out of it, just a call replaced by the shared helper. That way the refactor is surgical and reviewable.

2. Scene→page geometry

Agreed that's the only real work. For the CLI path (QPdfWriter at 96 dpi, zero margins, point-sized pages), the transform needs to be rebuilt from scratch — don't try to reuse the printer-based one. The helper should accept the transform as a parameter rather than computing it internally, so each caller (GUI print, CLI export) passes its own correct mapping.

3. Execution order

Do the shared helper extraction first as a standalone commit, then wire it into exportPdf() in a second commit. Keeps the diff reviewable and bisectable. The other pending CLI tools can follow after.

4. ProjectPrintWindow refactor scope

I'd rather not touch ProjectPrintWindow's internal structure beyond replacing the inline injection call with the helper call. No class hierarchy changes, no new virtual methods — just a free function in a new translation unit.

Go ahead on that basis.

scorpio810:
Not many users need the CLI mode, so for PDF export via the CLI, we can leave it as it is without the PDF hyperlink.
What do you think?

ispyisail:
> Not many users need the CLI mode, so for PDF export via the CLI, we can leave it as it is without the PDF hyperlink. What do you think?

This is going to change fast, People are going to want to automated access with AI and command line is the gateway

scorpio810:
So, ok.

scorpio810:
Yes I saw new skills for QET:
https://lobehub.com/skills/r2d2helm-claude-code-system-qelectrotech-skill
https://blink.new/p/qelectrosc-app-iq2mqeoz

PR #484, opened 2026-06-11
Update-UI-Chinese-translation

zi-mozhuang:
Completed and optimized all translations, including "<translation type="unfinished">". Please avoid overwriting them during the next export.


scorpio810:
hi @zi-mozhuang please provide for future translation .qm file.

PR #490, opened 2026-06-11
CLI: clickable cross-reference hyperlinks in PDF export

ispyisail:
Brings the GUI's clickable PDF cross-reference / folio navigation (cd76b6a) to the headless `--export-pdf` path, via a shared helper — as discussed on #483.

Two commits, bisectable per your suggested order:

**1. Extract `pdf_links.{cpp,h}` shared helper.** Moves the link-injection logic and `convertUriToGoTo()` post-processor out of `ProjectPrintWindow` into a standalone translation unit. The scene→page mapping is passed in as a `PageGeometry` (transform + `devToPdf` + source-rect lookup) so each caller supplies its own correct geometry — the helper never assumes a `QPrinter`. `ProjectPrintWindow` stays a pure caller: it builds its `PageGeometry` from the printer page layout exactly as before. No behavioural change to GUI print; no class-structure changes.

**2. Wire into CLI `exportPdf()`.** Builds the geometry from the `QPdfWriter` (96 dpi, zero margins, page sized to the diagram → scale ~1, no centering) — rebuilt from scratch, not reused from the printer path — and calls the helper per page, then post-processes after the painter closes.

**Verified** on `examples/industrial.qet` (50 pages): the exported PDF contains 304 `/GoTo` + `/FitR` internal link annotations (0 leftover `/URI`), well-formed. Builds clean on Qt5 + KF5 (Ubuntu).

Independent of the CLI tools in #489.

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -109,6 +109,8 @@ set(QET_RES_FILES
 set(QET_SRC_FILES
   ${QET_DIR}/sources/cli_export.cpp
   ${QET_DIR}/sources/cli_export.h
+  ${QET_DIR}/sources/pdf_links.cpp
+  ${QET_DIR}/sources/pdf_links.h
   ${QET_DIR}/sources/borderproperties.cpp
   ${QET_DIR}/sources/borderproperties.h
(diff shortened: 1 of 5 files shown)


scorpio810:
Very great thanks @ispyisail

Issue #421, opened 2026-02-03
[BUG] Menu and dialog language not changing


gwiethaus:
The menus are not all in the chosen language. I selected the system language or Brazilian Portuguese in the settings, but neither is taking effect. This is also happening in dialog boxes. The language seems to be in French and does not change.

I installed the new version 1.00

Blue marking indicates Brazilian Portuguese, red marking indicates French or another language. Example follows below.

<img width="706" height="603" alt="Image" src="https://github.com/user-attachments/assets/b0ac2e56-debf-403e-aa20-31b0c627fa64" />

scorpio810:
<img width="2956" height="2015" alt="Image" src="https://github.com/user-attachments/assets/f74b854d-a1a6-4f2a-9f10-3c7182c4e84e" />

This isn't translated in Brazilian Portuguese please work on it.

scorpio810:
Install qtcreator and get  repository https://github.com/qelectrotech/qelectrotech-source-mirror/archive/refs/heads/master.zip open https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/lang/qet_pt_BR.ts with qt linguist translate, save and publish, send me .ts and .qm files, thanks

https://www.qt.io/development/offline-installers
https://doc.qt.io/archives/qt-5.15/linguist-translators.html

gwiethaus:
> Install qtcreator and get repository https://github.com/qelectrotech/qelectrotech-source-mirror/archive/refs/heads/master.zip open https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/lang/qet_pt_BR.ts with qt linguist translate, save and publish, send me .ts and .qm files, thanks
> 
> https://www.qt.io/development/offline-installers https://doc.qt.io/archives/qt-5.15/linguist-translators.html

I apologize for taking so long to respond. The notification was hidden among many messages in my email. Yes, exactly. There's still no translation. I closed and reopened it, and it didn't work at that time.

gwiethaus:
> Install qtcreator and get repository https://github.com/qelectrotech/qelectrotech-source-mirror/archive/refs/heads/master.zip open https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/lang/qet_pt_BR.ts with qt linguist translate, save and publish, send me .ts and .qm files, thanks
> 
> https://www.qt.io/development/offline-installers https://doc.qt.io/archives/qt-5.15/linguist-translators.html

[...]

PR #498, opened 2026-06-11
Find translations when lang/ is beside bin/, not inside it (fixes #86)

### LINKED ISSUE #86: Language resets to Spanish after first restart of program and won't change in settings
Language is English after fresh install, but when I close program and restart it, it always changes to Spanish. However, if I managed to save project while in English GUI and I open that file, I get English properties names, but all menus and commands are still in Spanish. When I go to settings to change language, it refuses to do anything.
Tested with:
QElectroTech-0.70_x86_64-win64+git5967-1 and
QElectroTech-0.80-DEV_x86_64-win64+git6869-1
Similar to issue #75

ispyisail:
Fixes #86 (and the related #421).

**Root cause.** `QETApp::languagesPath()` defaults to `applicationDirPath() + "/lang/"`. The Windows installer puts the executable in a `bin/` subfolder while `lang/` sits *next to* `bin/` (i.e. `../lang`), so the default points at a non-existent `bin\lang\`. `qetTranslator.load("qet_en", …)` then fails and `setLanguage()` **silently falls back to French** (the source language) — which is the long-standing "language won't change / resets to French" symptom. It also explains why launching via the bundled `Lancer QET.bat` (which passes `--lang-dir=lang/`) works around it, while a normal Start-Menu/desktop/file-association launch does not.

I traced this on a live Windows install — its own log shows:
```
System language defined in QET configuration: "en"
language Path: "C:/Program Files/QElectroTech/bin/lang/"
```
…but on disk `…\bin\lang\` does not exist and the 320 `.qm` files are in `…\QElectroTech\lang\`. Full diagnosis in #86.

**Fix.** When the folder next to the binary (`bin/lang/`) doesn't exist, fall back to the sibling `../lang/` if it's present. Behaviour is unchanged for builds that already ship `lang/` next to the binary, and for the `QET_LANG_PATH` and `--lang-dir` code paths.

**Verified** on Qt5 (Ubuntu): built clean, and a small harness exercising the exact resolution logic against both layouts:

| layout | resolves to |
|--------|-------------|
| `lang/` beside the binary (normal) | `<bin>/lang/` (unchanged) |
| `bin/` exe with `../lang` (Windows installer) | `<root>/lang/` (the fix) |
| neither present | `<bin>/lang/` (unchanged default) |

This is the code-side fix; alternatively the installer could place the translations in `bin\lang\` — happy to go whichever way you prefer.

### DIFF
--- a/sources/qetapp.cpp
+++ b/sources/qetapp.cpp
@@ -1228,7 +1228,21 @@ QString QETApp::languagesPath()
 	 * en l'absence d'option de compilation, on utilise le dossier lang,
 	 *  situe a cote du binaire executable
 	 */
-	return(QCoreApplication::applicationDirPath() + "/lang/");
+	{
+		const QString bin_dir = QCoreApplication::applicationDirPath();
+		const QString next_to_bin = bin_dir + "/lang/";
+		// Some packagings (notably the Windows installer) put the binary in a
+		// "bin" subfolder while "lang" sits beside it (../lang). Fall back to
+		// that layout when the folder next to the binary is absent, so the
+		// translations are found without a --lang-dir argument. See issue #86.
+		if (!QDir(next_to_bin).exists()) {
+			const QString sibling_of_bin =
+				QDir::cleanPath(bin_dir + "/../lang") + "/";
+			if (QDir(sibling_of_bin).exists())
+				return(sibling_of_bin);
+		}
+		return(next_to_bin);
+	}
 #else
 	#ifndef QET_LANG_PATH_RELATIVE_TO_BINARY_PATH
 		/* the compilation option represents


scorpio810:
Hello @ispyisail, Have you tried the latest Windows packages created using CI/CD? -> https://github.com/qelectrotech/qelectrotech-source-mirror/releases/tag/nightly

scorpio810:
On ready-to-use (Portable version) user's need to launch Lancer QET .bat not the binary program in /bin directory
https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/misc/Lancer%20QET.bat

On MSI packaging I made some changes, and modified . bat for NSIS 3X installer.
https://github.com/qelectrotech/qelectrotech-source-mirror/tree/master/build-aux/windows

scorpio810:
https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/build-aux/windows/QET64.nsi
https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/build-aux/windows/QElectroTech.wxs
https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/build-aux/generate-page.py

https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/.github/workflows/windows-build.yml
https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/.github/workflows/windows-msi.yml

ispyisail:
Thanks @scorpio810 — and you're right about the launchers. I dug into `build-aux/windows/QET64.nsi` to be sure, and here's what it does:

- **Start-Menu + Desktop shortcuts** → `Lancer QET.bat` ✅ (these work — they pass `--lang-dir=lang/`)
- **`.qet` / `.elmt` file association** → `qelectrotech.exe` directly (`...qelectrotech.exe\shell\open\command`) — this **bypasses** the `.bat`
- `lang/` is installed to `$INSTDIR\lang` (beside `bin\`, not inside it)

So the gap is narrow but real: a user who **double-clicks a `.qet` project file** launches the binary directly, which looks in `bin\lang\` (absent) and silently falls back to French — and there's no `.bat` in that path to work around it. That's the only failing launch path; the shortcuts are fine.

I should also test the **nightly** packages you linked before concluding — happy to do that and report whether the file-association case still reproduces there.

This isn't really a "my PR vs. your way" thing — there are a few valid ways to close the gap, and it's your call which fits the project best. Laying them out:

| Option | What changes | Pros | Cons |
|---|---|---|---|
| **A. Code fallback (this PR)** | `languagesPath()` falls back to `../lang/` when `bin/lang/` is absent | Fixes **every** launch path (incl. file association & direct-exe) in one place; no installer change; zero regression for layouts that ship `lang/` beside the binary | Adds a small path-probe in app code; arguably compensating for a packaging layout |
| **B. Installer: put `lang/` in `bin\lang\`** | NSIS installs translations next to the binary | Matches the code's default (`applicationDirPath()+/lang/`); no app-code change; conceptually "correct" layout | Only fixes the **installer** build (portable/zip/other packagers can still hit it); changes install layout; needs every packaging kept in sync |
| **C. Installer: route file association through `Lancer QET.bat`** | `.qet`/`.elmt` `shell\open\command` → the `.bat` | Keeps one launch mechanism (the `.bat`) for everything | `.bat` adds a console flash / indirection on every file open; arg-passing of the file path through the `.bat` must be handled; still installer-only |
| **D. Status quo** | nothing | No work | Double-clicking a `.qet` still opens in French; the #86/#421 reports persist |

My honest read: **A** is the most robust because it fixes the symptom for *all* launch methods and *all* packagings (portable zip included), with no regression — but it does live in app code. **B** is the cleanest "correct layout" fix if you'd rather keep app code untouched and only support the installer. They're not mutually exclusive — A + B together is belt-and-braces.

Whatever you prefer, I'm happy to adjust: refine this PR, switch it to a `.qm`-file existence check instead of a directory check, or close it in favour of a packaging change. Just let me know which direction you want.

ispyisail:
**Correction to my previous comment — I had the file-association target wrong.**

I downloaded the nightly (`...readytouse.zip`, r8883) and checked the actual packaging, and I need to retract part of what I said above. Both launchers route through `Lancer QET.bat` — the file association does **not** bypass it:

- **Installer** (`QET64.nsi`): the association value is `$final_qet_exe`, and that is set to the `.bat`:
  ```
  StrCpy $final_qet_exe "$INSTDIR\Lancer QET.bat"
  ...
  HKEY_CLASSES_ROOT "qet_diagram_file\shell\open\command"  ""  '"$final_qet_exe" "%1"'
  ```
- **Portable** (`register_filetypes.bat`): same — `set expected_qet_exe=%current_dir%Lancer QET.bat`, and the `.qet`/`.elmt`/`.titleblock` `shell\open\command` all point at that `.bat`.

So you were right: with current packaging, **all intended launch paths — Start-Menu/Desktop shortcuts and file associations — go through `Lancer QET.bat`** and pass `--lang-dir=lang/`, so the language loads correctly. My earlier "double-clicking a `.qet` bypasses the `.bat`" was a misread of the `.nsi` on my part. Apologies for the noise.

That changes the standing of this PR. The remaining real issue is narrower:

- The default `languagesPath()` (`applicationDirPath()+"/lang/"`) still doesn't match this layout (`lang/` is beside `bin/`, not in it), so launching `bin\qelectrotech.exe` **directly** — bypassing the `.bat` and the association — silently falls back to French. The live-install log in #86 (`language Path: ".../bin/lang/"`) shows this did happen to at least one user, most likely an older install or a direct-exe launch.

So this PR is now **defense-in-depth** rather than a fix for the normal path: it makes the binary find `../lang/` on its own even when launched directly, so the fallback-to-French can't happen regardless of launcher. That's a "nice robustness" change, not a required one — entirely your call whether it's worth carrying.

If you'd rather not add a path-probe to app code for an unsupported launch method, closing this is completely reasonable and I won't be precious about it. If you do want it, I'm happy to tighten it (e.g. probe for `qet_en.qm` rather than the directory). Thanks for the patience.

scorpio810:
I don't know yet,  read this first https://qelectrotech.org/forum/viewtopic.php?pid=22820#p22820

I really did tear my hair out over that – and I don’t have much left to begin with...

ispyisail:
Thanks for the link — I read the whole thread, and the Windows packaging journey (MXE cross-build, 3-pass `ldd`, NSIS + portable + the WiX MSI, SignPath) is genuinely impressive. The hair loss is understandable.

That context actually pointed me at the missing piece, so I inspected the **MSI** itself (r8900) by reading its tables (`Shortcut`, `Registry`, `Directory`, `File`, `Component`). Here is what it does, verbatim from the package:

**Layout** — `bin` and `lang` are siblings under `INSTALLDIR`:
- exe: `[INSTALLDIR]\bin\qelectrotech.exe`
- translations: `[INSTALLDIR]\lang\` (`qet_en.qm` etc.)

**Shortcuts** (Desktop + Start Menu) — launch the exe directly *and* pass the dir args:
```
[INSTALLDIR]bin\qelectrotech.exe  --common-elements-dir="[INSTALLDIR]elements/" --common-tbt-dir="[INSTALLDIR]titleblocks/" --lang-dir="[INSTALLDIR]lang/" -style windowsvista
```

**`.qet` file association** (Registry table) — launches the exe with **only** the filename:
```
QElectroTech.Document\shell\open\command  =  "[INSTALLDIR]bin\qelectrotech.exe" "%1"
```

The association passes **no `--lang-dir`**. So that launch falls back to the default `languagesPath()` = `applicationDirPath()+"/lang/"` = `[INSTALLDIR]\bin\lang\`, which doesn't exist (translations are in `[INSTALLDIR]\lang\`), and `setLanguage()` silently falls back to French.

(Corroborating detail: the shortcuts bother to pass `--lang-dir` explicitly — which would be redundant if the exe could find `lang/` on its own from this layout. That the shortcuts need it is itself evidence the default resolution fails here, which is exactly the `bin/lang/` path the #86 log showed.)

**Logical conclusions:**

1. NSIS installer and the portable zip route file associations through `Lancer QET.bat` (which passes `--lang-dir`), so they're fine — my earlier "association bypasses the bat" was wrong *for those two*, as I corrected.
2. The **MSI is different**: it has no `.bat`, its shortcuts carry the args directly (good), but its `.qet` association does **not** — so **double-clicking a `.qet` file from an MSI install opens in French.** That's a normal user action on the newest/signed package.
3. The root cause is the same one #86 reported: the exe's default lang path assumes `lang/` sits *next to the binary*, but every current packaging puts it *beside `bin/`* (`../lang`).

**Two ways to fix the MSI case:**

- **Packaging-only:** add `--lang-dir="[INSTALLDIR]lang/"` to the `.qet`/`.elmt` association command in `QElectroTech.wxs` (mirroring the shortcuts). Minimal, MSI-local, but it patches one launch path and has to be kept in sync per packaging.
- **This PR (#498):** the exe finds `../lang` itself when `bin/lang` is absent — which fixes the MSI association **and** any direct-exe launch **and** any future launcher, in one place, with no regression for layouts that ship `lang/` beside the binary.

I'd lean to the code fix as the durable option, but they're not exclusive and it's entirely your call — happy to do either (or tighten this PR to probe for `qet_en.qm` instead of the directory). If useful, I can also install the MSI and screenshot the French-on-double-click to confirm end-to-end.

scorpio810:
Please try new Windows builds: https://qelectrotech.github.io/qelectrotech-source-mirror/

Issue #158, opened 2021-03-14
Position of edited Text in Elementeditor is wrong


Murmele:
1) edit an existing Element.
2) Add a new PartText (write into the text you would like to see)
3) Change size of the text
4) Save edited element
5) Close Element editor
6) Open Element editor for the edited element again
7) Check Position of the Text you set before. It is moved a little bit (I think it depends on the new size)

I found this:
QRectF PartText::boundingRect() const{	
    QRectF r = QGraphicsTextItem::boundingRect();	
    r.adjust(0.0, -1.1, 0.0, 0.0);	
    return(r);
}

why do I have to adjust the bounding Rect by a fixed factor?

moritz-fivi:
The bug seems to still be present in QElectroTech V 0.100.1-dev.

moritz-fivi:
I got the bug reproduced:
I first rotatet and scaled the Fuse1 text on the left side down and than copyed the hole block. On the copy on the ride the Fuse1 text seems to be rendered at the wrong position:

<img width="363" height="441" alt="Image" src="https://github.com/user-attachments/assets/48cda4d3-6cb1-484d-89cf-b0265f40add5" />

But in reality the left Fuse1 text was rendered wrong in the editor:

<img width="522" height="426" alt="Image" src="https://github.com/user-attachments/assets/5b5c00ce-457f-4df7-9adb-fb17a03e4f9a" />

The X position of all copyed elements are exact 20 px to the right 

<img width="1719" height="966" alt="Image" src="https://github.com/user-attachments/assets/92d907d4-8100-40a1-98d5-622b42ab4746" />

<img width="1766" height="936" alt="Image" src="https://github.com/user-attachments/assets/ffedda09-cfee-4c6c-94e5-b47358723812" />

ispyisail:
**Could not reproduce the position drift on Qt 5.15 — the save/reload round-trip is exact**

I instrumented `PartText` (logging pos / font / `margin()` / ascent at save and reload) and ran the repro — add text, change size, save, reopen — across font sizes 9, 12 and 48 pt. The position round-trips **bit-exact** every time:

```
SAVE x=5   y=16   -> reload -> pos=(5,16)
SAVE x=-30 y=-130 -> reload -> pos=(-30,-130)
SAVE x=140 y=20   -> reload -> pos=(140,20)
```

[...]

PR #501, opened 2026-06-12
PartText: keep text position stable across save/reopen on font-size change (#158)

ispyisail:
Fixes the static-text position drift reported in #158.

## Problem
A static text in the element editor shifts position after save/reopen when its font size was changed. The saved `(x,y)` is unchanged — only the *rendered* position differs.

## Cause
`PartText` is anchored by a transform of `-margin()`, where `margin().y = documentMargin + fontMetrics.ascent()` — so the rendered position depends on the font's **ascent**. `adjustItemPosition()` (which sets that transform) is only re-run on `blockCountChanged` / `contentsChanged`, **not** on a font/size change. After scaling, the transform keeps the *previous* font's ascent. On reopen, `setPlainText` re-triggers `adjustItemPosition()` with the saved font, recomputing the correct offset — hence the drift.

Verified with a probe in `toXml`: at size 4 the applied transform was `dy = -13` (from the original ~20 pt font) while the correct value is `-6` — a 7 px stale offset that vanishes on reload. This is independent of `real_font_size_`, which stays desynced but is not in the position path.

## Fix
Re-run `adjustItemPosition()` in `setFont()` so the anchor always matches the current font. After the fix the applied transform is `-6` at save (matching reload), and the drift is gone.

### DIFF
--- a/sources/editor/graphicspart/parttext.cpp
+++ b/sources/editor/graphicspart/parttext.cpp
@@ -313,6 +313,12 @@ void PartText::setPlainText(const QString &text) {
 void PartText::setFont(const QFont &font) {
 	if (font != this -> font()) {
 		QGraphicsTextItem::setFont(font);
+		// Re-anchor: the item's position transform is -margin(), and margin()
+		// depends on the font ascent. Without re-running this on a font change,
+		// the transform keeps the previous font's ascent — so the text renders
+		// at a different spot after save/reopen (the position recomputes from
+		// the saved font on load). See #158.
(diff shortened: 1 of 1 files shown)


scorpio810:
https://qelectrotech.org/forum/viewtopic.php?pid=22895#p22895
https://qelectrotech.org/forum/viewtopic.php?pid=22915#p22915
https://qelectrotech.org/forum/viewtopic.php?pid=22920#p22920

PR #504, opened 2026-06-15
[Draft / PoC] Database-driven from-to wiring list + "Add a wiring list" (RFC for #503)

ispyisail:
> ## ⚠️ DRAFT — PROOF OF CONCEPT — NOT FOR PRODUCTION USE
> This is a **demonstration / RFC** to make the idea in discussion #503 concrete
> and testable. It has **not** been through full review or QA. The schema, the
> `wire_list_view` SQL, the table format and the GUI wording are all **subject to
> change**. Please don't merge as‑is — it's here to gather feedback on direction.

Follows up on discussion **#503** (database‑driven from‑to wiring list). It adds, **additively** to `projectDataBase`:

### 1. A from‑to wiring list view
Two small tables (`terminal`, `conductor`) and a `wire_list_view`, rebuilt from the scene in `updateDB()` exactly like the existing tables (no stored state, existing tables/views untouched). Each conductor resolves to its two `element : terminal` endpoints + wire number + folio number, including the **cross‑folio** case:

| wire | from_point | from_folio | to_point | to_folio |
|------|------------|------------|----------|----------|
| 1001 | X2:4 | 10 | 10QF2 | 10 |
| 1011 | XV2:1 | 8 | 8F2 | 8 |

Terminals are keyed by element‑uuid + a per‑element index (a terminal's own uuid can be null/repeated), with a `Terminal*→id` map so conductor endpoints join to the exact row.

### 2. A "Add a wiring list" menu action
`Project → Add a wiring list` renders the view as a table on the current folio and spills onto new folios — mirroring the existing *Add a nomenclature* / *Add a summary*. It reuses QET's generic `ProjectDBModel`, so there's **no new rendering code**.

### Try it
1. Build this branch, open e.g. `examples/industrial.qet`.
2. `Project → Add a wiring list`.
3. Or `python3 docs/wiring-list-demo/make_demo.py` for a pre‑built demo project.

See `docs/wiring-list-demo/README.md` for details.

### Known limitations (it's a PoC)
- Coverage uses QET's existing element‑table filter (`Simple | Terminal | Master | Thumbnail`), so conductors ending on **folio‑report arrows** or **slave** contacts aren't resolved yet; element‑to‑element wires are complete.
- The wiring‑list query is fixed (no column‑picker dialog).
- Not performance‑reviewed; view schema is provisional.

### Commits
- `db:` terminal + conductor tables + `wire_list_view`
- `gui:` the "Add a wiring list" action
- `demo:` the support files in `docs/wiring-list-demo/`

Feedback on schema, naming and scope very welcome.

### DIFF
--- a/docs/wiring-list-demo/README.md
+++ b/docs/wiring-list-demo/README.md
@@ -1,0 +1,73 @@
+# Database‑driven from‑to wiring list — Proof of Concept
+
+> ## ⚠️ NOT FOR PRODUCTION USE
+> This branch is a **proof of concept / demonstration** to support discussion
+> [#503](https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/503).
+> It is **not production‑ready**, has **not** been through full review or QA, and
+> the schema, SQL view, table format and GUI wording are all **subject to change**.
+> Do not ship it, and do not rely on the `wire_list_view` schema staying stable.
+
+## What this demonstrates
+
+A **from‑to wiring list** built additively on QElectroTech's existing
+`projectDataBase`, plus a GUI action to drop it onto a folio:
+
+1. **`wire_list_view`** — for every conductor, resolves both ends to
+   `element : terminal` on a folio number, with the wire number:
+
+   | wire | from_point | from_folio | to_point | to_folio |
+   |------|------------|------------|----------|----------|
+   | 1001 | X2:4       | 10         | 10QF2    | 10       |
+   | 1011 | XV2:1      | 8          | 8F2      | 8        |
+
+   It is rebuilt from the scene in `updateDB()` like the existing tables
+   (no stored state), and **handles the cross‑folio case** (each end can be
+   on a different folio).
+
+2. **Project → "Add a wiring list"** — a menu action that renders the view as
+   a table on the current folio and spills onto new folios, exactly like the
+   existing *Add a nomenclature* / *Add a summary* actions. No new rendering
+   code: it reuses QET's generic `ProjectDBModel`.
+
+## What's in the branch
+
+| Commit | Contents |
+|--------|----------|
+| DB slice | `terminal` + `conductor` tables + `wire_list_view` in `projectdatabase.cpp/.h` (additive, existing tables untouched) |
+| GUI action | `m_add_wiring_list` action + `QetGraphicsTableFactory::createAndAddWiringList` |
+| Demo (this) | these support files |
+
+## How to try it
+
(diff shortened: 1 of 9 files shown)


Kellermorph:
Maybe I'm misunderstanding something, but there's already an export option for the wiring diagram. This is saved as a CSV file. You could simply add the wire numbers there.
Then you could insert this as a separate list directly into the schematic. But all the logic, including the merging of cross-references, is already included there.

ispyisail:
Thanks @Kellermorph — fair point, and you're right: `WiringListExport` already builds a wiring list, merges the cross-references, and can be inserted into the schematic. I should have referenced it.

The "Add a wiring list" action makes this look like a standalone wiring-list feature, which undersells it — that part is really just a **visible demo**. The actual proposal (discussion #503, which had a positive steer from the maintainers) is to grow `projectDataBase` into a **normalized, queryable connectivity layer** — `terminal` / `conductor` / `net` tables the scene populates, exposed as SQL views.

The reason for a DB layer rather than another exporter is that **one normalized model** could then drive a whole family of features from the same source: from-to lists, terminal plans, cable lists, BOM, named potentials, device cross-references — the database-backed tooling promis·e/EPLAN users expect. The wiring-list view here is just the **first, smallest slice** to make that concrete and testable — not the end goal.

Honest about where it stands today, though: on cross-references `WiringListExport` is the **more complete** of the two — this DB view is element-to-element only and doesn't yet resolve folio reports / slave contacts (≈ half the connections on a multi-folio project). That's scoped (it can reuse QET's own potential traversal) but deliberately not in this draft.

So the steer I'm really after, since this is an RFC, isn't "wiring list yes/no" — it's whether a **`projectDataBase` connectivity layer** is a foundation you'd want to invest in, with `WiringListExport` and the existing nomenclature/summary potentially becoming *consumers* of it (one model rather than several parsers). If yes, I'll build it out properly; if you'd rather keep things lean and just add wire numbers to `WiringListExport`, that's a perfectly good answer too.

ispyisail:
Closing this — it did its job.

This was always explicitly a draft/RFC ("please don't merge as-is"), written to make the idea in #503 concrete enough to argue about. That worked: the direction held up, and it has since been rebuilt properly as five reviewable slices, each standalone and independently revertable:

- #625 — give `Conductor` its own persisted uuid
- #628 — terminal and conductor tables in `projectDataBase`
- #629 — `wiring_list_view` over those tables
- #630 — wiring list dialog + excluded-conductor count
- #631 — BOM wire count per element

Keeping this open alongside them just adds noise to the queue and risks someone merging the prototype and the real thing. The discussion in #503 stays the reference for the design.

PR #509, opened 2026-06-16
Add Custom Guides

Kellermorph:
This PR implements a comprehensive system for **Custom Guides** within diagrams to improve element alignment and precision handling. Alongside the core drawing and management logic for these guides, this PR introduces new user configuration options to control the startup behavior of both the background grid and the custom guides.

Two new checkboxes have been added to the General Configuration page under a "behavior on startup" context. This allows users to independently define whether the grid and/or the custom guides should be visible by default when a new diagram or project is loaded (`grid_display_startup` and `guides_display_startup`). As stated in the UI, these startup preferences are applied upon the next application launch.


ispyisail:
Hi Kellermorph, I tried to send you a PM but i could find a link. 

I ran your code throught Claude 4.8 and it came up with some interesting results, I didn't want to post in public.

Kellermorph:
Hi ispyisail,
I'm not sure how you're supposed to send a private message on GitHub. Or what do you mean by a link? I can't find anything on your profile either.

But that doesn't sound quite right, does it?
The code worked for me. I did create it with Gemini, though.

ispyisail:
Hi @Kellermorph — I built the updated branch (Qt5, Linux) and went through it properly. It **compiles cleanly** and the design is in good shape; thanks for the work on this. Below is a full review: what I verified, then some points to discuss, roughly in priority order. Most are design/scope/style rather than bugs.

## What I checked and looks good
- **Builds clean** with no warnings introduced in the new code.
- **The `Diagram::undoStack()` refactor is correct.** The inline definition was removed from `diagram.h` and moved to `diagram.cpp`, so it still links. This is what makes the ~20 files need an explicit `#include "../qetproject.h"` (they previously got it transitively through `diagram.h`) — invasive, but legitimate.
- **XML serialization is backward-compatible and locale-safe.** Projects saved before this feature load fine (`m_default_guides` just stays empty when `<guides>` is absent, `qetproject.cpp:1693`), and orientation/position/color are written and read as plain int/double/name strings.
- **Startup behaviour is consistent.** Both the `Diagram` members (`draw_grid_`/`draw_guides_`, `diagram.cpp:77-78`) and the toolbar actions read the same `QSettings` keys, so the toolbar checkmarks match what's actually drawn on launch.
- **`setDefaultGuides()` is change-guarded, marks the project modified, and propagates to all open diagrams.**

## Points to discuss

**1. Are the guides meant to be interactive / snap-to? (main question)**
Right now they're visual dashed reference lines: elements snap to the grid but not to guides, and guides can only be created/edited numerically in *Preferences → Guides* — there's no drag-from-ruler or on-canvas handle. That's a perfectly reasonable first version, but the description frames them around "element alignment and precision," which usually implies snapping. Could you clarify the intended scope? If snapping / interactive placement is planned, it might be worth noting as a follow-up so reviewers know what to expect here.

**2. Indentation style.** `guidespropertieswidget.cpp/.h` use 4-space indentation, while the rest of QElectroTech uses tabs. Could you convert these to tabs for consistency?

**3. Integer vs. real positions.** `Guide::position` is a `qreal`, but the editor uses a `QSpinBox` and `setGuides()` does `qRound(g.position)`, so positions are truncated to integers. If sub-pixel precision is intended, a `QDoubleSpinBox` would preserve it.

**4. PR scope / diff size.** Two things make the diff larger than the feature itself:
- the `diagram.h` → `qetproject.h` decoupling that touches ~20 unrelated files, and
- a full `qet_de.ts` regeneration (+5090/−415).

Both are defensible, but splitting the header refactor into its own commit (and limiting the translation change to the new strings) would make this much easier to review and less likely to conflict with other open PRs.

**5. Smaller notes (non-blocking):**
- Guides are painted in `drawBackground()`, so they sit *behind* elements; guides in most tools float on top — minor, just flagging the choice.
- `m_draw_guides` uses `QIcon(":/ico/22x22/guides.png")` directly, whereas neighbouring actions use the `QET::Icons::` helpers.
- Guide colours are stored on the buttons via `property("color")` + a stylesheet; it works, just a bit indirect.
- Guide edits aren't on the undo stack (everything else in QET is undoable) — probably fine for a preference, mentioning for completeness.

Overall this is close. #2 and #3 are quick fixes; #1 is the main thing that decides whether this lands as "visual guides" (great as-is, just worth describing that way) or grows a snapping follow-up. Happy to re-test once you've had a look.

Kellermorph:
Hi ispyisail,

Thank you so much for taking the time to compile the code and for the extremely detailed and constructive review! This is highly appreciated.

To answer your points:

1. Interactive / Snap-to:
No, the guides are intentionally designed to be purely visual reference lines (similar to implementations in e.g. WSCAD). We specifically avoided snapping to prevent the canvas from becoming chaotic, where the grid-snap and guide-snap might conflict and cause elements to jump unpredictably. They serve to quickly verify consistent alignment and spacing across the diagram visually.

2. & 3. Indentation & QDoubleSpinBox:
You are absolutely right. I will convert the spaces to tabs in the widget files to match the coding guidelines. I will also change the UI to QDoubleSpinBox and remove qRound() to allow sub-pixel precision.

4. PR Scope (Diff size):
You correctly identified the diagram.h decoupling. We had to break a circular dependency regarding the undoStack() and QETProject, which resulted in the transitive include failing for those ~20 files. Regarding the qet_de.ts: The massive diff was automatically caused by lrelease / Qt Linguist re-sorting the XML nodes upon saving. I will see if I can revert the unnecessary sorting to keep the diff clean.

5. Minor notes:
Drawing them in drawBackground() was an intentional choice so they don't visually interfere with or obscure the actual components and wiring.

Critical finding on my end (Export Bug):
I just realized that the guides are currently being rendered during PDF/Image exports. I definitely need to fix this so they are suppressed during printing/exporting, just like the grid.

I will prepare a commit to fix the indentation, the DoubleSpinBox, and the export bug, and will push it to this branch shortly. Thanks again for the great feedback!

Kellermorph:
done

ispyisail:
Re-reviewed and rebuilt the updated branch (`0555cd9`, Qt5/Linux) — **compiles clean, 434/434**. The fixes all look good:

- **Spinboxes → `QDoubleSpinBox`** with `setDecimals(2)` and a `-10000…10000` range: sub-pixel precision preserved and negative positions now work. 👍
- **Indentation** converted to tabs in the guide widget. 👍
- **Export bug fix is nicely done** — the new `ExportProperties::draw_guides` (default false, persisted) is wired symmetrically with `draw_grid` across CLI export (toggle-and-restore in `renderDiagram`), printing (`projectprintwindow` forces false), and the GUI export path. Guides are now correctly suppressed in PDF/image/print output.

One leftover from point #4: `lang/qet_de.ts` is still a +5090/−415 diff from the Linguist re-sort. Could you reset it to master and re-add just the new guide strings (`git checkout origin/master -- lang/qet_de.ts lang/qet_de.qm`, then re-run lupdate for the new strings)? That'd shrink the diff a lot and avoid conflicts with other translation PRs. Not a blocker, just the last cleanup.

Minor (optional): the GUI Export dialog has a grid checkbox but none for guides — fine to leave, since guides arguably shouldn't print anyway.

Thanks for the quick turnaround on these!

Kellermorph:
it wasnt working.
I will do the translation with an other PR

ispyisail:
Thanks — the translation cleanup is exactly right. The `lang/qet_de` files now match master (handled separately), which drops the PR from ~5,600 changed lines to **~500 across 41 files** — much easier to review and far less likely to conflict with other PRs. 👍

Since these last two commits only touch the translation files, the source is identical to the version I built and tested earlier (compiles clean, 434/434, Qt5/Linux), so no re-verification needed. From my side this now covers everything: double spinboxes, tabs, the export-suppression fix, and the diff cleanup. Looks good to me.

Kellermorph:
Thank you so much for the thorough review, the testing, and your patience with the translation file quirks! I'm glad dropping them for now made the diff clean and manageable again. I'll take care of the German translations in a separate, dedicated PR later on. 

Thanks again for your guidance! 🚀

ispyisail:
Thank Claude, not me :) 

99% of the time your acutally talking to Claude im just siiting in the drivers seat pushing auto pilot

ispyisail:
I wonder how much AI is on your end we could have two AI's haveing a conversation? :)

Kellermorph:
Yeah, I'm doing this work with AI too :D Maybe the two of them talked a lot :D

scorpio810:
It's ready to push?

Kellermorph:
yes

scorpio810:
Features is disabled on export, print, or make PDF, actions?

scorpio810:
I’ll be heading back to the operating theatre soon for another health issue; I don’t have time to test everything....

Kellermorph:
Yes, that's turned off.
Get well soon

PR #396, opened 2025-08-05
add Flatpak build ci

Int-Circuit:
This MR add Continuous Integration (CI) for the flatpak package.
- This makes Github actions build an amd64 and arm64 package for each change.
- KDE Sdk version has been transitioned from  5.15-23.08 to 5.15-24.08, no bugs found during build and execution

Issue #299, opened 2024-07-14
options moving when opening "file", "edition" menus


Int-Circuit:
When using latest V0.100 Qelectrotech version, I get this visual bug
![image](https://github.com/user-attachments/assets/1c248689-cf28-495d-8af4-474131bccbe6)
menus are not glued to top of window

![image](https://github.com/user-attachments/assets/fc826f73-fc55-4cc2-bbd5-cca7a0e34653)

PR #510, opened 2026-06-18
ci: add Flatpak build workflow (x86_64 + aarch64) — rebase of #396

ispyisail:
Rebase of #396 onto current `master`, since that branch had drifted (its manifest still pinned `5.15-23.08`/`24.08`, while `master` is now `5.15-25.08`). Original work by @Int-Circuit.

### What this does
Adds a single workflow, `.github/workflows/build-flatpak.yml`, that builds the Flatpak package for **x86_64** and **aarch64** using the flathub-infra container and `flatpak/flatpak-github-actions/flatpak-builder@v6`. arm64 runs on a native `ubuntu-24.04-arm` runner (no qemu). The built `.flatpak` bundles are uploaded as artifacts.

It targets the current `5.15-25.08` runtime/SDK via the container tag, so **no manifest change is needed**.

### Scope kept deliberately minimal
- Only adds the workflow file — no manifest, `.gitignore`, or `packaging_script_Flatpak.sh` changes.
- **No publish/upload or GPG signing**, per @scorpio810's point about not giving GitHub Actions access to the QET server. This is build/CI only.

### Changes vs. the original #396 workflow
- Container tag `kde-5.15-24.08` → `kde-5.15-25.08` to match master's manifest.
- Per-arch `cache-key` (`…-${{ matrix.variant.arch }}-…`) so the two matrix legs don't share/overwrite one cache.
- `fail-fast: false` so one arch failing doesn't cancel the other.
- Added `pull_request` (path-filtered to flatpak files) and `workflow_dispatch` triggers, alongside the original `push: master`, so flatpak-affecting PRs get validated without burdening unrelated PRs.
- Whitespace/EOF cleanup.

### Tested
I ran this exact workflow on current master — both arches built green:

| Job | Result | Time | Bundle |
|-----|--------|------|--------|
| x86_64 (`ubuntu-24.04`) | ✅ | 15m 45s | `qelectrotech-x86_64.flatpak` (~16.8 MB) |
| aarch64 (`ubuntu-24.04-arm`) | ✅ | 11m 35s | `qelectrotech-aarch64.flatpak` (~16.5 MB) |

Run: https://github.com/ispyisail/qelectrotech-source-mirror/actions/runs/27746156993

Note: this verifies the **build** only (CI is headless), so it doesn't cover the Flatpak-only display issue #299.

### DIFF
--- a/.github/workflows/build-flatpak.yml
+++ b/.github/workflows/build-flatpak.yml
@@ -1,13 +1,10 @@
 name: Build Flatpak package
 
+# On-demand only: run manually from the Actions tab. This avoids spending
+# Actions minutes (arm64 is billed at a higher rate) on every push to master
+# for a build that publishes nothing. A release-tag publish step can be added
+# later once a distribution target is agreed.
 on:
-  push:
-    branches:
-      - master
-  pull_request:
-    paths:
-      - 'build-aux/flatpak/**'
-      - '.github/workflows/build-flatpak.yml'
   workflow_dispatch:
 
 jobs:
@@ -28,6 +25,9 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: 'recursive'
+          # The Flatpak build needs sources only, not the LFS-stored .qch docs.
+          # Keep this explicit so the workflow never draws on LFS bandwidth.
+          lfs: false
       - uses: flatpak/flatpak-github-actions/flatpak-builder@v6
         with:
           bundle: qelectrotech.flatpak
--- a/.github/workflows/build-flatpak.yml
+++ b/.github/workflows/build-flatpak.yml
@@ -1,0 +1,37 @@
+name: Build Flatpak package
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    paths:
+      - 'build-aux/flatpak/**'
+      - '.github/workflows/build-flatpak.yml'
+  workflow_dispatch:
+
+jobs:
+  flatpak:
+    container:
+      image: ghcr.io/flathub-infra/flatpak-github-actions:kde-5.15-25.08
+      options: --privileged
+    strategy:
+      fail-fast: false
+      matrix:
+        variant:
+          - arch: x86_64
+            runner: ubuntu-24.04
+          - arch: aarch64
+            runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.variant.runner }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: 'recursive'
+      - uses: flatpak/flatpak-github-actions/flatpak-builder@v6
+        with:
+          bundle: qelectrotech.flatpak
+          manifest-path: build-aux/flatpak/org.qelectrotech.QElectroTech.json
+          cache-key: flatpak-builder-${{ matrix.variant.arch }}-${{ github.sha }}
+          arch: ${{ matrix.variant.arch }}
+          verbose: true


scorpio810:
Hi @ispyisail, thanks for the clean rebase and the thorough write-up — the dual-arch native build (no QEMU for aarch64), the per-arch cache key fix, and the `fail-fast: false` are all exactly the right calls. The CI run is solid.

That said, I can't merge this in its current form. The workflow only validates that the build compiles and uploads temporary artifacts to GitHub — which burns roughly 15 minutes of Actions minutes per arch on every push to master, with nothing published at the end. Given that arm64 runners are billed at a higher rate, this would generate non-trivial costs for the project without delivering a distributable result for our users.

What I'd suggest:
- In the short term, switch the trigger to `workflow_dispatch` only, so the build can be run on demand without firing on every commit to master.
- In the longer term, if we can agree on a publish target (Flathub submission, a GitHub Release, or even the nightly page), this workflow becomes genuinely valuable and I'd be happy to merge it.

The foundation is good — let's get it to the finish line. Happy to discuss the publish step if you want to take it further!

scorpio810:
03/03/2026 10:18
You have used 100% of the Git LFS bandwidth included in your qelectrotech account for the qelectrotech account.

Your usage will reset in 29 days on April 01, 2026.

01/04/2026 14:38
Your plan includes 10.0 GB of Git LFS bandwidth per month at no extra cost. You have used 90% so far this billing cycle. for the qelectrotech account.

Your usage will reset in 30 days on May 01, 2026.

..............................

.............................



02/05/2026 09:40
Your plan includes 10.0 GB of Git LFS bandwidth per month at no extra cost. You have used 90% so far this billing cycle. for the qelectrotech account.

Your usage will reset in 30 days on June 01, 2026.


01/06/2026 22:21
Your plan includes 10.0 GB of Git LFS bandwidth per month at no extra cost. You have used 90% so far this billing cycle. for the qelectrotech account.

Your usage will reset in 30 days on July 01, 2026

02/06/2026 10:51
Your plan includes 10.0 GB of Git LFS bandwidth per month at no extra cost. You have used 100% so far this billing cycle. for the qelectrotech account.

Your usage will reset in 29 days on July 01, 2026.

etc etc... 


Unfortunately, I therefore had to pause the GitHub ‘Auto-build Doxygen docs’ actions and, consequently, the work done by @Int-Circuit, as this bandwidth issue was also holding up our work on https://github.com/flathub/org.qelectrotech.QElectroTech

I need to rework this file and delete the .qch files (990 MB) containing Qt help files for QET code, to reduce bandwidth usage whilst compiling LFS.

ispyisail:
Thanks Laurent — that all makes sense, and you're right that build-on-every-push isn't worth the minutes.

I've switched the trigger to **`workflow_dispatch` only** (just pushed), so it no longer fires on pushes to master — it only runs on demand from the Actions tab. I also set `lfs: false` explicitly on the checkout: this workflow builds from sources and doesn't pull the `.qch`/LFS objects, so it isn't adding to the LFS-bandwidth problem you're hitting (and good luck with the 990 MB `.qch` cleanup — that's clearly overdue).

For the publish step, a concrete proposal: trigger **on release tags** (in addition to `workflow_dispatch`) and attach the built x86_64 + aarch64 `.flatpak` bundles to the **GitHub Release**. That way it only runs on releases (cost stays negligible), produces a real downloadable artifact for users, and needs **no access to the QET server or GPG keys**. If you're happy with that target I'll add it as a follow-up commit; otherwise we can aim it at the Flathub submission (building on flathub#11) instead. Your call on the channel.

scorpio810:
Thanks, Shane. I’ve been updating my Flatpak QEmu/KVM VM for a few months now and have since published the new nightly builds to the QET Flathub repository: https://qelectrotech.org/forum/viewtopic.php?pid=11227#p11227
But I only work with x86_64, not aarch64, so it might be worth building a few packages from time to time using GitHub Actions; before that, I need to give some thought to updating the GitHub Pages.

ispyisail:
Nice — glad the VM's back and the x86_64 nightlies are flowing again.

That actually lines up well with where this PR landed: it's now `workflow_dispatch`-only, so it's exactly an "on-demand, build a few packages from time to time" tool — and since you're covered on x86_64 locally, the bit it uniquely adds is the native **aarch64** build you can't do in the VM.

Two thoughts:
- I can **drop x86_64 from the matrix** so it builds aarch64-only — that halves the run time and matches what you actually need from it. Or keep both for cross-checking; your call.
- On publishing: no rush — let's leave the workflow producing downloadable artifacts for now, and once you've sorted the GitHub Pages side I'm happy to add a step that drops the `.flatpak` onto a Release / the nightly page (no server or GPG access needed). Whenever you're ready.

So I think this is good to merge as the on-demand arm64 builder, and we wire up publishing as a follow-up once Pages is sorted.

scorpio810:
I prefer this suggestion: keep both so that I can compare them.

ispyisail:
Perfect, keeping both then — no change needed, the matrix already builds x86_64 + aarch64 side by side. So this is ready to merge as the on-demand builder whenever you like. Ping me once you've sorted the GitHub Pages side and I'll add the publish step (Release/nightly-page artifact, no server or GPG access). 👍

arummler:
I cherry-picked this into #722 which is a comprehensive CI/CD chain.

ispyisail:

PR #511, opened 2026-06-18
Fix #159: reset unused-element highlight when elements become used again

### LINKED ISSUE #159: Highlight unused will not be reset
1) Create new Project and add two different elements
2) Save the project and close it
3) Reopen the project
4) Remove on element and save
--> in Collections the element should get a red background, because it is unused
5) Add the element with the red background and save
--> Collections should update and the element should get a white background again.

I think for that in 
void ElementsCollectionModel::highlightUnusedElement()
the elements which are not unused should be reset to a white background?

ispyisail:
Fixes #159.

### Problem
`ElementsCollectionModel::highlightUnusedElement()` is add-only: it paints the currently-unused elements with a red `Dense4Pattern` background but never clears the background of items that are no longer unused. So in the reported flow — remove an element + save (it goes red), then re-add it + save — the function is called again on the existing model and re-reds the still-unused set, but the now-used element keeps its stale red highlight. It only clears when the whole model is rebuilt from scratch.

### Fix
Reset every item's background before re-applying the highlight to the current unused set:

```cpp
for (ElementCollectionItem *eci : items())
    eci->setBackground(QBrush());
```

`items()` already returns every `ElementCollectionItem` in the model, and resetting via `setBackground(QBrush())` is the same idiom already used in `elementscollectionwidget.cpp`. The reset loop is the same order of magnitude as the existing refresh, so no meaningful cost.

5-line change, no API/behaviour change beyond clearing stale highlights. Original report and root-cause analysis by @Murmele.

### DIFF
--- a/sources/ElementsCollection/elementscollectionmodel.cpp
+++ b/sources/ElementsCollection/elementscollectionmodel.cpp
@@ -538,10 +538,15 @@ QList<QETProject *> ElementsCollectionModel::project() const
 */
 void ElementsCollectionModel::highlightUnusedElement()
 {
-		//Reset the background of every item first, so elements that are no
-		//longer unused lose their previous red highlight (issue #159).
+		//Reset only the items currently highlighted in red, so elements that
+		//are no longer unused lose the highlight. Scoping to the red
+		//Dense4Pattern avoids touching other backgrounds (e.g. the amber
+		//"show this dir" highlight) and avoids needless updates on big
+		//collections (issue #159).
 	for (ElementCollectionItem *eci : items())
-		eci->setBackground(QBrush());
+		if (eci->background().style() == Qt::Dense4Pattern &&
+		    eci->background().color() == Qt::red)
+			eci->setBackground(QBrush());
 
 	QList <ElementsLocation> unused;
 
--- a/sources/ElementsCollection/elementscollectionmodel.cpp
+++ b/sources/ElementsCollection/elementscollectionmodel.cpp
@@ -538,6 +538,11 @@ QList<QETProject *> ElementsCollectionModel::project() const
 */
 void ElementsCollectionModel::highlightUnusedElement()
 {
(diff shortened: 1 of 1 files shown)


ispyisail:
Built locally to confirm it compiles: full clean build on Ubuntu/WSL (CMake + Ninja, Release, system Qt5) of this branch completed green — **433/433 targets, exit 0**, including the unit-test executables. The changed `sources/ElementsCollection/elementscollectionmodel.cpp` compiled with no errors or warnings and the app linked successfully.

Note this verifies the build only; the visual red→white reset behaviour from the report still wants a quick manual click-through on a desktop build.

plc-user:
In case of bigger projects: 
Would it perhaps be less CPU-intensive if we only changed the elements highlighted in red?
In that case, we could add a condition:
`	for (ElementCollectionItem *eci : items())`
` 		if ((eci->background().style() == Qt::Dense4Pattern) &&`
`		    (eci->background().color() == Qt::red)) {`
`		    eci->setBackground(QBrush());`
`		}`

ispyisail:
Good catch — adopted, thanks. I scoped the reset to only the items currently painted with the red `Dense4Pattern`:

```cpp
for (ElementCollectionItem *eci : items())
    if (eci->background().style() == Qt::Dense4Pattern &&
        eci->background().color() == Qt::red)
        eci->setBackground(QBrush());
```

Worth noting it's not only a CPU win: the unconditional reset also clears **other** backgrounds, in particular the amber "show this dir" highlight set in `elementscollectionwidget.cpp` (`QColor(255, 204, 0)`). Since `highlightUnusedElement()` runs on every save, the blanket version would wipe that highlight out. Restricting it to the red brush leaves it intact — so this is a correctness fix as well as avoiding the redundant updates on big collections.

Rebuilt locally to confirm it still compiles green (Ubuntu/WSL, Qt5).

PR #513, opened 2026-06-19
Import EPLAN Data Portal parts (.edz) into element collections

ispyisail:
# Import EPLAN Data Portal parts (`.edz`) into element collections

## What this adds

A new **Import an EPLAN part (.edz)…** action on writable element-collection
folders. The EPLAN Data Portal is the de-facto manufacturer-parts catalogue for
electrical CAD — ifm, Phoenix Contact, Siemens, Rittal, Weidmüller, Schneider and
others distribute their components as `.edz` — and QET previously had no way to
use them. This lets a user grab a part off the portal and drop it straight into a
project, with the manufacturer metadata wired into the BOM/nomenclature.

Right-click a writable collection folder → **Import an EPLAN part (.edz)…** →
choose the file. The generated `.elmt` is written into that folder and the panel
reloads.

## What is imported

A `.edz` is a 7-Zip archive containing a part definition (`*.part.xml`), an EPLAN
macro (`*.ema`), a product image and metadata. The importer reads only the
**portable** part data:

- **Terminals** — each `<functiontemplate>` with a physical connection
  designation becomes a QET terminal, labelled from its connection description,
  ordered by pin number.
- **Element information** — manufacturer, order number, designation and comment
  fill the element-information fields used by the BOM / nomenclature.
- **Localized names** — one `<name lang="xx">` per language the part provides, so
  the element labels itself in any UI language.
- **Symbol** — a generic symbol is generated: a body rectangle with one
  west-facing terminal per pin, with terminals grouped visually by designation.

### Deliberate scope: no macro geometry

The EPLAN `.ema` macro geometry is **not** reproduced. It is tokenised EPLAN PXF
that references EPLAN's own (proprietary) symbol libraries, so it cannot be
faithfully rendered from the package alone. A generic symbol is sufficient for
wiring and BOM, and the user can swap in a nicer symbol in the element editor
afterwards. Keeping this out is what makes the importer simple and reliable.

## Terminal naming and grouping

EPLAN parts can have multiple `<functiontemplate>` entries that share the same
`connectionDesignation` — for example a motor drive where both wire entries of
terminal "1" are both labelled "1", or a terminal block with double connections.
QET requires every terminal in an element to have a **unique name** for wiring and
terminal-diagram generation to work reliably.

`EdzElementBuilder` resolves duplicate designations in priority order:

1. **Unique designation** — kept as-is (no change for well-formed parts).
2. **Duplicated designation + distinct description** — combined as
   `designation_description`, e.g. `1_L_P`, `1_N`.
3. **Fallback** — numeric suffix: `1`, `1_2`, `1_3`, …

Terminals with the same base designation are also **visually grouped**: pins
are already sorted by designation (natural sort in `EdzPart::parse`), so
same-designation pins are always consecutive. A 5 px gap and a thin dashed
separator line are inserted between each group, mirroring the grouped-I/O
style found in typical manufacturer datasheets.

## Implementation

New, UI-decoupled module under `sources/import/edz/` (the non-UI classes are
testable headless):

| File | Role |
|------|------|
| `edzarchive.*` | Extract the `.edz` to a temp dir; locate `part.xml` |
| `edzsevenzip.*` | 7-Zip extraction via the bundled LZMA SDK |
| `edzpart.*` | Parse `part.xml` into a portable model |
| `edzelementbuilder.*` | Build the `.elmt` from the model |
| `edzimporter.*` | Orchestrate the above; write into a collection folder |
| `lzma/` | Vendored LZMA SDK (decode-only subset) |

The only UI change is one action wired into `ElementsCollectionWidget`
(modelled on the existing *New element* action). See
`sources/import/edz/README.md` for details.

## Bundled LZMA SDK

A `.edz` is a 7-Zip archive, and QET had no archive-handling code. Rather than add
a runtime dependency on an external 7-Zip, this vendors the **decode-only subset**
of Igor Pavlov's **LZMA SDK 23.01** under `sources/import/edz/lzma/` (18 `.c` +
18 `.h`, trimmed to the actual decode closure). It is **public domain** — see
`lzma/LZMA-SDK-LICENSE.txt`. Enabling the `C` language in CMake is the only build
change this requires.

Only 7-Zip-format `.edz` is supported; zip-format packages are detected by archive
magic and reported clearly ("not yet supported") rather than failing opaquely.
Corrupt/unrecognised input is likewise reported with a clear message.

## Testing

- Verified on three real ifm parts: **KG6000** (4-pin), **MFH200** (4-pin, pins
  listed out of order in the source — confirms pin-number ordering) and **R1D200**
  (5-pin, includes a Dutch name).
- Tested against the **ABB ACS880-04-330A-7+D150** `.edz` (raised in review) —
  duplicate terminal designations are now resolved to unique names and terminals
  appear in visually separated groups.
- The generated elements load and render correctly in the element editor.
- Error paths exercised: zip-format `.edz`, corrupt data, missing file.
- Builds clean on both MSVC (Windows) and gcc (Linux/WSL, system KF5).

## Not included (possible follow-ups)

- Faithful EPLAN macro geometry (see scope note above).
- Zip-format `.edz` support.
- Batch import of a folder of parts; writing QET parts-database entries.
- Geometry-driven grouping from `.ema` macro function groups (requires
  decoding the proprietary EPLAN PXF format).

## Notes for reviewers

- The large line count is almost entirely the vendored public-domain SDK; the
  QET-side code is ~1.1k lines across 14 files.
- The importer core is independent of the widget, so the archive/parse/build
  logic can be reasoned about (and tested) without the GUI.
- Copyright headers on all new files corrected to `2006-2026`.


plc-user:
So much code to read and try ... ;-)

Just one note before trying the code: 
The copyright year for new files should be the current one.

Will try your importer later!

ispyisail:
I suspect once opon a time opensouce project leads dreamed of getting lots of help, now AI has come, they are proably wishing for less help.

plc-user:
For the first shot: Impressive!

Source of my test-data:
zip-file of "ACS880-01 690V Eplan macro" from:
https://library.abb.com/r?dk=eplan%20data&q=macros
and from that zip-file the edz-file
ABB.ACS880-04-330A-7+D150.edz

results to: ACS880-04-330A-7_D150.elmt

Bildschirmfoto_2026-06-20_11-45-11

\
In my opinion, the terminal-names still need to be adjusted:
The text next to the terminals is correct, but there are, for example, several terminals named “1,” “2,” “3,” ...
I have no idea whether the information on grouping terminals is available in the EPlan files, but these are likely terminal blocks with multiple connections, which should at least be named correctly here. Perhaps using the text next to the connections, though spaces and special characters would need to be replaced appropriately.
Grouping terminals that belong together would be even better...!

ispyisail:
>  have no idea whether the information on grouping terminals is available in the EPlan files

From my understand that is one of the problems with EPLAN. Not all manufactues supply all information.

Is this feature any use to anybody?

plc-user:
> From my understand that is one of the problems with EPLAN.

QElectroTech is allowed to be better than proprietary software!  :wink:

It would be helpful to be able to see at a glance that, for example, the three wires of a motor are located right next to each other.
Or that the connections for an analog input are grouped together.

plc-user:
It should be clear to everyone that an importer like this cannot be perfect and cannot always fully meet all of the user’s wishes or needs. However, what we should definitely avoid are duplicate or even multiple terminal names. Otherwise, an automated system for creating wiring or terminal diagrams cannot function reliably.

plc-user:
It's definitely easier to follow when the I/O are right next to each other:

Bildschirmfoto_2026-06-20_12-32-04
Source: ABB

ispyisail:
Thank you for the thorough testing and the very clear feedback — two fixes just pushed:

**1. Copyright years**
All new files under `sources/import/edz/` updated to `2006-2026`. Good catch, apologies for missing that.

**2. Duplicate terminal names** (the core issue you identified)
EPLAN parts can have multiple `<functiontemplate>` entries with the same `connectionDesignation` — motor drives, terminal blocks with double connections, etc. QET requires every terminal in an element to have a unique `name`, otherwise wiring and terminal-diagram generation cannot work reliably (as you correctly pointed out).

`EdzElementBuilder::build()` now resolves duplicates in order:
- **Unique designation** → kept as-is (no change for well-formed parts)
- **Duplicated designation + distinct description** → `designation_description`, e.g. `1_L_P`, `1_N` — uses the text that was already shown next to the terminal
- **Fallback** → numeric suffix: `1`, `1_2`, `1_3`, … when descriptions are missing or still collide

For the ABB ACS880 example, the motor terminals (U2/V2/W2) and brake resistor connections that share designations with the power input should now each get a unique name derived from their function label.

Grouping visually (putting I/O clusters next to each other) would need knowledge of the internal function groups from the EPLAN macro, which isn't available in the portable `part.xml`. That remains a potential follow-up if the macro format can be decoded further.

ispyisail:
One more push — terminal grouping is now visual too.

Pins are already sorted by designation (natural sort in `EdzPart::parse`), so same-designation connections are already consecutive. The new change:

- Inserts a **5 px gap** between groups with different designations, so each group reads as a distinct block.
- Draws a **thin dashed horizontal line** through the gap — the same grouped-I/O style shown in your ABB screenshot.

The body rectangle and bounding box grow automatically to fit, so nothing is clipped. For the ACS880 example, the motor terminals (all sharing one designation), brake resistor connections, and power input connections should now each appear as a clearly separated cluster.

plc-user:
Thank you for the quick reply!

But: The grouping is wrong. See here:
grafik
\
At the moment the grouping is made by first character of the terminal-name, but the first part of the Terminal-name is the position of the terminal in that terminal-block!

Besides: The terminals should always be located in full ten-position slots.

ispyisail:
i see......must be a very big device, 11 termial blocks?

plc-user:
Quite normal for a servo-drive ...  :man_shrugging: 

24V Vcc
AC-IN
Motor Out
DC-Circuit
Breaking-Resistor
Feedback 
Analog-I/O 
Digital-I/O
...

ispyisail:
I only tested on 4 wire sensors, I've get back to this, My workstaion (Xeon 24 thread ECC) is busy looking for memory bugs.

ispyisail:
Fix pushed.

The grouping was using `connectionDesignation` (the pin position number: 1, 2, 3 …) as the group key, which is why it produced a separate group for every position number rather than one group per functional block.

The correct key is the `functiondefinition` attribute — which identifies the functional block (FINP, MOUT, DCBUS, BRES, …). It lives on `<functiontemplate>` directly in newer EPLAN versions, or on the parent `<function>` wrapper element in older ones.

Changes:
- Added `group` field to `EdzPin`, populated from `functiondefinition` (with fallback to the parent element)
- Pins now sort by group first (preserving XML appearance order), then by designation within each group via natural sort
- Group-break detection (separator lines + Y gaps) now compares `functiondefinition` blocks, with a fallback to designation for parts that carry no `functiondefinition` data

For the ACS880 this should give ~8 functional blocks with separator lines between them rather than 11+ groups by pin number. Could you test against your ABB macro and let me know if the grouping now matches the functional blocks?

plc-user:
There is a warning on compile:

```
../../sources/import/edz/edzelementbuilder.cpp: In static member function ‘static QDomDocument EdzElementBuilder::build(const EdzPart&)’:
../../sources/import/edz/edzelementbuilder.cpp:70:66: warning: missing initializer for member ‘EdzPin::group’ [-Wmissing-field-initializers]
   70 |                 pins.append(EdzPin{QStringLiteral("1"), QString()});
      |                                                                  ^
```

It looks like there is **no** change in resulting element:
grafik

Perhaps the information needed is not provided in that edz?
Maybe it'd be easier and less time-consuming, if you'd downloaded the edz-file yourself? :thinking: 
The way to find the data is already posted above.

ispyisail:
Thanks for the report @plc-user — you were right, terminals were being grouped by pin position number (designation) rather than by functional block.

**Root cause:** EPLAN 2022-style part.xml files (such as the IFM AL1122 tested here) use a numeric `functiondefgroup` attribute and do not include the text `functiondefinition` attribute that the previous grouping logic relied on. When `functiondefinition` was absent, the code fell back to grouping by `connectionDesignation`, producing symbols with all pin "1"s stacked together, then all "2"s, etc.

**Fix (commit 2aeeb74):**
- `terminalNr` is now read as the primary group key — it directly identifies the physical connector socket (e.g. `X01`, `X31` for an IFM AL1122).
- If `terminalNr` is absent (older EPLAN formats), the code falls back to the text `functiondefinition` block name (`FINP`, `MOUT`, …) as before.
- Pins within each connector group are still sorted by designation using natural sort.
- Removed the dashed inter-group separator lines; the 5 px gap between groups provides sufficient visual separation.

Result for IFM AL1122 (8-port IO-Link master): pins are now grouped per physical M12 socket — X01 shows pins 1, 2, 3, 5 together; X02 shows 1, 2, 3, 5; … X31 shows 1, 2, 3, 4 — with a gap between each connector group.

ispyisail:
I will get back to this.

plc-user:
Oh sorry, I only see a login-button at that page, but do not need to login to download ... strange ...

(removed file here: see below)

plc-user:
This may help sorting the Terminals:
In that file the terminal-blocks seem to be declared by
```
terminalNr="..."
```

inside the quotes you find the eleven terminal-blocks:
 - XAI
 - XAO
 - XD2D
 - XD24
 - XDI
 - XDIO
 - XPOW
 - XRO1
 - XRO2
 - XRO3
 - XSTO

All entries with `terminalNr` have numbers in `connectionDesignation`, which should be the position in the according terminal-block.

Additionally you find twelve named "connectionDesignation" like "L1/U1", "L2/V1", etc.
These can be sorted alphabetically, what even would not be fully correct because of `PE` and `PE'` that will belong to 3-phase-input or motor-output...

plc-user:
It's visually much better, when the lines and tags in the xml are sorted: 
EDIT: (File removed due to legal concerns.)
(unzip and view in editor with monospaced font)

Based on this file, I would suggest the terminal-names as a combination of `termnalNr` and `connectionDesignation` like "XDI.2" or "UDC+".
In that case the electrician can directly use an automated created wiring-list, as it is not guaranteed that signal-names (`connectiondescription`) are printed on the device. These should be used as text-label beside the terminal.

ispyisail:
Can you review now? 

image

ispyisail:
image

ispyisail:
**Terminal naming updated (commit 8268df8)** — implemented your suggestion exactly:

- Terminals with `terminalNr` → `terminalNr.connectionDesignation`, e.g. `XDI.2`, `XRO1.3`, `XPOW.1`
- Terminals without `terminalNr` (power/busbar connections) → designation as-is: `L1/U1`, `UDC+`, `PE`
- `connectiondescription` (signal name) stays as the human-readable label beside the terminal, not the name

The old duplicate-resolution logic (sanitise + numeric suffix + description appending) is replaced by the dot-notation which is unique by construction for all well-formed EPLAN data.

ispyisail:
Can you review now

plc-user:
We're almost there!  :smile: 

I made a few small changes in `edzelementbuilder.cpp`:

```
// add font-entry for label:
const QString LABEL_FONT =
    QStringLiteral("Liberation Sans,9,-1,5,50,0,0,0,0,0,Regular");

// always set terminals at positions n * 10
const int group_gap = 10; // extra px inserted between designation groups

// no gap between terminals with empty group:
for (int i = 1; i < n; ++i)
    is_group_break[i] = (groupKey(pins.at(i)) != groupKey(pins.at(i - 1))) && (!(pins[i].group.isEmpty()));

// move label according font-size:
dyn.setAttribute(QStringLiteral("y"), min_y - 9);

// use font for label:
dyn.setAttribute(QStringLiteral("font"), LABEL_FONT);
```

... to get this result:
grafik

This way the terminals are always located in full ten-position slots. and we do not have an extra-gap between L1, L2, L3 and can directly connect 3-phase AC-Source, for example.

plc-user:
Another Idea:
Maybe we can add extra-texts at each pin 1 of every (named) terminal-group with text of `terminalNr`?
Maybe in the gap above?

With something like this one can directly see what terminal-block to use. 
Here's the difference:
grafik

ispyisail:
Go for your life, im getting out of my depth

ispyisail:
## Review notes — EPLAN `.edz` Import

**Overall:** Well-structured PR. The pipeline architecture, error handling, and defensive guards are solid. Two things worth considering before merge.

---

### Architecture
The pipeline layers cleanly and each class is independently testable headless:

```
EdzArchive → EdzPart → EdzElementBuilder → EdzImporter → UI widget
```

Each layer carries its own `errorString()` and propagates failures upward. The UI widget calls just `importer.importToDirectory()` and refreshes — good decoupling.

---

### Things that look good
- **Magic-byte detection** in `edzarchive.cpp` — distinguishes 7z vs zip upfront with separate human-readable error messages rather than failing opaquely inside the LZMA SDK
- **Natural sort** for pin ordering (`natLess`) — electricians get "1, 2, 3, 10" not "1, 10, 2, 3"
- **Terminal name collision guard** — appends `_2`, `_3` suffix for malformed data; won't crash or silently alias pins
- **Localized name fallback chain** in `pickLang`: `en_EN` → `en_US` → `en_GB` → any `en_*` → first available
- **`LANGUAGES C CXX`** in `CMakeLists.txt` — correct, the LZMA SDK `.c` files need the C compiler registered
- **`QTemporaryDir` ownership** — destroyed after the `.elmt` is written, correct lifetime
- **Context menu guard** — `importEdz()` checks `!isCommonCollection()` and `isDir()` before proceeding

---

### Two things worth considering

**1. Silent overwrite — no collision protection**

In `edzimporter.cpp` the output file is opened with `QIODevice::WriteOnly`, so an existing `.elmt` with the same name is silently replaced. Users importing two variants of the same order number will lose the first without any warning. A pre-check and confirmation dialog (or `QIODevice::NewOnly` + a clear error) would be safer.

This is compounded by the filename sanitisation:
```cpp
base.replace(QRegularExpression("[^A-Za-z0-9._-]"), "_");
```
Two parts with order numbers `ABC/123` and `ABC 123` both produce `ABC_123.elmt`, so two different parts can silently collide to the same filename.

**2. Synchronous import on the UI thread**

The full pipeline (extract, decompress, parse, build, write) runs on the main thread. Most `.edz` files are small so it is probably fine in practice, but a large archive could freeze the UI noticeably. Worth at least a comment for future reference.

---

### Known limitation — noted, not a blocker

All generated terminals have `orientation="w"` (west-facing). The README is clear that EPLAN's `.ema` macro geometry uses proprietary PXF tokens that cannot be reproduced, and advises swapping in a nicer symbol afterwards. That is the right call — just flagging it so it is visible in the review thread.

scorpio810:
Hello Shane, thank you for your work, but is this really legal? See: https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/458

Thank you also for updating qelectrotech.pro. By the way, why didn’t you start straight away in the element editor? 
Like plc-user's QET_ElementScaler,  vadoola's dxf2elmt.

Edit : ok you start with add new element wizard.

You’ll have to use it anyway to fine-tune the symbol after import.
At the moment, I saw this PR conflicts with the master branch, but I haven’t had time these days to test it, sorry.

scorpio810:
Hey, really cool work on this — being able to pull EPLAN Data Portal parts directly into QET would be a huge win for anyone working with industrial components day to day. 🙌

One thing I wanted to flag before this goes further, just so it's on everyone's radar: I had a look at the EPLAN Data Portal Terms of Use, and there's a clause (§5.4) that says portal users may only incorporate .edz data "in connection with EPLAN and products acquired from EPLAN". So strictly speaking, importing .edz files into a non-EPLAN tool like QET could fall outside what the Terms of Use permit — not necessarily a code/copyright issue with this PR itself, more a contractual restriction on how the downloaded data can be used by whoever grabs it from the portal.

To be clear, I'm not a lawyer and this isn't legal advice — just something I found while poking around, and I think it's worth the project being aware of before merging/promoting this feature, especially since EPLAN explicitly states they hold the IP rights to the portal and its data.

Might be worth:

Adding a short note/disclaimer in the README or the import dialog itself, so users know the data usage restriction exists and it's on them to comply with EPLAN's ToS.
Maybe reaching out to EPLAN directly to ask if interoperability use like this is something they're fine with (some vendors are surprisingly relaxed about this in practice, others aren't).

Not trying to throw cold water on a genuinely useful feature — just flagging it now rather than after it ships. 🙂

scorpio810:
Following up on the licensing point above — I talked it through a bit more, and I think there might be a path that keeps this genuinely useful and keeps QET's official repo at arm's length from the ToS question.

Looking at the architecture you've already built, the core logic (EdzArchive → EdzPart → EdzElementBuilder → EdzImporter) is nicely decoupled from the QET UI — it's basically a standalone .edz → .elmt converter with one thin wiring point into ElementsCollectionWidget. That's great engineering, and it also happens to make this pretty easy to split out.

What I'd suggest: pull sources/import/edz/ into its own small repo (standalone CLI or minimal Qt GUI, no dependency on QET internals) rather than merging it into qelectrotech-source-mirror directly. The user runs the tool, gets an .elmt file, and imports it the normal way QET already supports (drop it into a writable collection folder / open it in the element editor). We could link to it from the wiki or docs as a community tool.

This isn't about the quality of the work — it's solid. It's more that:

It keeps the official QET project from being the one "incorporating EPLAN data outside the EPLAN ecosystem," which is the specific thing the Data Portal ToS restricts (§5.4).
It still gets the feature to everyone who wants it, just as an optional companion tool instead of baked into core — similar to how plc-user's QET_ElementScaler, dxf2elmt_vadoola or the terminal-table generator already live as separate repos in the org.
It sidesteps having to vendor/maintain the LZMA SDK inside the main codebase too, which is a nice side benefit.
Happy to help test against the standalone build if you want to go this route — and again, really appreciate the work that's gone into this, just trying to think about it from a "what happens if EPLAN's legal team emails us in six months" angle. 🙂

ispyisail:
I was also thinking about a plugin

https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/458

scorpio810:
Glad we're on the same page here, @ispyisail — and yeah, linking to #458 makes sense, that's the same concern I raised back in May .
To be honest the engineering itself isn't what worries me, it's clearly been done carefully — decoupled core, tested against real parts, scoped to avoid touching the .ema geometry. What I'd rather avoid is the official repo being the one that incorporates EPLAN Data Portal data outside their own ecosystem, since that's basically what §5.4 of their ToS restricts. If it's a separate tool the user installs and runs themselves, that risk sits with them, not with the project.

Splitting it out shouldn't be too painful given how it's structured already — EdzArchive/EdzPart/EdzElementBuilder/EdzImporter barely touch QET internals, so pulling sources/import/edz/ into its own repo as a small CLI or standalone GUI seems pretty doable. It would still spit out the same .elmt file, just imported through the normal collection flow instead of a built-in menu action. We could point to it from the wiki, same as we do for dxf2elmt or the terminal-table tool.

Bonus: that also gets the LZMA SDK out of the main codebase, which I wasn't thrilled about vendoring anyway.

Let me know if you want to go that route, happy to help test once it's split off.

ispyisail:
What about a plugin? End user installs into QET?

scorpio810:
>
> What about a plugin? End user installs into QET?

Yes

> I was also thinking about a plugin
> 
> #458

Yes, I' m think.

plc-user:
Thank you @ispyisail  :+1: 

With the latest changes users can take the created element "as is" or (when re-working with element-editor) can mark and drag complete terminal-blocks including the block-name to suit own needs!

To begin with: I am not a lawyer!

When I understand correctly, it is **not** allowed to use edz-files downloaded from eplan with other software than from eplan. Files from other sources (e.g from hardware-manufacturers) may have other license-conditions. 
Having that "disclaimer" in source-code is absolutely correct!
What do you think of an additional message-box in QET that pops-up once during a session, when first importing an edz-file?

ispyisail:
Thanks @plc-user — glad the drag-and-drop workflow landed well.

On the licensing question: the distinction I'd draw is between Eplan's own content and manufacturer content. The parts this importer is designed for — ifm, Phoenix Contact, Siemens, Weidmüller, etc. — are distributed by the manufacturers themselves via the EPLAN Data Portal specifically so engineers can use their product data in tooling. The EDZ is just the packaging format; the data belongs to the hardware manufacturer, who wants it in your BOM.

The importer also deliberately skips the `.ema` macro geometry for exactly this reason — that file contains Eplan-proprietary PXF references that can't be faithfully rendered without Eplan's own symbol libraries, and reproducing it would be the part most likely to raise a legitimate IP concern. The source-code disclaimer covers that reasoning.

On the message box: I'd gently push back on a per-session popup. In practice users click through those in under a second without reading them, so they carry real UX cost for very little actual benefit. If there's appetite for surfacing the concern at the UI level, I'd suggest a small static note in the import dialog instead — something like "Ensure you have the right to use files from this source." — which is readable without being disruptive and stays visible every time rather than once per session.

Happy to add that label if it would help get this merged.

plc-user:
My point here is to shift the legal responsibility onto the individual user. We’re simply providing a tool; the user is responsible for how they use it. 
A hammer is primarily designed to drive nails into something. Anyone who uses it for something else...

In my opinion, a clear hint or warning in the file-open dialog would be sufficient, but Laurent should weigh in on this before we merge the code.

scorpio810:
I don’t know yet; I’ve asked the question on the forum.  

https://qelectrotech.org/forum/viewtopic.php?pid=23031#p23031

- The translation file is cluttering up the PR. Please remove it.
- For the LZMA sources, a GitHub submodule be better, as it’s easier to maintain?
- The files are still missing from `qelectrotech.pro` for my Qmake packages builders workflows.

scorpio810:
[qelectrotech_pro_edz.patch](https://github.com/user-attachments/files/29656025/qelectrotech_pro_edz.patch)

scorpio810:
@plc-user : Happy to add that label if it would help get this merged. Ok

scorpio810:
Title: Warning — Importing an EPLAN file (.edz)

The .edz format can originate from two different sources:

  • The EPLAN Data Portal (dataportal.eplan.com), which is subject to
    the terms of use of the EPLAN Cloud environment;
  • A component manufacturer's own website (e.g. a manufacturer or
    distributor) that makes its .edz files available directly, under
    its own terms.

QElectroTech has no way of automatically determining the source of
the file you are importing, nor the terms that apply to it.

By importing this file, you confirm that:

  • you know its origin and are authorized to use it in this context,
    with regard to the terms applicable to that source;
  • this import is performed at your own risk and responsibility;
  • neither QElectroTech, its maintainers, nor its contributors can
    be held liable for any non-compliant use of this data.

☐ I have read and accept these terms.

                                    [ Cancel ]   [ Import ]

plc-user:
Sorry for the late reply: I unfortunately had to spend some time in hospital...

@scorpio810:
If that warning helps us keep the lawyers at bay, I’m obviously all for adding this feature to QET! 
But as already said: I’m not a lawyer myself, and that decision is up to you, Laurent!

plc-user:
@ispyisail: 
Did you already prepare a messagebox, window or whatever to display such a dialog?

ispyisail:
@plc-user: Not yet — I checked the current PR diff and there's no messagebox/dialog implementing @scorpio810's warning text. The only dialog in the import flow right now is the file picker; the `.edz` parsing/import logic (`edzarchive`, `edzimporter`, `edzelementbuilder`, `edzpart`) doesn't show any consent UI.

If the warning text as drafted is agreed, I'm happy to add a `QDialog` with the terms text, a checkbox ("I have read and accept these terms"), and Cancel/Import buttons, gated so Import stays disabled until the checkbox is ticked — shown once per import action before the file dialog opens. Let me know if that matches what you had in mind and I'll push it.

scorpio810:
@plc-user : Get well soon, and do get some rest – I, too, am regularly admitted to hospital.

> If the warning text as drafted is agreed, I'm happy to add a QDialog with the terms text, a checkbox ("I have read and accept these terms"), and Cancel/Import buttons, gated so Import stays disabled until the checkbox is ticked — shown once per import action before the file dialog opens. Let me know if that matches what you had in mind and I'll push it.

Yes, and if Eplan’s lawyers were to contact us and were not satisfied, we would withdraw this pull request without a second thought.

ispyisail:
@scorpio810 @plc-user Pushed. The consent dialog is in now: a `QDialog` shows the warning text before the file picker opens, with a checkbox ("I have read and accept these terms") and Cancel/Import buttons — Import stays disabled until the box is ticked. It only shows once per import action, same as discussed.

I built it locally to make sure it actually compiles and links cleanly before pushing (no errors, no new warnings from that file). Commit: 1b8dea3.

Appreciate you both weighing in on this — let me know if the wording needs any tweaks.

scorpio810:
https://qelectrotech.org/forum/viewtopic.php?pid=23106#p23106

plc-user:
Something I just noticed:
In `sources/import/edz/edzarchive.cpp`, some texts cannot be translated because they are statically implemented.

@ispyisail: 
Should we adjust this in another commit?

ispyisail:
> Something I just noticed: In `sources/import/edz/edzarchive.cpp`, some texts cannot be translated because they are statically implemented.
> 
> @ispyisail: Should we adjust this in another commit?

happy to go any direction you want to go

scorpio810:
@ispyisail LZMA SDK 23.01 here but  I saw new it taged 26.02 so for maintenance can you try a git submodule based on https://github.com/ip7z/7zip

For the 23.01 → 26.02 update: let's move to a `git submodule` against `ip7z/7zip` instead of re-vendoring static files each time — mainly for security/maintenance: a submodule pinned to a tag gives a clean, auditable diff on every version bump (`git log` between two commit hashes) instead of hand-diffing dropped-in files.

Steps:

1. Add the submodule and pin it to the tag:
```
git submodule add https://github.com/ip7z/7zip.git sources/import/edz/lzma-sdk
cd sources/import/edz/lzma-sdk
git fetch --depth 1 origin tag 26.02
git checkout 26.02
cd ../../../..
```

2. Restrict the submodule's working tree to just what we need — `ip7z/7zip` is the whole 7-Zip source tree now (GUI, other archive formats, encoders, AES/SHA...), not just the SDK. The `C/` folder alone has ~111 files vs. the 36 we currently vendor, so leaving it unrestricted would pull unrelated code into the tree:
```
git sparse-checkout init --cone
git sparse-checkout set C DOC
```

3. Patch `qelectrotech.pro`: the old glob (`$$files(sources/import/edz/lzma/*.c)`) can't just be repointed at the submodule's `C/` folder, since that glob would now sweep in all ~111 files instead of the 36 we actually use. I've replaced both globs with an explicit file list of the same 36 files, re-pathed to `sources/import/edz/lzma-sdk/C/`, and updated `INCLUDEPATH` accordingly. Patch attached.

4. Drop the old vendored copy: `git rm -r sources/import/edz/lzma`

Header compatibility checks out fine — `7z.h` is byte-identical between 23.01 and 26.02, and none of the `SzArEx_*` calls `edzsevenzip.cpp` uses have changed signature, so this isn't a risky API bump.

Patch is attached — can you apply it on your branch and confirm it builds cleanly on your end before we open the PR?
[lzma-sdk-submodule.patch](https://github.com/user-attachments/files/30370883/lzma-sdk-submodule.patch)

scorpio810:
Hum ..... Whait !!! https://github.com/ip7z/7zip/issues/233
Vulnerabilitiy CVE-2026-58052 in 26.2

scorpio810:
Please waiting.. https://sourceforge.net/p/sevenzip/bugs/search/?q=CVE

scorpio810:
https://korben.info/en/7-zip-booby-trapped-file-code-execution-vulnerability.html

**It affects all versions of 7-Zip from 21.07 through 26.01, and it's version 26.02, released on June 25, that patches the hole.**

https://www.it-connect.tech/7-zip-vulnerability-why-you-should-install-version-26-02-now/

scorpio810:
Looked into this one — CVE-2026-58052 doesn't apply to what we'd be pulling in.

The bug is a Mark-of-the-Web bypass in 7-Zip's Windows GUI: when extracting a crafted RAR5 archive, an alternate NTFS data stream named `:Zone.Identifier:$DATA` slips past the guard that normally strips the archive-supplied Zone.Identifier, so the extracted file loses its "downloaded from the internet" flag. That logic lives in the RAR5 archive handler and the GUI's on-disk extraction/Zone.Identifier propagation code — both outside `C/` `DOC`.

Our `edzsevenzip.cpp` only calls the portable SDK's `SzArEx_Open`/`SzArEx_Extract` for **7z**-format archives, never touches the RAR reader, and never goes through the GUI's MotW propagation path. So with the sparse-checkout scoped to `C DOC`, this CVE isn't reachable from our code regardless of whether it's fixed upstream yet.

@ispyisail @plc-user — let me know if you see things differently.
It would be useful to add a one-line comment to the submodule’s configuration to clarify that we are deliberately excluding RAR/GUI managers, so that this issue does not resurface every time a new CVE vulnerability relating to 7-Zip is reported.

scorpio810:
One more thing worth weighing before we commit to the submodule: `sources/import/edz/lzma` has barely changed since it landed — the only touch since the initial commits was the copyright-year fix. That's a very different update cadence from `pugixml` or `SingleApplication`, which get bumped regularly across the rest of the codebase. A submodule adds permanent overhead for every contributor (recursive clone, `git submodule update --init`, the sparse-checkout config to keep in sync) for a component that, going by history, might not move again for years.

That said, I don't think low churn makes it *useless* — it changes what the cost/benefit is about. The real argument for the submodule isn't "we update this often," it's "when a CVE like this week's drops out of nowhere, how fast can we tell if we're affected and pull a clean fix." Manually re-vendoring and diffing 36 files by hand under time pressure is exactly the kind of thing a submodule avoids — `git log <old-tag>..<new-tag> -- C/` gives you the whole picture in seconds.

Given that, would a middle ground work better here: keep the static vendored copy as-is, but add a short `UPSTREAM.md` (or a header comment) recording the exact upstream tag/commit it was sourced from? That gives most of the audit traceability without the recurring submodule overhead on every clone — and we can always revisit a full submodule if this SDK starts seeing more frequent security fixes going forward.

ispyisail:
@scorpio810 — agree on both counts, and I checked the details independently before replying.

**CVE-2026-58052 doesn't block us.** NVD's text is *"7-Zip for Windows through 26.01 fails to preserve the Mark-of-the-Web when extracting a crafted RAR5 archive."* Worth noting for anyone reading later that [ip7z/7zip#233](https://github.com/ip7z/7zip/issues/233) has a **title** saying "in 26.2" but a body that's only a bare NVD link — the title is what caused the scare. Your reachability analysis is right: RAR5 is `CPP/7zip/Archive/Rar*`, MotW propagation is `CPP/7zip/UI/`, neither is in `C/` or `DOC`, and `edzsevenzip.cpp` only ever calls `SzArEx_Open`/`SzArEx_Extract` on 7z-format input.

I also verified the compatibility claims by diffing both tags:

- `7z.h` is byte-identical 23.01 ↔ 26.02 ✔
- all 36 files still exist in 26.02's `C/`, and all still carry an explicit *"public domain"* statement ✔
- the 36-file subset is still self-contained — the only newly-referenced header outside it is `7zCrcEmu.h`, guarded by `#ifdef Z7_CRC_HW_USE`, which is commented out upstream. So the file list doesn't need to grow ✔

Three things on the patch itself before we open the follow-up PR:

**1. It breaks the CMake build.** `cmake/qet_compilation_vars.cmake` (lines 131–148) hard-lists all 18 `sources/import/edz/lzma/*.c` files, and `CMakeLists.txt:25` enables `LANGUAGES C CXX`. The patch only touches `.gitmodules` and `qelectrotech.pro`, so step 4 (`git rm -r sources/import/edz/lzma`) would leave CMake pointing at deleted sources and configure would fail for everyone building the default way. It needs the same re-path applied there, plus the include dir.

**2. `git sparse-checkout` doesn't propagate.** It writes to `.git/info/sparse-checkout` and local config, neither of which is tracked — so anyone running `git submodule update --init --recursive` gets the whole 7-Zip tree (`CPP/`, GUI, RAR handler and all). Step 2 trims your working copy, not other people's. What actually bounds the compiled surface is the explicit file list in step 3, which is correct and does the job. Happy to add the clarifying comment you suggested, but I'd word it around the file list rather than the sparse-checkout, so it doesn't document a guarantee we don't have.

One side effect worth naming, given how this PR's review went: `DOC/License.txt` puts `CPP/7zip/Compress/Rar*` under *"GNU LGPL with unRAR license restriction"*, which isn't GPL-compatible. It would sit in every contributor's checkout — uncompiled and unlinked, so I don't think it's a real problem, but it does mean trading 36 stated-public-domain files for a tree carrying a non-free subdirectory.

**3. Source tarballs.** This one affects you directly: `git archive` doesn't include submodules, so a release tarball cut that way ships a tree that can't compile, for every packager downstream. That works against the maintenance motivation.

Would this get you the same benefit? Keep the files vendored, and add `scripts/update-lzma-sdk.sh` that fetches a given `ip7z/7zip` tag, copies exactly those 36 files, and records the tag + upstream commit hash in a `VERSION` file next to them. Each bump becomes one commit whose `git diff` *is* the auditable diff — the same "clean diff on every version bump" you're after, with no tarball, clone-size or licence side effects, and nothing to change in either build system. If you'd still rather have the submodule I'll do it that way, just with the CMake side fixed.

---

**On the reason for bumping — I ran the numbers.**

I built our EDZ fuzz harness against master's vendored 23.01 and against 26.02, same compiler and flags, same 4023 inputs (the stock corpus plus 4000 seeded mutants that corrupt the next header and then repair both CRC gates, so they actually reach `SzReadHeader` instead of dying at the CRC check). 26.02 dropped in with no source changes and no new warnings.

Result: **23.01 crashes on 16 inputs, 26.02 on none, with zero regressions.**

All 16 are the same bug. `SzArEx_GetFileNameUtf16` (`7zArcIn.c:1729`) does:

```c
const size_t offs = p->FileNameOffsets[fileIndex];
```

with no NULL check. A 7z `kFilesInfo` section may legally declare `NumFiles > 0` and simply omit the `kName` property, which leaves `FileNameOffsets` NULL. 7-Zip's own tools always write names, so it never happens in practice — but nothing in the format forbids it, and it's reachable straight through `edzsevenzip.cpp:97` → `EdzArchive::extract` → `EdzImporter::importToDirectory`, i.e. the import action exactly as merged. I have a deterministic 354-byte reproducer; under ASan it's `SEGV on unknown address 0x000000000000`.

26.02 adds exactly the missing guard:

```c
const size_t * const offsets = p->FileNameOffsets;
if (!offsets) { if (dest) *dest = 0; return 1; }
```

and it's new in **26.02 specifically** — I checked 24.05, 24.09, 25.00, 26.00 and 26.01, none of them have it. Same story for a memory leak in `ReadTime()` when a crafted header repeats the time property: 23.01 overwrites `p->Defs` and leaks the earlier allocation, 26.02 rejects the duplicate with `if (p->Defs) return SZ_ERROR_ARCHIVE;`. Upstream's changelog for 26.02 says only *"Some bugs and vulnerabilities were fixed."*

So the case for 26.02 isn't the MotW CVE, which we've established is unreachable — it's that 23.01 has a crash our own import path hits.

Two things I'd add regardless of how the submodule question lands:

1. **A guard in `edzsevenzip.cpp`** — `if (db.NumFiles && !db.FileNameOffsets) { res = SZ_ERROR_ARCHIVE; }` before the extract loop. I verified this fixes the crash on 23.01 without touching the SDK at all, which matters for anyone building against an older or system copy.
2. **A bound on the declared unpack size.** `SzArEx_Extract` hands the header's declared unpack size to `ISzAlloc_Alloc` with only a truncation check, so a tiny `.edz` can request an absurd allocation (~9 EB in one of my mutants). **Neither 23.01 nor 26.02 fixes this.** To be clear it is *not* a crash — the SDK checks the allocation result and returns `SZ_ERROR_MEM` — but a small file can still drive real memory exhaustion at sizes `malloc` will actually satisfy, so a sanity check against the compressed input size is worth having.

Happy to open the follow-up with the SDK bump, both guards, and the crashing inputs added to the fuzz corpus as regression cases.

---

@plc-user — hope the recovery is going well. Your translation point is still open, and you were right: `edzarchive.cpp` has six user-facing error strings as `QStringLiteral` (file-not-found, cannot-read, zip-format, unrecognised-format, temp-dir, no-`part.xml`). `Q_DECLARE_TR_FUNCTIONS(EdzArchive)` + `tr()` sorts it. I'll fold that into the same PR unless you'd prefer it separate.

plc-user:
Thank you very much for your well-wishes: I’m on my way!

How we implement the changes regarding the missing translations shouldn’t really matter – the main thing is that we do it and at least offer texts for French, English and German! For other languages, it’s best if native speakers contribute to this.

I can’t offer much input on whether we should integrate the code separately or use it as a submodule, as I have no experience of how this can be done in detail.

Regarding the vulnerabilities: 
I don’t think we’re affected by them, as we only use the 7z unpacker. Our code does not process zip, xz or rar files. For zip format, there’s a separate message stating that it isn’t supported. Do we need further explicit messages for other archive formats, given that we already have the check for the 7z header in the archive? I don’t think so.

ispyisail:
@plc-user Good to hear you're on the mend.

**Translations:** agreed, the mechanism doesn't matter. I'll wrap the six strings in `edzarchive.cpp` with `Q_DECLARE_TR_FUNCTIONS` + `tr()` and supply EN/FR/DE in the follow-up. Draft strings below — if you'd correct the German I'd be grateful, you'll get it more idiomatic than I will.

| # | English (source) | Deutsch | Français |
|---|---|---|---|
| 1 | `File not found: %1` | `Datei nicht gefunden: %1` | `Fichier introuvable : %1` |
| 2 | `Cannot read %1` | `%1 kann nicht gelesen werden` | `Impossible de lire %1` |
| 3 | `This .edz is a zip-format package, which is not yet supported (only 7-Zip .edz files can be imported).` | `Diese .edz-Datei ist ein Paket im ZIP-Format, das noch nicht unterstützt wird (es können nur .edz-Dateien im 7-Zip-Format importiert werden).` | `Ce fichier .edz est un paquet au format zip, qui n'est pas encore pris en charge (seuls les fichiers .edz au format 7-Zip peuvent être importés).` |
| 4 | `Not a valid .edz package (unrecognised archive format).` | `Kein gültiges .edz-Paket (unbekanntes Archivformat).` | `Paquet .edz non valide (format d'archive non reconnu).` |
| 5 | `Could not create a temporary directory: %1` | `Temporäres Verzeichnis konnte nicht erstellt werden: %1` | `Impossible de créer un répertoire temporaire : %1` |
| 6 | `No *.part.xml found inside the .edz` | `Keine *.part.xml in der .edz-Datei gefunden` | `Aucun fichier *.part.xml trouvé dans le .edz` |

Two notes on those: the French entries use a space before `:` per French typographic convention, matching the rest of `qet_fr.ts`. And #2 reads oddly in English (`Cannot read %1`) — the German and French versions above put the filename first, which reads better; I can reword the English source string to `Cannot read file: %1` for consistency if nobody objects to touching the source text.

**Submodule vs vendored:** understood, no problem. That one's really Laurent's call. I've laid out the trade-offs and I'm happy either way, as long as the CMake file list gets fixed alongside whichever route we take.

**On the vulnerabilities — one distinction I want to draw, because I think two separate things are getting merged.**

You're right about CVE-2026-58052. It's a Mark-of-the-Web bypass in the RAR5 handler plus 7-Zip's Windows GUI extraction code, and we touch neither. That one is a non-issue for us, as scorpio810 concluded too.

But the crash I found isn't that class of bug. It's *in the 7z reader itself* — `SzArEx_GetFileNameUtf16` in `7zArcIn.c`, the code path we do use and the only one we use. The reproducer is a **genuine, well-formed 7z archive**: it carries the correct `7z BC AF 27 1C` magic, so the format check in `edzarchive.cpp` accepts it; it has valid CRCs on both the signature header and the next header, so `SzArEx_Open` proceeds normally. It then declares one file and simply omits the optional `kName` property. That is legal per the 7z format — 7-Zip's own tools always write names, so it never occurs naturally, but nothing forbids it — and 23.01 then dereferences a NULL `FileNameOffsets`. Under ASan: `SEGV on unknown address 0x000000000000`.

So the existing header check doesn't help here, because the file *is* a 7z file. Restricting ourselves to the 7z unpacker is exactly what exposes us to it. Out of 4000 mutated archives, 16 crashed 23.01 and none crashed 26.02.

**On extra format messages:** I agree with you, we don't need any. The magic check plus the specific zip message is the right amount. The gap isn't archives of *other* formats — it's malformed archives of *our* format, and that's addressed by the SDK bump plus the `edzsevenzip.cpp` guard rather than by more error strings.

scorpio810:
BTW, in  an long future 7-zip can be interesting for storing pictures, now image was encoded to base 64 in the qet.xml if user use so many pictures QET can lag  when decode large base64 codes.

> Modification of the int BACKUP_INTERVAL from 2 min to 20 min used by KautoSaveFile. On a large project with a 256 MB folio printed in A0 format, the graphical interface freezes for 30 seconds when KautoSaveFile writes this large amount of data to the disk every two minutes."


See https://qelectrotech.org/wiki_new/refonte_du_code_de_qelectrotech, sorry is in French.

ispyisail:
> See https://qelectrotech.org/wiki_new/refonte_du_code_de_qelectrotech, sorry is in French.

How much do you want this?

I did a quick review with AI, and it claimed it could do it. However, it suggested that the real limitation is getting enough real-world people to check, test, and approve it.

I don't really want to try if its going to sit as a PR for a few years

plc-user:
@ispyisail:
We can make the German Text No. 2 even shorter and in my opinion more readable: 
"Kann %1 nicht lesen."

plc-user:
> BTW, in an long future 7-zip can be interesting for storing pictures, now image was encoded to base 64 in the qet.xml if user use so many pictures QET can lag when decode large base64 codes.
> 
> > Modification of the int BACKUP_INTERVAL from 2 min to 20 min used by KautoSaveFile. On a large project with a 256 MB folio printed in A0 format, the graphical interface freezes for 30 seconds when KautoSaveFile writes this large amount of data to the disk every two minutes."
> 
> See https://qelectrotech.org/wiki_new/refonte_du_code_de_qelectrotech, sorry is in French.

This is of course an interesting topic, but I think it should be discussed in QET-Forum or in another PR-draft.

Issue #492, opened 2026-06-11
Latent use-after-free in QETProject async backup (writeBackup / QtConcurrent)


ispyisail:
While building the headless CLI tools I hit an intermittent segfault (~1 in 6 runs) on any command that loads a project — including the already-merged `--resave`. Tracked it down to the async crash-recovery backup.

`QETProject::writeBackup()` (called on construction and on the 20-min timer) does:
```cpp
QtConcurrent::run(QET::writeToFile, xml_project, &m_backup_file, nullptr);
```
It's fire-and-forget — the `QFuture` isn't stored, and `~QETProject` doesn't wait for it. The worker thread writes through `&m_backup_file` (a project member), so if the `QETProject` is destroyed before the write finishes, the thread touches freed memory → crash in `QET::writeToFile` during teardown. Backtrace:
```
QObject::connect → QTextStream::QTextStream → QET::writeToFile → QtConcurrent::RunFunctionTask::run
```
In a short-lived CLI process this races almost every run; in the GUI it's far less likely but theoretically possible on rapid open→close.

**For the CLI** I've worked around it by disabling the backup in headless mode (`QETProject::setBackupEnabled(false)` from `main()`) — a crash-recovery backup is meaningless for a one-shot command anyway. That fully fixes the CLI crash (verified 0/40).

**For the general case**, if you want to close the underlying race, the usual fix would be to store the `QFuture` and `waitForFinished()` in `~QETProject` (or have the write own its data rather than reference `m_backup_file`). That touches GUI teardown timing though, so I left it to you rather than change that behaviour — happy to do it whichever way you prefer.

PR #514, opened 2026-06-20
Fix SIGSEGV: cache QStandardPaths results in dataDir/configDir for thread safety

ispyisail:
## Problem

`QStandardPaths::writableLocation()` is **not thread-safe in Qt5**. `ElementsCollectionModel::reload()` launches:

```cpp
m_future = QtConcurrent::map(m_items_list_to_setUp, setUpData);
```

Each worker thread calls:
```
FileElementCollectionItem::setUpData()
  → collectionPath() / isCollectionRoot()
  → QETApp::userMacrosDir()
  → QETApp::dataDir()
  → QStandardPaths::writableLocation(AppDataLocation)  ← SIGSEGV
```

Valgrind confirms the crash:
```
==1== Process terminating with default action of signal 11 (SIGSEGV)
==1==  Access not within mapped region at address 0x0
==1==    at 0x5854BB6: ??? (in libQt5Core.so.5.15.3)
==1==    by 0x59FDDD2: QStandardPaths::writableLocation(...)
==1==    by 0x2B7709: QETApp::dataDir() (qetapp.cpp:914)
==1==    by 0x2B746C: QETApp::userMacrosDir() (qetapp.cpp:862)
==1==    by 0x449343: FileElementCollectionItem::isCollectionRoot()
==1==    by 0x449A4A: FileElementCollectionItem::setUpData()
==1==    by 0x42E3E9: setUpData(ElementCollectionItem*)
==1==    by 0x436977: QtConcurrent::MapKernel::runIteration(...)
```

The app crashes reliably when opening the element panel after startup triggers the concurrent collection load.

## Fix

Replace the bare `QStandardPaths::writableLocation()` calls in `dataDir()` and `configDir()` with a **C++11 static-local lambda**. The compiler guarantees the lambda body executes exactly once across all threads (magic statics, ISO C++11 §6.7). The first call (always on the main thread, via the `QETApp` constructor) populates the cache; all subsequent calls — including those from `QtConcurrent` workers — return the cached value lock-free.

```cpp
// configDir():
static const QString configdir = []() {
    QString d = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation);
    while (d.endsWith('/')) d.chop(1);
    return d;
}();
return configdir;

// dataDir():
static const QString datadir = []() {
    QString d = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
    while (d.endsWith('/')) d.chop(1);
    return d;
}();
return datadir;
```

The `QET_ALLOW_OVERRIDE_DD_OPTION` / `QET_ALLOW_OVERRIDE_CD_OPTION` CLI override paths are unaffected — they return before reaching the static.

## Notes

- Same QtConcurrent lifetime/safety pattern as the use-after-free fixed in PR #512 / issue #492.
- The Qt6 code path in `ElementsCollectionModel` is a TODO stub and not affected.
- Tested on Ubuntu 22.04, Qt 5.15.3, verified with Valgrind — crash no longer occurs.
- Variable names updated to `configdir` / `datadir` per reviewer feedback to match the original code style.

### DIFF
--- a/sources/qetapp.cpp
+++ b/sources/qetapp.cpp
@@ -889,12 +889,12 @@ QString QETApp::configDir()
 #endif
 	// C++11 static-local init runs exactly once across all threads — safe to
 	// call from QtConcurrent background threads (QStandardPaths is not).
-	static const QString cached = []() {
+	static const QString configdir = []() {
 		QString d = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation);
 		while (d.endsWith('/')) d.chop(1);
 		return d;
 	}();
-	return cached;
+	return configdir;
 }
 
 /**
@@ -916,12 +916,12 @@ QString QETApp::dataDir()
 #endif
 	// C++11 static-local init runs exactly once across all threads — safe to
 	// call from QtConcurrent background threads (QStandardPaths is not).
-	static const QString cached = []() {
+	static const QString datadir = []() {
 		QString d = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
 		while (d.endsWith('/')) d.chop(1);
 		return d;
 	}();
-	return cached;
+	return datadir;
 }
 
 /**
--- a/sources/qetapp.cpp
+++ b/sources/qetapp.cpp
@@ -887,11 +887,14 @@ QString QETApp::configDir()
 #ifdef QET_ALLOW_OVERRIDE_CD_OPTION
 	if (config_dir != QString()) return(config_dir);
 #endif
-	QString configdir = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation);
-	while (configdir.endsWith('/')) {
-		configdir.remove(configdir.length()-1, 1);
-	}
-	return configdir;
+	// C++11 static-local init runs exactly once across all threads — safe to
+	// call from QtConcurrent background threads (QStandardPaths is not).
+	static const QString cached = []() {
+		QString d = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation);
+		while (d.endsWith('/')) d.chop(1);
+		return d;
+	}();
+	return cached;
 }
 
 /**
@@ -911,11 +914,14 @@ QString QETApp::dataDir()
 #ifdef QET_ALLOW_OVERRIDE_DD_OPTION
 	if (data_dir != QString()) return(data_dir);
 #endif
-	QString datadir = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
-	while (datadir.endsWith('/')) {
-		datadir.remove(datadir.length()-1, 1);
-	}
-	return datadir;
+	// C++11 static-local init runs exactly once across all threads — safe to
+	// call from QtConcurrent background threads (QStandardPaths is not).
+	static const QString cached = []() {
+		QString d = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
+		while (d.endsWith('/')) d.chop(1);
+		return d;
+	}();
+	return cached;
 }
 
 /**


ispyisail:
## Verification — Valgrind memcheck (Debug build)

Rebuilt a Debug Docker image with the fix applied and ran Valgrind memcheck to confirm the SIGSEGV is eliminated.

**Before fix:** Valgrind reported a null-pointer dereference inside `QStandardPaths::writableLocation()` called from a `QtConcurrent::map()` worker thread — the `QStandardPaths` static cache was being initialised concurrently from multiple threads, racing on the underlying `QStandardPaths` global state.

**After fix:** No SIGSEGV. The static-local lambda in `dataDir()` and `configDir()` ensures the result is computed exactly once on whichever thread first reaches it (C++11 §6.7 magic-static guarantee), and all subsequent callers (including background threads) read the already-cached `QString` — no Qt global state touched after that point.

A second Valgrind run combining this fix with PR #515 also confirmed no regression — both fix sets co-exist cleanly.

ispyisail:
## What this fix does (plain English)

When QElectroTech starts up, it runs some tasks in the background at the same time as the main window is loading. One of those background tasks was asking the operating system "where should I save my data files?" — but Qt has a rule that this question can only be asked safely from the main thread. Asking it from a background thread at the same time as the main thread is doing other things caused a crash (the app tried to read memory that wasn't ready yet).

**The fix:** We now ask that question once, right at startup on the main thread, and cache the answer. After that, every part of the app — background threads included — just reads the saved answer instead of asking Qt again.

---

## Verification

After applying the fix, we rebuilt the app in Debug mode and ran it under **Valgrind** (a tool that watches every memory access and flags anything unsafe).

- **Before the fix:** Valgrind reported a crash caused by a background thread touching Qt memory it wasn't allowed to touch.
- **After the fix:** No crash. Valgrind confirms the unsafe memory access is gone. A second combined run with PR #515 also showed no new problems introduced.

ispyisail:
## Explication en français

Au démarrage de QElectroTech, certaines tâches s'exécutent en arrière-plan en même temps que le chargement de la fenêtre principale. L'une de ces tâches demandait au système d'exploitation « où dois-je enregistrer mes fichiers de données ? » — mais Qt impose une règle : cette question ne peut être posée en toute sécurité que depuis le fil d'exécution principal (le « main thread »). La poser depuis un fil d'exécution secondaire, en même temps que le fil principal effectue d'autres opérations, provoquait un plantage : l'application tentait de lire une zone mémoire qui n'était pas encore prête.

**La correction :** On pose cette question une seule fois, dès le démarrage et depuis le fil principal, puis on mémorise la réponse. Ensuite, toutes les parties de l'application — y compris les fils d'exécution en arrière-plan — lisent simplement la réponse déjà enregistrée, sans interroger Qt à nouveau.

---

## Vérification

Après application du correctif, nous avons recompilé l'application en mode Debug et l'avons exécutée sous **Valgrind** (un outil qui surveille chaque accès mémoire et signale tout accès non sécurisé).

- **Avant le correctif :** Valgrind signalait un plantage causé par un fil d'exécution secondaire accédant à une zone mémoire Qt à laquelle il n'avait pas le droit de toucher.
- **Après le correctif :** Plus aucun plantage. Valgrind confirme que l'accès mémoire non sécurisé a disparu. Une deuxième exécution combinée avec la PR #515 n'a révélé aucun nouveau problème.

sources/qetapp.cpp
```diff
@@ -887,11 +887,14 @@ QString QETApp::configDir()
...
 	if (config_dir != QString()) return(config_dir);
 #endif
-	QString configdir = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation);
-	while (configdir.endsWith('/')) {
-		configdir.remove(configdir.length()-1, 1);
-	}
-	return configdir;
+	// C++11 static-local init runs exactly once across all threads — safe to
+	// call from QtConcurrent background threads (QStandardPaths is not).
+	static const QString cached = []() {
```
plc-user: It's just a small thing, but for the look:
I'd prefer to see the original variable names here instead of “cached.”

sources/qetapp.cpp
```diff
@@ -911,11 +914,14 @@ QString QETApp::dataDir()
...
 	if (data_dir != QString()) return(data_dir);
 #endif
-	QString datadir = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
-	while (datadir.endsWith('/')) {
-		datadir.remove(datadir.length()-1, 1);
-	}
-	return datadir;
+	// C++11 static-local init runs exactly once across all threads — safe to
+	// call from QtConcurrent background threads (QStandardPaths is not).
+	static const QString cached = []() {
```
plc-user: It's just a small thing, but for the look:
I'd prefer to see the original variable names here instead of “cached.”

sources/qetapp.cpp
ispyisail: Good catch — updated to use the original variable name `configdir` to match the existing code style. Thank you!

sources/qetapp.cpp
ispyisail: Good catch — updated to use the original variable name `datadir` to match the existing code style. Thank you!

PR #515, opened 2026-06-20
Fix three uninitialised-value bugs found by Valgrind

ispyisail:
Follow-up to PR #514. Three more Valgrind findings from running `--tool=memcheck` on Ubuntu 22.04 / Qt 5.15.3.

## 1. `machine_info.h` — uninitialised `Screen` struct members

`Max_width`, `Max_height`, `count`, `width[]` and `height[]` were bare `int32_t` with no initialiser. The comparisons in `init_get_Screen_info()` read them before any write:

```
==1== Conditional jump or move depends on uninitialised value(s)
==1==    at 0x2A236C: MachineInfo::init_get_Screen_info() (machine_info.cpp:271)
==1==  Uninitialised value was created by a heap allocation
==1==    at 0x4849013: operator new(unsigned long)
==1==    by 0x21D568: MachineInfo::instance() (machine_info.h:39)
```

**Fix:** Add `= 0` / `= {}` in-class initialisers to all `Screen` members.

## 2. `main.cpp` — `MachineInfo` singleton constructed on background thread

`MachineInfo::instance()` was first called inside `QtConcurrent::run()`, so its constructor (which calls `qApp->screens()`) ran on a background thread. `QScreen` methods are not thread-safe in Qt5:

```
==1== Conditional jump or move depends on uninitialised value(s)
==1==    at 0x2A236C: MachineInfo::init_get_Screen_info() (machine_info.cpp:271)
==1==    by 0x2A67B9: main::{lambda()#1}::operator()() (main.cpp:251)
==1==    by 0x2A70A7: QtConcurrent::StoredFunctorCall0<void,...>::runFunctor()
```

**Fix:** Call `MachineInfo::instance()` once on the main thread before `QtConcurrent::run()`. The singleton is then fully built; the worker just returns the cached pointer.

## 3. `qetdiagrameditor.h` — `m_first_show` read before initialisation

C++ initialises members in declaration order. `m_first_show` was declared at line 256, after the `QActionGroup` members at line 168. During construction of `m_row_column_actions_group(this)`, Qt dispatches a `QObject` parent-change event that reaches `QETDiagramEditor::event()`, which reads `m_first_show` before it has been initialised:

```
==1== Conditional jump or move depends on uninitialised value(s)
==1==    at 0x2D93AB: QETDiagramEditor::event(QEvent*) (qetdiagrameditor.cpp:928)
==1==    by 0x4B7D712: QApplicationPrivate::notify_helper(QObject*, QEvent*)
==1==    by 0x4B7AD66: QActionGroup::QActionGroup(QObject*)
==1==    by 0x2CEDA4: QETDiagramEditor::QETDiagramEditor(...) (qetdiagrameditor.cpp:64)
==1==  Uninitialised value was created by a heap allocation
==1==    by 0x2B43C8: QETApp::QETApp() (qetapp.cpp:142)
```

**Fix:** Move `m_first_show` to the top of the first `private:` block so it is initialised before any member that can trigger events.

### DIFF
--- a/sources/qetapp.cpp
+++ b/sources/qetapp.cpp
@@ -889,12 +889,12 @@ QString QETApp::configDir()
 #endif
 	// C++11 static-local init runs exactly once across all threads — safe to
 	// call from QtConcurrent background threads (QStandardPaths is not).
-	static const QString cached = []() {
+	static const QString configdir = []() {
 		QString d = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation);
 		while (d.endsWith('/')) d.chop(1);
 		return d;
 	}();
-	return cached;
+	return configdir;
 }
 
 /**
@@ -916,12 +916,12 @@ QString QETApp::dataDir()
 #endif
 	// C++11 static-local init runs exactly once across all threads — safe to
 	// call from QtConcurrent background threads (QStandardPaths is not).
-	static const QString cached = []() {
+	static const QString datadir = []() {
 		QString d = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
 		while (d.endsWith('/')) d.chop(1);
 		return d;
 	}();
-	return cached;
+	return datadir;
 }
 
 /**
--- a/sources/qetapp.cpp
+++ b/sources/qetapp.cpp
@@ -887,11 +887,14 @@ QString QETApp::configDir()
 #ifdef QET_ALLOW_OVERRIDE_CD_OPTION
 	if (config_dir != QString()) return(config_dir);
 #endif
-	QString configdir = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation);
-	while (configdir.endsWith('/')) {
-		configdir.remove(configdir.length()-1, 1);
-	}
-	return configdir;
+	// C++11 static-local init runs exactly once across all threads — safe to
+	// call from QtConcurrent background threads (QStandardPaths is not).
+	static const QString cached = []() {
+		QString d = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation);
+		while (d.endsWith('/')) d.chop(1);
+		return d;
+	}();
+	return cached;
 }
 
 /**
@@ -911,11 +914,14 @@ QString QETApp::dataDir()
 #ifdef QET_ALLOW_OVERRIDE_DD_OPTION
 	if (data_dir != QString()) return(data_dir);
 #endif
-	QString datadir = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
-	while (datadir.endsWith('/')) {
-		datadir.remove(datadir.length()-1, 1);
-	}
-	return datadir;
+	// C++11 static-local init runs exactly once across all threads — safe to
+	// call from QtConcurrent background threads (QStandardPaths is not).
+	static const QString cached = []() {
+		QString d = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
+		while (d.endsWith('/')) d.chop(1);
+		return d;
+	}();
+	return cached;
 }
 
 /**
--- a/sources/machine_info.h
+++ b/sources/machine_info.h
@@ -75,11 +75,11 @@ class MachineInfo
 	{
 		struct Screen
 		{
-			int32_t count;
-			int32_t width[10];
-			int32_t height[10];
-			int32_t Max_width;
-			int32_t Max_height;
+			int32_t count     = 0;
+			int32_t width[10] = {};
+			int32_t height[10]= {};
+			int32_t Max_width = 0;
+			int32_t Max_height= 0;
 		}screen;
 		struct Built
 		{
--- a/sources/main.cpp
+++ b/sources/main.cpp
@@ -241,6 +241,11 @@ QGuiApplication::setHighDpiScaleFactorRoundingPolicy(QetSettings::hdpiScaleFacto
 	QObject::connect(&app, &SingleApplication::receivedMessage,
 			 &qetapp, &QETApp::receiveMessage);
 
+	// Pre-initialise on the main (GUI) thread: the constructor calls
+	// qApp->screens() which is not thread-safe in Qt5 — calling instance()
+	// here guarantees the singleton is fully built before the worker runs.
+	MachineInfo::instance();
+
 	QtConcurrent::run([=]()
 	{
 		// for debugging
--- a/sources/qetdiagrameditor.h
+++ b/sources/qetdiagrameditor.h
@@ -80,6 +80,11 @@ class QETDiagramEditor : public QETMainWindow
 	  protected:
 		bool event(QEvent *) override;
 	private:
+		// Declared first so it is initialised before any member whose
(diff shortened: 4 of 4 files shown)


ispyisail:
## Verification — Valgrind memcheck (Debug build, all fixes applied)

Rebuilt a Debug Docker image with all three fixes applied (alongside PR #514) and ran Valgrind memcheck to confirm the targeted warnings are eliminated.

**Results:**

| Fix | Valgrind pattern checked | Result |
|---|---|---|
| `machine_info.h` — `Screen` struct zero-init | `init_get_Screen_info` in any error stack | **Absent — fixed** |
| `main.cpp` — pre-init `MachineInfo` on main thread | `MachineInfo::instance()` called from non-main thread | **Absent — fixed** |
| `qetdiagrameditor.h` — `m_first_show` declaration order | `m_first_show` / `event()` uninitialised read | **Absent — fixed** |

**Remaining Valgrind noise (not QElectroTech bugs):**
- `libdrm_nouveau` / Mesa GPU driver uninitialised-byte warnings — third-party, not fixable here
- `waitid(NULL)` flagged inside Qt 5.15's `QProcess::waitForFinished()` — known Qt internal behaviour
- "Still reachable" memory at exit — normal for Qt apps (Qt does not destroy all globals at shutdown)

No SIGSEGV (PR #514 fix still holds). All three `Uninitialised value` warnings from the original Valgrind report are confirmed eliminated.

ispyisail:
## What these fixes do (plain English)

**Fix 1 — Screen information had no starting value**

When QElectroTech collects information about your monitors at startup, several variables (screen count, width, height) were created empty — they had no "starting value" of zero. If anything read those variables before they were properly filled in, it would get random garbage data from memory. We added `= 0` starting values so they always begin at zero, which is what you'd expect.

**Fix 2 — Monitor information was being collected in the wrong place**

The code that reads your screen size and number of monitors was running in a background thread. Qt has a rule that anything to do with screens and displays must only run on the main thread. Running it in the background could cause crashes or incorrect values on some machines. We moved that one call so it runs on the main thread first, before the background work starts.

**Fix 3 — A true/false flag was used before it was set up**

C++ initialises class members in the order they are listed in the header file. There was a flag called `m_first_show` (used to track whether the main window has been shown for the first time) that was listed near the bottom of the class, after some other members whose setup triggered a window event. That event checked the flag before it had been given its starting value of `true`. Moving the flag to the top of the list means it is always ready before anything can ask about it.

---

## Verification

We rebuilt the app in Debug mode and ran it under **Valgrind** (a tool that watches every memory access and flags anything unsafe).

| Fix | Was the warning still there after the fix? |
|---|---|
| Screen variables starting value | No — confirmed gone |
| Screen info on wrong thread | No — confirmed gone |
| `m_first_show` used too early | No — confirmed gone |

The only warnings left in Valgrind come from the graphics driver (Nouveau/Mesa) and Qt's own internals — nothing that can be fixed in QElectroTech itself.

ispyisail:
## Explication en français

**Correction 1 — Les informations d'écran n'avaient pas de valeur initiale**

Au démarrage, QElectroTech recueille des informations sur vos moniteurs (nombre d'écrans, largeur, hauteur). Plusieurs variables étaient créées sans valeur de départ — elles contenaient des données aléatoires issues de la mémoire. Si quelque chose lisait ces variables avant qu'elles soient correctement renseignées, il obtenait des valeurs incohérentes. Nous avons ajouté une valeur initiale de `0` afin qu'elles commencent toujours à zéro, ce qui est le comportement attendu.

**Correction 2 — Les informations sur les écrans étaient collectées au mauvais endroit**

Le code qui lit la taille et le nombre de vos écrans s'exécutait dans un fil d'exécution secondaire (arrière-plan). Qt impose une règle : tout ce qui concerne les écrans et l'affichage doit s'exécuter uniquement depuis le fil principal. L'exécuter en arrière-plan pouvait provoquer des plantages ou des valeurs incorrectes sur certaines machines. Nous avons déplacé cet appel pour qu'il s'exécute sur le fil principal, avant le démarrage des tâches en arrière-plan.

**Correction 3 — Un indicateur vrai/faux était utilisé avant d'être initialisé**

En C++, les membres d'une classe sont initialisés dans l'ordre où ils sont déclarés dans le fichier d'en-tête. Un indicateur appelé `m_first_show` (qui permet de savoir si la fenêtre principale a déjà été affichée) était déclaré en bas de la classe, après d'autres membres dont l'initialisation déclenchait un événement de fenêtre. Cet événement consultait l'indicateur avant qu'il ait reçu sa valeur de départ (`true`). En déplaçant sa déclaration en haut de la classe, on s'assure qu'il est toujours prêt avant que quoi que ce soit puisse l'interroger.

---

## Vérification

Nous avons recompilé l'application en mode Debug et l'avons exécutée sous **Valgrind** (un outil qui surveille chaque accès mémoire et signale tout accès non sécurisé).

| Correction | L'avertissement était-il encore présent après le correctif ? |
|---|---|
| Valeur initiale des variables d'écran | Non — confirmé disparu |
| Informations d'écran sur le mauvais fil | Non — confirmé disparu |
| `m_first_show` utilisé trop tôt | Non — confirmé disparu |

Les seuls avertissements restants dans Valgrind proviennent du pilote graphique (Nouveau/Mesa) et des composants internes de Qt — rien qui puisse être corrigé dans QElectroTech lui-même.

sources/qetapp.cpp
```diff
@@ -887,11 +887,14 @@ QString QETApp::configDir()
...
 	if (config_dir != QString()) return(config_dir);
 #endif
-	QString configdir = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation);
-	while (configdir.endsWith('/')) {
-		configdir.remove(configdir.length()-1, 1);
-	}
-	return configdir;
+	// C++11 static-local init runs exactly once across all threads — safe to
+	// call from QtConcurrent background threads (QStandardPaths is not).
+	static const QString cached = []() {
```
plc-user: It's just a small thing, but for the look:
I'd prefer to see the original variable names here instead of “cached.”

sources/qetapp.cpp
```diff
@@ -911,11 +914,14 @@ QString QETApp::dataDir()
...
 	if (data_dir != QString()) return(data_dir);
 #endif
-	QString datadir = QStandardPaths::writableLocation(QStandardPaths::AppDataLocation);
-	while (datadir.endsWith('/')) {
-		datadir.remove(datadir.length()-1, 1);
-	}
-	return datadir;
+	// C++11 static-local init runs exactly once across all threads — safe to
+	// call from QtConcurrent background threads (QStandardPaths is not).
+	static const QString cached = []() {
```
plc-user: It's just a small thing, but for the look:
I'd prefer to see the original variable names here instead of “cached.”

sources/qetapp.cpp
ispyisail: Fixed — rebased onto the updated PR #514 branch which uses `configdir` throughout. Thanks for the consistent feedback across both PRs!

sources/qetapp.cpp
ispyisail: Fixed — rebased onto the updated PR #514 branch which uses `datadir` throughout. Thanks for the consistent feedback across both PRs!

PR #516, opened 2026-06-20
Fix data race in setUpData() and duplicate static QString constants in header

ispyisail:
## Summary

Two original bugs found via Valgrind memcheck + static analysis, plus two additional bugs found via ThreadSanitizer (TSan):

### Bug 1 — Data race in `setUpData()` (FileElementCollectionItem & XmlProjectElementCollectionItem)

`ElementsCollectionModel::loadCollections()` uses `QtConcurrent::map()` to call `setUpData()` on items **already inserted into the live `QStandardItemModel`**. Inside `setUpData()`, calls to `setText()`, `setFlags()`, `setData()`, and `setToolTip()` all trigger `dataChanged` signals through the model — these are **not safe from background threads** and can race with the main thread reading the model to paint the view.

The comment in `loadMacrosCollection()` already acknowledges the problem (*"Load the macros collection synchronously to avoid thread-collisions"*) but the fix was only applied to macros.

**Fix:** Add `computeDisplayName()` — a pure-computation helper that reads files/XML but never touches the `QStandardItem` API. Restructure `setUpData()` so:
- Expensive I/O runs in the worker thread (unchanged parallelism)
- `setText` / `setFlags` / `setData` / `setToolTip` are dispatched to the main thread via `QMetaObject::invokeMethod(..., Qt::BlockingQueuedConnection)`

### Bug 2 — `static QString` constants duplicated per translation unit (`qetinformation.h`)

The 80 `static QString` variables in `qetinformation.h` give **every `.cpp` that includes the header its own private copy** of the string data (C++ internal-linkage rule). This wastes memory proportional to the number of translation units and causes Valgrind to report them as "possibly lost" at shutdown.

**Fix:** Change to `inline const QString` with `QStringLiteral` (C++17). All translation units share one definition; strings are initialised at compile time.

### Bug 3 — Silent deadlock when `setUpData()` is called from the GUI thread

`loadMacrosCollection()` calls `addMacrosCollection(true)` → `setUpData()` directly from the **main thread**. The `BlockingQueuedConnection` invoke in Bug 1's fix deadlocks when the caller is already the target thread — Qt prints `QMetaObject::invokeMethod: Dead lock detected` and silently skips the invoke, leaving macros tree items with empty display names, wrong flags, and no search data.

**Fix:** Check `QThread::currentThread() == m->thread()` before invoking — call directly if already on the model's thread, use `BlockingQueuedConnection` only from background threads. Applied to both `FileElementCollectionItem::setUpData()` and `XmlProjectElementCollectionItem::setUpData()`.

### Bug 4 — `loadMacrosCollection()` modifies the model while concurrent map is still running

`loadMacrosCollection()` called `appendRow()` on the model without waiting for the in-flight `QtConcurrent::map` (from a previous reload) to finish. Background threads were still iterating model internals via `setUpData()` while the main thread modified the child list — a TSan-confirmed data race that can corrupt model state.

**Fix:** Call `m_future.waitForFinished()` at the top of `loadMacrosCollection()` before touching the model.

## Test plan
- [x] Builds cleanly (Debug + all prior patches applied)
- [x] No C++ compiler errors or warnings on changed files
- [x] Run under ThreadSanitizer — deadlock warning gone, collection loads cleanly in ~2.7 s
- [x] Remaining TSan warnings are Qt-internal false positives (futex synchronisation invisible to TSan)

### DIFF
--- a/sources/ElementsCollection/elementscollectionmodel.cpp
+++ b/sources/ElementsCollection/elementscollectionmodel.cpp
@@ -311,6 +311,12 @@ void ElementsCollectionModel::loadCollections(bool common_collection,
  */
 void ElementsCollectionModel::loadMacrosCollection()
 {
+	// Wait for any in-flight concurrent setUpData map to finish before
+	// modifying the model — appendRow() races with background threads
+	// that are still reading model internals via setUpData().
+	if (!m_future.isFinished())
+		m_future.waitForFinished();
+
 	m_items_list_to_setUp.clear();
 	addMacrosCollection(true);
 }
--- a/sources/ElementsCollection/fileelementcollectionitem.cpp
+++ b/sources/ElementsCollection/fileelementcollectionitem.cpp
@@ -23,6 +23,8 @@
 #include "elementslocation.h"
 
 #include <QDir>
+#include <QMetaObject>
+#include <QStandardItemModel>
 
 /**
 	@brief FileElementCollectionItem::FileElementCollectionItem
@@ -110,6 +112,47 @@ bool FileElementCollectionItem::isElement() const
 }
 
 /**
+ * @brief FileElementCollectionItem::computeDisplayName
+ * Compute the display name without calling setText() — safe to call from
+ * any thread.  localName() and setUpData() both delegate to this.
+ */
+QString FileElementCollectionItem::computeDisplayName() const
+{
+	if (isCollectionRoot()) {
+		QString macrosPath = QETApp::userMacrosDir();
+		if (macrosPath.endsWith('/')) macrosPath.chop(1);
+
+		if (m_path == QETApp::commonElementsDirN())
+			return QObject::tr("Collection QET");
+		if (m_path == QETApp::companyElementsDirN())
+			return QObject::tr("Collection Company");
+		if (m_path == QETApp::customElementsDirN())
+			return QObject::tr("Collection utilisateur");
+		if (m_path == macrosPath)
+			return QObject::tr("Makros");
+		return QObject::tr("Collection inconnue");
+	}
+	if (isDir()) {
+		const QString str = fileSystemPath() % "/qet_directory";
+		pugi::xml_document docu;
+		if (docu.load_file(str.toStdString().c_str())) {
+			if (QString(docu.document_element().name()) == "qet-directory") {
+				NamesList nl;
+				nl.fromXml(docu.document_element());
+				return nl.name();
+			}
+		}
+		return {};
+	}
+	// Element
+	ElementsLocation loc(collectionPath());
+	QString name = loc.name();
+	if (name.endsWith(".qetmak"))
+		name.remove(".qetmak");
+	return name;
+}
+
+/**
  * @brief FileElementCollectionItem::localName
  * @return the located name of this item
  */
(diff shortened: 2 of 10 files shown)


ispyisail:
## What these fixes do (plain English)

**Fix 1 — Background threads were changing the elements panel while it was being displayed**

When QElectroTech loads its elements collection (the panel on the left showing all the symbols), it uses several background threads to load them in parallel for speed. However, those background threads were also directly updating the display — setting the name, icon flags, and tooltip on each element entry at the same time as the main window was drawing them.

Qt has a strict rule: anything that changes what is shown on screen must only happen from the main thread. Doing it from a background thread can cause random crashes or display corruption, and the existing code even had a comment admitting the problem: *"Load the macros collection synchronously to avoid thread-collisions"* — but only the macros collection was fixed.

The fix keeps the slow part (reading files from disk and parsing XML) running in the background threads as before, but routes the screen-update calls back to the main thread so they arrive safely.

**Fix 2 — The same string was being stored dozens of times in memory**

There is a header file (`qetinformation.h`) that defines 80 text constants — things like the names of element properties (`"label"`, `"description"`, `"manufacturer"`, etc.). Because of a C++ rule, every source file that includes this header gets its own private copy of all 80 strings, meaning they can be stored 30–40 times in memory at once.

Changing `static` to `inline const` (a C++17 feature) tells the compiler to share one copy across all source files. The strings are also now created at compile time rather than at program startup, which is slightly faster.

---

## Verification

- Builds cleanly in Debug mode with no compiler errors or warnings on the changed files
- Running under Valgrind to confirm the threading warnings are eliminated (in progress)

ispyisail:
## Explication en français

**Correction 1 — Des fils d'exécution en arrière-plan modifiaient le panneau des éléments pendant son affichage**

Lorsque QElectroTech charge sa collection d'éléments (le panneau à gauche affichant tous les symboles), il utilise plusieurs fils d'exécution secondaires pour les charger en parallèle, ce qui est plus rapide. Cependant, ces fils d'exécution modifiaient aussi directement l'affichage — en définissant le nom, les indicateurs et l'infobulle de chaque entrée — en même temps que la fenêtre principale les dessinait.

Qt impose une règle stricte : tout ce qui modifie ce qui est affiché à l'écran doit uniquement se faire depuis le fil principal. Le faire depuis un fil secondaire peut provoquer des plantages aléatoires ou des corruptions d'affichage. Le code existant contenait d'ailleurs un commentaire reconnaissant le problème : *« Charger la collection de macros de manière synchrone pour éviter les collisions de fils »* — mais seule la collection de macros avait été corrigée.

La correction conserve la partie lente (lecture des fichiers sur le disque et analyse XML) dans les fils secondaires comme auparavant, mais renvoie les appels de mise à jour de l'affichage vers le fil principal, afin qu'ils arrivent en toute sécurité.

**Correction 2 — La même chaîne de caractères était stockée des dizaines de fois en mémoire**

Un fichier d'en-tête (`qetinformation.h`) définit 80 constantes textuelles — les noms des propriétés des éléments comme `"label"`, `"description"`, `"manufacturer"`, etc. En raison d'une règle du C++, chaque fichier source qui inclut cet en-tête obtient sa propre copie privée des 80 chaînes, ce qui signifie qu'elles peuvent être stockées 30 à 40 fois simultanément en mémoire.

Changer `static` en `inline const` (une fonctionnalité C++17) indique au compilateur de partager une seule copie entre tous les fichiers source. Les chaînes sont désormais créées à la compilation plutôt qu'au démarrage du programme, ce qui est légèrement plus rapide.

---

## Vérification

- Compilation propre en mode Debug sans erreurs ni avertissements sur les fichiers modifiés
- Exécution sous Valgrind pour confirmer l'élimination des avertissements de threading (en cours)

scorpio810:
This PR fixes a genuine race condition — background threads were touching the Qt model while the main thread was painting it — plus cleans up some duplicated string constants in memory. The fix itself is sound on the Qt side.

The catch: it uses BlockingQueuedConnection, which blocks the worker thread while the main thread handles the call. That could re-serialize part of the loading that used to run in parallel. And the only performance number given was measured under ThreadSanitizer, which slows everything down anyway — so it tells us nothing about actual startup time.

Nobody's tested it, no reviews yet. Before merging, someone just needs to measure load time before/after on a large collection in a normal release build. Otherwise we're merging blind on a critical startup path.

ispyisail:

PR #519, opened 2026-06-20
Fix four memory leaks found by AddressSanitizer

ispyisail:
## Summary

Four distinct memory leaks identified and fixed using AddressSanitizer (ASan). Together they eliminate **536 KB** of leaked memory per session (603 KB → 67 KB total after fix; residual is third-party: libdbus, libfontconfig).

### StyleEditor — orphaned `QVBoxLayout` (`sources/editor/styleeditor.cpp`)

`new QGridLayout(this)` internally calls `setLayout()` on the StyleEditor widget, claiming the layout slot. The subsequent explicit `setLayout(main_layout)` silently fails — Qt's `setLayout` returns early if a layout is already installed — permanently orphaning `main_layout`. Every editor type that embeds a `StyleEditor` (Arc, Line, Ellipse, Polygon, Rectangle) leaked one instance per element editor window opened.

**Fix:** `new QGridLayout()` with no parent argument, so `setLayout(main_layout)` succeeds and `main_layout` is properly owned by the widget.

### ExportDialog — `ExportDiagramLine` objects never freed (`sources/exportdialog.cpp`)

`diagram_lines_` holds `ExportDiagramLine*` heap objects. The QWidget members inside each line are parented to the dialog and cleaned up via Qt's parent-child chain, but the `ExportDiagramLine` C++ wrapper structs themselves were never deleted. `~ExportDialog()` was empty.

**Fix:** `qDeleteAll(diagram_lines_)` in the destructor.

### GenericPanel — orphaned `QTreeWidgetItem` (`sources/genericpanel.cpp`)

`getItemForDiagram(diagram)` called without the `bool* created` parameter would still call `makeItem()` and allocate a parentless `QTreeWidgetItem` when the diagram wasn't in the panel cache. Callers (e.g. in `qetdiagrameditor.cpp`) used the returned item only for `setCurrentItem()` then discarded it — the item was never added to any tree and never freed.

**Fix:** Early-return `nullptr` when `created == nullptr` and the item isn't found in the cache. All call sites already guard with `if (item)`.

### ElementScene — paste area never freed (`sources/editor/elementscene.cpp`)

`m_paste_area` (a `QGraphicsRectItem`) is created in `initPasteArea()` and temporarily added/removed from the scene during XML paste operations. At destruction time it is not in the scene, so `~QGraphicsScene()` does not delete it, and `~ElementScene()` did not either.

**Fix:** `delete m_paste_area` in `~ElementScene()`, guarded by `!m_paste_area->scene()` to avoid a double-free if it happens to still be in the scene.

## Verification

Built and run under ASan (`-fsanitize=address -g -O1 -fno-omit-frame-pointer`) in an isolated Docker container on Ubuntu 22.04 / Qt 5:

| | Direct leaks (QET source) | Total leaked |
|---|---|---|
| Before | Multiple (StyleEditor ×5, ExportDiagramLine, QTreeWidgetItem, m_paste_area) | 603,544 bytes |
| After | **0** | 67,448 bytes (libdbus + libfontconfig only) |

### DIFF
--- a/sources/editor/elementscene.cpp
+++ b/sources/editor/elementscene.cpp
@@ -107,6 +107,9 @@ ElementScene::~ElementScene()
 
 	if (m_decorator)
 		delete m_decorator;
+
+	if (m_paste_area && !m_paste_area->scene())
+		delete m_paste_area;
 }
 
 /**
--- a/sources/exportdialog.cpp
+++ b/sources/exportdialog.cpp
@@ -112,6 +112,7 @@ ExportDialog::ExportDialog(
 */
 ExportDialog::~ExportDialog()
 {
+	qDeleteAll(diagram_lines_);
 }
 
 /**
--- a/sources/genericpanel.cpp
+++ b/sources/genericpanel.cpp
@@ -321,9 +321,11 @@ QTreeWidgetItem *GenericPanel::getItemForDiagram(Diagram *diagram,
 		if (created) *created = false;
 		return(diagram_qtwi);
 	}
-	
+
+	if (!created) return(nullptr);
+
 	diagram_qtwi = makeItem(QET::Diagram);
-	if (created) *created = true;
+	*created = true;
 	return(diagram_qtwi);
 }
 
--- a/sources/editor/styleeditor.cpp
+++ b/sources/editor/styleeditor.cpp
@@ -386,7 +386,7 @@ StyleEditor::StyleEditor(QETElementEditor *editor, CustomElementGraphicPart *p, 
 
 	outline_color->setSizeAdjustPolicy(QComboBox::AdjustToContents);
 	filling_color->setSizeAdjustPolicy(QComboBox::AdjustToContents);
-	auto grid_layout = new QGridLayout(this);
+	auto grid_layout = new QGridLayout();
 	grid_layout->addWidget(new QLabel(tr("Contour :")), 0,0, Qt::AlignRight);
 	grid_layout->addWidget(outline_color, 0, 1);
 	grid_layout->addWidget(new QLabel(tr("Remplissage :")), 1, 0, Qt::AlignRight);


scorpio810:
Tested this under ASan (`qmake CONFIG+=debug QMAKE_CXXFLAGS+="-fsanitize=address -g -O1 -fno-omit-frame-pointer" QMAKE_LFLAGS+="-fsanitize=address"`). My test scenario only opened a project file (57 diagrams) and closed the app, which exercises the parsing/loading path but not `StyleEditor`, `ExportDialog`, or the paste-area in `ElementScene`, so I can't confirm those three leak fixes yet from this run — I'll redo it with a scenario that opens an element editor, exports a diagram, and does an XML paste.

That said, the run did catch something else: a separate, unrelated leak of about 700 KB, almost entirely from `ElementPictureFactory::parseElement` and `parseText` (`elementpicturefactory.cpp` lines 258, 503, 545), called from `Element::buildFromXml` during diagram loading. It looks like `QGraphicsSimpleTextItem` instances and their associated `QFont`/`QPen`/`QBrush` are allocated to render element previews but never freed — the leak scales with the number of elements in the diagram. Worth a separate look, but flagging here since it showed up in the same ASan pass.

scorpio810:
Manual test checklist to exercise the four fixed paths under ASan (since opening a project alone, as in my previous comment, doesn't cover any of them):

**1. `StyleEditor` (Arc, Line, Ellipse, Polygon, Rectangle)**
- Open the element editor (right-click an existing element > Edit, or create a new element).
- For each primitive type covered by the PR — Arc, Line, Ellipse, Polygon, Rectangle — select the corresponding tool in the editor toolbar, which brings up the style panel (`StyleEditor`) in the right-hand panel.
- Just clicking each primitive once is enough to instantiate a `StyleEditor`; no need to actually draw anything on the canvas.
- Ideally, open/close several element editor windows in a row (5-6 times) to amplify the leak if the fix didn't work.
- Close the element editor (not just minimize it).

**2. `ExportDialog`**
- From the main diagram editor (not the element editor), `File > Export`.
- The dialog should show a list of rows (`ExportDiagramLine`) — one per diagram in the project.
- No need to go through with the actual export; opening the dialog and hitting Cancel is enough to instantiate and destroy the `ExportDiagramLine` objects.
- Repeat 2-3 times (open/cancel) to amplify.

**3. `GenericPanel` (`getItemForDiagram`)**
- Already partly covered by the initial test via `diagramActivated`/`syncElementsPanel` (`qetdiagrameditor.cpp` lines ~1857/1870), so it's enough to:
  - Click on several different diagrams in the left-hand tree to trigger `setCurrentItem` multiple times.
  - Switch project tabs if several projects are open, to force `syncElementsPanel`.
- Testing on a freshly opened project (cold cache) is more representative than one already browsed through, since the fix concerns the case where the item isn't cached yet.

**4. `ElementScene` (`m_paste_area`)**
- In the element editor, select one or more primitives drawn on the canvas.
- Copy (Ctrl+C), then paste (Ctrl+V) — this triggers the XML paste path that initializes `m_paste_area` via `initPasteArea()`.
- Close the element editor afterward (the leak only shows up when `ElementScene` is destroyed).
- Repeat the copy-paste 2-3 times before closing, to amplify.

After these four steps, quit the application cleanly (not a kill) to trigger the LeakSanitizer report, then check for:

`grep -A3 "styleeditor\|exportdialog\|genericpanel\|elementscene" asan_output.txt`

If the patch is correct, none of these four classes should show up in the final report — only the known third-party leaks (libfontconfig, libwayland-client) and the already-flagged `ElementPictureFactory` leak should remain.

scorpio810:
Update: ran through the full manual checklist above on the patched branch, on top of the ASan build. Summary:

| Fix | Action performed | Result |
|---|---|---|
| `StyleEditor` | Drawing tool (Arc/Line/Ellipse/Polygon/Rectangle) selected in the element editor — confirmed `PartPolygon` constructed in the trace | No `StyleEditor` leak |
| `ExportDialog` | Real export of 55 diagrams | No `ExportDiagramLine` leak |
| `GenericPanel` (`getItemForDiagram`) | Several recent-file openings exercising this path | No `GenericPanel` leak |
| `ElementScene` (`m_paste_area`) | Copy/paste of a primitive inside the element editor | No `ElementScene`/`m_paste_area` leak |

No crashes across any of these runs (I did hit one crash early on while exploring the element editor, but it didn't reproduce on later attempts and doesn't show up as an ASan error in any of the logs, so it looks unrelated to this patch).

This is in addition to the separate `ElementPictureFactory` leak flagged earlier, which is unrelated to this PR and still present.

LGTM from my side on the leak fixes themselves.

scorpio810:
[asan-compare.sh](https://github.com/user-attachments/files/29409652/asan-compare.sh)


`
./asan-compare.sh -r ~/Qet-svn/git/qet -b master -p 519 -f /chemin/vers/projet.qet -t 30

laurent@debian:~/Qet-svn/git/qet/build-asan-master$ ASAN_OPTIONS=detect_leaks=1 ./qelectrotech ~/Documents/LAURENT/schemas/2022-7-23_Vibrafloor.qet > /media/backup6/download/run.log 2>&1

laurent@debian:~/Qet-svn/git/qet/build-asan-master$ ASAN_OPTIONS=detect_leaks=1 ./qelectrotech /chemin/vers/projet.qet 2>&1 | tee /media/backup6/download/run.log
`

sources/exportdialog.cpp
```diff
@@ -112,6 +112,7 @@ ExportDialog::ExportDialog(
 */
 ExportDialog::~ExportDialog()
 {
+	qDeleteAll(diagram_lines_);
```
plc-user, line 115: When I read this when searching for "qDeleteAll()": 
"_qDeleteAll() leaves the container populated with dangling (freed) pointers. You must explicitly call .clear() on the container afterward to avoid crashes._"
It seems that we also need 
`diagram_lines_.clear();`
to be completely safe...

sources/editor/elementscene.cpp
```diff
@@ -107,6 +107,9 @@ ElementScene::~ElementScene()
 
 	if (m_decorator)
 		delete m_decorator;
+
+	if (m_paste_area && !m_paste_area->scene())
+		delete m_paste_area;
```
plc-user, line 112: looks fine to me!

sources/genericpanel.cpp
```diff
@@ -321,9 +321,11 @@ QTreeWidgetItem *GenericPanel::getItemForDiagram(Diagram *diagram,
 		if (created) *created = false;
 		return(diagram_qtwi);
 	}
-	
+
+	if (!created) return(nullptr);
+
 	diagram_qtwi = makeItem(QET::Diagram);
-	if (created) *created = true;
+	*created = true;
```
plc-user, line 328: It's too warm to get a clear thought, but:
Don't we change the logic with these modifications? 🤔

ispyisail (commented):
## Review

This PR plugs four distinct leaks totalling ~536 KB per session: an orphaned `QVBoxLayout` in `StyleEditor`, leaked `ExportDiagramLine` wrapper structs in `ExportDialog`, a parentless `QTreeWidgetItem` created in `GenericPanel::getItemForDiagram`, and a `QGraphicsRectItem` that outlived the `ElementScene`. The ASAN before/after numbers are credible and the verification methodology is sound.

---

### Findings

**`sources/genericpanel.cpp:323` — public method contract change not documented**

The new `if (!created) return(nullptr)` silently changes what callers without the `created` output parameter receive: previously a freshly-allocated (but unregistered, parentless, useless) `QTreeWidgetItem`; now `nullptr`. All current external callers in `qetdiagrameditor.cpp` (lines 1857, 1870) and `elementspanelwidget.cpp` (lines 273–375) already guard the return value with `if (item)`, so there is no crash. The behavioral change is strictly correct — the old code was leaking an item that could never do useful work since it wasn't registered in `diagrams_` and not inserted into the tree. However, the function's doc comment still says nothing about returning `nullptr` when `created == nullptr` and the diagram isn't cached, and the sibling `getItemForProject` (protected) still uses the old `if (created) *created = true;` pattern — making them inconsistent in a way that will confuse future callers of this public method. The doc comment on `getItemForDiagram` should be updated to describe this behaviour.

**`sources/editor/elementscene.cpp:110` — `m_paste_area` has no null initializer**

`m_paste_area` is declared as a raw pointer (`QGraphicsRectItem *m_paste_area;`) with no `= nullptr` in-class initializer and is not listed in the constructor's initializer list; it is set by `initPasteArea()` inside the constructor body. The PR's added destructor guard (`if (m_paste_area && ...)`) is written as if `m_paste_area` could be null, but it can only be null if the constructor partially failed before `initPasteArea()` ran — an edge case not guarded against. In practice `initPasteArea()` always runs and the code is safe today, but adding `= nullptr` to the member declaration would make the defensive check accurate and protect against future refactoring. This is a pre-existing nit the PR surface-exposes.

---

### Other three fixes

**`styleeditor.cpp`** — correct: dropping `(this)` from `new QGridLayout(this)` prevents the auto-install that was silently blocking `setLayout(main_layout)`. `main_layout->addItem(grid_layout)` then takes ownership, and `setLayout(main_layout)` installs the full tree. Clean.

**`exportdialog.cpp`** — correct: `ExportDiagramLine::~ExportDiagramLine()` is empty (verified in source), so `qDeleteAll` frees only the plain struct memory without touching widgets owned by Qt's parent-child chain. No double-free risk.

**`elementscene.cpp`** — correct: the `!m_paste_area->scene()` guard correctly distinguishes items the scene destructor will handle from items that have been `removeItem()`-d and need manual cleanup.

---

*🤖 Generated with [Claude Code](https://claude.com/claude-code)*

scorpio810:
Please try it every which way and several times.
I don’t want to push a PR that could jeopardise our users’ professional work. 
### By the way, it’s becoming urgent to release a stable version soon and refocus on the Qt6 framework.

scorpio810:
And as plc-user says, it’s far too hot at the moment to think, work on things, compiling and test them; here in the room where my machines are, it’s over 35°C ...

plc-user:
See this comment as a remark!

The FreeCAD team has published guidelines on the use of AI in PullRequests, which I support and would like to see adopted by QET as well!

I don’t want to “discuss” code snippets with AI systems that just spout flowery language — because they merely string together words that always sound nice and correct, but aren't always what they seem...


Quote from: https://blog.freecad.org/2026/06/29/ai-policy-update/
"In par­tic­u­lar, the dev team expects con­trib­u­tors to dis­close the use of AI in the pull request descrip­tion, take respon­si­bil­i­ty for the code they are sub­mit­ting, and com­mu­ni­cate with code review­ers with­out the use of chatbots."

Quote from: https://github.com/FreeCAD/FreeCAD/blob/main/AI_POLICY.md
"We ask the contributor in the Pull Request template to mark a required checkbox that says _“This PR is not unverified AI output, I take responsibility for it, and all communication from my side in this PR is done by me personally.”_."

ispyisail:
Noted, will adjust my procedures. Will still use AI to find stuff but be more careful with replys

plc-user:
Just a final word to the AI-discussion.

A few days ago, a magazine published an interview with a Siemens manager:
_AI is “brilliant” at automating and speeding things up. However, it is not suited to replacing an engineer’s judgement. Or, as he puts it: “AI should never replace the professional judgement of engineers.”_

plc-user:
If the change to the logic in sources/genericpanel.cpp, line 323 onwards, does what is intended in the function, it makes sense to accept this pull request and plug the memory leaks.

ispyisail:
@plc-user Yes, it does what it's meant to.

The old code always made a new tree item on a cache miss, even when the caller didn't pass a `created` pointer to take ownership of it. That happens on 9 call sites (in `elementspanelwidget.cpp` and `qetdiagrameditor.cpp`) that just want to check "does this diagram already have an item?" — they never registered or parented that item, so it just leaked.

The fix makes it return `nullptr` in that case instead of creating an orphan item. `addDiagram()` still works the same as before since it always passes `&creation_required` and properly registers/parents the item. The other call sites already treat a `nullptr` return as "nothing to select," so nothing changes for them behaviorally — the leak is just gone.

Good to merge from my side too.

Issue #409, opened 2025-11-30
moving terminals on terminal block manager is not working


FeLiiix:
Cannot move independent terminals to terminal blocks.

Both drag and drop is not working as well as button on the right.

<img width="1097" height="308" alt="Image" src="https://github.com/user-attachments/assets/7cc34fe2-a08b-4a59-9f79-fc0f44cb0f26" />

I'm using version 0.100 (DEV) macOS Apple Silicon arm64 on macOS 26.1 (25B78)

ispyisail:
Traced this — found a real fragility, though I want to flag upfront that I couldn't fully pin down which of two mechanisms is the actual trigger without a live repro (no Mac here, and I ran out of confidence chasing it further blind).

### What's there

`FreeTerminalEditor::on_m_move_pb_clicked()` (`sources/TerminalStrip/ui/freeterminaleditor.cpp:247`) reads the terminal strip to move into from `ui->m_move_in_cb->currentData()`. That combo box is populated **only** inside `reload()` (`:105`), and `reload()` runs at exactly three points: navigating to the Free Terminal page (`TerminalStripEditorWindow::updateUi()`), clicking Reset, and at the end of `apply()`.

Separately, the constructor wires:

```cpp
connect(m_model, &FreeTerminalModel::dataChanged, this, [=] {
    this->setDisabledMove();     // default arg = true
});
```

so **any** `dataChanged` on the free-terminal table disables the Move button and the "move to" combo — and nothing re-enables them except one of those three `reload()` triggers. There's no path that re-enables Move after a routine edit is cleanly committed without the user also hitting Apply/Reset or switching strips in the tree.

### Two candidate triggers, not disambiguated

1. **A cell edit (e.g. the LED column) fires `dataChanged` and Move stays disabled** until Apply/Reset/switch-strips — plausible if editing any field before attempting the move is a normal part of the workflow.
2. **Something benign in the free-terminal table fires `dataChanged` without the user intending an edit at all** (e.g. a delegate touching a cell on selection/click rather than a deliberate edit) — this would make Move appear broken on essentially every attempt, matching "always reproducible" with no edit mentioned in the report.

Both point at the same underlying fragility: `setDisabledMove()` is a blunt "any change → disabled until a full reload" gate rather than something that re-checks whether there's an actual unapplied/pending edit (`m_model->modifiedModelRealTerminalData()`, already used in `apply()`, would be the right thing to check instead of blindly disabling).

[...]

PR #526, opened 2026-06-21
fix(terminal-strip): free-terminal move button disabled and unresponsive (#409)

ispyisail:
## Root cause

Three bugs stacked to produce the symptom reported in #409 (move button does nothing / appears broken):

**Bug 1 — `selectionChanged()` declared but never implemented or connected**
`freeterminaleditor.h` declares `void selectionChanged()` as a private slot but the body was never written and it was never connected to the selection model. The move button had no awareness of whether a terminal was actually selected, so it could appear enabled with nothing selected and then silently return early inside `on_m_move_pb_clicked()` at the `real_t_vector.isEmpty()` guard.

**Bug 2 — `dataChanged → setDisabledMove(true)` was a one-way trap**
Any cell edit — including accidentally opening and closing a Type/Function combo, or toggling the LED combo back to its existing value — permanently disabled the move button until `reload()` was called. There was no tooltip explaining why the button was greyed out, so the user had no way to understand or recover.

**Bug 3 — LED cell missing a change guard in `setData()`**
`LABEL_CELL` checks `label != value` before marking the cell modified. `LED_CELL` did not, so clicking the LED combo while it was already set to the same value still emitted `dataChanged` and triggered the disable.

## Fix

- Implement `selectionChanged()`: enable the move controls only when ≥1 row is selected **and** there are no pending (yellow) edits.
- Connect it to both `selectionModel::selectionChanged` and `model::dataChanged` so button state is always consistent.
- Route `reload()`'s re-enable through `selectionChanged()` instead of `setDisabledMove(false)` directly, so selection state is respected immediately after a successful move.
- Add a tooltip to `m_move_pb` that explains the disabled state: *"Apply or discard pending edits before moving"*.
- Add the missing change guard to `LED_CELL` in `setData()`.

## Not included

Drag-and-drop between the free-terminal list and the strip tree (#409 also mentions D&D). Neither `FreeTerminalModel` nor `TerminalStripModel` has `Qt::ItemIsDragEnabled` or any `mimeData()`/`dropMimeData()` implementation — D&D was never built for this panel. That is a separate feature request.

## Test plan

- [ ] Open Terminal Block Manager with at least one independent terminal and one terminal strip
- [ ] Confirm move button is **disabled** with nothing selected
- [ ] Select a terminal row — confirm button becomes **enabled**
- [ ] Edit a Type/Function/LED cell (row turns yellow) — confirm button becomes **disabled** with explanatory tooltip
- [ ] Click Apply to commit the edit — confirm button becomes **enabled** again (if row still selected)
- [ ] Select a row, choose a target strip, click move — confirm terminal disappears from the free list and appears in the strip
- [ ] Toggle LED combo to its current value — confirm row does **not** turn yellow and button remains enabled

### DIFF
--- a/sources/TerminalStrip/ui/freeterminaleditor.cpp
+++ b/sources/TerminalStrip/ui/freeterminaleditor.cpp
@@ -46,10 +46,11 @@ FreeTerminalEditor::FreeTerminalEditor(QETProject *project, QWidget *parent) :
         connect(m_project, &QObject::destroyed, this, &FreeTerminalEditor::reload);
     }
 
-		//Disabled the move if the table is currently edited (yellow cell)
-	connect(m_model, &FreeTerminalModel::dataChanged, this, [=] {
-		this->setDisabledMove();
-	});
+		//Disable the move button while cells are edited (yellow), re-evaluate on every change
+	connect(m_model, &FreeTerminalModel::dataChanged, this, &FreeTerminalEditor::selectionChanged);
+
+	connect(ui->m_table_view->selectionModel(), &QItemSelectionModel::selectionChanged,
+			this, &FreeTerminalEditor::selectionChanged);
 
 	connect(ui->m_table_view, &QAbstractItemView::doubleClicked, this, [=](const QModelIndex &index)
 	{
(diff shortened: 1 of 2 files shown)


scorpio810:
Wait's Joshua say and think about this PR.

ispyisail:

PR #505, opened 2026-06-15
Fix Qt6 build: cmake KF6 support + C++ source compatibility

ispyisail:
Fixes #402 — Qt6 build currently broken with both qmake and cmake.

## Summary

This PR fixes the two issues reported in #402 and all downstream C++ compilation errors that surface once the build system is corrected. The qmake path is not fixable (KF6 dropped qmake support entirely); this PR focuses on the CMake build.

Tested on **Ubuntu 25.04 / Qt 6.8.3 / KDE Frameworks 6** inside Docker. Clean build, binary confirmed Qt6-linked via `ldd`, startup confirmed via `QT_QPA_PLATFORM=offscreen --version` and `--help`.

---

## Commit 1 — cmake: add Qt6 and KDE Frameworks 6 support

**`CMakeLists.txt`**
- `find_package(QT NAMES Qt6 Qt5 ...)` — try Qt6 before falling back to Qt5
- Replace `qt5_create_translation` / `qt5_add_translation` with the version-agnostic `qt_create_translation` / `qt_add_translation` aliases
- Scope `lupdate` source scan to `sources/` — Qt6's stricter `lupdate` rejects the modern JS static-field syntax in the `doxygen-awesome-css` submodule
- Update `target_link_libraries` to use `${KF_PRIVATE_LIBRARIES}` (works for both KF5 and KF6)

**`cmake/fetch_kdeaddons.cmake`**
- Rewrote to auto-detect Qt major version (probes early if `QT_VERSION_MAJOR` is not yet set) and select KF5 (Qt5 builds) or KF6 (Qt6 builds) accordingly
- Accepts both `BUILD_WITH_KF5` (legacy) and `BUILD_WITH_KF` flag names
- `BUILD_KF=OFF` uses system-installed KF6 packages; `BUILD_KF=ON` fetches from KDE git via FetchContent

---

## Commit 2 — fix(qt6): C++ source incompatibilities

| File | Change |
|---|---|
| `sources/qgimanager.h` / `.cpp` | Qt6 makes `QVector<T>` an alias for `QList<T>`, so the deprecated `QList<>` overloads of `manage()` / `release()` become duplicate signatures. Guarded under `#if QT_VERSION < 6`. |
| `sources/diagramview.cpp` | `QTextStream::setCodec()` removed in Qt6. Qt6 defaults to UTF-8. Guarded out under Qt6. |
[...]

PR #533, opened 2026-07-05
Fix Qt-only build when BUILD_WITH_KF5=OFF

geri1701:
While packaging QElectroTech for nixpkgs, I noticed that configuring with `-DBUILD_WITH_KF5=OFF` still enters the KF5 setup path and later fails on the remaining KDE color widgets.

This fixes that build mode by:

- respecting `BUILD_WITH_KF5=OFF` in `fetch_kdeaddons.cmake`
- adding small Qt-only replacements for the `KColorButton` and `KColorCombo` APIs used by the existing `.ui` files
- only adding those replacements and their include path when `BUILD_WITH_KF5=OFF`

The default KF5 build path is unchanged.

Tested locally:

- `BUILD_WITH_KF5=OFF`: configure and build `qelectrotech` succeeds
- `BUILD_WITH_KF5=ON` with system KF5: configure and build `qelectrotech` succeeds

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,6 +148,14 @@ target_include_directories(
     ${QET_DIR}/sources/svg
 )
 
+if(NOT BUILD_WITH_KF5)
+    target_include_directories(
+        ${PROJECT_NAME}
+        PRIVATE
+        ${QET_DIR}/sources/ui/nokde
+    )
+endif()
+
 install(TARGETS ${PROJECT_NAME})
 
 if (NOT MINGW)
--- a/cmake/fetch_kdeaddons.cmake
+++ b/cmake/fetch_kdeaddons.cmake
@@ -16,7 +16,7 @@
 
 message(" - fetch_kdeaddons")
 
-if(DEFINED BUILD_WITH_KF5)
+if(BUILD_WITH_KF5)
   Include(FetchContent)
 
   option(BUILD_KF5 "Build KF5 libraries, use system ones otherwise" YES)
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -726,6 +726,8 @@ set(QET_SRC_FILES
 
 if(NOT BUILD_WITH_KF5)
   list(APPEND QET_SRC_FILES
+    ${QET_DIR}/sources/ui/nokde/kautosavefile.cpp
+    ${QET_DIR}/sources/ui/nokde/kautosavefile.h
     ${QET_DIR}/sources/ui/nokde/kcolorbutton.cpp
     ${QET_DIR}/sources/ui/nokde/kcolorbutton.h
     ${QET_DIR}/sources/ui/nokde/kcolorcombo.cpp
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -724,6 +724,15 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/xml/terminalstriplayoutpatternxml.h
   )
 
+if(NOT BUILD_WITH_KF5)
+  list(APPEND QET_SRC_FILES
+    ${QET_DIR}/sources/ui/nokde/kcolorbutton.cpp
+    ${QET_DIR}/sources/ui/nokde/kcolorbutton.h
+    ${QET_DIR}/sources/ui/nokde/kcolorcombo.cpp
+    ${QET_DIR}/sources/ui/nokde/kcolorcombo.h
+  )
+endif()
+
 set(TS_FILES
   ${QET_DIR}/lang/qet_ar.ts
   ${QET_DIR}/lang/qet_ca.ts
--- a/sources/qetapp.cpp
+++ b/sources/qetapp.cpp
@@ -47,6 +47,7 @@
 #include <QProcessEnvironment>
 #include <QRegularExpression>
 #ifdef BUILD_WITHOUT_KF5
+#	include "ui/nokde/kautosavefile.h"
 #else
 #	include <KAutoSaveFile>
 #endif
@@ -2500,9 +2501,6 @@ void QETApp::buildSystemTrayMenu()
 */
 void QETApp::checkBackupFiles()
 {
-#ifdef BUILD_WITHOUT_KF5
-	return;
-#else
 	QList<KAutoSaveFile *> stale_files = KAutoSaveFile::allStaleFiles();
 
 	//Remove from the list @stale_files, the stales file of opened project
@@ -2577,7 +2575,6 @@ void QETApp::checkBackupFiles()
 			delete stale;
 		}
 	}
-#endif
 }
 
 /**
--- a/sources/qetdiagrameditor.cpp
+++ b/sources/qetdiagrameditor.cpp
@@ -49,6 +49,7 @@
 #include "ui/terminalnumberingdialog.h"
 #include <QDebug>
 #ifdef BUILD_WITHOUT_KF5
+#	include "ui/nokde/kautosavefile.h"
 #else
 #	include <KAutoSaveFile>
 #endif
@@ -1948,8 +1949,6 @@ bool QETDiagramEditor::drawGrid() const
 	return m_draw_grid->isChecked();
 }
 
-#ifdef BUILD_WITHOUT_KF5
-#else
 /**
 	@brief QETDiagramEditor::openBackupFiles
 	@param backup_files
@@ -1980,7 +1979,6 @@ void QETDiagramEditor::openBackupFiles(QList<KAutoSaveFile *> backup_files)
 		DialogWaiting::dropInstance();
 	}
 }
-#endif
 /**
 	met a jour le menu "Fenetres"
 */
--- a/sources/qetdiagrameditor.h
+++ b/sources/qetdiagrameditor.h
@@ -44,10 +44,7 @@ class ElementsCollectionWidget;
 class AutoNumberingDockWidget;
 class TerminalNumberingDialog;
 
-#ifdef BUILD_WITHOUT_KF5
-#else
 class KAutoSaveFile;
-#endif
 /**
 	This class represents the main window of the QElectroTech diagram editor and,
 	ipso facto, the most important part of the QElectroTech user interface.
@@ -72,10 +69,7 @@ class QETDiagramEditor : public QETMainWindow
 		ProjectView *currentProjectView() const;
 		QETProject *currentProject() const;
 		bool drawGrid() const;
-#ifdef BUILD_WITHOUT_KF5
-#else
 		void openBackupFiles (QList<KAutoSaveFile *> backup_files);
-#endif
 
 	  protected:
 		bool event(QEvent *) override;
--- a/sources/qetproject.cpp
+++ b/sources/qetproject.cpp
@@ -93,8 +93,6 @@ QETProject::QETProject(const QString &path, QObject *parent) :
 	init();
 }
 
-#ifdef BUILD_WITHOUT_KF5
-#else
 /**
 	@brief QETProject::QETProject
 	@param backup : backup file to open, QETProject take ownership of backup.
@@ -129,7 +127,6 @@ QETProject::QETProject(KAutoSaveFile *backup, QObject *parent) :
 
 	init();
 }
-#endif
 
 /**
 	@brief QETProject::~QETProject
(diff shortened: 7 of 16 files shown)


scorpio810:
https://qelectrotech.org/forum/viewtopic.php?pid=22479#p22479
https://qelectrotech.org/forum/viewtopic.php?pid=22445#p22445
https://qelectrotech.org/forum/viewtopic.php?pid=21213#p21213

scorpio810:
Thanks for tackling this, the DEFINED vs value fix is correct and matches the reported bug.

One concern before this can be merged: with BUILD_WITH_KF5=OFF, KCoreAddons is dropped entirely, which also removes KAutoSaveFile. That class backs QET's crash-recovery/autosave mechanism — without it (or an equivalent), a Qt-only build loses the safety net that recovers unsaved work after a crash. That's a real regression for users on that build path, not just a cosmetic one.

Could you either:
1. Add a minimal Qt-only replacement for the autosave/lock-file behavior (similar in spirit to what you did for KColorButton/KColorCombo), or
2. Make it explicit in the build output / docs that BUILD_WITH_KF5=OFF disables crash recovery, so packagers and users know what they're trading off?

Happy to help review a Qt-only autosave replacement if you want to take a shot at it (QLockFile + a simple recovery-file convention would likely cover it).

geri1701:
Thanks, that context is useful.

I agree with the concern there: the main Qt/KF build should keep using the KDE Frameworks widgets instead of reimplementing them locally.

That is also why this PR keeps the default `BUILD_WITH_KF5=ON` path unchanged. The Qt-only replacements are only added for the already existing `BUILD_WITH_KF5=OFF` build mode, where the KDE widgets are intentionally unavailable.

So the intended scope here is just to make that explicit no-KF5 build option work again, without changing the normal Qt/KF direction.

scorpio810:
Is this driven by nixpkgs dropping/deprecating the KF5 kcoreaddons/kwidgetsaddons packages specifically, or just packaging convenience?

scorpio810:
Understood on scope, and agreed — the default KF5 path should keep using the real KDE widgets, no argument there.

But that wasn't quite my question. My concern is specifically about KAutoSaveFile, not KColorButton/KColorCombo. Right now the PR replaces the color widgets for the KF5=OFF path, but as far as I can see it doesn't touch KAutoSaveFile — so in that mode, crash recovery just silently disappears with no replacement and no warning.

Before I can merge this, I need one of:
1. A minimal Qt-only stand-in for the autosave/recovery behavior in the KF5=OFF path, or
2. A clear, visible warning (build-time message and/or README note) that BUILD_WITH_KF5=OFF ships without crash recovery.

Without either, someone packaging or building Qt-only ends up with an app that silently can't recover unsaved schematics after a crash, and they won't know it until it's too late. That's the piece I need resolved either way.

geri1701:
Thanks, understood.

My previous comment was only about the Qt/KF widget scope and crossed with your KAutoSaveFile concern.

For the motivation: this is driven by nixpkgs trying to avoid reintroducing KF5 into this package while the Qt6/KF6 work is still ongoing. It is not meant as a statement against the normal Qt/KF build path.

You are right about the crash-recovery path. I had focused on the build failure and the remaining color widgets, but `BUILD_WITH_KF5=OFF` also disables the KAutoSaveFile-backed recovery code today. That should not be silent.

I will not treat this PR as ready until that is addressed. I will look at the Qt-only autosave/recovery option first, and if that turns out too large for this PR, I will at least make the limitation explicit as you suggested.

scorpio810:
Before you build a KAutoSaveFile replacement from scratch: take a look at the qt6-cmake branch (https://github.com/qelectrotech/qelectrotech-source-mirror/tree/qt6-cmake). elevatormind already did work there to get rid of the last KF6 addons dependency, including a QStandardPaths-based rework of the autosave path handling (QETApp::autosaveDir() and friends) as part of removing KAutoSaveFile. It's from the Qt6/KF6 side rather than Qt5/KF5, so it won't drop in directly, but the approach should transfer and save you from re-deriving it independently.

scorpio810:
One more thing before we go further on the Qt-only path: there's another PR (#505) that adds Qt6/KF6 support directly to master. From what you said, the actual driver here is nixpkgs avoiding KF5 specifically (which is being actively deprecated/removed there), not KDE Frameworks as a whole. If that's accurate, KF6 wouldn't have that problem — it's the actively maintained version.

Would you be able to test #505 in depth on your end — building QET with Qt6 + KF6 through it, and actually exercising the app (not just build + launch): opening/saving projects, crash-recovery/autosave, PDF export, the title block editor, terminal strips? If that path turns out solid, it might cover your nixpkgs use case without needing the KF5=OFF / no-KF mode from #533 at all. And if it doesn't (missing something, or you specifically need zero KDE Frameworks dependency regardless of version), that tells us #533 stays necessary as a separate track. Either way, real-world testing from someone with your nixpkgs packaging use case would be really valuable here.

Heads up if you do: I noticed #505 still uses `if(DEFINED BUILD_WITH_KF5 OR DEFINED BUILD_WITH_KF)` in fetch_kdeaddons.cmake, which has the same DEFINED-vs-value issue as the bug you fixed in #533 — so passing OFF explicitly won't behave as expected there yet either.

geri1701:
Thanks for the pointers. I pushed an update keeping this PR scoped to the existing `BUILD_WITH_KF5=OFF` option.

The no-KF path now has a small Qt-only `KAutoSaveFile`-compatible fallback using `QLockFile`, so the existing crash-recovery flow stays active. The KF5 path still uses KDE's implementation.

I agree #505 is the separate Qt6/KF6 track. I did not fold that broader testing into this PR.

Verified locally with both build modes. I also ran a small standalone smoke test for the Qt-only autosave helper.

scorpio810:
Nice, QLockFile is a solid choice — keeps it Qt-only with no new dependency, and glad you kept #505 out of scope.

Before I merge: could you walk through the actual end-to-end recovery scenario manually (not just the standalone smoke test)? Concretely: open a project in the KF5=OFF build, make unsaved changes, hard-kill the process (kill -9), relaunch, and confirm the recovery file is offered/restored correctly. Also worth double-checking how the QLockFile-based helper behaves if the lock is left behind by a real crash (stale lock detection) — want to make sure a crash doesn't leave someone locked out instead of recovered.

Last thing: would you be up for folding that smoke test into the existing test suite (tests/googletest or tests/catch) instead of keeping it standalone? This is exactly the kind of path that's worth having a permanent regression test for.

geri1701:
Thanks, I tested that path manually now and pushed a small regression test.

Manual check with the `BUILD_WITH_KF5=OFF` build:
- opened a project from an isolated config/data home
- added an unsaved folio
- waited for the normal recovery backup pass to write the updated snapshot
- killed QET with `SIGKILL`
- relaunched; the recovery dialog was offered despite the leftover lock file
- accepted recovery; the restored project contained the added folio while the original file stayed unchanged

I also folded the helper smoke test into `tests/catch` for the no-KF5 build. It covers stale-file discovery, stale-lock recovery/opening, reading, and cleanup.

Locally verified both build modes, plus the no-KF5 `C_unittests` build and full `C_unittests` run.

scorpio810:
Thanks for the update — keeping this scoped to `BUILD_WITH_KF5=OFF` and using `QLockFile` for the fallback is the right call. I went through the patch in detail.

**Blocking issue: `staleFiles()` doesn't check if the lock is still held**

```cpp
QList<KAutoSaveFile *> KAutoSaveFile::staleFiles(...)
{
    for (const auto &file : findAllStaleFiles(applicationName)) {
        const auto managed_file = readManagedFileMetadata(file);
        ...
        stale_files << stale_file;  // added without checking the lock!
    }
    return stale_files;
}
```

`findAllStaleFiles()` lists every `*.qetautosave` file in the directory without ever consulting its associated `.lock`. As written, an autosave file actively in use by another running QET session (two projects open in parallel, or a second instance if `SingleApplication` is bypassed) would be offered as "recoverable" the same way a genuine crash file would.

The existing filtering in `qetapp.cpp` ("Remove from the list @stale_files, the stales file of opened project") only protects against projects open *within the same QET instance* — not against a file locked by an external process. Since the KDE version of `KAutoSaveFile` normally does this filtering via the lock internally, this is a real regression in the API contract, not just a nitpick.

**Suggested fix:** in `staleFiles()`, before adding a candidate, instantiate a temporary `QLockFile` on `file + ".lock"` and only keep the file if `tryLock()` succeeds (then `unlock()` immediately) — this is exactly the staleness-detection mechanism `QLockFile` already provides natively.

**Two minor points, non-blocking:**

1. **Orphaned `.lock` files after a real crash.** The `.qetautosave` and its `.lock` correctly survive a crash (that's the point), but once the file is recovered or dismissed on the next launch, nothing cleans up the leftover `.lock` — it lingers indefinitely in the autosave directory. Cosmetic (a stale lock will eventually be overwritten/detected by a future `tryLock()`), but worth flagging.

2. **Windows paths not explicitly tested.** The `storedManagedFile()` / `managedFileFromStorage()` roundtrip handles Unix paths (`/home/...`) fine, but for a Windows path (`C:/Users/...`), `QUrl::fromLocalFile()` produces something like `/C:/Users/...` — the roundtrip through the `.path` sidecar file doesn't look tested against this case. Since QET ships Windows nightlies, this deserves a dedicated test before considering the PR mergeable across platforms.

**What's done well:** the stable per-instance filename generation (created once, reused across periodic autosaves), clean lock release + file removal on normal close via `releaseLock()`, and the clean KF5/no-KF5 separation via `#ifdef` — all solid.

Once the lock-check issue is addressed, this looks mergeable to me.

scorpio810:
Thanks for testing the crash-recovery path — that confirms the basic flow works. But I looked at the actual `staleFiles()` implementation, and the lock-check issue is still present:

```cpp
QList<KAutoSaveFile *> KAutoSaveFile::staleFiles(
    const QUrl &url,
    const QString &applicationName)
{
    const auto managed_file_filter = normalizeManagedFile(url);
    QList<KAutoSaveFile *> stale_files;

    for (const auto &file : findAllStaleFiles(applicationName)) {
        const auto managed_file = readManagedFileMetadata(file);
        if (managed_file.isEmpty()) {
            continue;
        }
        if (!managed_file_filter.isEmpty()
            && managed_file != managed_file_filter) {
            continue;
        }

        auto *stale_file = new KAutoSaveFile(managed_file);
        stale_file->setFileName(file);
        stale_file->m_managed_file_name_changed = false;
        stale_files << stale_file;   // <- no lock check here
    }

    return stale_files;
}
```

There's no `QLockFile`/`tryLock()` check anywhere in this loop. The only place a lock gets checked is in `open()`:

```cpp
if (!m_lock) {
    m_lock = std::make_unique<QLockFile>(
        fileName() + QStringLiteral(".lock"));
    m_lock->setStaleLockTime(60 * 1000);
}

if (m_lock->isLocked() || m_lock->tryLock()) {
    return true;
}
```

...which only runs if/when the caller actually opens the returned `KAutoSaveFile`. `staleFiles()` itself will still list — and offer as "recoverable" — an autosave file that's currently locked by another live QET session.

Your new regression test doesn't catch this either: the child process is fully `waitpid()`-ed (i.e. exited, lock released) *before* `allStaleFiles()` is called, so by the time the discovery runs, there's no active lock left to conflict with. It's a good test for the "genuine crash" path, but it doesn't exercise the "second session still running" case.

**What's still needed:** in `staleFiles()`, before adding a candidate, try to acquire a temporary `QLockFile` on `file + ".lock"` and skip the entry if the lock is currently held by a live process (i.e. `tryLock()` fails and the lock isn't stale). A second Catch test case would help too: fork a child that holds the lock open (don't let it exit before checking), then assert that `allStaleFiles()` does *not* return that file while the child is still alive.

Everything else — metadata roundtrip, cleanup on close, KF5/no-KF5 split — still looks solid. This is the one piece left before I'd merge.

geri1701:
Thanks, you are right — the previous test covered the crash/stale-lock path, but not an actively held lock.

I pushed a focused follow-up. `staleFiles()` now checks the candidate's `QLockFile` before returning it and skips files whose lock is still held by a live process.

I also extended the Catch test so the child process keeps the autosave file open and locked while the parent calls `allStaleFiles()`: it now asserts that no file is returned while the child is alive, then kills the child and verifies the same file becomes recoverable. I checked that this updated test fails against the previous PR head and passes with the fix.

Re-verified locally:
- full no-KF5 `C_unittests` run
- no-KF5 `qelectrotech` build
- KF5 `qelectrotech` build
- quick GUI stale-lock recovery-dialog smoke after `SIGKILL`

scorpio810:
Thanks for this, @geri1701 — nice work overall.

The `KColorButton`/`KColorCombo` replacements are clean and idiomatic. The `KAutoSaveFile` reimplementation is the more sensitive piece (crash-recovery logic touches user data), but it's well isolated behind `BUILD_WITH_KF5=OFF` and doesn't affect the default KF5 path we ship on Windows/macOS, so it looks safe to merge for the Linux/nix packaging use case it targets. Appreciate the added test coverage too.

Merging this — thanks for the contribution!

Edit:
Thanks for this, @geri1701 — clean, minimal fix.

The `KColorButton`/`KColorCombo` Qt-only replacements are idiomatic and well scoped behind `BUILD_WITH_KF5=OFF`, and the `fetch_kdeaddons.cmake` fix is correct. Doesn't touch the default KF5 path we ship on Windows/macOS.

Merging this — thanks for the contribution!

PR #538, opened 2026-07-11
Complete German translation of unfinished terminal-label strings

DieterMayerOSS:
qet_de.ts had 20 unfinished entries in the PartTerminal and TerminalEditor contexts: 12 empty ones (shown in French at runtime, since French is the source language of the tr() literals) and 8 that already carried German text but were still flagged unfinished.

Translate the 12 empty strings, matching the terminology already established in these contexts (borne/terminal -> "Anschluss", label -> "Beschriftung", cadre -> "Rahmen", police -> "Schriftart"), and mark the 8 existing drafts as finished. lrelease now reports 2621 finished and 0 unfinished translations for de_DE.


scorpio810:
Please add .qm files too.

PR #539, opened 2026-07-12
Add compiled qet_de.qm for the completed German translation

DieterMayerOSS:
Regenerate lang/qet_de.qm from lang/qet_de.ts with lrelease so the committed binary matches the strings completed in #538 (2621 finished, 0 unfinished for de_DE).


scorpio810:
Thanks !

PR #548, opened 2026-07-17
Find translations relative to the binary and in the build dir

DieterMayerOSS:
On Windows QET_LANG_PATH is "./l10n/", which languagesPath() returned verbatim: a path relative to the current working directory, so the binary only found its translations when launched from the right folder (the UI silently fell back to the French source language otherwise). Anchor a relative configured path to the executable's directory instead (absolute paths, as on Linux installs, stay unchanged), and add a fallback to the executable's own folder where qt_add_translation drops the qet_*.qm, so a build run straight from the build directory is localized too.

### DIFF
--- a/sources/qetapp.cpp
+++ b/sources/qetapp.cpp
@@ -1295,6 +1295,13 @@ QString QETApp::languagesPath()
 			if (QDir(sibling_of_bin).exists())
 				return(sibling_of_bin);
 		}
+		// Running straight from a build directory: the build emits the
+		// qet_*.qm right next to the binary (no "lang" subfolder). Use the
+		// binary's own folder when the compiled translations sit there, so
+		// a dev build runs localized without first assembling a lang/
+		// folder.
+		if (QDir(bin_dir).exists(QStringLiteral("qet_en.qm")))
+			return(bin_dir + "/");
 		return(next_to_bin);
 	}
 #else
@@ -1304,7 +1311,18 @@ QString QETApp::languagesPath()
 		 * l'option de compilation represente
 		 *  un chemin absolu ou relatif classique
 		 */
-		return(resolveConfiguredDataPath(QUOTE(QET_LANG_PATH)));
+	{
+		const QString configured =
+			resolveConfiguredDataPath(QUOTE(QET_LANG_PATH));
+		if (QDir(configured).exists())
+			return(configured);
+		// Same build-directory fallback as above: nothing configured
+		// exists, but the freshly compiled qet_*.qm sit beside the binary.
+		const QString bin_dir = QCoreApplication::applicationDirPath();
+		if (QDir(bin_dir).exists(QStringLiteral("qet_en.qm")))
+			return(bin_dir + "/");
+		return(configured);
+	}
 	#else
 		/* the compilation option represents a path relative
 		 *  to the folder containing the executable binary


plc-user:
That looks quite plausible. And if it doesn’t mess up the language search on the other systems – which is what it looks like at first glance – we should accept the PR!
However, I can’t compile or test the changes on win.

ispyisail:
@DieterMayerOSS — apologies, this one is on me.

You opened this on 17 July and @plc-user reviewed it favourably on the 19th. On 26 July I opened #559 fixing the same root cause without checking whether anyone already had a PR open for it, and it was merged the same day. That is why this now shows as conflicting: my change rewrote `languagesPath()` underneath you. A one-minute look at the open PRs would have avoided it, and I should have built on your work rather than restating it.

For the record, since it may be useful when you rebase — here is how the two compare against master as it stands now:

| | #548 | master (after #559) |
|---|---|---|
| Anchor a relative `QET_LANG_PATH` to the binary directory | ✅ | ✅ |
| Absolute paths passed through unchanged | ✅ | ✅ |
| Same treatment for the elements and title-block directories | — (lang only) | ✅ |
| **Build-directory fallback (`qet_en.qm` beside the binary)** | ✅ | ❌ |
| MSI file association given the `--lang-dir` etc. arguments | — | ✅ |
| cmake `QET_LANG_PATH` `l10n/` → `lang/` | — | ✅ |

So the packaging side is covered now, **but the build-directory fallback is not, and I think it is still worth landing**:

```cpp
// Running straight from a build directory: qt_add_translation emits the
// qet_*.qm right next to the binary (no "lang" subfolder).
if (QDir(bin_dir).exists(QStringLiteral("qet_en.qm")))
    return(bin_dir + "/");
```

`resolveConfiguredDataPath()` looks for a *directory*, so it never picks up loose `.qm` files sitting beside the binary — meaning a build run straight from the build tree still comes up in the source language. That is a real developer-experience gap, it applies to the `#ifndef QET_LANG_PATH` branch as well (so Linux dev builds too, not just Windows), and nothing in master addresses it.

Rebasing this down to just that piece would be a small diff on top of what is there now. Happy either way:

- if you would rather do it yourself, it is your idea and your PR;
- if you would rather not spend more time on a PR that has already been through a review cycle, say the word and I will prepare the rebase and hand it to you to check.

Either way I would rather this did not get closed as "superseded", because that part of it is not.

DieterMayerOSS:
No apology needed — with the pace this repo has had over the last week, two people fixing the same root cause within days was bound to happen, and #559 covering the packaging side (file association arguments, the cmake rename) went further than my original PR did anyway. Your comparison table is exactly right.

Rebased it down myself, as offered: the PR is now a single commit on top of current master adding only the missing piece — the loose `qet_*.qm`-beside-the-binary fallback, in both the `QET_LANG_PATH` and the no-compile-option branch, with `resolveConfiguredDataPath()` still doing the directory resolution first. Compile-checked on Qt 6.11.1/MinGW.

Thanks for flagging it instead of letting it die as "superseded".

ispyisail:
:)

ispyisail:
Went back over this one carefully before it merges, since the rebase changed what's actually being proposed — the current diff is just the new fallback in `languagesPath()`:

```cpp
const QString configured = resolveConfiguredDataPath(QUOTE(QET_LANG_PATH));
if (QDir(configured).exists())
    return(configured);
const QString bin_dir = QCoreApplication::applicationDirPath();
if (QDir(bin_dir).exists(QStringLiteral("qet_en.qm")))
    return(bin_dir + "/");
return(configured);
```

i.e.: if nothing else resolves, check whether `qet_en.qm` sits **directly** in the same folder as the binary (no `lang/` subfolder at all). I tried to reproduce the scenario this fixes and, after quite a bit of digging, I don't think it can currently happen anywhere in this project's own build or packaging setup — writing this up in case I'm missing your actual use case, since I'd rather ask than assume.

**Where I looked:**

1. **Built it and ran it.** Compiled current `master` + this PR with CMake in a normal out-of-tree build, ran the binary directly (no install). Identical "failed to load qet_en" warning, identical resolved `languagesPath()`, whether the new fallback code was present or not.
2. **Why: `.qm` files never land beside the binary in a CMake build.** `CMakeLists.txt` has `set_source_files_properties(${TS_FILES} PROPERTIES OUTPUT_LOCATION "${QET_DIR}/lang")` — unconditional, no matter where the build directory is. Compiled translations always land in the **source tree's** `lang/` folder, never in the build/binary directory. So `bin_dir` (what this fallback checks) can't contain `qet_en.qm` in any CMake build.
3. **There used to be a second mechanism that would have made this moot** — `QMFILES_AS_RESOURCE`, embedding `.qm` as Qt resources (`:/lang`) directly in the binary, bypassing filesystem search entirely. I checked: it was introduced in `373fcfb4e`, survived into an experimental Qt6-only CMake rewrite, and was dropped when the Qt5/Qt6 CMake build got unified (`a8e2a7acf`). It's gone from current `master`. So filesystem search really is the only path today.
4. **The project's own Dockerfile** does `cmake -B build` in-tree (`/src` → `/src/build`). For that layout I checked directly: the binary's directory is `/src/build`, and `bin_dir/../lang` resolves to `/src/lang` — which exists. That's the **already-merged #559** fallback (`sibling_of_bin`), not this one.
5. **The real, shipped Windows CI package** (`.github/workflows/windows-build.yml`) assembles the layout explicitly:
   ```bash
   cp -r "$GITHUB_WORKSPACE/lang" "$FILES/lang" || true
   find "$GITHUB_WORKSPACE/build" -name "*.qm" -exec cp {} "$FILES/lang/" \;
   ...
   for d in ico elements lang titleblocks fonts examples bin; do
   ```
   `bin/` and `lang/` end up as **sibling** folders under `$FILES` — again the pattern #559 already covers, not a loose `.qm` sitting next to the exe.

So across dev CMake builds (in-tree or out-of-tree), the Dockerfile's reference build, and the actual CI-built/shipped Windows package, I can't find a layout this project produces where a `.qm` file ends up directly beside the binary with no `lang/` folder involved — every path either already works via #559 or genuinely doesn't have the file where this checks for it.

**What I might be missing:** your own local Windows/MinGW workflow. If there's a specific way you build (a particular qmake invocation, an IDE's shadow-build behavior, a manual copy step you do) where `qet_en.qm` really does end up loose beside `qelectrotech.exe`, that's the scenario this PR is for — I just haven't been able to reconstruct it from the repo's own build/CI configuration. Could you share the exact steps/layout that showed you this was needed? Happy to be wrong here, I'd just like to understand the concrete case before it merges.

DieterMayerOSS:
Thanks for doing this level of due diligence — you're right, and I'd rather close this than merge a fallback with no reproducible scenario behind it.

The concrete case behind the original PR predates the rebase: during my early Qt6 testing on Windows I had hand-copied `.qm` files sitting loose beside `qelectrotech.exe` — a manual layout, not something any build system of the project produces. Since #559 (sibling `../lang`) and the `resolveConfiguredDataPath()` change landed, and since my own dev workflow switched to the `bin_dir/lang/` convention (which `languagesPath()` already finds), that scenario no longer exists — exactly as your audit across the CMake build, Dockerfile and CI packaging shows.

Closing — one speculative branch less in `languagesPath()`. Thanks for catching it before it merged.

Issue #553, opened 2026-07-24
Qt 6 opportunities: what could be improved today, and what Qt 6.12 (LTS) will bring


DieterMayerOSS:
This is a collection issue about what QElectroTech could gain from Qt 6 — things that are possible today with Qt 6.5–6.11 behind `QT_VERSION_CHECK` guards (keeping the Qt 5 build working, so nothing here requires dropping Qt 5), plus a short look at what Qt 6.12 (the next LTS) will bring in September.

I have prototyped several of these on a fork (Qt 6.11.1 / Windows 11 / MinGW build) and am happy to submit the ones you find useful as patches.

## Prototyped and working (patches available)

- **XML parse errors with line/column when loading fails.** A small wrapper around `QDomDocument::setContent()` that uses the structured `QDomDocument::ParseResult` on Qt ≥ 6.5 and the legacy out-parameters below that. Wired into project loading (`QETProject::openFile`) and the element editor, so a broken `.qet`/`.elmt` now reports e.g. *"line 4, column 10: tag mismatch"* instead of only *"not a QElectroTech project file"*.
- **PDF bookmarks (one per folio) and document title metadata in the PDF export.** Qt still has no public outline API, so this extends the existing post-processing in `pdf_links.cpp` (same rewrite-and-rebuild-xref technique as the GoTo link conversion): a flat `/Outlines` tree referenced from the catalog with `/PageMode /UseOutlines`, titles as UTF-16BE hex strings. Works from both the GUI print window and the CLI export.
- **Optional dark color scheme** (combo "Système / Sombre" on the Apparence config page). The UI chrome (docks, menus, panels, title bar) goes dark while the diagram canvas and the element trees keep their light, print-like colors — the element icons are black strokes and would disappear on a dark base.

## A finding you probably want even without any of the above

`QETApp::useSystemPalette()` currently sets this application stylesheet:

```cpp
qApp->setStyleSheet(
        "QAbstractScrollArea#mdiarea {"
        "background-color -> setPalette(initial_palette_);"
        "}"
        );
```

The rule is syntactically broken (looks like C++ pasted into CSS during a merge), so it styles nothing — but the mere presence of a non-empty application stylesheet wraps every widget in `QStyleSheetStyle`, which silently disables per-widget `QWidget::setStyle()` application-wide. Replacing it with `qApp->setStyleSheet(QString())` restores normal behavior.

## Dark mode: why I did NOT use QStyleHints::setColorScheme

[...]

PR #558, opened 2026-07-26
Clear the application stylesheet when using system colors

ispyisail:
Picks up the stylesheet finding reported by @DieterMayerOSS in #553, split out as a standalone one-line change so it can be reviewed independently of the Qt 6 discussion in that issue.

## The problem

`QETApp::useSystemPalette(true)` installs an application stylesheet whose only declaration is invalid CSS (`sources/qetapp.cpp:1693`):

```cpp
qApp->setStyleSheet(
        "QAbstractScrollArea#mdiarea {"
        "background-color -> setPalette(initial_palette_);"
        "}"
        );
```

## Where it came from

I traced it, because the history explains why nobody noticed. It is **not** a bad merge — it is an unfinished edit from 2014. Commit `e6c32bc0`, *"Background set to use System Palette"*:

```diff
-		"	background-color:#D5D2D1;"
+		"	background-color -> setPalette(initial_palette_);"
```

A hardcoded grey was replaced with a note-to-self describing the intended change, and the note was committed as if it were code.

It went unnoticed for twelve years for a good reason: Qt's CSS parser **silently skips invalid declarations**, and at the time the same rule still carried valid `background-image` / `background-repeat` / `background-position` properties. The block kept working, so nothing looked wrong. Those properties were dropped in a later restructure, leaving a rule with no valid declarations at all.

The selector itself is still live — `qetdiagrameditor.cpp:100` does `m_workspace.setObjectName("mdiarea")` — so the rule targets a real widget and does nothing to it.

## Why it is worth fixing

Mostly it is dead code, but not entirely inert: a **non-empty** application stylesheet wraps every widget in `QStyleSheetStyle`, which overrides per-widget `QWidget::setStyle()`.

To be precise about the impact, since #553 states it slightly more strongly: **QET does not call `QWidget::setStyle()` anywhere today** — every `setStyle(` call site in `sources/` is `QPen::setStyle` or `QBrush::setStyle` — so nothing is visibly broken right now. It does block that API for future work, which is how it was found (prototyping a palette-based dark mode).

## The change

```cpp
qApp->setStyleSheet(QString());
```

plus a comment recording the above so it does not get "restored" later.

**Behavior is unchanged.** The `use == true` branch already discarded whatever `style.css` had loaded — by overwriting it with the inert rule — and `qApp->setPalette(initial_palette_)` on the line above is what actually supplies the system colors, which is exactly what the 2014 note was asking for. The intent is already satisfied; the rule is vestigial.

**The `style.css` path (`use == false`) is untouched**, so existing custom themes behave exactly as before.

## Testing

- `sources/qetapp.cpp` compiles clean, no new warnings
- full build relinks; resulting binary runs (`--version` → `0.100.1-dev`)

Verified on Linux / Qt 5.15.18 / GCC.

### DIFF
--- a/sources/qetapp.cpp
+++ b/sources/qetapp.cpp
@@ -1690,11 +1690,15 @@ void QETApp::invertMainWindowVisibility(QWidget *window) {
 void QETApp::useSystemPalette(bool use) {
 	if (use) {
 		qApp->setPalette(initial_palette_);
-		qApp->setStyleSheet(
-				"QAbstractScrollArea#mdiarea {"
-				"background-color -> setPalette(initial_palette_);"
-				"}"
-				);
+		// Drop any stylesheet previously loaded from style.css: with system
+		// colors requested, the palette set just above is what provides them.
+		//
+		// This used to install a one-rule stylesheet whose only declaration
+		// was invalid CSS ("background-color -> setPalette(initial_palette_);",
+		// a note-to-self committed in e6c32bc0, 2014). It styled nothing, but
+		// a non-empty application stylesheet still wraps every widget in
+		// QStyleSheetStyle, which overrides per-widget QWidget::setStyle().
+		qApp->setStyleSheet(QString());
 	} else {
 		QFile file(configDir() + "/style.css");
 		if (file.open(QFile::ReadOnly)) {


scorpio810:
https://github.com/qelectrotech/qelectrotech-source-mirror/commit/764fe5b75ea79dd65badd4e272ef773b54294ea8
https://github.com/qelectrotech/qelectrotech-source-mirror/commit/e6c32bc0574d3e1010f5c99a19786ddd420025c5

PR #559, opened 2026-07-26
Fix language (and data paths) when opening a .qet by double-click on Windows

### LINKED ISSUE #554: Language change failed when openning file in windows
Hello,

I have a problem with language switching on Windows.
Version: V 0.100.1‑dev 64‑bit, installed from the MSI package.

I've noticed that when I run QElectroTech from the Programs menu, the language changes correctly. However, when I double‑click a .qet file, the language is always French.

So I checked the shortcut in the Programs menu and saw these command‑line parameters:

```
--common-elements-dir="C:\Program Files\QElectroTech\elements/"
--common-tbt-dir="C:\Program Files\QElectroTech\titleblocks/"
--lang-dir="C:\Program Files\QElectroTech\lang/"
-platform windows:fontengine=freetype
```
After adding these parameters to the
KEY_CLASSES_ROOT\QElectroTech.Document\shell\open\command  
registry key, everything works correctly and the language is applied as expected.

Reg file's contents bellow:
```
Windows Registry Editor Version 5.00

[HKEY_CLASSES_ROOT\QElectroTech.Document\shell\open\command]
@="\"C:\\Program Files\\QElectroTech\\bin\\qelectrotech.exe\"  --common-elements-dir=\"C:\\Program Files\\QElectroTech\\elements/\" --common-tbt-dir=\"C:\\Program Files\\QElectroTech\\titleblocks/\" --lang-dir=\"C:\\Program Files\\QElectroTech\\lang/\" -platform windows:fontengine=freetype \"%1\""
```

ispyisail:
Fixes #554 — the interface always coming up in French when a `.qet` is opened by double-click, while the Start Menu shortcut works.

@mr-rfh diagnosed this correctly and the registry value he built by hand is exactly right. Three commits, smallest first, so the first can be taken on its own if you'd rather not touch the rest.

## What is happening

**The MSI contradicts itself.** `QElectroTech.wxs` registers both shortcuts with the arguments QET needs, under a comment reading *"Point directly to qelectrotech.exe with all required arguments"* (lines 64 and 79). The file association, ~20 lines further down, was written without them:

```xml
Value="&quot;[INSTALLDIR]bin\qelectrotech.exe&quot; &quot;%1&quot;"
```

**Why that matters:** on Windows the compiled-in data paths are *relative* (`./elements/`, `./titleblocks/`, `./lang/`), and `commonElementsDir()` / `commonTitleBlockTemplatesDir()` / `languagesPath()` return them verbatim — so they resolve against the **process working directory**. Explorer sets that to the folder holding the document, so QET looks for its data next to the user's file and finds nothing. The shortcuts hide this by passing the three paths explicitly; the file association didn't.

**Why French specifically:** no translation loads, and QET's source strings *are* French (`<source>Importer une pièce EPLAN…</source>` → `<translation>Import an EPLAN part…</translation>`). "No translation" and "French UI" are the same state.

Worth noting the common element collection and title blocks are resolved the same way, so they are very likely failing on double-click too — @mr-rfh, does the element panel look right when you open a project that way *without* your registry fix?

## The commits

**1. `QElectroTech.wxs` — give the file association the same arguments as the shortcuts.** Immediate fix, matches the reporter's verified workaround.

**2. `qetapp.cpp` — resolve relative compiled paths from the binary rather than the CWD.** Adds `resolveConfiguredDataPath()`: absolute paths pass through untouched; a relative one is tried against the working directory **first** (so anything relying on today's behaviour keeps working), then next to the executable, then in its parent — the packaging layout where the binary is in `bin/` with data beside it.

That last fallback already exists for issue #86, but only in the branch taken when the compile-time option is *absent* — it was unreachable whenever `QET_LANG_PATH` is set, which is always on Windows.

**3. `cmake/paths_compilation_installation.cmake` — `l10n/` → `lang/`.** `QET_LANG_PATH` was `"l10n/"` on WIN32, a string that appears nowhere else in the tree: the shortcuts pass `--lang-dir="[INSTALLDIR]lang/"` and the CI workflow copies the `.qm` files into `files/lang/`. The compiled default pointed at a directory no packaging creates, which is why commit 2 needs this to find anything on Windows.

## Testing

- Full build clean, no new warnings; binary runs (`--version` → `0.100.1-dev`)
- `QElectroTech.wxs` verified well-formed after the edit
- `resolveConfiguredDataPath()` exercised standalone from a **foreign working directory**, covering: absolute unchanged; empty unchanged; relative found in CWD (back-compat); relative found next to the binary; relative found as a sibling of `bin/` (the MSI layout, the case that was broken); unresolvable returned unchanged. All pass.

## Caveats

Verified on Linux/GCC — I can't build the MSI or test on Windows. Commit 1 is a literal transcription of the arguments already on lines 64/79 and of the registry value @mr-rfh confirmed working, but **someone with a Windows toolchain should confirm the built MSI before this is closed.**

Commits 2 and 3 change path resolution, so they deserve a look from whoever owns the packaging — particularly commit 3, in case any downstream build expects `l10n/`. Dropping commit 3 leaves commits 1 and 2 valid; commit 1 alone also stands on its own.

### DIFF
--- a/build-aux/windows/QElectroTech.wxs
+++ b/build-aux/windows/QElectroTech.wxs
@@ -97,8 +97,13 @@
         <RegistryValue Root="HKCR" Key="QElectroTech.Document" Type="string" Value="QElectroTech Project" />
         <RegistryValue Root="HKCR" Key="QElectroTech.Document\DefaultIcon" Type="string"
                        Value="[INSTALLDIR]ico\qelectrotech.ico" />
+        <!-- Pass the same arguments as the shortcuts above: opening a .qet from
+             Explorer sets the working directory to the document's folder, so a
+             bare command line leaves QET unable to locate its data. Without the
+             lang-dir argument the UI falls back to the source language, which
+             is French. See issue #554. -->
         <RegistryValue Root="HKCR" Key="QElectroTech.Document\shell\open\command" Type="string"
-                       Value="&quot;[INSTALLDIR]bin\qelectrotech.exe&quot; &quot;%1&quot;" />
+                       Value="&quot;[INSTALLDIR]bin\qelectrotech.exe&quot; --common-elements-dir=&quot;[INSTALLDIR]elements/&quot; --common-tbt-dir=&quot;[INSTALLDIR]titleblocks/&quot; --lang-dir=&quot;[INSTALLDIR]lang/&quot; $(var.QtPlatformArgs) &quot;%1&quot;" />
       </Component>
     </ComponentGroup>
 
--- a/cmake/paths_compilation_installation.cmake
+++ b/cmake/paths_compilation_installation.cmake
@@ -58,7 +58,10 @@ if(WIN32)
   set(QET_BINARY_PATH             "./")
   set(QET_COMMON_COLLECTION_PATH  "elements/")
   set(QET_COMMON_TBT_PATH         "titleblocks/")
-  set(QET_LANG_PATH               "l10n/")
+  # "lang/" and not "l10n/": that is where every Windows packaging actually
+  # puts the .qm files (see build-aux/windows/QElectroTech.wxs and the
+  # windows-build workflow), and what the shortcuts pass as --lang-dir.
+  set(QET_LANG_PATH               "lang/")
   set(QET_LICENSE_PATH            "./")
   # Liste des ressources Windows
 #RC_FILE = qelectrotech.rc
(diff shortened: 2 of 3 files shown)


scorpio810:
It 's related to #548 ?

scorpio810:
@ispyisail I push  Windows actions builder, please check it when MSI packages will finished.

PR #560, opened 2026-07-27
Log project load times, split by phase

ispyisail:
Picks up @scorpio810's request in #553:

> A new timer in the logs to measure project load times would be useful for comparing the improvements made by Qt6 compared to Qt5. If anyone fancies coding it and taking it on?

QET already reports how long the *elements collection* takes to load (`ElementsCollectionWidget::reload`, the line that produces the "1.3s to load 9000 elements" figure). This adds the equivalent for opening a *project*, following the same `QElapsedTimer` + `qInfo()` idiom so it appears in the normal log without a debug build.

## Why it reports phases and not just a total

This is the part worth a second of review. A single total would have been less useful and possibly misleading: reading the XML and constructing the objects is largely Qt-version-independent (QDom is QDom), whereas refreshing the diagrams is graphics-scene work. Mixed into one number, a real improvement in the graphics path could be diluted to invisibility — or a Qt version could look like it "made no difference" when the difference simply isn't in the phase being measured.

Measured on the bundled example projects, the split is lopsided enough to matter:

| Project | Size | Total | XML parse | Diagrams | Refresh | Database |
|---|---|---|---|---|---|---|
| `741.qet` | 123 KiB | 0.213 s | 0.007 | 0.192 | 0.006 | 0.006 |
| `Habitat-Schemas_developpes.qet` | 209 KiB | 0.619 s | 0.013 | 0.499 | 0.097 | 0.008 |
| `weneedpolonez-…_wiring_diagram.qet` | 3399 KiB | 1.505 s | 0.110 | 1.153 | 0.196 | 0.033 |

XML parsing is 3–7 % of the time; diagram construction is 77–83 %. So a timer wrapped only around the parse — the intuitive place to put one — would have reported almost nothing useful.

## What it logs

```
Project content built in 1.391 seconds (elements collection 0.009, diagrams 1.153,
  terminal strips 0, refresh 0.196, database 0.033)
Project "weneedpolonez-Polonez_MR89_wiring_diagram.qet" (3399 KiB) opened in 1.505 seconds
  (xml parsing 0.11, content 1.395)
```

Two lines: `QETProject::openFile()` reports the total with the parse/build split, and `readProjectXml()` reports the build phases. The file size is included so numbers are comparable between machines and projects.

## Scope

`readProjectXml()` has exactly one caller, and `openFile()` runs only from the two project-opening constructors, so this fires on project open (including backup recovery) and nowhere else — no extra output on save, import or export.

## Testing

Built and run against three example projects of increasing size, output as in the table above. Full build clean, no new warnings.

Verified on Linux/Qt 5.15.18. The point of the change is a Qt5-vs-Qt6 comparison, so the interesting numbers are the ones @scorpio810 or @DieterMayerOSS can now produce from a Qt6 Windows build against the same file — that comparison is what this is for, and I can't run it here.

### DIFF
--- a/sources/qetproject.cpp
+++ b/sources/qetproject.cpp
@@ -35,6 +35,7 @@
 #include "qetxml.h"
 #include "qetversion.h"
 
+#include <QElapsedTimer>
 #include <QHash>
 #include <QTimer>
 #include <QtConcurrentRun>
@@ -227,6 +228,9 @@ void QETProject::init()
 */
 QETProject::ProjectState QETProject::openFile(QFile *file)
 {
+	QElapsedTimer load_timer;
+	load_timer.start();
+
 	bool opened_here = file->isOpen() ? false : true;
 	if (!file->isOpen()
 			&& !file->open(QIODevice::ReadOnly
@@ -245,9 +249,18 @@ QETProject::ProjectState QETProject::openFile(QFile *file)
 		}
 		return XmlParsingFailed;
 	}
+	const qint64 xml_parse_ms = load_timer.elapsed();
 
 		//Build the project from the xml
 	readProjectXml(xml_project);
+
+	const qint64 total_ms = load_timer.elapsed();
+	qInfo().nospace().noquote()
+			<< "Project \"" << fi.fileName() << "\" ("
+			<< fi.size() / 1024 << " KiB) opened in "
+			<< total_ms / 1000.0 << " seconds (xml parsing "
+			<< xml_parse_ms / 1000.0 << ", content "
+			<< (total_ms - xml_parse_ms) / 1000.0 << ")";
 
 	if (!fi.isWritable()) {
 		setReadOnly(true);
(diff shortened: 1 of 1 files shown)


DieterMayerOSS:
Nice — this supersedes the simpler variant I had offered in #553 (total + parse/build split only); I'll drop mine in favor of this, the phase split inside `readProjectXml()` is clearly the right call, especially separating the graphics-scene work from the rest.

First Qt6 data point, since you can't run that side: Qt 6.11.1 / Windows 11 / MinGW, my own 12-folio project (726 KiB), measured with my coarser variant so only the top-level split is comparable:

```
total 1.344 s — xml parsing 0.038, content 1.306
```

Same lopsided ratio as your table (~3 % parse), which independently confirms the design decision.

Caveat: my machine ≠ your machine, so these numbers say nothing Qt5-vs-Qt6 by themselves. For the comparison this PR exists for, the best-placed setup is probably @scorpio810's, since he produces both the Qt5 releases and the Qt6 MSI on the same machine — with this merged, running one large project through both builds gives the #553 answer directly. Happy to feed more Qt6/Windows numbers for any project file people want measured.

DieterMayerOSS:
Ran your patch on the Qt6 side — it applied cleanly on my branch. Same three example files, Qt 6.11.1 / Windows 11 / MinGW, median of 3 runs each, via the headless CLI (`offscreen` platform):

| Project | Size | Total | XML parse | Diagrams | Refresh | Database |
|---|---|---|---|---|---|---|
| 741.qet | 123 KiB | 0.297 s | 0.006 | 0.264 | 0.001 | 0.005 |
| Habitat-Schemas_developpes.qet | 209 KiB | 0.470 s | 0.009 | 0.434 | 0.001 | 0.007 |
| weneedpolonez-…_wiring_diagram.qet | 3399 KiB | 1.346 s | 0.089 | 1.214 | 0.005 | 0.016 |

Two observations:

1. **Same shape as your Qt5 table** — diagrams dominate (~90 %), parsing is noise. The phase split does exactly what you built it for.
2. **Method caveat on the refresh column:** my runs are headless (offscreen QPA), and refresh comes out near-zero here vs. 0.097–0.196 s in your table. That difference may be platform *or* method (offscreen likely cheapens the refresh work). Whoever compares Qt5 vs Qt6 should use the same code path on both sides — GUI vs GUI or CLI vs CLI, not mixed.

Since these are different machines, none of this is Qt5-vs-Qt6 evidence by itself — but the tool works, the numbers are stable across runs (±2 %), and the phase design proves itself. LGTM from my side.

scorpio810:
Thanks @ispyisail @DieterMayerOSS merged, and pushed action Github Windows builds CI/CD.

DieterMayerOSS:
Heads-up for anyone testing the fresh Windows CI builds: in a GUI session the new timer lines currently go to stderr, not the daily log file — `qInstallMessageHandler()` runs in a startup worker that is scheduled after QETApp's constructor, i.e. after the whole startup including project loading has already happened. On Windows there is no console, so the lines are simply invisible. Fix is ready (install the handler synchronously before `SingleApplication`) and comes as a PR today; until then the headless CLI with `QT_FORCE_STDERR_LOGGING=1` shows them.

PR #562, opened 2026-07-27
PLC Manager

Kellermorph:
PLC-Master/PLC-Slave System for QElectroTech
Overview
Implementation of a PLC-Master/PLC-Slave linking system that allows users to define PLC I/O modules as master elements with an I/O table, which can be linked to slave elements in schematics. Includes Excel-compatible copy/paste, automatic variable propagation, and a dedicated linking GUI.

New Features
PLC-Master Element:

Master type "PLC" with I/O table (Type, Address, Function text, Comment, Cross-ref)
I/O table displayed as an in-circuit table with configurable column widths, fonts, column order, and headers
Multiple break positions for long tables (up to 4 page breaks)
Slave position updates in real-time when slaves are moved
I/O count up to 128 entries per PLC-Master
PLC-Slave Element:

Dedicated linking GUI (PlcLinkWidget) with tree view: PLC-Master > I/O entries
Each I/O entry shows Type, Address, Function, Comment, and Terminal count
Only one slave per I/O entry — already-linked entries are struck through and greyed out
PLC variables (type, address, function, comment, cross-ref) are automatically propagated from master to slave
Label is set on link and removed on unlink
Excel-Compatible Paste:

Copy/paste I/O data from/to the table via Ctrl+C / Ctrl+V
Supports both vertical and horizontal paste modes from clipboard
Bug Fixes
Cross-ref on PLC-Slave now shows the master's position (previously showed the slave's own position)
Cross-ref is properly cleared on unlink — fixed incorrect kindInformations()["type"] check (returned "commutator" instead of "plc")
PLC-Master/Slave no longer appear in the normal Master/Slave linking dialog — separate GUI prevents incorrect linking between PLC and non-PLC elements
MasterElement::unlinkElement() fixed — all kindInformations()["type"] == "plc" checks replaced with m_data.m_master_type == ElementData::PLC
Cross-ref for PLC-Master table column now correctly uses XRefProperties — configurable under Settings > New Folio > Cross References

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -331,6 +331,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/editor/graphicspart/partpolygon.h
   ${QET_DIR}/sources/editor/graphicspart/partrectangle.cpp
   ${QET_DIR}/sources/editor/graphicspart/partrectangle.h
+  ${QET_DIR}/sources/editor/graphicspart/partplctable.cpp
+  ${QET_DIR}/sources/editor/graphicspart/partplctable.h
   ${QET_DIR}/sources/editor/graphicspart/partterminal.cpp
   ${QET_DIR}/sources/editor/graphicspart/partterminal.h
   ${QET_DIR}/sources/editor/graphicspart/parttext.cpp
@@ -673,6 +675,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/ui/marginseditdialog.h
   ${QET_DIR}/sources/ui/masterpropertieswidget.cpp
   ${QET_DIR}/sources/ui/masterpropertieswidget.h
+  ${QET_DIR}/sources/ui/plclinkwidget.cpp
+  ${QET_DIR}/sources/ui/plclinkwidget.h
   ${QET_DIR}/sources/ui/multipastedialog.cpp
   ${QET_DIR}/sources/ui/multipastedialog.h
   ${QET_DIR}/sources/ui/potentialselectordialog.cpp
--- a/sources/autoNum/assignvariables.cpp
+++ b/sources/autoNum/assignvariables.cpp
@@ -275,6 +275,12 @@ namespace autonum
 
 		str.replace("%{void}", QString());
 
+		str.replace("%{plc_type}", dc.value("plc_type").toString());
+		str.replace("%{plc_address}", dc.value("plc_address").toString());
+		str.replace("%{plc_function}", dc.value("plc_function").toString());
+		str.replace("%{plc_comment}", dc.value("plc_comment").toString());
+		str.replace("%{plc_crossref}", dc.value("plc_crossref").toString());
+
 		return str;
 	}
 
--- a/sources/dataBase/projectdatabase.cpp
+++ b/sources/dataBase/projectdatabase.cpp
@@ -114,6 +114,11 @@ QSqlQuery projectDataBase::newQuery(const QString &query) {
 */
 void projectDataBase::addElement(Element *element)
 {
+	if (!element || !element->diagram()) {
+		qDebug() << "projectDataBase::addElement: null element or diagram";
+		return;
+	}
+
 	m_insert_elements_query.bindValue(":uuid", element->uuid().toString());
 	m_insert_elements_query.bindValue(":diagram_uuid", element->diagram()->uuid().toString());
 	m_insert_elements_query.bindValue(":pos", element->diagram()->convertPosition(element->scenePos()).toString());
--- a/sources/editor/elementscene.cpp
+++ b/sources/editor/elementscene.cpp
@@ -31,6 +31,7 @@
 #include "graphicspart/partline.h"
 #include "graphicspart/partpolygon.h"
 #include "graphicspart/partrectangle.h"
+#include "graphicspart/partplctable.h"
 #include "graphicspart/partterminal.h"
 #include "graphicspart/parttext.h"
 #include "ui/qetelementeditor.h"
@@ -84,12 +85,12 @@ ElementData ElementScene::elementData() {
 
 void ElementScene::setElementData(ElementData data)
 {
-	bool emit_ = m_element_data.m_informations != data.m_informations;
+	bool emit_info = (m_element_data != data);
 	bool type_changed = m_element_data.m_type != data.m_type;
 
 	m_element_data = data;
 
-	if (emit_)
+	if (emit_info)
 		emit elementInfoChanged();
 	if (type_changed)
 		emit elementTypeChanged();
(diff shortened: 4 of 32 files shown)


ispyisail:
Had a proper look through this — it is a lot of careful work, and it builds cleanly here (CMake, Qt 5.15.18, Release): **0 errors and 0 new warnings**. The four files that do warn were already warning on `master`.

One thing I would call blocking, plus a couple of questions. Everything else I checked looked sound.

## Blocking: unvalidated index in the column-names parser

`sources/properties/elementdata.cpp`, reading `<columnNames>`:

```cpp
int index = xml_col.attribute(QStringLiteral("index")).toInt();
while (m_plc_master_data.columnNames.size() <= index)
    m_plc_master_data.columnNames.append(QString());
m_plc_master_data.columnNames[index] = xml_col.text();
```

`columnNames` is a `QStringList` and `index` comes straight from the file with no validation:

- **`index="-1"`** — `while (0 <= -1)` is false, so the guard never runs and `columnNames[-1]` is written **out of bounds**.
- **`index="1000000000"`** — the loop appends a billion empty `QString`s (roughly 24 GB) before reaching the assignment.

I reproduced the parse verbatim in a standalone test under AddressSanitizer:

| Build | Result for `index="-1"` |
|---|---|
| asserts enabled | `ASSERT failure in QList<T>::operator[]: "index out of range"` |
| release | `ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000` |

Both QET build systems define `QT_NO_DEBUG` in release, so `Q_ASSERT_X` is compiled out and a shipping build gets the segfault rather than the assert. It is reachable from any `.elmt` or `.qet` carrying a PLC element — hand-edited, truncated, or simply received from someone else, which is how element collections normally circulate.

A bounds check in the same style you already use elsewhere is enough:

```cpp
int index = xml_col.attribute(QStringLiteral("index")).toInt();
if (index < 0 || index >= COL_COUNT) continue;   // or whatever the sane max is
```

**To be clear this reads as one spot that slipped, not a habit** — every `columnOrder` consumer is properly guarded:

```cpp
if (logical >= 0 && logical < COL_COUNT && …)     // partplctable.cpp:75
if (data_idx < 0 || data_idx > 4) continue;       // elementpropertieseditorwidget.cpp:916
```

and the *second* copy of this same parser, in `elementpicturefactory.cpp:1181`, uses `QMap<int, QString>` and is safe by construction. It is just this one site that got a `QStringList` and missed the check.

## Two questions

1. The description mentions *"I/O count up to 128 entries per PLC-Master"*, but I could not find that enforced anywhere — the only `128` I see is a spinbox maximum for break positions (`elementpropertieseditorwidget.cpp:742`). Is the limit intended? If so it would be worth applying it on the XML read path too, which is where it matters most.
2. `columnOrder` is appended from the file without a cap. The consumers all bounds-check the *values*, so nothing unsafe happens, but a crafted file could still make the list arbitrarily long. Probably worth a size cap alongside the fix above.

- **Ownership** — the 11 `new` calls in `plclinkwidget.cpp` are all either Qt-parented or pushed to the undo stack, and `associatedUndo()` follows the existing `PropertiesEditorWidget` contract, so the returned-pointer case is consistent with the rest of the codebase.
- **Backward compatibility** — the old `breakAfter` attribute is handled explicitly on read, which is easy to forget.
- `colWidths` / `colVisible` are `QMap<int, …>`, so unvalidated indices there are memory-safe; at worst a crafted file bloats them.

Happy to re-check once the index validation is in.

scorpio810:
Hallo Thomas, I dont time to check it, but can you provide an video howto?

Kellermorph:
> Had a proper look through this — it is a lot of careful work, and it builds cleanly here (CMake, Qt 5.15.18, Release): **0 errors and 0 new warnings**. The four files that do warn were already warning on `master`.
> 
> One thing I would call blocking, plus a couple of questions. Everything else I checked looked sound.
> 
> ## Blocking: unvalidated index in the column-names parser
> `sources/properties/elementdata.cpp`, reading `<columnNames>`:
> 
> ```c++
> int index = xml_col.attribute(QStringLiteral("index")).toInt();
> while (m_plc_master_data.columnNames.size() <= index)
>     m_plc_master_data.columnNames.append(QString());
> m_plc_master_data.columnNames[index] = xml_col.text();
> ```
> 
> `columnNames` is a `QStringList` and `index` comes straight from the file with no validation:
> 
> * **`index="-1"`** — `while (0 <= -1)` is false, so the guard never runs and `columnNames[-1]` is written **out of bounds**.
> * **`index="1000000000"`** — the loop appends a billion empty `QString`s (roughly 24 GB) before reaching the assignment.
> 
> I reproduced the parse verbatim in a standalone test under AddressSanitizer:
> 
> Build	Result for `index="-1"`
> asserts enabled	`ASSERT failure in QList<T>::operator[]: "index out of range"`
> release	`ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000`
> Both QET build systems define `QT_NO_DEBUG` in release, so `Q_ASSERT_X` is compiled out and a shipping build gets the segfault rather than the assert. It is reachable from any `.elmt` or `.qet` carrying a PLC element — hand-edited, truncated, or simply received from someone else, which is how element collections normally circulate.
> 
> A bounds check in the same style you already use elsewhere is enough:
> 
> ```c++
> int index = xml_col.attribute(QStringLiteral("index")).toInt();
> if (index < 0 || index >= COL_COUNT) continue;   // or whatever the sane max is
> ```
> 
> **To be clear this reads as one spot that slipped, not a habit** — every `columnOrder` consumer is properly guarded:
> 
> ```c++
> if (logical >= 0 && logical < COL_COUNT && …)     // partplctable.cpp:75
> if (data_idx < 0 || data_idx > 4) continue;       // elementpropertieseditorwidget.cpp:916
> ```
> 
> and the _second_ copy of this same parser, in `elementpicturefactory.cpp:1181`, uses `QMap<int, QString>` and is safe by construction. It is just this one site that got a `QStringList` and missed the check.
> 
> ## Two questions
> 1. The description mentions _"I/O count up to 128 entries per PLC-Master"_, but I could not find that enforced anywhere — the only `128` I see is a spinbox maximum for break positions (`elementpropertieseditorwidget.cpp:742`). Is the limit intended? If so it would be worth applying it on the XML read path too, which is where it matters most.
> 2. `columnOrder` is appended from the file without a cap. The consumers all bounds-check the _values_, so nothing unsafe happens, but a crafted file could still make the list arbitrarily long. Probably worth a size cap alongside the fix above.
> 
> ## Things I checked that are fine
> * **Ownership** — the 11 `new` calls in `plclinkwidget.cpp` are all either Qt-parented or pushed to the undo stack, and `associatedUndo()` follows the existing `PropertiesEditorWidget` contract, so the returned-pointer case is consistent with the rest of the codebase.
> * **Backward compatibility** — the old `breakAfter` attribute is handled explicitly on read, which is easy to forget.
> * `colWidths` / `colVisible` are `QMap<int, …>`, so unvalidated indices there are memory-safe; at worst a crafted file bloats them.
> 
> Happy to re-check once the index validation is in.

Thanks for the thorough review!

Blocking fix (columnNames parser): Fixed. Added bounds check if (index < 0 || index >= 5) continue; before the while-loop.

Question 1 (128 I/O limit): Yes, the limit is intentional and covers even the largest DDC systems. We've now enforced it on the XML read path as well: if (ios.size() >= 128) break; prevents loading more than 128 entries from any file.

Question 2 (columnOrder cap): Added a cap of 5 entries matching the fixed column count: if (columnOrder.size() >= 5) break;.

All three fixes are in sources/properties/elementdata.cpp. Happy for another look!

Kellermorph:
> Hallo Thomas, I dont time to check it, but can you provide an video howto?

yeah sure i will do if u guys say everything is fine.

ispyisail:
Re-checked the blocking issue against the new commit (3db2989) — it's fixed, verified rather than just read.

```cpp
int index = xml_col.attribute(QStringLiteral("index")).toInt();
if (index < 0 || index >= 5)
    continue;
```

- `5` matches `COL_COUNT` used elsewhere in this PR (`partplctable.h`/`.cpp`), so it's not an arbitrary number — consistent with the existing guarded consumers I compared against in my first pass.
- Re-ran the exact scenarios from my original PoC through the fixed logic (still under ASan): `index=-1` and `index=1000000000` now both no-op cleanly instead of crashing/allocating, while valid indices (0-4) still write correctly and `index=5` (one past the bound) is correctly rejected too. No more OOB write, no more unbounded-allocation path.
- The same commit also caps `columnOrder` at 5 entries and `ios` at 128 (matching the "up to 128 entries" design limit already mentioned in the PR description) — good additional hardening against unbounded growth from untrusted XML, beyond just the one path I'd flagged.
- Rebuilt `elementdata.cpp` standalone with `-Wall -Wextra`: zero warnings.

This resolves the blocking issue from my last review. Thanks for the quick turnaround, @Kellermorph — nothing else outstanding from my side.

ispyisail (approved):
Blocking issue from my earlier review is resolved and re-verified (see comment above). Approving.

Kellermorph:
https://qelectrotech.org/forum/viewtopic.php?pid=23124#p23124

scorpio810:
Thanks you Thomas.

Issue #552, opened 2026-07-23
[Feature Request] UI enhancement for displaying longer functional element descriptions in the Element Panel


joaocoutont:
### 🚀 Summary
We would like to propose a UI/visualization enhancement in QElectroTech's Element Panel (tree view/browser) to improve how longer, functionally-descriptive element names are rendered and previewed.
---
### 💡 Motivation & Background
In large element libraries (e.g., thousands of components across various manufacturers), element names are increasingly formatted to describe the component's **technical function** rather than relying solely on raw manufacturer part codes (e.g., *"Disjuntor-motor de proteção termomagnética (3P)"* instead of just *"GV2P"*).
While functional descriptions significantly improve component discoverability and usability for electrical engineers, longer element names can currently be truncated in the standard Element Panel tree view, making it harder to read the full description at a glance.
---
### 🛠️ Proposed UI Enhancements
To enable rapid identification of components by their technical function, we suggest considering the following improvements to QElectroTech's Element Panel:
1. **Enhanced Tooltips on Hover**: Displaying the full element name, category, and functional description in a multi-line tooltip when hovering over an item in the element tree.
2. **Flexible / Resizable Tree Columns**: Allowing users to expand or auto-fit the Element Panel width to easily read longer names without constant manual scrolling.
3. **Text Wrapping or Multi-line Item Rendering**: Option to wrap text or display multi-line titles in the element tree view.
4. **Full-Name Preview in Element Properties Panel**: A prominent full-name preview area at the bottom of the element selector panel.
---
###🎯 Expected Benefits
* **Faster Component Identification**: Engineers can instantly locate components by their function rather than searching by obscure part numbers.
* **Better Usability in Large Libraries**: Dramatically improves workflow efficiency when working with large or custom element collections.
* **Enhanced Multilingual Support**: Accommodates languages with longer technical descriptions (e.g., Portuguese, Spanish, German, French).
Thank you for your incredible work on QElectroTech! We appreciate your continuous dedication to making this project the best open-source CAD tool for electrical engineering.

ispyisail:
Thanks for the detailed write-up. The underlying need — finding components by function rather than by part code — is a real one, and it gets more real the bigger the collection.

I had a look at how the panel is currently built, and the four suggestions land quite differently. Two of them are largely available today, one is a small fix, and one I'd be cautious about. Ordering by value for effort rather than as a set:

## 2. Resizable panel — this already works

[...]

PR #571, opened 2026-07-29
Element panel: show name and element information in the tooltip

DieterMayerOSS:
Implements the tooltip fix from the discussion in #552 (thanks @ispyisail for the triage and the invitation to take it).

The element tooltip showed only the collection path — the least useful string exactly when a long descriptive name is truncated in the tree. It now shows the **localized name, description, manufacturer and manufacturer reference** (each only when set), with the collection path kept as the last line. Directories and `.qetmak` entries keep the plain path tooltip.

Implementation reuses the location/context that `setUpData()` already parses right above for the search index, so there is **no additional file access or parsing** — the concern about per-item cost at scale does not apply here.

GUI-verified against a library where every element carries these fields; hovering an element shows e.g.:

```
MEAN WELL HDR-100-24 - 1-reihig (Klemmenreihe)
MEAN WELL HDR-100-24 - Hutschienennetzteil 85-264VAC auf 24VDC, 92W, Schutzklasse II (kein PE)
MEAN WELL
HDR-100-24
custom://MeanWell/hdr_100_24/hdr_100_24_komplett_1reihig.elmt
```

Elements without the metadata degrade gracefully to name + path — which also makes the missing metadata visible, per the discussion.

### DIFF
--- a/sources/ElementsCollection/fileelementcollectionitem.cpp
+++ b/sources/ElementsCollection/fileelementcollectionitem.cpp
@@ -327,6 +327,26 @@ void FileElementCollectionItem::setUpData()
 			{ search_list.append(context.value(key).toString()); }
 			search_list.append(localName(loc));
 			setData(search_list.join(" "));
+
+			// Tooltip: show what a truncated tree label can't - the full
+			// localized name and the descriptive element information.
+			// Reuses the location/context already parsed for the search
+			// index above; the collection path stays as the last line
+			// (it used to be the whole tooltip).
+			QStringList tip;
+			tip << localName(loc);
+			for (const auto &key : { QStringLiteral("description"),
+									 QStringLiteral("manufacturer"),
+									 QStringLiteral("manufacturer_reference") })
+			{
+				const QString value = context.value(key).toString();
+				if (!value.isEmpty())
+					tip << value;
+			}
+			tip << collectionPath();
+			tip.removeDuplicates();
+			setToolTip(tip.join(QLatin1Char('\n')));
+			return;
 		}
 	}
 


ispyisail (approved):
Reviewed this against `master` and a local build.

- The early `return;` is scoped correctly: it only sits inside the "real element" branch (after the `.qetmak` check), so it bypasses the old unconditional `setToolTip(collectionPath())` at the bottom of `setUpData()` only for genuine elements — dirs and `.qetmak` entries still fall through to the plain path tooltip, exactly as described.
- `"description"`, `"manufacturer"`, `"manufacturer_reference"` are the correct `DiagramContext` keys (match `ELMT_DESCRIPTION`/`ELMT_MANUFACTURER`/`ELMT_MANUFACTURER_REF` in `qetinformation.h`, and the same raw-literal convention used in `assignvariables.cpp`).
- `DiagramContext::value()` returns a default `QVariant` for a missing key, so `.toString()` is safely empty for elements lacking these fields — no crash risk, confirms the "degrades gracefully" claim.
- No added file access or parsing, confirmed — `context`/`loc` are the same objects already built for the search index just above.
- Compiled `fileelementcollectionitem.cpp` standalone against this branch (Qt5/pugixml headers wired up locally): zero errors, zero warnings.

Looks correct and well-scoped. Approving.

DieterMayerOSS:
Thanks for the thorough review - especially for double-checking the early-return scoping and compiling it against Qt5, that's exactly the compatibility concern I wanted covered. Glad the `DiagramContext` reuse holds up; no extra parsing was the main design goal here.

PR #585, opened 2026-07-31
Diagram: Tab/Shift+Tab item-selection cycling + select-all-conductors/text-fields

ispyisail:
Implements the second pillar of #574: keyboard-driven selection on the diagram canvas.

**What it does**

- `Tab` / `Shift+Tab` select the next / previous item on the current diagram, cycling through `items()` (z-order) and wrapping at either end. If nothing is selected, `Tab` selects the first item and `Shift+Tab` the last. Skipped while a text item has focus, for the same reason arrow-key movement already guards on `!focusItem()`.
- Candidates use the same "what counts as a real selectable diagram item" filter (`QetGraphicsItem` / `DiagramTextItem` / `Conductor`) already established by `Diagram::invertSelection()`, so the cycling order always matches what a user could reach by clicking.
- New `Diagram::selectAllConductors()` / `selectAllTextFields()`, wired up as two new actions in the existing `select_all` / `select_nothing` / `select_invert` action group in `qetdiagrameditor.cpp` — they appear in the Edit menu and go through the same `QAction` → `data()` → `selectGroupTriggered()` dispatch as the existing selection commands.

**Why getting Tab to actually reach the scene needed two separate fixes**

Both were found by testing, not assumption — the feature silently did nothing on the first build, and I dug into *why* rather than guessing:

1. `QWidget` (`DiagramView`) intercepts Tab/Backtab for widget focus-chain traversal before a key event is even generated. Overriding `DiagramView::focusNextPrevChild()` to return `false` disables that.
2. `QGraphicsScene` (`Diagram`) has its own, *separate* item-focus-chain traversal, checked before `keyPressEvent()` is ever reached. The obvious fix — overriding `Diagram::focusNextPrevChild()` the same way — silently does nothing on Qt 5, because `QGraphicsScene::focusNextPrevChild()` only becomes virtual in Qt 6 (guarded by the `QT6_VIRTUAL` macro). A compile error surfaced this immediately when I tried it directly, rather than shipping something that worked on Qt 6 and silently no-opped on Qt 5. Intercepting `QEvent::KeyPress` in `Diagram::event()` instead is virtual on every Qt version and sidesteps the scene's internal traversal entirely.

**Testing**

Beyond compiling clean (full CMake build, zero new warnings), verified end-to-end in a real running session (Xvfb + xdotool) against a multi-transistor schematic:
- `Tab`/`Shift+Tab` correctly move a single selection forward/backward through elements and text fields — confirmed via the properties panel updating to each new item and the visual selection box moving on canvas.
- `Tab`/`Shift+Tab` from no selection correctly select the first/last item.
- "Select all conductors" and "Select all text fields" each correctly select every matching item and deselect everything else.


scorpio810:
Please add a video to explain this.
This is to keep our users up to date with the latest news and to promote the QET YouTube channel.

ispyisail:
@scorpio810 apologies for the long silence on your video request — that was on me, not on the branch.

Before the video, I have brought the PR up to date and fixed something that went stale underneath it.

## Updated

- **Merged current `master` in.** The branch was 382 commits behind and had not been touched since 1 August. It merges clean, builds clean, and the 22 loadable projects in `examples/` still load and export with no crash or hang.
- **Registered the two new actions with `ShortcutManager`.** This branch was cut on 31 July, one day before `ShortcutManager` landed in 5275fb44f, so *Select all conductors* and *Select all text fields* were written before that convention existed. They were the only members of the selection group not registered — `select_all`, `select_nothing` and `select_invert` all are. Without it they never appeared in the shortcuts configuration page and no key could be bound to them.

  Both are registered with an empty default sequence: they are menu actions and neither has an obvious default worth claiming, so the point is simply that a user *can* bind one.

## What the PR does, in case a written description helps

**Tab / Shift+Tab cycles the selection** through the items on the current folio, forwards and backwards. It only does so when nothing else has focus, so it does not interfere with typing in a text field.

Two implementation notes that may be of interest, since both were traps:

- `QGraphicsScene` has its own Tab focus-chain traversal that runs *before* `keyPressEvent()`, so the key had to be intercepted in `Diagram::event()`. Overriding `QGraphicsScene::focusNextPrevChild()` would have been the obvious route, but it is only virtual from Qt6 onward, so on a Qt5 build it would have silently done nothing.
- `QWidget` also consumes Tab for focus traversal before any key event, hence `DiagramView::focusNextPrevChild()` returning false. One consequence worth stating plainly: Tab no longer moves keyboard focus *out* of the diagram view. F6 still cycles the docks, so there is still a keyboard route out, but if you would rather Tab kept its usual focus behaviour and the cycling moved to another key, that is an easy change and I would rather know now than after merge.

**Select all conductors** and **Select all text fields** are two new entries in the selection menu: each selects every item of that kind on the folio and deselects everything else. The text-field one covers independent texts, conductor labels and dynamic element texts.

## On the demo

I am not able to put a video together for this one, sorry. The written walkthrough above is what I can offer — happy to expand any part of it, add screenshots of the two new menu entries, or describe the behaviour in more detail if that would help explain it to users.

ispyisail:

PR #586, opened 2026-07-31
Add Ctrl+G "jump to element" quick-open popup

ispyisail:
Implements the third pillar of #574: a keyboard-driven quick-open for locating an element on a diagram by typing part of its label, type name, or other info.

**What it does**

- `Ctrl+G` opens a small, transient `JumpToElementDialog` over the current diagram — a `QLineEdit` above a live-filtered `QListWidget`.
- Typing filters the list against every element on the diagram, case-insensitively, matching against the element's instance label (`elementInformations().value("label")`), its type name (`Element::name()`), and every other `elementInformations()` value — so partial matches on any of those surface the element.
- `Up`/`Down` move through the filtered results (wrapping at either end), `Enter` selects the highlighted element on the diagram, calls `ensureVisible()` to scroll it into view, and closes the popup; `Escape` cancels without changing the current selection.
- The dialog is disposable — built fresh from `m_diagram->items()` each time it's opened, so it always reflects the diagram's current state with no persistent list to keep in sync.

**Design notes**

- Candidates are built once on open (`buildCandidates()`) rather than re-scanning `items()` on every keystroke, so filtering a live-typed query stays cheap even on larger diagrams.
- `QPointer<Element>` guards each candidate against the (unlikely but possible) case of an element being deleted out from under the dialog while it's open; `activateCurrentItem()` and `updateFilteredList()` both check it before dereferencing.
- Reuses `QDialog`'s standard `eventFilter()` override on the line edit rather than subclassing `QLineEdit`, keeping the arrow/enter/escape handling colocated with the rest of the dialog's logic in one file.

**Testing**

Beyond compiling clean (full CMake build, zero new warnings), verified end-to-end in a real running session (Xvfb + xdotool) against a multi-transistor schematic:
- `Ctrl+G` opens the popup with the full element list and focus on the text field.
- Typing live-filters the list correctly against label, type name, and element info fields.
- `Up`/`Down` move the highlighted row and wrap at both ends.
- `Enter` selects the correct element on the diagram, scrolls it into view, and closes the popup.
- `Escape` closes the popup without altering the diagram's prior selection.

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -697,6 +697,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/ui/importelementdialog.h
   ${QET_DIR}/sources/ui/importelementtextpatterndialog.cpp
   ${QET_DIR}/sources/ui/importelementtextpatterndialog.h
+  ${QET_DIR}/sources/ui/jumptoelementdialog.cpp
+  ${QET_DIR}/sources/ui/jumptoelementdialog.h
   ${QET_DIR}/sources/ui/inditextpropertieswidget.cpp
   ${QET_DIR}/sources/ui/inditextpropertieswidget.h
   ${QET_DIR}/sources/ui/linksingleelementwidget.cpp
--- a/sources/qetdiagrameditor.cpp
+++ b/sources/qetdiagrameditor.cpp
@@ -39,6 +39,7 @@
 #include "qetmessagebox.h"
 #include "recentfiles.h"
 #include "ui/bomexportdialog.h"
+#include "ui/jumptoelementdialog.h"
 #include "ui/diagrampropertieseditordockwidget.h"
 #include "ui/backupdialog.h"
 #include "ui/dialogwaiting.h"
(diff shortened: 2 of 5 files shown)


scorpio810:
@ispyisail can you add an little video howto?

scorpio810:
You could use also CTRL + F

ispyisail:
If you accecpt my other PR you can select what ever key you want :)

scorpio810:
https://youtu.be/eOuN7lGf5xQ

https://youtu.be/AJjCRItDieo

scorpio810:
This search and replace function already exists in a more comprehensive form; that’s what I meant. ;-)

ispyisail:
@scorpio810 fair point, and you're right that on raw search coverage, "Chercher/remplacer" is a superset of what this dialog matches — it also picks up dynamic text items and text-group names, and it searches every folio in the project, not just the current one.

Where I think this still earns its place is interaction cost rather than search coverage. To jump to one element today: open the search panel, expand the right category node in the tree (folio → simple/master/slave/report/terminal), scan for it, double-click. `Ctrl+G` skips all of that — type a few characters, arrow to the match, `Enter`, done, and the popup closes itself.

That's the same split most editors make deliberately: a "Go to Symbol / Quick Open" alongside a heavier "Find in Project," even though the heavier tool can technically find anything the quick one can — because they serve different reflexes (a few keystrokes vs. a deliberate search-and-replace session).

If that distinction doesn't feel worth a second search UI to you, I understand closing it — just wanted to lay out the actual reasoning rather than leave it at "it's a bit different."

scorpio810:
Just enter ctrl + F and enter an text, and label, or number like 29/04 it' s found all conductors quickly in the project.
You could try it with my example with breaks fonts.

BTW, i will download the patch and try it.

PR #588, opened 2026-07-31
Fix-PLC-Manager

Kellermorph:
Summary of 5 Fixes
1. Terminal count and values not propagated to slave
Problem: terminalCount and terminals (T1/T2/T3/T4) were not transferred from PLC-Master to PLC-Slave.

Fix (5 files):

qetinformation.h — New constants ELMT_PLC_TC, ELMT_PLC_T1–T4
assignvariables.cpp — Variables %{plc_tc}, %{plc_t1}–%{plc_t4} available
dynamictextfieldeditor.cpp — New keys in PLC-Slave dropdown
linkelementcommand.cpp — Propagate terminal values to slave when linking
element.cpp — Propagate terminal values to slaves on master update
2. Terminal name resets when editing master
Problem: MasterPropertiesWidget::plcUpdateDisplaySettings() rebuilt IOs from the table, but the table had no terminalCount or terminals columns. New PlcIO objects got terminalCount=0 and empty terminals → all terminal data was lost on every edit.

Fix:

masterpropertieswidget.cpp — Preserve terminalCount and terminals from the original IO before reading from table
3. BMK/label not updating on slave
Problem: setElementInformations() had no PLC-slave propagation. When the user only changed the label (not IOs), setElementData() was not called and the change never reached the slave. Also, sorting by name() broke after linking because it returned master labels instead of terminal names.

Fix (3 files):

element.cpp::setElementInformations() — Propagate label to all linked PLC-slaves
linkelementcommand.cpp — Sort using baseName() instead of name() to avoid master-label conflict
terminal.h/.cpp — New baseName() method always returns the original name regardless of master label
4. Column width limited and no text wrapping
Problem: Column width capped at 200mm, long text disappeared instead of wrapping.

Fix (2 files):

elementpropertieseditorwidget.cpp — Increased maximum from 200 to 500
element.cpp::drawPlcTable() — Added Qt::TextWordWrap to drawText() call
5. Terminal label alignment not working
Problem: In Terminal::paint(), alignment was hardcoded: non-rotated path used AlignLeft|AlignTop, rotated path centered the text rect regardless of alignment. Since most PLC-slave terminals are rotated 270°, alignment never worked.

Fix:

terminal.cpp — Both paths (rotated + non-rotated) now calculate text rect position based on m_label_halignment / m_label_valignment

### DIFF
--- a/sources/autoNum/assignvariables.cpp
+++ b/sources/autoNum/assignvariables.cpp
@@ -280,6 +280,11 @@ namespace autonum
 		str.replace("%{plc_function}", dc.value("plc_function").toString());
 		str.replace("%{plc_comment}", dc.value("plc_comment").toString());
 		str.replace("%{plc_crossref}", dc.value("plc_crossref").toString());
+		str.replace("%{plc_tc}", dc.value("plc_tc").toString());
+		str.replace("%{plc_t1}", dc.value("plc_t1").toString());
+		str.replace("%{plc_t2}", dc.value("plc_t2").toString());
+		str.replace("%{plc_t3}", dc.value("plc_t3").toString());
+		str.replace("%{plc_t4}", dc.value("plc_t4").toString());
 
 		return str;
 	}
--- a/sources/dataBase/projectdatabase.cpp
+++ b/sources/dataBase/projectdatabase.cpp
@@ -388,9 +388,15 @@ void projectDataBase::createElementNomenclatureView()
 						 "ei.supplier_auxiliary4 AS supplier_auxiliary4,"
 						 "ei.quantity_auxiliary4 AS quantity_auxiliary4,"
 						 "ei.unity_auxiliary4 AS unity_auxiliary4,"
-						 "ei.exclude_from_bom AS exclude_from_bom,"
-						
-						 "d.pos AS diagram_position,"
+					 "ei.exclude_from_bom AS exclude_from_bom,"
+					 
+					 "ei.plc_type AS plc_type,"
+					 "ei.plc_address AS plc_address,"
+					 "ei.plc_function AS plc_function,"
+					 "ei.plc_comment AS plc_comment,"
+					 "ei.plc_crossref AS plc_crossref,"
+					
+					 "d.pos AS diagram_position,"
 						 "e.type AS element_type,"
 						 "e.sub_type AS element_sub_type,"
 						 "di.title AS title,"
--- a/sources/dataBase/ui/elementquerywidget.cpp
+++ b/sources/dataBase/ui/elementquerywidget.cpp
@@ -47,6 +47,7 @@ ElementQueryWidget::ElementQueryWidget(QWidget *parent) :
 	m_button_group.addButton(ui->m_coil_cb, 4);
 	m_button_group.addButton(ui->m_protection_cb, 5);
 	m_button_group.addButton(ui->m_thumbnail_cb, 6);
+	m_button_group.addButton(ui->m_plc_cb, 7);
 #if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)	// ### Qt 6: remove
 	connect(&m_button_group, static_cast<void (QButtonGroup::*)(int)>(&QButtonGroup::buttonClicked), [this](int id)
 #else
@@ -76,7 +77,7 @@ ElementQueryWidget::ElementQueryWidget(QWidget *parent) :
 		else
 		{
 			int checked = 0;
-			for (int i=1 ; i<7 ; ++i) {
+			for (int i=1 ; i<8 ; ++i) {
 				if (m_button_group.button(i)->isChecked()) {++checked;}
 			}
 
@@ -85,7 +86,7 @@ ElementQueryWidget::ElementQueryWidget(QWidget *parent) :
 				case 0 :
 					check_box->setCheckState(Qt::Unchecked);
 					break;
-				case 6:
+				case 7:
 					check_box->setCheckState(Qt::Checked);
 					break;
 				default:
@@ -187,12 +188,16 @@ void ElementQueryWidget::setQuery(const QString &query)
 				if (ui->m_protection_cb) {
 					++c;
 				}
-				ui->m_thumbnail_cb->setChecked  (str_type.contains(ElementData::typeToString(ElementData::Thumbnail)) ? true : false);
-				if (ui->m_thumbnail_cb->isChecked()) {
-					++c;
-				}
+			ui->m_thumbnail_cb->setChecked  (str_type.contains(ElementData::typeToString(ElementData::Thumbnail)) ? true : false);
+			if (ui->m_thumbnail_cb->isChecked()) {
+				++c;
+			}
+			ui->m_plc_cb->setChecked       (str_type.contains(ElementData::masterTypeToString(ElementData::PLC)) ? true : false);
+			if (ui->m_plc_cb->isChecked()) {
+				++c;
+			}
 
-				if (c == 6) {
+			if (c == 7) {
 					ui->m_all_cb->setCheckState(Qt::Checked);
 				} else if (c > 0) {
 					ui->m_all_cb->setCheckState(Qt::PartiallyChecked);
@@ -366,6 +371,11 @@ QString ElementQueryWidget::queryStr() const
 	if (ui->m_protection_cb->isChecked()) {
 		if (b) where +=" OR";
 		where +=  QStringLiteral(" element_sub_type = '") += ElementData::masterTypeToString(ElementData::Protection) += "'";
+		b = true;
+	}
+	if (ui->m_plc_cb->isChecked()) {
+		if (b) where +=" OR";
+		where +=  QStringLiteral(" element_sub_type = '") += ElementData::masterTypeToString(ElementData::PLC) += "'";
 	}
 	where.append(")");
 
--- a/sources/dataBase/ui/elementquerywidget.ui
+++ b/sources/dataBase/ui/elementquerywidget.ui
@@ -234,86 +234,96 @@
        <number>0</number>
       </property>
       <item>
-       <widget class="QGroupBox" name="groupBox_3">
-        <property name="title">
-         <string>Type d'éléments</string>
-        </property>
-        <property name="alignment">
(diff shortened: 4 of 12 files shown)


ispyisail:
Reviewed the diff against the current call graph (traced `setElementData()`/`setElementInformations()` callers and the PLC IO table's row-management actions). Three issues, in order of severity:

## 1. Terminal data preservation is keyed by table row *position*, not IO *identity* — breaks on reorder/delete

`MasterPropertiesWidget::plcUpdateDisplaySettings()`:

```cpp
if (row < ed.plcMasterData().ios.size()) {
    const auto &orig_io = ed.plcMasterData().ios.at(row);
    io.terminalCount = orig_io.terminalCount;
    io.terminals = orig_io.terminals;
}
```

`ed` is the master's currently-persisted data, fetched fresh each call. `plcMoveRowUp()`/`plcMoveRowDown()`/`plcRemoveRow()` reorder or remove rows by swapping/removing `QTableWidgetItem`s in place, and none of them touch `ed.plcMasterData().ios`'s order. Since `m_plc_table`'s `cellChanged` is wired directly to `plcIOCellChanged()` → `plcUpdateDisplaySettings()`, every `setItem()` call inside a row swap re-triggers this preserve logic mid-move and copies `terminalCount`/`terminals` **by position** onto whatever IO now occupies that row.

Concretely: click "Move Row Up" on IO #2 — its address/function/comment move up correctly, but its T1–T4 terminal data stays behind and gets attached to IO #1 instead (and vice versa). Deleting a non-last row causes the same off-by-one shift for every row below it. This silently corrupts the exact terminal↔IO association that bug #2 in this PR set out to fix — it only holds for pure in-place field edits, not for row reordering/removal, both of which are exposed UI actions in this same dialog.

## 2. Word-wrapped cell text overflows its fixed-height row

`Element::drawPlcTable()`:

```cpp
QRectF cr(cx, ry, col_widths[col], row_h);   // row_h is one constant, same for every row
...
painter->drawText(text_rect, Qt::AlignLeft | Qt::AlignVCenter | Qt::TextWordWrap, cell_text);
```

`row_h` is never recomputed from wrapped line count. Combined with the column-width cap going from 200→500mm in this same PR, a moderately long comment/function string now wraps to 2+ lines whose combined height exceeds `row_h`; `AlignVCenter` then centers that taller block within the (too-short) rect, so the overflow bleeds both upward and downward into neighboring rows' borders/text instead of being clipped. This trades bug #4's original "text disappears" for "text overlaps the adjacent row."

## 3. `plc_changed` / `label_changed` are treated as mutually exclusive in `setElementData()`

```cpp
if (plc_changed) { ... /* ELMT_LABEL is no longer added here */ ... }
else if (label_changed) { ctx.addValue(ELMT_LABEL, actualLabel()); ... }
```

Pre-patch, the label was unconditionally included in the propagated `ctx` whenever `plc_changed`. This PR moved that into an `else if`, so a future (or currently-unexercised) call path that changes IOs and the label in the same `setElementData()` call would silently stop propagating the label — regressing the exact bug this PR fixes. I checked every current caller (`MasterPropertiesWidget`, `ChangeElementDataCommand`/TerminalStrip editors) and none currently hits this combination, so it's not user-reproducible today, but it's a one-line latent trap worth closing (e.g. two independent `if`s instead of `if`/`else if`).

**Suggested priority**: #1 is the one I'd block merge on — it's a real data-corruption path in the exact dialog whose job is managing/reordering these rows. #2 is a visible regression worth fixing before merge (clip, or size rows to wrapped content). #3 is a latent issue, fine to note but not blocking.

scorpio810:
@Kellermorph @ispyisail Ask me when ready to merge.

ispyisail (changes requested):
Reviewed the full diff plus the surrounding code paths that exercise it (traced call sites via `Kellermorph/qelectrotech-source-mirror@Fix-PLC-Manager`). The five fixes are individually well-motivated, but three of the changes have concrete correctness gaps:

## 1. Label update silently dropped when label + PLC data change together
`sources/qetgraphicsitem/element.cpp`, `Element::setElementData()`:

```cpp
if (plc_changed) {
    ... // propagates PLC_TYPE/ADDRESS/FUNCTION/COMMENT/CROSSREF/TC/T1-T4
    slave->setElementInformations(ctx);   // no ELMT_LABEL added here
}
else if (label_changed) {
    ctx.addValue(QETInformation::ELMT_LABEL, actualLabel());
    slave->setElementInformations(ctx);
}
```

Before this PR, the label was added unconditionally alongside the PLC fields in this block. Now it's mutually exclusive with `plc_changed` — any `setElementData()` call that changes IOs *and* the label in the same call (e.g. `ChangeElementDataCommand::redo()`/`undo()`, which swaps in a full pre/post `ElementData` snapshot, used by `freeterminaleditor.cpp` and `terminalstripeditor.cpp`) will update the slave's PLC fields but silently skip the label update. Suggest two independent `if`s rather than `if/else if`, with the `ELMT_LABEL` line present in both branches.

## 2. Terminal-name preservation keyed to raw table row index breaks under row reorder/remove
`sources/ui/masterpropertieswidget.cpp`, `plcUpdateDisplaySettings()`:

```cpp
if (row < ed.plcMasterData().ios.size()) {
    const auto &orig_io = ed.plcMasterData().ios.at(row);
    io.terminalCount = orig_io.terminalCount;
    io.terminals = orig_io.terminals;
}
```

This reads terminal data from the last-saved `ios` list by positional index. But `plcMoveRowUp()`/`plcMoveRowDown()` (pre-existing) swap cell widgets/items between rows directly in the `QTableWidget`, without touching the element's stored `ios` order. The next time `plcUpdateDisplaySettings()` runs, it reads the table in its new order but attaches terminal data from the old order at the same index — reattaching T1–T4 names to the wrong IO row after any move (or after `plcRemoveRow()`, which shifts every row below the removed one up by one). This reintroduces essentially the bug fix #2 is meant to solve, just via a different trigger (reorder/remove instead of a plain edit).

## 3. `Qt::TextWordWrap` added without growing row height to match
`sources/qetgraphicsitem/element.cpp`, `drawPlcTable()`:

Row height is a fixed constant (`row_h`, default 8.0) used both for each cell's border rect and for the next row's y-offset (`ry = pos.y() + header_h + row * row_h`). Wrapped multi-line text in a cell now renders past the bottom of that fixed-height rect — `drawText(rect, ...)` doesn't clip or expand the rect — so long comments/function text will visually overlap the row below instead of being contained. Raising the column-width cap 200→500 reduces how often this triggers but doesn't eliminate it.

---

Minor, non-blocking: the `t < 4` terminal cap is duplicated as a magic number in three places (`element.cpp` ×2, `linkelementcommand.cpp`) tied to the four new `ELMT_PLC_T1..T4` constants — fine for now, but worth a named constant if a 5th terminal is ever needed.

Kellermorph:
Thanks for the thorough review! All three points are valid. Here's the status:

1. Label dropped when PLC + label change together — Already fixed in our working copy. The else if (label_changed) has been changed to if (label_changed), and ELMT_LABEL is now set in both the plc_changed and label_changed branches. Will be included in the next push.
2. Terminal-name preservation keyed to raw table row index — Good catch. The m_plc_updating guards in plcMoveRowUp/Down/Remove prevent spurious cellChanged signals, but you're right that the positional index lookup orig_io = ed.plcMasterData().ios.at(row) will still reattach T1–T4 names to the wrong IO after a row reorder or removal. We need to match terminal data by a stable identifier (e.g. address) rather than by index. This fix is still in progress.
3. TextWordWrap without growing row height — Acknowledged. The setClipRect fix prevents visual overflow into adjacent rows, but wrapped text gets clipped rather than displayed fully. The row height is a fixed constant used for layout calculations throughout drawPlcTable, so growing it dynamically would be a larger change. For now the clipping approach is intentional — long text is contained within its cell. If full word-wrap display is desired later, we'd need to pre-calculate the wrapped text height per row in a layout pass before drawing.

I also need to add the PLC to the CSV export and the materials list. I had forgotten to do that.

ispyisail:
I will wait

Kellermorph:
Done. I hope everthing is fine now.

ispyisail (approved):
Re-reviewed the latest push (`4c1f82f8` → `e91bab14`). All three original findings are addressed:

1. **Label drop on combined edits** — fixed: `else if (label_changed)` → `if (label_changed)`, plus `ELMT_LABEL` is now also set directly in the `plc_changed` branch. Both paths update the label now.
2. **Row-height/wordwrap overlap** — fixed via `painter->setClipRect(text_rect, Qt::IntersectClip)` around the `drawText` call. Long text now gets cropped to the cell instead of bleeding into the row below — a reasonable resolution (truncation over corruption).
3. **Terminal-name misattribution on reorder/remove** — fixed, and well: `plcUpdateDisplaySettings()` now keys the lookup by `io.address` via a fresh `addr_to_orig_idx` map instead of raw row position, and `plcMoveRowUp/Down`/`plcRemoveRow` now call it after the swap/removal so the resync actually happens.

One residual, non-blocking edge case worth being aware of: the new address-based lookup assumes `address` is unique per row. `plcAddRow()` leaves a new row's address blank by default, so two freshly-added-but-not-yet-configured rows (or any two rows a user leaves with the same address) will collide in `addr_to_orig_idx` — the last one inserted wins, so the earlier row silently inherits the wrong terminal data on the next edit-triggered resync. Not a regression from this PR's baseline (previous behavior was strictly worse - guaranteed-wrong on any reorder), so I wouldn't block on it, but worth a follow-up if it comes up in practice — e.g. falling back to row-position matching only among rows that share a blank/duplicate address.

Minor nitpick, also non-blocking: in `ElementQueryWidget::queryStr()`, the new PLC checkbox branch is missing the `b = true;` that every sibling branch (Coil, Protection, etc.) sets after appending its `OR` clause. Currently harmless since PLC is the last condition in the chain, but would silently produce a malformed `WHERE` (missing `OR`) if another filter is ever added after it.

Also confirmed the PLC-nomenclature-view / query-widget changes (exposing `plc_type`/`plc_address`/`plc_function`/`plc_comment`/`plc_crossref` in `element_nomenclature_view` and the new PLC filter checkbox) are in-scope, not stray — they extend the same PLC data flow the rest of the PR fixes.

Nice work addressing all three - approving.

PR #582, opened 2026-07-30
Write font descriptions in a version-stable format (#553)

DieterMayerOSS:
This is the write-side half of the font-format problem analyzed in #553 (complementing @ispyisail's proposed read-side `fromString()` check): `QFont::toString()` changed to a 19-field format in Qt 6.11 which `QFont::fromString()` of Qt 5.x and Qt <= 6.10 rejects, so projects saved by a Qt 6.11+ build broke on older builds.

This PR adds `QETUtils::fontToString()`, composing the legacy 10/11-field description ourselves (weight mapped back to the legacy scale using the same closest-match table Qt applies when parsing), and uses it at every site that stores a font description — project XML (static/dynamic/independent texts, terminal labels, project-db tables), element editor XML, table config JSON and the shared QSettings. Every Qt version from 5.15 through 6.12.0-beta1 parses this form correctly, verified with standalone probes and an end-to-end check: on a Qt 6.11 build, autosave of a real project now writes 53/53 font attributes in the 11-field form (it was the 19-field form before).

On Qt 5 builds the helper simply forwards to `QFont::toString()`, which already emits this format.

PR #592, opened 2026-08-01
Report salvaged and unreadable font descriptions after opening a project

DieterMayerOSS:
Follow-up to #582, implementing what @scorpio810 asked for in https://github.com/qelectrotech/qelectrotech-source-mirror/pull/582#issuecomment-5140089732: font problems used to surface only as console warnings most users never see, so nobody learned that their texts silently lost their formatting.

`QETUtils::fontFromString()` now counts per project load how many descriptions were salvaged from a foreign or corrupt format and how many stayed unreadable, and after opening a project a message box reports both — salvaged descriptions get rewritten in the stable format on the next save, unreadable ones fall back to the default font. Projects without font issues open exactly as before, and non-interactive opens only log the counters.

Verified with a Qt 5.15 build on a project carrying 52 19-field and one 21-field description: the dialog reports "53 description(s) de police écrite(s) dans un format étranger ou corrompu ont été restaurée(s)…". The same file opened with a Qt 6.11 build shows no dialog, since that Qt parses those formats natively.

### DIFF
--- a/sources/qetdiagrameditor.cpp
+++ b/sources/qetdiagrameditor.cpp
@@ -43,6 +43,7 @@
 #include "ui/backupdialog.h"
 #include "ui/dialogwaiting.h"
 #include "undocommand/addelementtextcommand.h"
+#include "utils/qetutils.h"
 #include "undocommand/rotateselectioncommand.h"
 #include "undocommand/rotatetextscommand.h"
 #include "diagram.h"
@@ -1168,6 +1169,11 @@ bool QETDiagramEditor::openAndAddProject(
 	//Create the project
 	DialogWaiting::instance(this);
 
+		//Per-project window for the font counters reported below; the folios
+		//(and with them the stored font descriptions) are built between here
+		//and the end of addProject().
+	QETUtils::resetFontRestorationCounters();
+
 	QETProject *project = new QETProject(filepath);
 	if (project -> state() != QETProject::Ok)
 	{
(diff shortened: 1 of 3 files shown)


ispyisail:
Reviewed the diff against the actual call graph (traced every `fontFromString()` call site and how `DialogWaiting` drives the event loop during project load). One thing worth crediting, one finding.

## Good catch: the empty-description guard is load-bearing, not decorative

```cpp
if (description.trimmed().isEmpty()) {
    return false;
}
```

Several call sites pass `xml_element.attribute("font")` straight through with no default — `independenttextitem.cpp:68`, `dynamicelementtextitem.cpp:171`, `parttext.cpp:129`, `partdynamictextfield.cpp:218` — and an absent `font` attribute (the common case: most elements don't override their font) yields an empty string there. Without this guard, `QFont::fromString("")` fails, falls through the salvage branches, and increments `unreadable_font_count` — meaning ordinary projects with zero real font issues would report dozens of false "unreadable font description" hits. Good that this was caught before it shipped.

## Finding: the counters aren't safe against re-entrant `openAndAddProject()` calls, and this codebase's own event-pump pattern makes that plausible

`salvaged_font_count`/`unreadable_font_count` are anonymous-namespace globals, reset at the top of `openAndAddProject()` and read at the bottom. That's safe only if project loads never interleave — but they can, on the same thread:

- `DialogWaiting` calls `qApp->processEvents()` on every progress update (`dialogwaiting.cpp:52`).
- `QETProject`'s XML parser calls `dlgWaiting->setProgressBar(...)` once per diagram/folio while loading (`qetproject.cpp:294`, `:1586`).

So loading a multi-folio project pumps the event loop many times over its duration, and any of those pumps can dispatch a *second* `openAndAddProject()` call to completion (drag-and-drop onto a different editor window, a queued "Open Recent" click, etc.) — entirely nested inside the first call's pump. When the first call resumes and reads the counters at the bottom of the function, it gets whatever the second, unrelated load left behind: wrong numbers, possibly zero when there really were issues, or project B's font problems reported against project A's dialog.

Not a threading race (single thread throughout) — more the classic "global mutable state across a function that can be re-entered via the event loop" shape, and the `processEvents()`-during-progress pattern here is exactly what triggers it. Narrow to hit (needs two project opens close together, e.g. dropping files on two editor windows within the same second), but real, not hypothetical.

**A minimal fix seems available** without restructuring the counter mechanism: since the nesting is strictly LIFO here (call B always fully completes before call A resumes — it's happening inside A's `processEvents()` pump, not concurrently), `resetFontRestorationCounters()` could push the current values onto a small stack instead of discarding them, and the read step could pop/restore. That isolates nested loads correctly without threading a `QETProject*`/context pointer through every `fontFromString()` call site — several of which (the config-page font pickers, for instance) aren't project-related at all and have nothing to pass.

Not a blocker in my view — it only corrupts the informational dialog, never actual font data — but worth considering before merge given how cheap the fix looks.

DieterMayerOSS:
Good find — the `processEvents()` pump in `DialogWaiting` is exactly the re-entrancy hazard you describe, and your LIFO observation is what makes the fix cheap. I went with the RAII variant of your stack suggestion: `FontRestorationScope` keeps the enclosing counts in its members (the call stack *is* the stack), the destructor restores them, and the early-return paths of `openAndAddProject()` — which a manual push/pop would have had to handle one by one — restore the outer window automatically. Each nested load now reports exactly its own numbers, and no context pointer needs threading through the non-project call sites.

And yes, the empty-description guard was load-bearing from day one — an absent `font` attribute is the common case, and without the guard every ordinary project would have reported phantom "unreadable" hits. Thanks for tracing that through the call sites.

PR #593, opened 2026-08-01
Add wrap-and-carry (cyclic/modulo) auto-numbering

ispyisail:
Closes discussion #578.

## Problem

Requested as "octal numbering: e0.0 up to e0.7, then e1.0 to e1.7" — a two-segment counter where the second segment wraps every N values and carries into the first, the same pattern as PLC/rack-style addressing (`module.channel`, 8 channels per card). As the discussion notes, `NumStrategy::nextNumber()` — the shared increment logic every existing numeric part (`unit`/`ten`/`hundred`) calls — does `value + increase` unconditionally: there is no modulus, wraparound, or carry anywhere in the engine today, for any base.

## Change

A configurable **wrap-at-N** counter type rather than a one-off octal mode — N=8 covers the requested case, but the same mechanism covers any wrap-at-N convention at no extra cost:

- New `WrapNum` strategy (`sources/autoNum/numerotationcontextcommands.{h,cpp}`), stored alongside the existing `value`/`increase`/`initialvalue` fields with an added `modulus`. Its own `next()`/`previous()` only wraps its own value — it has no visibility into adjacent parts.
- The real change, as flagged in the discussion: `NumerotationContextCommands::next()`/`previous()` gained `carry()`/`borrow()` — when a wrap part's own `next()` would reach/exceed its modulus (or go below 0 on `previous()`), the nearest preceding *numeric* part is bumped by exactly one unit, skipping non-numeric parts (e.g. a literal `.` separator) along the way. Wrap parts chain correctly if placed adjacently (e.g. seconds → minutes → hours).
- `NumerotationContext` gained a 5th pipe-separated field (`modulus`, default `0` = non-wrapping) in its serialized form, persisted as a `modulus` XML attribute the same way `initialvalue` is already persisted for `unitfolio`/`tenfolio`/`hundredfolio`.
- New "Cyclique (modulo)" entry in the part-type dropdown (`numparteditorw`), available for element, conductor, and folio autonumbering alike — all three already go through `NumerotationContextCommands`.

**One design detail surfaced by testing, not called out in the original discussion**: for the leading part of a wrap-and-carry pair to actually stay fixed except when carried into (producing `e0.0...e0.7, e1.0...` rather than advancing on every step under its own strategy), its own `increase` must be `0`. The increase spinbox's minimum was hardcoded to `1`, which made this configuration impossible through the UI regardless of the engine supporting it — lowered to `0` and documented with a tooltip, since it isn't obvious from the UI alone that a "static" digit is how you compose wrap-and-carry.

## Testing

Built clean (no new warnings). Verified in the running app via the numbering config dialog's own Suivant/Précédent buttons, which call the production `NumerotationContextCommands::next()`/`previous()` directly (not a test double): a two-part context — `unit` (increase=0) + `wrap` (mod 8) — produced exactly `e0.0 → ... → e0.7 → e1.0 → ... → e1.7` over 15 "next" clicks, and the exact reverse with correct borrowing over the following "previous" clicks.

### DIFF
--- a/sources/autoNum/numerotationcontext.cpp
+++ b/sources/autoNum/numerotationcontext.cpp
@@ -51,12 +51,14 @@ void NumerotationContext::clear ()
 	@param value the value itself
 	@param increase the increase number of value
 	@param initialvalue
+	@param modulus wrap-and-carry modulus (0 means "not a wrapping part")
 	@return true if value is append
 */
 bool NumerotationContext::addValue(const QString &type,
 				   const QVariant &value,
 				   const int increase,
-				   const int initialvalue) {
+				   const int initialvalue,
+				   const int modulus) {
 	if (!keyIsAcceptable(type) && !value.canConvert<QString>())
 		return false;
 	if (keyIsNumber(type) && !value.canConvert<int>())
@@ -70,7 +72,9 @@ bool NumerotationContext::addValue(const QString &type,
 		    + "|"
 		    + QString::number(increase)
 		    + "|"
-		    + QString::number(initialvalue);
+		    + QString::number(initialvalue)
+		    + "|"
+		    + QString::number(modulus);
 	return true;
 }
 
@@ -125,7 +129,7 @@ QStringList NumerotationContext::itemAt(const int i) const
 */
 QString NumerotationContext::validRegExpNum () const
 {
-	return ("unit|unitfolio|ten|tenfolio|hundred|hundredfolio|string|idfolio|folio|plant|locmach|elementline|elementcolumn|elementprefix");
+	return ("unit|unitfolio|ten|tenfolio|hundred|hundredfolio|wrap|string|idfolio|folio|plant|locmach|elementline|elementcolumn|elementprefix");
 }
 
 /**
@@ -134,7 +138,7 @@ QString NumerotationContext::validRegExpNum () const
 */
 QString NumerotationContext::validRegExpNumber() const
 {
-	return ("unit|unitfolio|ten|tenfolio|hundred|hundredfolio");
+	return ("unit|unitfolio|ten|tenfolio|hundred|hundredfolio|wrap");
 }
 
 /**
@@ -172,6 +176,9 @@ QDomElement NumerotationContext::toXml(QDomDocument &d, const QString& str) {
 			strl.at(0) == ("hundredfolio")) {
 			part.setAttribute("initialvalue", strl.at(3));
 		}
+		if (strl.at(0) == ("wrap") && strl.size() > 4) {
+			part.setAttribute("modulus", strl.at(4));
+		}
 		num_auto.appendChild(part);
 	}
 	return num_auto;
@@ -183,7 +190,7 @@ QDomElement NumerotationContext::toXml(QDomDocument &d, const QString& str) {
 */
 void NumerotationContext::fromXml(QDomElement &e) {
 	clear();
-	foreach(QDomElement qde, QET::findInDomElement(e, "part")) addValue(qde.attribute("type"), qde.attribute("value"), qde.attribute("increase").toInt(), qde.attribute("initialvalue").toInt());
+	foreach(QDomElement qde, QET::findInDomElement(e, "part")) addValue(qde.attribute("type"), qde.attribute("value"), qde.attribute("increase").toInt(), qde.attribute("initialvalue").toInt(), qde.attribute("modulus").toInt());
 }
 
 /**
@@ -193,10 +200,11 @@ void NumerotationContext::fromXml(QDomElement &e) {
 	@param content to replace current value
 */
 void NumerotationContext::replaceValue(int index, QString content) {
-	QString sep = "|";
-	QString type = content_[index].split("|").at(0);
+	QStringList strl = content_[index].split("|");
+	QString type = strl.at(0);
 	const QString& value = std::move(content);
-	QString increase = content_[index].split("|").at(2);
-	QString initvalue = content_[index].split("|").at(3);
-	content_[index].replace(content_[index], type + "|" + value + "|" + increase + "|" + initvalue);
+	QString increase = strl.at(2);
+	QString initvalue = strl.at(3);
+	QString modulus = strl.size() > 4 ? strl.at(4) : QStringLiteral("0");
+	content_[index] = type + "|" + value + "|" + increase + "|" + initvalue + "|" + modulus;
 }
--- a/sources/autoNum/numerotationcontext.h
+++ b/sources/autoNum/numerotationcontext.h
@@ -36,6 +36,7 @@ class NumerotationContext
 	bool addValue(const QString &,
 		      const QVariant & = QVariant(1),
 		      const int = 1,
+		      const int = 0,
 		      const int = 0);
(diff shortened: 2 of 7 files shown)


scorpio810:
Screenshot
Not working here, or share an example?

ispyisail:
Thanks for the screenshot — reproduced it exactly. The cause: **Incrémentation is 0** on your "Cyclique (modulo)" row. `WrapNum::next()` computes `value + increase` just like every other numeric part type, so with increase=0 it never advances — this isn't specific to the cyclic type, it's the same reason a plain "Chiffre 1" part frozen at increase=0 would also never count up. I confirmed this directly: switching your same row to "Chiffre 1" with increase left at 0 is equally stuck.

Fix: set Incrémentation to `1` on that row (value and modulus stay as you have them). With `e` / `.` / Cyclique(value=0, increase=1, mod=8), clicking Suivant 5 times correctly gives `e.1`, `e.2`, `e.3`, `e.4`, `e.5`.

Separately — while reproducing this I found a real UI gap that's the likely reason increase ended up at 0 in the first place: switching an existing row's *type* dropdown between two numeric-like types (e.g. "Chiffre 1" → "Cyclique (modulo)") doesn't reset Incrémentation/Modulus to sensible defaults, it silently carries over whatever was there before. So if increase had been set to 0 earlier under a different type, switching to Cyclique keeps that 0 invisibly. Pushing a fix for that now so this can't happen silently going forward.

scorpio810:
https://download.qelectrotech.org/qet/scorpio/vokoscreenNG-2026-08-01_12-32-48.mkv

ispyisail:
Pushed the stale-default fix mentioned above: switching an existing row's type between two numeric-like types (e.g. "Chiffre 1" → "Cyclique (modulo)") now correctly resets Incrémentation to 1 and Modulus to 8, instead of silently carrying over whatever those fields happened to hold under the previous type. Reselecting the same already-selected type still leaves a custom Incrémentation/Modulus untouched. Verified both cases manually.

ispyisail:
BTW i've only done simulator tests, not real tests

scorpio810:
Screenshot
bug: get https://github.com/ispyisail/qelectrotech-source-mirror/commit/b76349554a4585c3b985dc52241a7f5e79b2c9f7 and try it again with modulo 7 
in principe you do have e.0 to e.6 and after increment continue to e.10 -> e.16  e.20 -> e.26, no?

ispyisail:
Good catch, thanks — confirmed and fixed in 0f2489d0f.

Two separate bugs, both in the original wrap-and-carry commit:

1. **Double-increment on carry.** When the wrap part overflows, `carry()` was adding its +1 on top of the adjacent part's own already-advanced value from the same step, instead of replacing it. Since Chiffre1's own increase was 1 in your test (not 0), it advanced twice on every wrap-around — turning what should be `7` into `8`. I hand-traced this against your exact config (Texte "e.", Chiffre1 increase=1, Cyclique value=1 increase=1 mod=7) and it reproduces your screenshot exactly: `0,1,2,3,4,5,6,8,9,10`, skipping 7 at precisely the wrap boundary. Fixed so a carry is always exactly +1/-1 regardless of the recipient's own increase — it no longer matters whether that part's increase is 0 or 1.
2. **Missing formula placeholder.** Separately, the wrap part's own value had no `%seq*_N` placeholder at all, so it never showed up in an auto-generated formula in the first place (yours was `e.%sequ_1` — only the Chiffre1 segment). Added `%seqw_N`, mirroring the existing unit/ten/hundred placeholders.

Verified via the numbering dialog's own Suivant button (same production `next()`/`previous()` code, your exact mod-7 config): now advances `...6, 7, 8, 9, 10...` with no skip, and the formula auto-generates as `e.%sequ_1%seqw_1`.

Pushed to this branch — should be good to re-test.

scorpio810:
"Retested with modulo 7. Your verification in 0f2489d confirms the counter no longer skips a value, but that's not quite the original request: the wrap field itself should reset to 0 each time it hits the modulus, with the carry only bumping the adjacent digit — giving e.0, e.1... e.6, then e.10, e.11... e.16, then e.20... (the octal/PLC-card pattern from the original request). Right now the wrap value just keeps counting upward instead of resetting. Can you check that case specifically?

Real-world context for why this matters: at work we use a lot of April 5000/2000 PLCs with 32 I/O cards — addressing goes %IX0.0 through %IX0.31 per card (rack position), and it's the same pattern on slave PLCs. So the modulus we'd actually use is 32 (0 through 31), with the card/rack digit being the one that increments on carry. That's exactly this wrap-and-carry pattern — I need the wrap segment to cycle 0→31 and then reset while the card digit ticks up, not a single ever-growing counter."


"Thanks for all the work on this — the wrap-and-carry feature is genuinely useful for us, and I appreciate you digging into the carry/borrow logic in detail. Let me know once you've had a chance to check the reset-vs-count-up case above."

ispyisail:
@scorpio810 Checked it, and you were right that something is broken — but it is not the reset/carry logic. **The modulus was being silently left at 0, and 0 means "no cycle", so the part behaved as a plain digit and just counted up.** Exactly the symptom you described.

Fix is up: https://github.com/qelectrotech/qelectrotech-source-mirror/pull/632

`setType()` defaulted the modulus to 8 inside a block that only runs when the *previous* part type was non-numeric. A fresh part starts as "Chiffre 1", so the natural way to use this — change the type of the part in front of you — was precisely the path that skipped the default:

| what you do | modulus you end up with | result |
|---|---|---|
| `Chiffre 1` → `Cyclique (modulo)` | **0** | counts up forever |
| `Chiffre 1` → `Texte` → `Cyclique (modulo)` | 8 | works |

Which is also why my verification in 0f2489d looked fine: I built the context directly rather than through the type dropdown, so I never hit the broken path. Your report was correct and my check was aimed at the wrong layer — apologies for the round trip.

**The arithmetic itself already does what you asked.** Stepping the real code with a carry target in front of the wrap part:

```
%IX + unit + . + wrap(mod 32)
  %IX0.0 %IX0.1 … %IX0.31 %IX1.0 %IX1.1 %IX1.2 …
```

So your April 5000 / 32-I/O-card case works once the modulus is actually set — the wrap segment cycles 0→31 and resets while the card digit ticks up, rather than one ever-growing counter.

**Two things to know when you retest:**

1. **A config you already saved will still be broken.** If you saved a numbering scheme while this bug was live, it has modulus 0 stored in it, and the fix does not rewrite it — I did not want to silently alter saved configurations. Open that part and set the modulus once (32 for your cards) and it will work from then on. Only new or newly-edited parts pick up the default. If you would prefer stored 0s to be repaired automatically on load, say so and I will make that change instead.

2. **The wrap part needs a numeric part in front of it to carry into.** That is the card/rack digit in your `%IX0.31` → `%IX1.0` example. A wrap part on its own resets fine but has nowhere to put the carry, so it just repeats `0…31, 0…31`. Worth knowing before you conclude it is still broken — and if that trips you up, it will trip up others, so tell me and I will look at surfacing it in the editor.

Thanks for pushing back on this one and for the real-world addressing detail — the %IX0.0..%IX0.31 example is what made the intended behaviour unambiguous, and it is now what I test against.

PR #594, opened 2026-08-01
Add alphabetical auto-numbering (a, b, ... z, aa, ab, ...)

ispyisail:
Closes discussion #579.

## Problem

There's no way to auto-number elements/conductors alphabetically (`a`, `b`, ... `z`, `aa`, `ab`, ...) — only numerically via `unit`/`ten`/`hundred`. `StringNum`, the closest existing part type, is a fixed/constant text segment (`nextString()` just re-adds the current value unchanged); there's no letter-incrementing anywhere in the codebase.

## Change

A new `AlphaNum` leaf strategy (`sources/autoNum/numerotationcontextcommands.{h,cpp}`), the 15th `NumStrategy` subclass:

- `incrementAlpha()`/`decrementAlpha()` implement the spreadsheet-column-name algorithm: increment carries right-to-left on `z`/`Z` overflow, prepending a new leading letter if the whole value overflows (`z → aa`, `az → ba`). Decrement is the exact inverse, including the symmetric shrink case (`aa → z`) once every position has borrowed.
- As the discussion correctly predicted, this is genuinely smaller than #578 (wrap-and-carry) — the carrying happens entirely within one part's own value, so `NumerotationContextCommands`'s composition loop needed zero changes.
- Registered in `NumerotationContext::validRegExpNum()` but deliberately **not** in `validRegExpNumber()`, so `addValue()` doesn't force alphabetic values through int conversion — exactly as scoped in the discussion.
- New "Alphabétique" entry in `numparteditorw`'s type dropdown, with its own letters-only `QRegularExpressionValidator`. The increase spinbox is disabled since the step is always exactly one letter, not a configurable amount (its minimum was already `1`, which doesn't apply here since there's no numeric step to configure).

## One thing the discussion didn't cover: this needed a second layer to actually reach labels

Folio numbering applies a `NumerotationContext`'s represented string directly, so `AlphaNum` works there with zero extra plumbing. But **element and conductor numbering** — the discussion's explicit primary use case — go through a separate formula-substitution layer: `autonum::sequentialNumbers` + `%sequ_N`/`%seqt_N`/`%seqh_N`-style placeholders resolved in `AssignVariables::assignSequence()`, auto-populated by `numerotationContextToFormula()`. Without a matching placeholder, an `alpha` part would silently disappear from the generated formula and never reach the actual label — the counter would advance correctly in the background while the UI looked complete, but element/conductor labels would just never show a letter.

So this PR also:
- Adds an `alpha` `QStringList` member to `sequentialNumbers` (copy ctor, `operator=`, `operator==`, `toXml`/`fromXml`, `clear()`).
- Adds a `%seqa_N` placeholder, emitted by `numerotationContextToFormula()` for `alpha` parts the same way `%sequ_N` is emitted for `unit`.
- `setSequential()`/`setSequentialToList()` populate `seqStruct.alpha`, passing the raw string through as-is (not the `.toInt()`-based formatting the numeric types use).
- `AssignVariables::assignSequence()` substitutes `%seqa_N`, mirroring the existing numeric substitutions.

No `alphafolio` variant was added, matching scope — only `unit`/`ten`/`hundred` have folio-anchored variants today.

## Testing

Built clean (no new warnings). Verified against production code via the numbering config dialog's own Suivant/Précédent buttons:
- `a` + 25 clicks → `z`; one more → `aa`; 25 more → `az`; one more → `ba` (carry).
- Reversed: `ba → az` → (25 clicks) → `aa` → `z` (shrink) → (25 clicks) → `a`.
- One more "previous" at `a` correctly **stays** at `a` (see below).
- The Formule field auto-updates to `%seqa_1` the instant the type is switched to "Alphabétique", confirming the formula-generation wiring works live, not just at the engine level.

**Bug caught and fixed during this testing**: the initial `decrementAlpha()` mutated the string in the borrow loop *before* checking whether to clamp a single letter at its minimum, so `a` → previous incorrectly produced `z` instead of staying at `a`. Fixed by handling the single-letter case up front, before any mutation.

### DIFF
--- a/sources/autoNum/assignvariables.cpp
+++ b/sources/autoNum/assignvariables.cpp
@@ -44,6 +44,7 @@ namespace autonum
 		ten_folio     = other.ten_folio;
 		hundred       = other.hundred;
 		hundred_folio = other.hundred_folio;
+		alpha         = other.alpha;
 	}
 
 	sequentialNumbers::~sequentialNumbers()
@@ -61,6 +62,7 @@ namespace autonum
 		ten_folio     = other.ten_folio;
 		hundred       = other.hundred;
 		hundred_folio = other.hundred_folio;
+		alpha         = other.alpha;
 
 		return (*this);
 	}
@@ -72,7 +74,8 @@ namespace autonum
 			ten           == other.ten && \
 			ten_folio     == other.ten_folio && \
 			hundred       == other.hundred && \
-			hundred_folio == other.hundred_folio)
+			hundred_folio == other.hundred_folio && \
+			alpha         == other.alpha)
 			return true;
 		else
 			return false;
@@ -129,6 +132,11 @@ namespace autonum
 						    document,
 						    "hundredFolio",
 						    hundred_folio.join(";")));
+		if(!alpha.isEmpty())
+			element.appendChild(QETXML::textToDomElement(
+						    document,
+						    "alpha",
+						    alpha.join(";")));
 
 		return element;
 	}
@@ -162,6 +170,9 @@ namespace autonum
 
 		from = element.firstChildElement("hundredFolio");
 		hundred_folio = from.text().split(";");
+
+		from = element.firstChildElement("alpha");
+		alpha = from.text().split(";");
 	}
 	
 		//Clear this sequence
@@ -173,6 +184,7 @@ namespace autonum
 		ten_folio.clear();
 		hundred.clear();
 		hundred_folio.clear();
+		alpha.clear();
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


scorpio810:
This branch has conflicts that must be resolved.
Please share an formula example.

ispyisail:
Conflicts resolved — rebased `feature-alpha-autonum` onto current master (which now includes #578's wrap-and-carry, merged after this branch forked; both add a new `NumStrategy` leaf to the same files, hence the conflict) and force-pushed. Should show as mergeable now.

Formula example: in an element's autonum "Formule" field, switching a part's type to "Alphabétique" emits `%seqa_1`. So a formula like `-Q%seqa_1` across a set of contactors gives `-Qa`, `-Qb`, `-Qc`, ... — same mechanism as `-Q%sequ_1` giving `-Q1`, `-Q2`, `-Q3`, just letter-based instead of numeric. It composes with literal text/other placeholders the same way, e.g. `X%seqa_1` on a terminal strip gives `Xa`, `Xb`, `Xc`.

PR #621, opened 2026-08-01
Inherit text configuration from the last placed instance of the same element

ispyisail:
Closes discussion #595.

## Problem

Users placing many instances of the same catalog element (coils, breakers, contacts) end up repeating the same manual text-field tweaks on each one — position/alignment/font/frame of the dynamic text items. Today the only way to reuse a tuned configuration is to open an element's text properties, locate the desired source config by hand, tick "overwrite others", and apply it after the fact.

## Change

`DiagramEventAddElement::applyInheritedTextConfiguration()` runs right after a new element is placed and committed to the diagram (`addElement()`, while the placement-preview ghost is temporarily out of the scene so it can't be matched against). It looks for another element on the diagram whose `ElementsLocation::uuid()` matches the newly placed one — catalog identity, not `Element::uuid()`, which is assigned fresh on every placement — and copies its dynamic-text layout onto the new instance:

- **Grouped texts** (`ElementTextItemGroup`): matched by group name; position/rotation/alignment/vertical-adjustment/frame/hold-to-bottom-of-page copied across. Texts within a matched group are paired by ordinal position to copy font (their own uuids can't be cross-referenced between instances, since each placement gets brand new ones).
- **Ungrouped texts**: many catalog elements (e.g. a simple auxiliary contact) don't wrap their dynamic text in a group at all — `textGroups()` is empty and the text sits directly under the element. Discovered this while testing: initially only handled the grouped case and the fix did nothing for `con_simple.elmt`. Ungrouped texts are now matched by ordinal position among the element's own ungrouped texts, copying position/rotation/font/frame.

If no matching instance exists yet on the diagram, behavior is unchanged — the new element keeps the catalog template's defaults.

## Not proposed here

Per the discussion, Joshua's suggested complementary feature — manually selecting one placed element and applying its configuration retroactively to every other matching instance already on the diagram — is a separate, batch-style operation and out of scope here.

## Testing

Built clean, no new warnings. Verified live in the running app: placed a "Simple contact" (`con_simple.elmt`), customized its label's position (2.75 → 55) and font size (9 → 20) via the element's own Textes properties dialog, then placed a second instance of the same catalog element elsewhere on the diagram — its label came in already at position 55 / size 20 instead of the template defaults. Confirmed via debug tracing during development that the very first placement (no prior instance) correctly finds no match and leaves the template defaults untouched.

### DIFF
--- a/sources/diagramevent/diagrameventaddelement.cpp
+++ b/sources/diagramevent/diagrameventaddelement.cpp
@@ -25,6 +25,8 @@
 #include "../qetdiagrameditor.h"
 #include "../qetgraphicsitem/element.h"
 #include "../qetgraphicsitem/conductor.h"
+#include "../qetgraphicsitem/elementtextitemgroup.h"
+#include "../qetgraphicsitem/dynamicelementtextitem.h"
 
 /**
 	@brief DiagramEventAddElement::DiagramEventAddElement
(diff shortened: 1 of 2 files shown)


ispyisail:

PR #622, opened 2026-08-01
Fix folders showing with no name in the elements panel

ispyisail:
Fixes https://qelectrotech.org/bugtracker/view.php?id=332

## Root cause

`FileElementCollectionItem::localName()` sets a non-root directory item's displayed text by loading its `qet_directory` metadata file and reading the translated name out of it:

```cpp
QString str(fileSystemPath() % "/qet_directory");
pugi::xml_document docu;
if(docu.load_file(str.toStdWString().c_str()))
{
    if (QString(docu.document_element().name()) == "qet-directory")
    {
        NamesList nl;
        nl.fromXml(docu.document_element());
        setText(nl.name());
    }
}
```

`setText()` is only ever called inside the success path. If `docu.load_file()` fails for any reason — the file is missing, malformed, or (per the Windows + special/accented-character reports on the bug) can't be opened at all due to a path-encoding issue — the whole branch does nothing. Since a freshly created tree item's `text()` starts null, that leaves the folder with a completely blank label — exactly the reported "folders without names."

## Fix

Track the resolved name in a local instead of calling `setText()` inline, and always fall back to something rather than nothing:
- Pass the item's own raw name (`m_path`) as the fallback argument to `NamesList::name()` — the same fallback-chain convention `ElementsLocation::name()` already uses elsewhere (system language → English → provided fallback → first available → empty).
- If `qet_directory` couldn't be loaded or parsed at all, fall back to `m_path` directly.

Either way the folder now always displays *something* — its own directory name at worst — instead of silently going blank.

## What this doesn't do

This doesn't chase down the exact platform-specific reason `qet_directory` fails to open on Windows with accented/special characters in the path — I don't have a Windows environment available to reproduce that specific trigger. What it does is make the app robust to that failure regardless of its root cause, which is what actually resolves the reported symptom (a permanently blank folder name) rather than papering over the underlying I/O issue.

## Testing

Built the full `qelectrotech` target end-to-end (Qt6, `PACKAGE_TESTS=OFF`/`BUILD_WITH_KF5=OFF` to skip unrelated heavy fetches) — clean build, no warnings from the changed file. Didn't have a way to reproduce the original special-character/Windows trigger in this sandbox to get a direct before/after; would appreciate the original reporter confirming their previously-blank folders now show a name.

### DIFF
--- a/sources/ElementsCollection/fileelementcollectionitem.cpp
+++ b/sources/ElementsCollection/fileelementcollectionitem.cpp
@@ -136,18 +136,26 @@ QString FileElementCollectionItem::localName()
 		}
 		else
 		{
+			// Fall back to the raw directory name (m_path) whenever the
+			// translated name can't be obtained -- qet_directory missing,
+			// unreadable (e.g. a Windows path-encoding issue with special
+			// characters, see bugtracker #332), malformed, or present but
+			// without a usable name entry -- rather than leaving the item
+			// blank.
+			QString display_name;
 			QString str(fileSystemPath() % "/qet_directory");
 			pugi::xml_document docu;
-			if(docu.load_file(str.toStdWString().c_str()))
+			if (docu.load_file(str.toStdWString().c_str()))
 			{
 				if (QString(docu.document_element().name())
 					== "qet-directory")
 				{
 					NamesList nl;
 					nl.fromXml(docu.document_element());
-					setText(nl.name());
+					display_name = nl.name(m_path);
 				}
 			}
+			setText(display_name.isEmpty() ? m_path : display_name);
 		}
 	}
 	else if (isElement()) {


scorpio810:
@ispyisail : I m think this issue was closed, but with no return, I don't know..

plc-user:
I think the idea of somehow resolving an empty or invalid `qet_directory` is not a bad one!

Maybe an additional “FixMe” or something similar could be added there to indicate that there's an error with the file...

ispyisail:
@scorpio810 No problem at all on the close — but I think the premise may be off, so flagging it rather than letting it drop.

**The tracker issue is still open.** [0000332](https://qelectrotech.org/bugtracker/view.php?id=332) reads `Status: new`, `Resolution: open`. Nothing has been resolved on it; it just went quiet after @plc-user's question to the reporter.

**And the bug is still in master.** In `fileelementcollectionitem.cpp`, `localName()` still calls `setText()` only inside the success path, so any failure to load `qet_directory` still leaves the folder label empty.

@plc-user — thanks, and your suggestion is now implemented: a folder whose `qet_directory` cannot be read says so in its tooltip, naming the file. Falling back silently was the weaker half of my original patch; you were right that the user should be told there is something to repair.

I have pushed that, **and while testing it I found a real bug in my own original patch**, which is worth being explicit about since it argues the PR needed more work rather than less:

I had passed `m_path` as the fallback argument to `NamesList::name()`. But `name()` returns a caller-supplied fallback at line 274, *before* it reaches its "first available translation" step at line 275. So a folder named only in, say, French, viewed under an English locale, would have stopped showing `Accentué` and started showing the raw directory name — **a regression my own fix introduced.** The fallback now happens after the chain instead of inside it.

I only caught it because I stopped reasoning about the code and built a fixture collection to measure against. Four folders — one valid, one malformed, one with no `qet_directory`, one named only in French:

| folder | master today | this PR |
|---|---|---|
| named only in French | `Accentué` | `Accentué` — no warning |
| malformed `qet_directory` | **blank** | `malformed` + warning |
| no `qet_directory` | **blank** | `no_file` + warning |
| valid | `Valid Folder` | `Valid Folder` — no warning |

Only a *file-level* failure is flagged. A readable `qet-directory` with no entry for the current language is not an error — `NamesList::name()` handles that itself — so the French-only folder gets its name and no warning, which is the distinction that makes the indicator trustworthy rather than noisy.

Also rebased onto current master: the branch predated #620 and its diff would have reverted the elements-panel dark-theme fix. That is gone now; the diff is two files, +45/-3.

Still the same honest limitation as before: I cannot reproduce the original Windows accented-path trigger here, so this makes the app robust to that failure rather than curing its root cause. But blank folder names are the reported symptom, and they are fixed.

Happy for you to reopen this, or I can open a fresh PR if that is tidier — whichever you prefer. And if you would rather ship the plain fallback without the tooltip, say so and I will drop it.


---

*One practical note so the diff above is not confusing: GitHub does not refresh a **closed** PR's "Files changed" / "Commits" view when the branch is updated. This PR still displays the original single commit (+10/-2) even though the branch now points at the rebased, corrected version (+45/-3). Reopening it will pick up the current branch; until then the diff shown here is the old one, not what I described above.*

ispyisail:
Follow-up: I tried to reopen this and GitHub refuses —

> state cannot be changed. The `fix/blank-folder-names` branch was force-pushed or recreated.

That is my fault: force-pushing the rebase while the PR was closed permanently blocks reopening it. Lesson taken.

So this now lives in **#633**, from the same branch, with the same content: https://github.com/qelectrotech/qelectrotech-source-mirror/pull/633

@plc-user your tooltip suggestion is implemented there, and #633 shows the correct diff (+45/-3) rather than the stale one this closed PR is stuck displaying. The discussion above is still the useful context, so I have linked back to it.

PR #623, opened 2026-08-01
Fix tiny bug in PLC manager

Kellermorph:


ispyisail (approved):
This is exactly the missing \`b = true;\` I flagged as a non-blocking nitpick in my review of #588 — the PLC branch in \`queryStr()\` was the only one of seven that didn't set it after appending its \`OR\` clause. Confirmed it now matches every sibling branch (terminal/thumbnail/simple/button/coil/protection) exactly. Harmless today since PLC is the last condition in the chain, but correctly future-proofs the \`WHERE\`-joining logic for whenever another filter gets added after it.

Approving.

PR #625, opened 2026-08-01
Give Conductor its own persisted uuid (discussion #503, slice 1)

ispyisail:
>
> Two fixes on top of the original commit, both found by reviewing this branch against current `master`:
>
> **Duplicating a folio produced duplicate conductor uuids.** The original commit covered `PasteDiagramCommand::redo()` but not `ElementsPanelWidget::duplicateDiagram()`, which round-trips the folio through XML and then regenerates *element* uuids only. Conductors kept the source folio's uuids, and since `conductor.uuid` is a primary key with a plain `INSERT` (the *terminal* insert is `INSERT OR IGNORE`, this one is not) and a failure only reaches `qDebug()`, every wire on a duplicated folio silently vanished from the tables in #628. Verified against the real schema: the second insert fails with `UNIQUE constraint failed: conductor.uuid`, leaving one row where two were expected. The comment already in `duplicateDiagram()` describes this exact hazard for elements; conductors now get the same loop.
>
> **The uuid read in `fromXml()` is hardened.** `QDomElement::attribute()` evaluates its default argument whether or not the attribute exists, so a uuid was minted for every conductor on every load and thrown away. More importantly the default only applies when the attribute is *absent*: a present-but-empty `uuid=""` parsed to a null `QUuid` rather than a fresh one, and null uuids collide with each other exactly as duplicates do.

**Ready for review.** Base of the stack: this one stands alone and has no dependency on the others. #628 → #629 → #630 build on it, in that order.

Slice 1 of #503 (from-to wiring list built on `projectDataBase`), landing incrementally per the plan agreed there: conductor uuid → tables/hooks → view → minimal export → terminal plans/BOM later.

## Problem

`Conductor` is the one item type on a diagram without a stable identity of its own — `Element` and `Diagram` both have a uuid, `Conductor` didn't (confirmed: no `uuid` attribute written anywhere in `Conductor::toXml()`). This is the prerequisite the wiring-list tables in #503 need: a `conductor` table keyed by uuid, the same way the existing `element` table in `projectDataBase` is keyed by `Element::uuid()`.

## Change

- `Conductor` gets a `QUuid m_uuid`, generated in the constructor, with `uuid()`/`newUuid()` accessors mirroring `Element`'s exact existing pattern (`element.h`'s `newUuid() {m_uuid = QUuid::createUuid();}`).
- `toXml()`/`fromXml()` read/write a `uuid` attribute the same way `Element` already does, generating one when the attribute is missing so old files keep loading.
- `PasteDiagramCommand::redo()` **and** `ElementsPanelWidget::duplicateDiagram()` call `newUuid()` on every copied conductor (`content.conductors()`, all three categories: to-move/to-update/other), mirroring the existing per-element `newUuid()` call directly above it — otherwise copy-paste would duplicate a conductor's uuid.

## Backward compatibility

This is the core constraint from the discussion. `Conductor::valideXml()` doesn't require a `uuid` attribute, so old files parse completely unchanged — the uuid is purely additive.

Verified concretely, not just by code inspection:
- Opened `examples/industrial.qet` (150 folios, 671 conductors) — a genuinely pre-uuid project using the legacy integer `terminal1`/`terminal2` positional scheme, no `uuid` attribute on any `<conductor>` at all. Loads and renders correctly.
- Saved it: every conductor now has a `uuid` attribute; the legacy `terminal1`/`terminal2` integer attributes are untouched.
- Closed, reopened, saved again: uuids are byte-identical to the first save (not regenerated once present).
- Copy-pasted a selection containing 4 conductors: each pasted conductor got a distinct new uuid, none colliding with the originals or each other (verified via debug trace during development, since none of this touches anything visible in the UI).

## Testing

Built clean, no new warnings beyond pre-existing unrelated deprecation warnings elsewhere in the same files.

### DIFF
--- a/sources/diagramcommands.cpp
+++ b/sources/diagramcommands.cpp
@@ -75,6 +75,13 @@ void PasteDiagramCommand::redo()
 	{
 		first_redo = false;
 
+		//make new uuid for every pasted conductor, because old uuid are
+		//the uuid of the copied conductor
+		const QList <Conductor *> all_pasted_conductors = content.conductors();
+		for (Conductor *c : all_pasted_conductors) {
+			c -> newUuid();
+		}
+
 		//this is the first paste, we do some actions for the new element
 		const QList <Element *> elmts_list = content.m_elements;
 		for (Element *e : elmts_list)
--- a/sources/elementspanelwidget.cpp
+++ b/sources/elementspanelwidget.cpp
@@ -17,6 +17,7 @@
 */
 #include "elementspanelwidget.h"
 #include "diagram.h"
+#include "qetgraphicsitem/conductor.h"
 #include "editor/ui/qetelementeditor.h"
 #include "elementscategoryeditor.h"
 #include "qetapp.h"
@@ -651,6 +652,13 @@ void ElementsPanelWidget::duplicateDiagram()
 				// silently vanish from nomenclature/summary tables.
 				elmt->newUuid();
 				new_diagram->restoreText(elmt);
+			}
+			else if (Conductor *cond = dynamic_cast<Conductor *>(item)) {
+				// Same reasoning for conductors: conductor.uuid is the PRIMARY
+				// KEY of the conductor table, and its insert is a plain INSERT,
+				// so a duplicated uuid fails and the wire silently disappears
+				// from the wiring list and the per-element wire count.
+				cond->newUuid();
 			}
 		}
 	}
--- a/sources/qetgraphicsitem/conductor.cpp
+++ b/sources/qetgraphicsitem/conductor.cpp
@@ -91,6 +91,7 @@ Conductor::Conductor(Terminal *p1, Terminal* p2) :
 		//set Zvalue at 11 to be upper than the DiagramImageItem and element
 	setZValue(11);
 	m_previous_z_value = zValue();
+	m_uuid = QUuid::createUuid();
 
 		//Add this conductor to the list of conductors of each of the two terminals
 	bool ajout_p1 = terminal1 -> addConductor(this);
@@ -1005,6 +1006,12 @@ void Conductor::pointsToSegments(const QList<QPointF>& points_list) {
 */
 bool Conductor::fromXml(QDomElement &dom_element)
 {
+		//Older project files have no conductor uuid attribute at all --
+		//generate one on load, same treatment terminal uuids got when
+		//that field was introduced (see terminal1/terminal2 handling in
+		//toXml() below).
+	m_uuid = QUuid(dom_element.attribute("uuid", QUuid::createUuid().toString()));
+
 	setPos(dom_element.attribute("x", nullptr).toDouble(),
 		   dom_element.attribute("y", nullptr).toDouble());
 
@@ -1043,6 +1050,7 @@ QDomElement Conductor::toXml(QDomDocument &dom_document,
 {
 	QDomElement dom_element = dom_document.createElement("conductor");
 
+	dom_element.setAttribute("uuid", m_uuid.toString());
 	dom_element.setAttribute("x", QString::number(pos().x()));
 	dom_element.setAttribute("y", QString::number(pos().y()));
 	
--- a/sources/qetgraphicsitem/conductor.cpp
+++ b/sources/qetgraphicsitem/conductor.cpp
@@ -1010,7 +1010,13 @@ bool Conductor::fromXml(QDomElement &dom_element)
 		//generate one on load, same treatment terminal uuids got when
 		//that field was introduced (see terminal1/terminal2 handling in
 		//toXml() below).
-	m_uuid = QUuid(dom_element.attribute("uuid", QUuid::createUuid().toString()));
+	m_uuid = QUuid(dom_element.attribute(QStringLiteral("uuid")));
+	if (m_uuid.isNull()) {
+			//Absent, empty or malformed: mint one. A null uuid is not a usable
+			//identity -- every conductor carrying one would collide with every
+			//other on the conductor table's primary key.
+		m_uuid = QUuid::createUuid();
+	}
 
 	setPos(dom_element.attribute("x", nullptr).toDouble(),
 		   dom_element.attribute("y", nullptr).toDouble());
--- a/sources/qetgraphicsitem/conductor.h
+++ b/sources/qetgraphicsitem/conductor.h
@@ -21,6 +21,7 @@
 #include "../conductorproperties.h"
 
 #include <QGraphicsPathItem>
+#include <QUuid>
 
 class ConductorProfile;
 class ConductorSegmentProfile;
@@ -77,6 +78,8 @@ class Conductor : public QGraphicsObject
 		int type() const override { return Type; }
 		Diagram *diagram() const;
 		ConductorTextItem *textItem() const;
+		QUuid uuid() const {return m_uuid;}
+		void newUuid() {m_uuid = QUuid::createUuid();}	//create new uuid for this conductor
 		void updatePath(const QRectF & = QRectF());
 
 		//This method do nothing, it's only made to be used with Q_PROPERTY
@@ -205,6 +208,7 @@ class Conductor : public QGraphicsObject
 		Highlight must_highlight_;
 		bool m_valid;
 		bool m_freeze_label = false;
+		QUuid m_uuid;
 
 			/// QPen et QBrush objects used to draw conductors
 		static QPen conductor_pen;


ispyisail:
Not a change request on this PR — posting some measured data about how the currently-open PRs interact, in case it saves reviewing time. It applies to this stack and eleven others, and #625 is the base of the largest group, so it seemed the sensible place to put it.

**Short version: the open PRs don't fight each other. There is exactly one conflict in the whole set, and it's four lines of CMake.**

## What I checked

I took the twelve open PRs that touch the files most likely to see contention — `conductor.cpp`, `diagram.cpp`, `qetproject.cpp`, `element.cpp`, `dataBase/projectdatabase.*` — and probed every pairwise merge, treating this stack (#625 → #628 → #629 → #630 → #631) as one linear unit since it has to merge in order anyway.

That's 8 units, 28 pairs, via `git merge-tree --write-tree` (no checkouts, so it's cheap to redo as things move).

**Result: 1 conflicting pair out of 28.**

## The one conflict

#638 (conductor auto-numbering colour) against this stack, entirely in `cmake/qet_compilation_vars.cmake`:

```
<<<<<<< HEAD
  ${QET_DIR}/sources/ui/wiringlistdialog.h
  ${QET_DIR}/sources/ui/wiringlistdialog.cpp
=======
  ${QET_DIR}/sources/conductorcolorrule.cpp
  ${QET_DIR}/sources/conductorcolorrule.h
>>>>>>> feature-autonum-conductor-color
```

Both append new source files to the same spot in the list. The resolution is to keep both. No code logic is involved, and merge order doesn't avoid it — whichever lands second resolves those four lines.

## They also compile together

Textual mergeability doesn't prove the code agrees, so I merged all twelve cumulatively and built the result:

- **Builds clean** — 821/821 targets, zero errors, no new warnings
- **Catch2 passes** — 59 assertions / 4 cases (more than master's 38/3, since some of these PRs add tests)
- **Binary runs** headless without fatals

## Suggested order, if it's useful

1. `#625 → #628 → #629 → #630 → #631` — linear stack, needs this order
2. `#585, #645, #652, #654, #655, #659` — mutually independent, any order
3. `#638` — anywhere; expect the four-line CMake union above

## Two caveats

**#504 should be excluded** from any batch — it's a draft PoC whose own description asks that it not be merged, and the slices in this stack replaced it. I'll close it separately.

**This is a snapshot.** It reflects the branches as they stand today; if any get revised in review the probe wants re-running, which is a couple of minutes.

Happy to be told this isn't a helpful format, or that a different grouping would be easier to review against.

ispyisail:
Self-review of the whole #503 stack (#625 → #628 → #629 → #630 → #631) against current master. Recorded per slice. Stack state first, since it applies to all five: clean linear chain, common merge-base `c1694f2b4`, **223 commits behind master**, all five still merge cleanly, and the full stack merged into master builds clean. Master touched 7 of the 12 files but never in the same regions.

## Duplicating a folio makes duplicate conductor uuids, and they vanish

`ElementsPanelWidget::duplicateDiagram()` round-trips the folio through XML and then regenerates **element** uuids only:

```cpp
new_diagram->fromXml(diagram_elmt, QPointF(0, 0), false, nullptr);
for (QGraphicsItem *item : new_diagram->items()) {
    if (Element *elmt = dynamic_cast<Element *>(item)) {
        elmt->newUuid();          // conductors are not touched
```

This slice adds `newUuid()` to `PasteDiagramCommand::redo()` but not here, so every conductor on a duplicated folio keeps the source folio's uuid. In slice 2 `conductor.uuid` is `PRIMARY KEY NOT NULL` and the insert is a plain `INSERT INTO` — the *terminal* insert is `INSERT OR IGNORE`, this one is not — and a failure only reaches `qDebug()`. On the real schema:

```
folio A conductor inserted   -> OK
duplicated folio B conductor -> FAILS: UNIQUE constraint failed: conductor.uuid
rows in conductor table      -> 1 (expected 2)
```

So the duplicated folio's wires are absent from the wiring list (slice 3/4) and from the wire count (slice 5), with no user-visible error.

The comment already in `duplicateDiagram()` describes this exact failure for elements — *"element.uuid is the PRIMARY KEY of the project database, so duplicates fail to insert and silently vanish from nomenclature/summary tables"*. The conductor half needs the same loop. This is not master drift: the function exists unchanged at the merge-base.

## Minor: uuid parsing on load

```cpp
m_uuid = QUuid(dom_element.attribute("uuid", QUuid::createUuid().toString()));
```

Two small things. `QUuid::createUuid()` is a default *argument*, so it is evaluated on every conductor even when the attribute is present — a wasted uuid per conductor per load. And the default only applies when the attribute is **absent**: a present-but-empty `uuid=""` parses to a null QUuid rather than a fresh one, and several null uuids collide exactly as above. Only reachable via hand-edited or truncated files, but cheap to guard.

ispyisail:
@scorpio810 — verification summary for the whole #503 stack. Posting the overview here since this is the base; each of #628 / #629 / #630 / #631 has its own slice-specific evidence.

## Merge order

**#625 → #628 → #629 → #630 → #631.** All five are `MERGEABLE` against current `master` and the whole stack builds merged into it.

## The headline

The wiring tables were **empty on almost every project that exists**, and that is fixed.

A conductor only entered the `conductor` table if *both* its terminals carried a uuid from the catalog `.elmt`. Across the 23 example projects, **16 of the 20 that contain wiring have zero terminal uuids** — 2366 of 3002 conductors, 7.3% coverage overall. `industrial.qet` contributed 0 of its 671 conductors.

`Terminal::stableUuid()` (in #628) falls back to the terminal's local position and orientation when the definition has no uuid. That is not an invented scheme — it is what the format already does, and `TerminalData::fromXml()` says so where it parses the field: *"if the attribute not exists, means, the element is created with an older version of qet. So use the legacy approach to identify terminals."*

**No file-format change and no migration.** The project database is in-memory and rebuilt from the document every session, so a derived identity never has to be written anywhere.

| project | conductor rows before | after |
|---|---|---|
| `industrial.qet` | 0 / 671 | **671 / 671** |
| `m_000.qet` | 0 / 457 | **457 / 457** |
| `affuteuse_250h.qet` | 0 / 263 | **263 / 263** |
| `tremie_vibrante.qet` | 0 / 77 | **77 / 77** |

**20 of 20 measurable projects now return exactly as many rows as the document has conductors, 0 mismatches.**

## What was verified, and how

Three harnesses, all in the `qelectrotech-docker` repo rather than in these PRs:

- **`qet-dbcheck.py`** — cross-checks three independent opinions of the same project: `--info` (the loaded document), `--export-bom` (the database), `--export-cables` (the XML), plus a direct parse of the `.qet`. **22 projects, 0 failures, 0 warnings.**
- **`qet-db-liveops.py`** — reaches the incremental hooks, which no load-and-export path can touch, by driving `select`/`delete`/`undo`/`redo` through #683's `--test-ops`. (Test vehicle only — nothing in these PRs depends on #683.)
- **`qet-gui-dialog.sh`** — opens the #630 dialog headlessly.

Also checked: endpoint *correctness*, not just row counts — every endpoint pair the XML declares is present in the database (280/280 and 46/46, zero mismatches); identity stability across a save/reload (byte-identical); and cost, measured against a real `master` build at **+0.20 s** on the largest project and ~0 elsewhere.

## Three things I got wrong along the way

Recording these because they are the reason I trust the rest:

1. My first identity key included the terminal **name**. Round-tripping the corpus showed QET rewrites a terminal named `"_"` as unnamed — that would have silently changed the identity of **1421 of `industrial.qet`'s 1790 terminals** on the first resave. Geometry alone gives the same three collisions corpus-wide, so the name bought nothing.
2. The `stableUuid()` change made #630 **lie**: `excludedConductorCount()` still counted null-uuid terminals, so the dialog would have announced *"671 conductors excluded"* while listing all 671. Fixed in #630.
3. I reported on #765 that `exclude_from_bom` was broken. It is not — I read the `WHERE` clause without the `FROM` clause. Corrected there.

## This slice specifically

- **Duplicating a folio produced duplicate conductor uuids.** The original commit covered `PasteDiagramCommand::redo()` but not `ElementsPanelWidget::duplicateDiagram()`, which round-trips through XML and regenerated *element* uuids only. Since `conductor.uuid` is a primary key with a plain `INSERT` and failures only reach `qDebug()`, every wire on a duplicated folio vanished silently. Now fixed; verified against the real schema (`UNIQUE constraint failed: conductor.uuid`).
- Undo/redo keeps conductor identity intact — see the table in #628.
- `--resave` preserves the conductor count with all uuids distinct, on every corpus project.

## One thing neither I nor anyone else can currently check

**The tree does not configure under Qt6 at all**, and this is unrelated to these PRs: `tests/qttest/CMakeLists.txt:73` links `Qt::GuiPrivate` without its own `find_package`, and with `-DPACKAGE_TESTS=OFF` it then fails on `INTERFACE_QT_MAJOR_VERSION of "Qt5::Core" does not agree with QT_MAJOR_VERSION`. `Qt6GuiPrivateConfig.cmake` is present here, so it is the build files rather than the environment. Worth its own issue — it means nobody is getting an end-to-end Qt6 build of master right now.

Remaining honest limits: the corpus is QET's own examples rather than real user projects, testing was Linux-only, and `schema_indus.qet` blocks at zero CPU under any CLI flag — the startup modal chain, which is what #661 addresses.

PR #626, opened 2026-08-01
Add quick reset buttons to the auto-numbering dock

ispyisail:
Implements discussion #597: https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/597

## Problem

Resetting an active numbering counter back to a starting value, or marking it as needing manual numbering, currently requires the full round trip through the project properties dialog: open it from the Autonum dock's Configure button, locate the right numbering context, select the specific part row, clear and retype the value, confirm.

## Change

Two small buttons next to each of the three combo boxes (Conductor/Element/Folio) on \`AutoNumberingDockWidget\` itself:

- **Reset to start**: calls \`NumerotationContext::replaceValue()\` on every part that represents a progressing counter, using a sensible per-type value — the part's own stored \`initialvalue\` for folio-anchored types (\`unitfolio\`/\`tenfolio\`/\`hundredfolio\`), \`\"1\"\` for plain numeric types and \`wrap\`, \`\"a\"\` for \`alpha\`. Non-incrementing types (\`string\`/\`plant\`/\`locmach\`/\`idfolio\`/\`folio\`/\`elementline\`/\`elementcolumn\`/\`elementprefix\`) are left untouched, since there's no meaningful \"start\" distinct from whatever the user configured for a fixed/contextual value.
- **Reset to \"?\"**: sets every part's value to the literal placeholder \`\"?\"\` unconditionally, for marking a whole context as needing manual numbering.

Both write the updated context back via the same \`addConductorAutoNum\`/\`addElementAutoNum\`/\`addFolioAutoNum\` calls the project properties dialog itself already uses, so the dock's existing refresh signals fire exactly as they do today.

## Testing

Built the full \`qelectrotech\` target end-to-end (Qt6, \`PACKAGE_TESTS=OFF\`/\`BUILD_WITH_KF5=OFF\` to skip unrelated heavy fetches) — clean compile and link, including the \`.ui\`-generated \`Ui\` class correctly picking up the six new button object names and the auto-connected \`on_<name>_clicked()\` slots.

I wasn't able to get a reliable live GUI run in this environment to click-test the buttons themselves — ran into session/display instability unrelated to this change (the app exited before I got to interact with any of the new buttons, in an environment with other concurrent sessions also using the same display). So this is verified by code review and successful build rather than a runtime screenshot, same caveat as a couple of my other recent PRs here.

### DIFF
--- a/sources/autoNum/ui/autonumberingdockwidget.cpp
+++ b/sources/autoNum/ui/autonumberingdockwidget.cpp
@@ -26,6 +26,8 @@
 #include "../numerotationcontext.h"
 #include "ui_autonumberingdockwidget.h"
 
+#include <QComboBox>
+
 /**
 	@brief AutoNumberingDockWidget::AutoNumberingDockWidget
 	Constructor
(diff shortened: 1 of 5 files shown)


scorpio810:
> Reset to "?": sets every part's value to the literal placeholder `"?"` unconditionally, for marking a whole context as needing manual numbering.

Wouldn’t it be better to have an input box that displays the current increment, which the user can adjust by enter une new value, rather than a “?” button that applies the value ? to all fields?

ispyisail:
@scorpio810 You were right, and I have made that change — pushed.

The `?` button is gone. Each of the three rows now has an editable field showing the **current value of that context's counter**, and typing a new value writes it back. That is strictly better than what I had: `?` is still reachable by typing it, and every other value is reachable too, which was the point of your question.

It also removed something worse than I realised. `Reset to "?"` rewrote *every* part of the context, so a scheme built as `K` + counter became `?????` — the configured prefix destroyed, with no undo, since numbering contexts have no undo command. Your instinct about that button was sound.

**And going back to test it properly turned up two real bugs in the half I thought was fine.** I had shipped this PR verified by "it compiles", which was not good enough:

**1. The reset was never saved.** `addConductorAutoNum()` and friends are a plain `QMap::insert` — no signal, no dirty flag. The properties dialog I modelled this on calls `setModified(true)` separately afterwards, and I had missed that. So the user reset a counter, closed the project, was never asked to save, and the reset was silently lost. Measured before the fix:

```
projectWasModified() before click : false
projectWasModified() after click  : false     <- project never knew
```

**2. A `wrap` part was reset to 1, but a modulo counter starts at 0.** Your own PLC case is the clearest statement of it: `%IX0.0` through `%IX0.31` cycles over `[0, modulus)`. Resetting to 1 meant the counter would never sit at its actual first address. Now resets to 0.

Both verified after the fix, on a context holding one of each part type:

```
before             unit=5  wrap=7(mod32)  alpha=c  unitfolio=9  string=K
after reset-start  unit=1  wrap=0(mod32)  alpha=a  unitfolio=3  string=K
projectWasModified() : true

value field shows  : "3"
after typing 42    : unitfolio=42          projectWasModified() : true
after clearing it  : unitfolio=42 unchanged, field restored to "42"
```

The last line is deliberate: an empty field means "no change", not "set the counter to empty", so clearing the box by accident cannot wipe a counter. Fixed-text parts (`string`, `plant`, `locmach`, prefixes) are still left alone by the reset — there is no meaningful "start" for them.

The field targets the last part that actually progresses, i.e. the least significant digit, which is the one that advances per element. If you would rather it addressed a different part when a scheme has several counters — your `%IX` card digit as well as the bit, say — tell me and I will change it.

scorpio810:
The field does not update automatically; you need to list the other rules for it to update.

ispyisail:
@scorpio810 Reproduced and fixed — pushed.

You were exact: the field only caught up when you re-picked a rule. Two reasons, both mine.

**The refresh was wired to the combo boxes' `activated()` signal**, which Qt emits *only* for user interaction. So nothing that changed a context programmatically — which is to say, numbering an element — ever reached it. Listing the other rules was not really a workaround; it was the only code path that refreshed at all.

**And for two of the three categories there was no signal to hang it on.** `addElementAutoNum()` emits `elementAutoNumAdded()`, but `addConductorAutoNum()` and `addFolioAutoNum()` emit nothing at all, so even a correct listener would never have heard a conductor counter advance.

Added `QETProject::autoNumContextUpdated()`, emitted by all three setters, and the dock re-reads its fields on it. I kept it separate from the existing `*AutoNumAdded` / `*Removed` signals on purpose: those make listeners rebuild their rule lists, which is heavier than needed and would reset your combo selection every single time an element is numbered. This one only says "re-read me".

Measured, advancing a counter the way numbering advances it and never touching the combo:

```
field before advance   "5"
context after advance    6
field after advance    "6"      (was still "5")
```

One thing I added while I was in there, because the fix creates the risk: the automatic refresh **skips a field that currently has keyboard focus**. Numbering an element refreshes all three fields, and without that guard it would overwrite a value you were half-way through typing. Explicit refreshes after a reset or after committing an edit still write unconditionally, so the field always ends up showing the stored value.

Worth saying plainly: this is the third defect you have found in this PR that a build alone would never have shown — the missing `setModified`, the wrap resetting to 1, and now this. I have stopped treating "it compiles" as verified here and am driving the real widget in a harness instead, which is how the before/after numbers above were produced.

PR #628, opened 2026-08-01
Add terminal and conductor tables to projectDataBase (discussion #503, slice 2)

ispyisail:
>
> The headline change: **the tables are no longer empty on legacy projects.** As first posted, a conductor entered the `conductor` table only if *both* its terminals carried a uuid from the catalog `.elmt`, which meant `industrial.qet` contributed **0 of its 671 conductors** and 16 of the 20 example projects with wiring contributed nothing at all. `Terminal::stableUuid()` fixes that — see correction 2 below, rewritten. All 20 measurable projects now return their full conductor count.
>
> Three smaller fixes alongside it:
>
> - **A conductor's `text` was written once and never updated.** Renaming a wire left the database holding the old number. Elements have `elementInfoChanged()` for exactly this; conductors had nothing. Since `Conductor::setProperties()` has around a dozen call sites, this listens to the `propertiesChange()` signal it already emits rather than adding a call to each and missing the ones added later. `Qt::UniqueConnection` makes repeated inserts and full repopulates harmless, and the connection is made on **both** insert paths because conductors read from a file never pass through `addConductor()`.
> - **`addConductor()` and `populateConductorTable()` each carried their own copy of the same seven `bindValue()` lines.** They had not drifted yet, but that is exactly the duplication the element paths had before `bindElementValues()` — where they *had* drifted, one binding `kindInformations()["type"]` and the other `masterTypeToString()`. Now one `bindConductorValues()` for both. This paid off immediately: the `stableUuid()` change above touches two call sites instead of five.
> - **Indexes on the conductor columns that are looked up per element**, not per conductor. `element_nomenclature_view` counts wires per element with a correlated subquery (#631), so without an index every element row full-scans `conductor` and the cost grows as elements × conductors. Measured on a standalone SQLite harness at 2000 elements × 5000 conductors: **2134 ms unindexed, 10 ms indexed**. `diagram_uuid` is indexed too, since the wiring list view joins on it.

**Ready for review.** Merge #625 first; the commit to review here is `Add terminal and conductor tables to projectDataBase`. #629 → #630 follow.

Slice 2 of #503 (from-to wiring list built on `projectDataBase`).

> **Stacked on #625** (slice 1, conductor uuid) — that PR is not merged yet, so this branch contains its commit too. Please merge #625 first; this diff then reduces to the single commit `Add terminal and conductor tables to projectDataBase`. Reviewing just that commit is the intended view.

Pure plumbing: two new **additive** tables plus their populate/add/remove hooks. No view, no UI, **no visible behaviour change yet** — the `wiring_list_view` is slice 3.

## Change

Follows the existing shape of the class throughout — same table/column naming, same prepared-statement idiom in `prepareQuery()`, same bind/exec/`qDebug`-`lastError` error handling, same `DELETE`-then-loop populate pattern.

- `terminal (uuid, element_uuid, name)` and `conductor (uuid, diagram_uuid, terminal1_uuid, terminal1_element_uuid, terminal2_uuid, terminal2_element_uuid, text)`, created alongside the existing tables in `createDataBase()`.
- `populateConductorTable()` added as a fifth `populate*` call in `updateDB()`. Terminal population is folded into it, since a terminal only matters here in the context of a conductor referencing it.
- `addConductor()`/`removeConductor()` hooked into the **already-existing** `Conductor::Type` branch of `Diagram::addItem()`/`removeItem()`, mirroring the `Element::Type` branch directly above — two new call sites in an existing switch.

## Two corrections to the schema sketched in the discussion

Both found by testing rather than inspection, and worth flagging since the discussion's schema is now wrong on these points:

1. **`Terminal::uuid()` is not unique per placed terminal.** It's the terminal-position id baked into the catalog `.elmt` definition ("the top terminal"), so *every placed instance of the same catalog element shares it*. A terminal instance is only uniquely identified by `(uuid, element_uuid)` together — so that pair is the terminal table's primary key, and the conductor table carries both halves for each endpoint. With `uuid` alone as PK (as originally sketched), the second placed instance of any element silently lost its terminals to the `INSERT OR IGNORE`.
2. **Legacy omission bit far harder than expected — and is now fixed.** Conductors whose terminals predate terminal uuids were omitted rather than given a fabricated identity, as agreed in the discussion. But `examples/industrial.qet` has 1790 terminals and **zero** terminal uuids, so **all 671 of its conductors were omitted** — a wiring list for that project would have been completely empty. Across the 23 example projects, 16 of the 20 that contain conductors have no terminal uuids at all: 2366 of 3002 conductors, 7.3% overall coverage. That is not an ageing project-file problem that re-saving fixes; it is a property of the element catalog definitions.

   Rather than ship tables that are empty on almost every project that exists, `Terminal::stableUuid()` now returns the terminal's own uuid when it has one and otherwise derives one from its **local position and orientation** inside its element. That is not an invented scheme — it is what the project format already does, and `TerminalData::fromXml()` says so where it parses the field: *"if the attribute not exists, means, the element is created with an older version of qet. So use the legacy approach to identify terminals."* `m_pos` comes from the definition and is not touched by moving the element on a folio, so the identity survives loads, saves and folio moves. Derived values are UUID v5 in a fixed namespace, so they are reproducible without being written to the file and cannot collide with the v4 uuids the element editor generates. **No file-format change, no migration.**

   Two things it deliberately does not key on, both settled by measurement rather than argument:

   - **The terminal name**, because it is not stable. QET rewrites a terminal named `"_"` as unnamed, which would have silently changed the identity of 1421 of `industrial.qet`'s 1790 terminals on their first resave — caught by round-tripping the corpus. Dropping it costs nothing: geometry alone yields exactly the same three collisions across the whole corpus, and it means renaming a terminal no longer changes what it is.
   - **Uniqueness against a definition that declares two terminals at the same point and orientation.** Three such cases exist in the entire corpus. They merge to one terminal row, which is harmless — two terminals identical in position and orientation are indistinguishable in every observable respect, and every conductor on either still resolves to the right element and terminal name. Both affected projects (`industrial`, `perceuse`) return their full conductor count.

   The only conductor still skipped is one whose terminal has no parent element, which has no identity to key on under any scheme.

## Testing

All live, in the running app:

| Path | Result |
|---|---|
| Incremental add (`Diagram::addItem`) | fresh project, two aligned contacts → autoconnect → **2 terminals, 1 conductor** |
| Incremental remove (`Diagram::removeItem`) | delete that conductor → **1 → 0** |
| Undo | Ctrl+Z after delete → **0 → 1**, no duplicate-PK error (same `Conductor` object keeps its uuid) |
| Bulk populate (`updateDB`) | `weneedpolonez-Polonez_MR89_wiring_diagram.qet`, 366 conductors → **366 conductor rows** (was 280 before `stableUuid()`) |
| Join correctness | `conductor` → `terminal` (composite key) → `element_info` resolves real from-to rows with real element labels |
| Legacy-only project | `industrial.qet` → **671 of 671 conductors** (was 0), loads and renders fine, no crash |
| Whole corpus | 20 of 20 measurable projects return exactly as many conductor rows as the document has conductors, 0 mismatches |
| Identity stability | resave `industrial.qet` → the derived-identity key multiset is byte-for-byte unchanged across the round trip |

No SQL errors logged in any of the above. Built clean, no new warnings.

## Known limitation

`removeDiagram()` does not cascade-delete that diagram's conductor rows — exactly as it already does not cascade to `element`/`element_info`. A full `updateDB()` rebuild clears them, and the slice-3 view selects `FROM conductor`, so orphan terminal rows never surface. Left consistent with existing behaviour rather than changing it in a plumbing PR.

### DIFF
--- a/sources/dataBase/projectdatabase.cpp
+++ b/sources/dataBase/projectdatabase.cpp
@@ -271,13 +271,8 @@ void projectDataBase::addConductor(Conductor *conductor)
 	insertTerminal(conductor->terminal1);
 	insertTerminal(conductor->terminal2);
 
-	m_insert_conductor_query.bindValue(":uuid", conductor->uuid().toString());
-	m_insert_conductor_query.bindValue(":diagram_uuid", conductor->diagram()->uuid().toString());
-	m_insert_conductor_query.bindValue(":terminal1_uuid", conductor->terminal1->uuid().toString());
-	m_insert_conductor_query.bindValue(":terminal1_element_uuid", conductor->terminal1->parentElement()->uuid().toString());
-	m_insert_conductor_query.bindValue(":terminal2_uuid", conductor->terminal2->uuid().toString());
-	m_insert_conductor_query.bindValue(":terminal2_element_uuid", conductor->terminal2->parentElement()->uuid().toString());
-	m_insert_conductor_query.bindValue(":text", conductor->properties().text);
+	watchConductor(conductor);
+	bindConductorValues(m_insert_conductor_query, conductor, conductor->diagram());
 	if (!m_insert_conductor_query.exec()) {
 		qDebug() << "projectDataBase::addConductor insert error : " << m_insert_conductor_query.lastError();
 	} else {
@@ -297,6 +292,85 @@ void projectDataBase::removeConductor(Conductor *conductor)
 	} else {
 		emit dataBaseUpdated();
 	}
+}
+
+/**
+	@brief projectDataBase::updateConductor
+	Refresh the mutable columns of an already-inserted conductor.
+
+	Only the text (the wire number) can change without the conductor being
+	removed and re-added: its endpoints are fixed for its lifetime. Without
+	this, renaming a wire left the database holding the old number and the
+	wiring list showed a stale value until the next full repopulate.
+	@param conductor
+*/
+void projectDataBase::updateConductor(Conductor *conductor)
+{
+	if (!conductor) {
+		return;
+	}
+
+	m_update_conductor_query.bindValue(QStringLiteral(":uuid"), conductor->uuid().toString());
+	m_update_conductor_query.bindValue(QStringLiteral(":text"), conductor->properties().text);
+	if (!m_update_conductor_query.exec()) {
+		qDebug() << "projectDataBase::updateConductor update error : " << m_update_conductor_query.lastError();
+	}
+
+		//Deliberately no dataBaseUpdated() here, unlike add/remove. The only
+		//column this touches is the wire text, which no view watched by
+		//ProjectDBModel displays -- the nomenclature shows elements, and its
+		//wire_count changes when a conductor appears or disappears, not when
+		//it is renamed. Emitting would make every ProjectDBModel re-run its
+		//query, and auto-numbering renames every conductor in the project in
+		//one pass.
+}
+
+/**
+	@brief projectDataBase::watchConductor
+	Keep this conductor's row in step with its properties.
+
+	Conductor::setProperties() has a dozen call sites (auto-numbering, the
+	properties dialog, element moves, deletion re-links...), so listening to
+	the signal it already emits is the only way to catch them all -- and the
+	only way to catch the ones added later. Qt::UniqueConnection makes a
+	repeated insert or a full repopulate harmless.
+	@param conductor
+*/
+void projectDataBase::watchConductor(Conductor *conductor)
+{
+	connect(conductor, &Conductor::propertiesChange,
+			this, &projectDataBase::conductorPropertiesChanged,
+			Qt::UniqueConnection);
+}
+
+/**
+	@brief projectDataBase::conductorPropertiesChanged
+*/
+void projectDataBase::conductorPropertiesChanged()
+{
+	if (auto *conductor = qobject_cast<Conductor *>(sender())) {
+		updateConductor(conductor);
+	}
+}
+
+/**
+	@brief projectDataBase::bindConductorValues
+	One binder for both insert paths, so a conductor added to a live diagram
+	and one read from a file can never drift apart -- the same reason
+	bindElementValues() exists for elements.
+	@param query
+	@param conductor
+	@param diagram : the diagram the conductor belongs to
+*/
+void projectDataBase::bindConductorValues(QSqlQuery &query, Conductor *conductor, Diagram *diagram)
+{
+	query.bindValue(QStringLiteral(":uuid"), conductor->uuid().toString());
+	query.bindValue(QStringLiteral(":diagram_uuid"), diagram->uuid().toString());
+	query.bindValue(QStringLiteral(":terminal1_uuid"), conductor->terminal1->uuid().toString());
+	query.bindValue(QStringLiteral(":terminal1_element_uuid"), conductor->terminal1->parentElement()->uuid().toString());
+	query.bindValue(QStringLiteral(":terminal2_uuid"), conductor->terminal2->uuid().toString());
+	query.bindValue(QStringLiteral(":terminal2_element_uuid"), conductor->terminal2->parentElement()->uuid().toString());
+	query.bindValue(QStringLiteral(":text"), conductor->properties().text);
 }
 
 /**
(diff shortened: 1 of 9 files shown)


ispyisail:
Self-review, slice 2 of the #503 stack. Context in #625.

## A conductor's text is written once and never updated

The `conductor` table carries `text`, populated at insert from `conductor->properties().text`. There is no update path for it anywhere in the stack — I grepped the full merged tree at slice 5 for `UPDATE conductor`, `conductorInfoChanged`, `updateConductor`: nothing. The only hooks are add and remove, both in `diagram.cpp`:

```
1677:  m_project->dataBase()->addConductor(conductor);
1708:  m_project->dataBase()->removeConductor(conductor);
```

Elements have the equivalent (`elementInfoChanged` → `m_update_element_query`); conductors do not. So renaming a wire mid-session leaves the database holding the old number, and the wiring list in slice 4 shows the stale value — that dialog builds its model straight off the database and never calls `updateDB()`, so nothing masks it. It only corrects itself on the next full repopulate.

## The insert is duplicated, which is the thing slice 4 fixes for elements

`addConductor()` and `populateConductorTable()` each carry their own copy of the same seven `bindValue` lines. Slice 4 ("Insert an element the same way whether it is added live or on load") removes exactly this duplication for elements via `bindElementValues()` / `bindElementInfoValues()`, and the commit message is right that it mattered — the two element paths had genuinely diverged, one binding `kindInformations()["type"]` and the other `masterTypeToString()`. The conductor pair has not diverged yet, but it is the same shape of hazard and deserves the same treatment while the code is fresh.

## Checked and correct

- `populateConductorTable()` does `DELETE FROM conductor` and `DELETE FROM terminal` before repopulating, so a repeated `updateDB()` does not collide on the primary key.
- `updateDB()` correctly calls `populateConductorTable()`.
- The composite `PRIMARY KEY (uuid, element_uuid)` on `terminal` with `INSERT OR IGNORE` is the right call, and the reasoning in the commit message about `Terminal::uuid()` not being unique per placed terminal checks out.

ispyisail:

This is the slice that changed most, because as first posted the tables it adds were **empty on almost every project that exists**.

## Coverage: the reason for `Terminal::stableUuid()`

A conductor entered the `conductor` table only if *both* terminals carried a uuid from the catalog `.elmt`. Across the 23 example projects, **16 of the 20 containing wiring have none at all** — 2366 of 3002 conductors; 7.3% coverage overall.

| project | before | after |
|---|---|---|
| `industrial.qet` | 0 / 671 | **671 / 671** |
| `m_000.qet` | 0 / 457 | **457 / 457** |
| `affuteuse_250h.qet` | 0 / 263 | **263 / 263** |
| `tremie_vibrante.qet` | 0 / 77 | **77 / 77** |

**20 of 20 measurable projects match the document exactly, 0 mismatches.**

The fallback keys on the terminal's local position and orientation, which is what the format itself matches on — `TerminalData::fromXml()` says so in its own comment. Values are UUID v5 in a fixed namespace: reproducible without being stored, and unable to collide with the v4 uuids the element editor generates. **No file-format change, no migration** — the database is in-memory and rebuilt each session.

## Endpoints are correct, not merely present

Row counts alone would not catch a wrong join. Comparing the database's endpoint element pairs against the XML's own `element1`/`element2` declarations:

```
weneedpolonez : 280 XML endpoint pairs, 280 matched, 0 missing
tableau_domestique : 46 XML endpoint pairs, 46 matched, 0 missing
```

## The live hooks, including undo

`addConductor()` / `removeConductor()` / `updateConductor()` only run while editing, so no load-and-export path reaches them. Driven here through #683's `--test-ops` (a test vehicle only — this PR does not depend on it). Deleting an element cascades to its conductors, so delete → undo → redo covers remove, add, remove:

| project | baseline | after delete | after undo | after redo |
|---|---|---|---|---|
| weneedpolonez | 366 | 314 | **366** | 314 |
| schema_unifilaire_voltaique2 | 76 | 70 | **76** | 70 |
| photovoltaique | 64 | 57 | **64** | 57 |

`COUNT(DISTINCT uuid) == COUNT(*)` at every step. The undo row is the important one: the same `Conductor` objects are re-inserted carrying the uuids they already had, which is exactly where a broken identity scheme fails on the primary key.

## Two deliberate limits, both settled by measurement

- **The terminal name is not part of the key.** It is not stable: QET rewrites a terminal named `"_"` as unnamed, which would have changed the identity of **1421 of `industrial.qet`'s 1790 terminals** on the first resave — caught by round-tripping the corpus, not by reading. Dropping it costs nothing (geometry alone yields the same three collisions corpus-wide) and it means renaming a terminal no longer changes what it is.
- **Two terminals declared at the same point and orientation merge to one row.** Three such cases exist in the whole corpus. Harmless: such terminals are indistinguishable in every observable respect, and both affected projects (`industrial`, `perceuse`) return their full conductor count.

Identity is stable across a save cycle: resaving `industrial.qet` leaves the derived-key multiset byte-for-byte unchanged.

## Cost

Measured against a real `master` build, best of three, load + `updateDB()` + BOM query:

| project | master | this stack | delta |
|---|---|---|---|
| `industrial.qet` (671 conductors, 50 folios) | 3.91 s | 4.11 s | **+0.20 s** |
| `m_000.qet` | 3.01 s | 3.01 s | +0.00 |
| `weneedpolonez` | 1.20 s | 1.21 s | +0.01 |

Also in this slice: a conductor's `text` is now refreshed when it changes (it was written once at insert), the two insert paths share one binder so they cannot drift, and the conductor columns the nomenclature subquery scans are indexed — 2134 ms → 10 ms at 2000 elements × 5000 conductors.

scorpio810:
> industrial.qet (671 conductors, 50 folios)	3.91 s	4.11 s	+0.20 s
> m_000.qet	3.01 s	3.01 s	+0.00
> weneedpolonez

These are just a few small examples; at work, we have managed some very large-scale projects…
An example Enzo use one folio with 280MB...

PR #629, opened 2026-08-02
Add wiring_list_view: from-to wiring list over the conductor tables (discussion #503, slice 3)

ispyisail:
>
> **`--export-wiring` is added here.** The view was reachable only through the GUI, which meant the one thing worth proving about it — that it still describes the project — could not be checked without a person clicking. This is the same shape as the existing `--export-bom`, which reads `element_nomenclature_view`: it writes `wiring_list_view` as CSV. Two consequences:
>
> - the view becomes verifiable in CI, and
> - this slice is useful on its own, rather than waiting on the dialog in #630.
>
> The overlap with the existing `--export-cables` is deliberate. That one builds the same logical list from the document XML instead, so running both and diffing them is a direct check that the cache and the document still agree — nothing else in the codebase can do that. Measured on the corpus, they also differ in what they can fill in. Rows carrying any endpoint data:
>
> | project | `--export-cables` (XML) | `--export-wiring` (DB) |
> |---|---|---|
> | `industrial.qet` | 0 / 671 | **541 / 671** |
> | `m_000.qet` | 0 / 457 | **362 / 457** |
> | `affuteuse_250h.qet` | 0 / 263 | **197 / 263** |
> | `tremie_vibrante.qet` | 0 / 77 | **61 / 77** |
> | `tableau_domestique.qet` | 58 / 130 | **104 / 130** |
>
> Both return one row per conductor; the XML-derived one leaves the component and terminal columns empty on the older projects. The cause is precise and worth recording: `WiringListExport` resolves endpoints from `element1`/`element2` **attributes on the conductor XML**, and legacy files do not write them — `industrial.qet` has them on **0 of 671** conductors, which is exactly its score. The database path resolves them because at runtime the terminals are attached to real elements. That is not an argument for removing `--export-cables`: it carries columns the view does not, and being independent is what makes it a useful second opinion.
>
> **The `diagram` join is now `LEFT`.** The comment below argued carefully for the two joins that could drop rows, then ended with an inner join to `diagram` it never mentioned. Feeding the real schema a conductor whose `diagram_uuid` has no `diagram` row returned 2 view rows for 3 conductors; as a `LEFT JOIN` it returns 3, with a null folio instead of a missing wire. In practice it should never fire — `QETProject::diagramAdded` is wired to `addDiagram()` — but an inner join makes that an assumption the view enforces silently, and of everything this view can get wrong, dropping a wire from a wiring list is the one that matters most.

**Ready for review.** Merge #625 → #628 first; the commit to review here is `Add wiring_list_view: from-to wiring list over the conductor tables`. #630 follows.

Slice 3 of #503. One row per conductor, each endpoint resolved to its element label and terminal name — the `F1:4 -> M200:U1` shape from the original prototype.

> **Stacked on #628** (slice 2), which is itself stacked on #625 (slice 1). Neither is merged yet, so this branch contains all three commits. Merge order: #625 → #628 → this. The commit to review here is just `Add wiring_list_view: from-to wiring list over the conductor tables` (2 files, +52).

## The view

```sql
CREATE VIEW wiring_list_view AS SELECT
  c.uuid AS conductor_uuid,
  c.text AS wire_number,
  t1.element_uuid AS from_element_uuid, ei1.label AS from_element_label, t1.name AS from_terminal,
  t2.element_uuid AS to_element_uuid,   ei2.label AS to_element_label,   t2.name AS to_terminal,
  d.pos AS diagram_position
FROM conductor c
  JOIN terminal t1 ON c.terminal1_uuid = t1.uuid AND c.terminal1_element_uuid = t1.element_uuid
  JOIN terminal t2 ON c.terminal2_uuid = t2.uuid AND c.terminal2_element_uuid = t2.element_uuid
  LEFT JOIN element_info ei1 ON t1.element_uuid = ei1.element_uuid
  LEFT JOIN element_info ei2 ON t2.element_uuid = ei2.element_uuid
  LEFT JOIN diagram d ON c.diagram_uuid = d.uuid;
```

## Why it differs from the SQL sketched in the discussion

Both changes exist because the sketched version **silently loses wires**:

- **No join to the `element` table.** A terminal row already carries its `element_uuid`, so joining `element` back just to read the same uuid adds nothing. Worse, it *filters*: `populateElementTable()` only inserts elements matching `Simple|Terminal|Master|Thumbnail` (`ElementData::Slave` is `16`, not in that mask). So `Slave` elements — relay contacts and the like, **extremely common at the end of a wire** — and report elements are simply absent from that table after a project load, and an inner join through it drops their conductors.
- **`element_info` is LEFT joined** for the same reason. A wire whose endpoint element has no info row still belongs in a wiring list; it comes back with an empty label rather than vanishing. Losing a wire from a wiring list is a worse failure than showing one with a blank end.

**This only bites after a save/reload**, which is why it's easy to miss: the incremental `addElement()` path does *not* apply the type filter, so a slave element placed live is present in `element`/`element_info` and an inner join looks perfectly fine. It's the bulk repopulate on project load that drops it. Testing only the live-editing path would have shown nothing wrong.

> **Note added once slice 4 was finished.** The type-filter behaviour described just above is accurate *at this point in the stack*, but #630 goes on to remove it: it turned out that `element`/`element_info` had two insert paths disagreeing with each other, so `populateElementTable()` now inserts every element type and the `Simple|Terminal|Master|Thumbnail` restriction moves into `element_nomenclature_view`.
>
> The view below is deliberately **unchanged** by that. Its shape never depended on the filter, and the `LEFT JOIN element_info` is the right call either way — a wire whose endpoint carries no label still belongs in a wiring list. Flagging this so the reasoning here doesn't read as stale to whoever reviews after #630 lands.

## Measured

Both variants built from the same tables in the same session and counted side by side:

| project | conductors | `wiring_list_view` | inner-join variant |
|---|---|---|---|
| `weneedpolonez-Polonez_MR89_wiring_diagram.qet` | 280 | 280 | 280 |
| two slave contacts, **after save + reload** | 1 | **1** | **0** |

Polonez happens to have no slave elements at conductor ends, so both agree there and the problem is invisible — that project alone would have "confirmed" the sketched view. The second case is the minimal reproduction: place two `Simple contact` elements (`link_type="slave"`) so autoconnect wires them, save, reload — the sketched view returns **zero rows for a project that plainly has a wire in it**.

Acceptance criterion held throughout: **`wiring_list_view` row count == `conductor` row count**, i.e. the view itself drops nothing. Also verified the join produces real from-to rows with real element labels (e.g. `36 dipped beam relay`) and correct folio positions.

Conductors are no longer excluded for want of a terminal uuid: `Terminal::stableUuid()` in #628 derives an identity from the terminal's geometry, so `industrial.qet` now yields 671 rows rather than none. The only conductor still absent is one whose endpoint has no parent element at all, and #630 surfaces a count for that.

Built clean, no new warnings; no SQL errors on project load.

### DIFF
--- a/sources/cli_export.cpp
+++ b/sources/cli_export.cpp
@@ -70,6 +70,7 @@ const QHash<QString, QString> &exportFlags()
 		{"--export-cables", "cables"},
 		{"--export-wires", "wires"},
 		{"--export-bom", "bom"},
+		{"--export-wiring", "wiring"},
 		{"--export-nets", "nets"},
 		{"--export-links", "links"},
 		{"--info", "info"},
@@ -525,6 +526,54 @@ QHash<Element *, int> folioIndex(QETProject &project)
 	return folio;
 }
 
+/// From-to wiring list: one row per conductor, each endpoint resolved to its
+/// element label and terminal name.
+///
+/// Reads wiring_list_view out of the project database. --export-cables produces
+/// the same logical list from the document XML instead, and the two are meant
+/// to agree: running both and diffing them is a direct check that the database
+/// still describes the project, which is otherwise only observable through the
+/// GUI.
+int exportWiring(QETProject &project, const QString &output)
+{
+	// The project database is built lazily; force a (re)build before querying.
+	project.dataBase()->updateDB();
+
+	static const QStringList columns {
+		"wire_number", "from_element_label", "from_terminal",
+		"to_element_label", "to_terminal", "diagram_position", "conductor_uuid"
+	};
+
+	QSqlQuery query = project.dataBase()->newQuery(
+		"SELECT " % columns.join(", ") %
+		" FROM wiring_list_view ORDER BY diagram_position, wire_number");
+	if (!query.exec()) {
+		err << "Wiring list query failed: " << query.lastError().text() << "\n";
+		return 1;
+	}
+
+	QString csv = columns.join(";") % "\n";
+	int rows = 0;
+	while (query.next()) {
+		QStringList values;
+		for (int i = 0; i < columns.size(); ++i)
+			values << csvField(query.value(i).toString());
+		csv += values.join(";") % "\n";
+		++rows;
+	}
+
+	QFile file(output);
+	if (!file.open(QIODevice::WriteOnly | QIODevice::Text)) {
+		err << "Cannot open '" << output << "' for writing.\n";
+		return 1;
+	}
+	QTextStream fout(&file);
+	fout << csv;
+	file.close();
+	out << "Exported " << rows << " conductor(s) -> " << output << "\n";
+	return 0;
+}
+
 /// Electrical nets: groups of terminals joined into one potential.
 /// Walks QET's own potential graph, so each net is a connected component
 /// of terminals across all folios. The ground truth for connectivity.
@@ -823,6 +872,8 @@ int run(const QStringList &args)
 		return exportCsv(project, format, output);
 	if (format == "bom")
 		return exportBom(project, output);
+	if (format == "wiring")
+		return exportWiring(project, output);
 	if (format == "nets")
 		return exportNets(project, output);
 	if (format == "links")
--- a/sources/cli_export.h
+++ b/sources/cli_export.h
@@ -48,6 +48,7 @@ namespace CLIExport {
 		  qelectrotech --export-cables  <project.qet> <output.csv>
 		  qelectrotech --export-wires   <project.qet> <output.csv>
 		  qelectrotech --export-bom     <project.qet> <output.csv>
+		  qelectrotech --export-wiring  <project.qet> <output.csv>
 		  qelectrotech --export-nets    <project.qet> <output.json>
 		  qelectrotech --export-links   <project.qet> <output.csv>
 		  qelectrotech --info           <project.qet> [output.json]
@@ -60,6 +61,11 @@ namespace CLIExport {
 		cables: wiring list (one row per conductor) as CSV.
 		wires: list of distinct wire numbers as CSV.
 		bom: bill of materials (one row per element) as CSV.
+		wiring: from-to wiring list (one row per conductor) as CSV, read from
+		        the project database. Same logical list as `cables`, which
+		        reads the document XML instead; the two are meant to agree,
+		        so diffing them checks that the database still describes the
+		        project.
 		nets: electrical nets (connected-terminal groups) as JSON.
 		links: element cross-references (coil/contact) as CSV, with
 		       unresolved links flagged.
--- a/sources/dataBase/projectdatabase.cpp
+++ b/sources/dataBase/projectdatabase.cpp
@@ -271,13 +271,8 @@ void projectDataBase::addConductor(Conductor *conductor)
 	insertTerminal(conductor->terminal1);
 	insertTerminal(conductor->terminal2);
 
-	m_insert_conductor_query.bindValue(":uuid", conductor->uuid().toString());
-	m_insert_conductor_query.bindValue(":diagram_uuid", conductor->diagram()->uuid().toString());
-	m_insert_conductor_query.bindValue(":terminal1_uuid", conductor->terminal1->uuid().toString());
-	m_insert_conductor_query.bindValue(":terminal1_element_uuid", conductor->terminal1->parentElement()->uuid().toString());
-	m_insert_conductor_query.bindValue(":terminal2_uuid", conductor->terminal2->uuid().toString());
-	m_insert_conductor_query.bindValue(":terminal2_element_uuid", conductor->terminal2->parentElement()->uuid().toString());
-	m_insert_conductor_query.bindValue(":text", conductor->properties().text);
+	watchConductor(conductor);
+	bindConductorValues(m_insert_conductor_query, conductor, conductor->diagram());
 	if (!m_insert_conductor_query.exec()) {
 		qDebug() << "projectDataBase::addConductor insert error : " << m_insert_conductor_query.lastError();
 	} else {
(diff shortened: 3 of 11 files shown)


ispyisail:
Self-review, slice 3 of the #503 stack. Context in #625.

## The row-count claim in the comment is not quite true

The doc block argues carefully for two choices — no inner join to `element`, and `LEFT JOIN element_info` — and concludes:

> The result is that this view returns exactly as many rows as the conductor table holds ... and nothing new is dropped here.

Both of those choices are right, and I verified the `LEFT JOIN` behaves as claimed: a conductor whose endpoint element has no `element_info` row comes back with a null label rather than disappearing. But the view ends with a third join the comment never mentions:

```sql
JOIN diagram d ON c.diagram_uuid = d.uuid
```

That one is inner, so it can drop rows. Building the schema and feeding it a conductor whose `diagram_uuid` is not in the `diagram` table:

```
conductor rows   : 3
wiring_list_view : 2
dropped by view  : ['{c3}']
```

In practice this should not fire — `QETProject::diagramAdded` is wired to `projectDataBase::addDiagram()`, so a diagram exists before anything can be drawn on it — so I would not call it a live bug. But the comment currently promises something the SQL does not guarantee, and it is the kind of promise a future reader will rely on. Either make it `LEFT JOIN diagram` (and accept a null folio column in the impossible case) or say in the comment that the row-count guarantee is conditional on the diagram row existing.

## Checked and correct

- No row multiplication: `element_info.element_uuid` is the primary key, so neither `LEFT JOIN` can fan out.
- Both `terminal` joins are satisfied for every conductor row, because `addConductor()` and `populateConductorTable()` both `insertTerminal()` each endpoint before inserting the conductor.

ispyisail:

## The view is now checkable without a GUI

`--export-wiring` writes `wiring_list_view` as CSV, the same shape as the existing `--export-bom` writing `element_nomenclature_view`. Two consequences: the view can be verified in CI, and this slice is useful on its own rather than waiting on the dialog in #630.

`qet-dbcheck.py` uses it to confirm, on every corpus project and with no special build, that the database returns one row per conductor in the document, that it agrees with `--export-cables` reading the same list out of the XML, and that every `conductor_uuid` is distinct.

**22 projects, 0 failures, 0 warnings.**

## Which of the two wiring exports actually has data

Keeping both is the point — they are independent, so diffing them checks the cache against the document. But they are not equivalent in what they can fill in. Rows carrying any endpoint data:

| project | `--export-cables` (XML) | `--export-wiring` (DB) |
|---|---|---|
| `industrial.qet` | 0 / 671 | **541 / 671** |
| `m_000.qet` | 0 / 457 | **362 / 457** |
| `affuteuse_250h.qet` | 0 / 263 | **197 / 263** |
| `tremie_vibrante.qet` | 0 / 77 | **61 / 77** |
| `tableau_domestique.qet` | 58 / 130 | **104 / 130** |

Both return one row per conductor. The cause of the difference is specific rather than a quality judgement: `WiringListExport` resolves endpoints from `element1` / `element2` **attributes on the conductor XML**, and older files never write them — `industrial.qet` carries them on **0 of 671** conductors, which is exactly its score. The database path resolves them because at runtime the terminals are attached to real elements.

This is **not** an argument for removing `--export-cables`. It carries columns the view does not (tension/protocol, colour, section, function), and being independent is what makes it useful as a second opinion. It does mean the database path is the one with the data on the projects people actually have.

Worth noting for expectations: the *terminal name* columns come back empty on most projects, and that is absent source data rather than loss in transit — `tableau_domestique.qet` has no terminal name on any of its 457 terminals, and `industrial.qet` stores the `"_"` placeholder on 1421 of 1790.

## One correctness fix in the view

The `diagram` join was inner. The comment above the view argued carefully for the two joins that could drop rows, then ended with a third it never mentioned. Feeding the real schema a conductor whose `diagram_uuid` has no `diagram` row returned **2 view rows for 3 conductors**; as a `LEFT JOIN` it returns 3, with a null folio instead of a missing wire.

In practice it should never fire, since `QETProject::diagramAdded` is wired to `addDiagram()`. But an inner join makes that an assumption the view enforces silently, and of everything this view can get wrong, dropping a wire from a wiring list is the one that matters most.

ispyisail:
@scorpio810 — following up now that #625 and #628 have landed, since this one (slice 3) and #630 (slice 4) are the rest of the #503 stack and both are ready.

To make this easy to pick up: #629 is clean against current master (no rebase needed), and I spent some extra time today re-verifying it independently rather than just standing on the earlier self-review comments above:

- Re-checked `Terminal::stableUuid()`'s core claim — that `d->m_pos` is a local, definition-scoped position, set once at construction and never mutated by moving/rotating the element on the folio. Confirmed directly (`mapToScene(d->m_pos)` elsewhere in the file requires it to be local; no other assignment site exists outside the constructor), so the derived-uuid fallback is genuinely stable the way the doc comment claims.
- Re-verified the `LEFT JOIN diagram` fix from the self-review above is actually in the current SQL (it is).
- One small doc nit, not worth a re-push on its own: the PR body says `Terminal::stableUuid()` is "in #628," but it's actually introduced in this PR's own diff (#628 is merged and doesn't touch `terminal.cpp`). Just flagging so it doesn't confuse whoever reads it after merge — happy to fix if you'd rather not have it linger.

#630 is queued right behind this one and is also clean against master — I did the same independent pass there too (checked `allElementTypes()` covers the full `ElementData::Type` enum with no gaps, and that `element_nomenclature_view`'s new `e.type IN (...)` filter reproduces the exact strings `typeToString()` used to filter on in C++, so the nomenclature view's own output is provably unchanged by the refactor).

No new code needed on my end for either — just wanted to give you a clean, re-verified starting point rather than asking you to re-derive all of the above yourself.

ispyisail:
@scorpio810 — now that I have write access, I want to check with you rather than assume.

These two are the remaining half of the #503 stack: **#629** (`wiring_list_view` + `--export-wiring`) and **#630** (the wiring list dialog). #625 and #628 landed on 24 August; these have been sitting since then.

## They are still good

I re-verified today rather than relying on the August self-reviews, because master has moved a long way since:

- **373 commits behind master**, and both still merge **clean** — no rebase needed
- the full stack **builds clean** against current master
- `--export-wiring examples/industrial.qet` exports **671 conductors**, matching the figure the dialog reported in my August verification on #630 — so the view and the dialog still agree with each other, and with what they said three weeks ago
- `qet-dbcheck --round-trip` over the whole corpus: **23 projects, 0 failures, 0 warnings**, including "database wiring list covers all N conductors" on every project that has any

#630 contains #629, so merging #630 takes both; or #629 can go on its own, since `--export-wiring` makes it useful without the dialog.

## What I am actually asking

I could merge these myself now, but I would rather not do that as the first use of the access you have given me. They are 1,254 lines across 16 files in `projectDataBase`, which is central enough that I would want your eyes on it — or at least your explicit "go ahead".

So: happy to merge them myself if you are content, happy to wait if you would rather look first, and happy to close them if you have changed your mind about the direction. No rush from my side — they have waited three weeks and can wait longer.

If it helps, the two self-review comments above already flag the two things I found wrong in my own work — the row-count claim in this PR's doc block, and `QSqlQueryModel::rowCount()` under-reporting on projects over 256 wires in #630 — both fixed in the commits that followed.

PR #630, opened 2026-08-02
Wiring list dialog + excluded-conductor count (discussion #503, slice 4)

ispyisail:
>
> **The excluded-conductor count no longer describes reality, and has been realigned.** It counted conductors whose terminals had no uuid — the right rule when that was why they were dropped. `Terminal::stableUuid()` in #628 means those conductors are now *in* the table, so left unchanged this dialog would have announced **"671 conductors excluded" on `industrial.qet` while listing all 671 of them** — a worse failure than the one the count exists to prevent, because it undermines a list that is now correct. Both the count and its message now describe the case that actually remains: an endpoint attached to no element at all.
>
> **The "N conducteur(s) listé(s)" figure was wrong above 256.** `QSqlQueryModel` fetches lazily, so `rowCount()` straight after `setQuery()` returns the rows fetched so far, not the size of the query. Measured with Qt's own QSQLITE driver: a 1000-row view reports 256 until the model is drained, then 1000. The test project quoted in #628 has 280 conductors, so this was already displaying 256 on our own data. The irony is that this line exists to *stop* a misleading number. The model is now drained before the count is read.
>
> **The dialog refreshes the database before building its model.** It queries the database rather than the diagrams, so anything not yet written through was invisible here.

Slice 4 of #503. **Ready for review** — the gap that originally made this a draft is closed; see "The gap that made this a draft" below for what changed and what a reviewer should sign off on.

> Stacked on #629 → #628 → #625. Merge in that order. The commits to review here are the last three: `a0fcfb8`, `07377c4`, `2cc02ec`.

## What it adds

`Projet > Liste de câblage (base de données)` opens a read-only table of `wiring_list_view`, headed by a line saying how many conductors are listed and — when non-zero — how many were **excluded and why**.

`projectDataBase::excludedConductorCount()` counts, from the live scene, the conductors absent from the `conductor` table. Counted from the scene precisely because the database is where those conductors *are not*.

| project | listed | excluded |
|---|---|---|
| `examples/industrial.qet` | **671** | **0** |
| two contacts with uuid-bearing terminals | 1 | 0 |

## Deliberately *not* another exporter

While building this I checked what already ships: `Projet > Exporter le plan de câblage` and `--export-cables` (`WiringListExport`) walk the project XML and already produce a from-to CSV.

**Correction to an earlier revision of this description.** It said my impression that that exporter produced blank output was wrong, and blamed unlabelled test projects. Measuring properly says otherwise: on the *same* project, `--export-cables` fills endpoint data on **0 of 671** rows of `industrial.qet` while the database view fills **541 of 671**. The cause is specific — `WiringListExport` resolves endpoints from `element1`/`element2` attributes on the conductor XML, and legacy files never write them (0 of 671 conductors in `industrial.qet` carry one). So it is not the projects being unlabelled; it is that the XML exporter has nothing to resolve on files that use the older positional scheme, while the database path resolves them because at runtime the terminals are attached to real elements.

That still is not an argument for a competing CSV export, and `--export-cables` keeps columns the view does not have. It *is* the argument for this dialog: on the projects people actually have, the database path is the one with the data in it.

## The gap that made this a draft — now closed

The draft flagged that after a save/reload the component columns came back blank for slave elements, and offered three ways out. Two of them were workarounds. The last two commits take the third route and fix the cause, because the cause turned out to be worth fixing on its own merits.

**The actual cause was not the type filter.** It was that `element`/`element_info` had *two* insert paths that disagreed with each other:

- incremental — `addElement()`, via `Diagram::addItem()`, applied **no** type filter;
- bulk — `populateElementTable()`/`populateElementInfoTable()`, via `updateDB()`, applied `Simple|Terminal|Master|Thumbnail`.

So the same project produced a different `element` table depending on whether you had just drawn it or just opened it. That is why the symptom was invisible while editing and appeared only after reload — and it was never specific to the wiring list.

Two commits close it:

- **`07377c4` — populate every element type; move the nomenclature filter into its view.** The bulk path now inserts all eight types, and the `Simple|Terminal|Master|Thumbnail` restriction moves into `element_nomenclature_view`, where it is a property of the nomenclature rather than of the table everything else reads. BOM contents are unchanged by construction: the same filter, applied one layer later.
- **`2cc02ec` — insert an element the same way whether it is added live or on load.** Both paths now bind through shared `bindElementValues()`/`bindElementInfoValues()` helpers, modelled on the existing `bindDiagramInfoValues()` idiom, so the two paths cannot drift apart again. This is the part that makes the fix structural rather than a second filter kept in sync by hand.

## Verification

| check | result |
|---|---|
| `--export-bom` vs pre-change baseline, 19 projects | identical content on 19/19; line-for-line on 18/19 |
| wiring list end-to-end, after save + reload | `K2 → K1`, re-confirmed on the final commit |
| live vs reloaded `element` table | identical (`slave/ ×1` in both) |
| excluded-conductor count | 0 on `industrial.qet` and on a uuid-bearing project alike |
| load cost, `industrial.qet` (150 folios) | 0.210 s → 0.233 s |
| build | clean, no new warnings |

On the one project that is not line-for-line identical (`photovoltaique`), three **byte-identical** rows appear in a different order. They tie on `ORDER BY label`, and the existing query never defined an order among ties — both orderings are equally valid outputs of the unchanged query. No row was added, removed or altered.

## What a reviewer should decide consciously

**In a live session, nomenclature contents now match what a reload produces.** Before, the two could disagree; they now agree by construction. The saved-and-reloaded result was already the one users saw in a fresh session, so this aligns the live case to it rather than the other way round — but it *is* a behaviour change in the live case, and it belongs to you, not to me.

## Known limitations, deliberately not addressed here

- `removeDiagram()` does not cascade-delete that diagram's `conductor` rows. This matches its existing behaviour for `element`/`element_info`; a full `updateDB()` clears them, and the view selects `FROM conductor`, so orphans never surface in output. Worth fixing, but as its own change to the existing pattern.
- Terminal-name columns are blank for catalog elements that define no terminal names. That is source data, not a defect.

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -155,6 +155,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/conductornumexport.cpp
   ${QET_DIR}/sources/wiringlistexport.h
   ${QET_DIR}/sources/wiringlistexport.cpp
+  ${QET_DIR}/sources/ui/wiringlistdialog.h
+  ${QET_DIR}/sources/ui/wiringlistdialog.cpp
   ${QET_DIR}/sources/conductornumexport.h
   ${QET_DIR}/sources/conductorprofile.cpp
   ${QET_DIR}/sources/conductorprofile.h
--- a/sources/cli_export.cpp
+++ b/sources/cli_export.cpp
@@ -70,6 +70,7 @@ const QHash<QString, QString> &exportFlags()
 		{"--export-cables", "cables"},
 		{"--export-wires", "wires"},
 		{"--export-bom", "bom"},
+		{"--export-wiring", "wiring"},
 		{"--export-nets", "nets"},
 		{"--export-links", "links"},
 		{"--info", "info"},
@@ -525,6 +526,54 @@ QHash<Element *, int> folioIndex(QETProject &project)
 	return folio;
 }
 
+/// From-to wiring list: one row per conductor, each endpoint resolved to its
+/// element label and terminal name.
+///
+/// Reads wiring_list_view out of the project database. --export-cables produces
+/// the same logical list from the document XML instead, and the two are meant
+/// to agree: running both and diffing them is a direct check that the database
+/// still describes the project, which is otherwise only observable through the
+/// GUI.
+int exportWiring(QETProject &project, const QString &output)
+{
+	// The project database is built lazily; force a (re)build before querying.
+	project.dataBase()->updateDB();
+
+	static const QStringList columns {
+		"wire_number", "from_element_label", "from_terminal",
+		"to_element_label", "to_terminal", "diagram_position", "conductor_uuid"
+	};
+
+	QSqlQuery query = project.dataBase()->newQuery(
+		"SELECT " % columns.join(", ") %
+		" FROM wiring_list_view ORDER BY diagram_position, wire_number");
+	if (!query.exec()) {
+		err << "Wiring list query failed: " << query.lastError().text() << "\n";
+		return 1;
+	}
+
+	QString csv = columns.join(";") % "\n";
+	int rows = 0;
+	while (query.next()) {
+		QStringList values;
+		for (int i = 0; i < columns.size(); ++i)
+			values << csvField(query.value(i).toString());
+		csv += values.join(";") % "\n";
+		++rows;
+	}
+
+	QFile file(output);
+	if (!file.open(QIODevice::WriteOnly | QIODevice::Text)) {
+		err << "Cannot open '" << output << "' for writing.\n";
+		return 1;
+	}
+	QTextStream fout(&file);
+	fout << csv;
+	file.close();
+	out << "Exported " << rows << " conductor(s) -> " << output << "\n";
+	return 0;
+}
+
 /// Electrical nets: groups of terminals joined into one potential.
 /// Walks QET's own potential graph, so each net is a connected component
 /// of terminals across all folios. The ground truth for connectivity.
@@ -823,6 +872,8 @@ int run(const QStringList &args)
 		return exportCsv(project, format, output);
 	if (format == "bom")
 		return exportBom(project, output);
+	if (format == "wiring")
+		return exportWiring(project, output);
 	if (format == "nets")
 		return exportNets(project, output);
 	if (format == "links")
--- a/sources/cli_export.h
+++ b/sources/cli_export.h
@@ -48,6 +48,7 @@ namespace CLIExport {
 		  qelectrotech --export-cables  <project.qet> <output.csv>
 		  qelectrotech --export-wires   <project.qet> <output.csv>
 		  qelectrotech --export-bom     <project.qet> <output.csv>
+		  qelectrotech --export-wiring  <project.qet> <output.csv>
 		  qelectrotech --export-nets    <project.qet> <output.json>
 		  qelectrotech --export-links   <project.qet> <output.csv>
 		  qelectrotech --info           <project.qet> [output.json]
@@ -60,6 +61,11 @@ namespace CLIExport {
 		cables: wiring list (one row per conductor) as CSV.
 		wires: list of distinct wire numbers as CSV.
 		bom: bill of materials (one row per element) as CSV.
+		wiring: from-to wiring list (one row per conductor) as CSV, read from
+		        the project database. Same logical list as `cables`, which
+		        reads the document XML instead; the two are meant to agree,
+		        so diffing them checks that the database still describes the
+		        project.
 		nets: electrical nets (connected-terminal groups) as JSON.
 		links: element cross-references (coil/contact) as CSV, with
 		       unresolved links flagged.
--- a/sources/dataBase/projectdatabase.cpp
+++ b/sources/dataBase/projectdatabase.cpp
@@ -113,14 +113,17 @@ QSqlQuery projectDataBase::newQuery(const QString &query) {
 
 /**
 	@brief projectDataBase::excludedConductorCount
-	@return how many conductors of the project are deliberately absent from
-	the conductor table because at least one of their terminals has no uuid.
+	@return how many conductors of the project are absent from the conductor
+	table because an endpoint has no parent element to key on.
 
 	Counted from the live scene rather than from the database, precisely
-	because the database is where these conductors are *not*. See
-	addConductor() for why they are omitted: a terminal uuid comes from the
-	catalog .elmt definition, so an element whose definition predates that
-	field yields terminals with no stable identity to key on.
+	because the database is where these conductors are *not*.
+
+	This used to count conductors whose terminals had no uuid, which was most
+	of them on most projects. Terminal::stableUuid() now derives an identity
+	from the terminal's geometry when the definition provides no uuid, so that
+	is no longer a reason to exclude anything, and this counts only the case
+	that remains genuinely unkeyable.
 
 	This is what lets a caller tell the user "N wires are missing and here
 	is why", instead of silently presenting a short list as if it were
@@ -138,8 +141,10 @@ int projectDataBase::excludedConductorCount() const
 		const auto conductor_list = diagram->conductors();
 		for (auto *conductor : conductor_list)
 		{
-			if (conductor->terminal1->uuid().isNull()
-				|| conductor->terminal2->uuid().isNull()) {
+				//Must match addConductor()'s guard exactly, or this reports
+				//wires as missing that the list is in fact showing.
+			if (!conductor->terminal1->parentElement()
+				|| !conductor->terminal2->parentElement()) {
 				++count;
 			}
 		}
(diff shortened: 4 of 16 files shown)


ispyisail:
Self-review, slice 4 of the #503 stack. Context in #625.

## The "N conducteur(s) listé(s)" figure is wrong on any project over 256 wires

```cpp
const int listed = model->rowCount();
```

`QSqlQueryModel` fetches lazily. Straight after `setQuery()` it has only pulled the first batch, so `rowCount()` reports that batch, not the query's true size. Measured with Qt's own QSQLITE driver rather than assumed:

```
rows actually in the view       : 1000
rowCount() right after setQuery : 256
rowCount() after fetchMore loop : 1000
```

The count only becomes correct once a user scrolls far enough to force the remaining fetches, and it is read here before the view is even shown. Worth noting the test project quoted in slice 2 has 280 conductors, so this would already be displaying 256 on our own test data — plausible enough that nobody would notice.

The irony is that this line exists to stop a misleading number, since the surrounding comment explains the count matters so an empty list is distinguishable from a fully-excluded one. As written it introduces one. Either drain the model first:

```cpp
while (model->canFetchMore()) model->fetchMore();
```

or ask the database directly with a `SELECT COUNT(*) FROM wiring_list_view`, which avoids materialising rows nobody has scrolled to.

## Also: the dialog never refreshes the database

The model is built directly on `m_project->dataBase()->database()` with no `updateDB()` call. Combined with there being no update path for `conductor.text` (see #628), a wire renamed during the session shows its old number here. Refreshing on open would fix both halves at once.

## Checked and correct

The element-insert unification is right, and it fixes a real latent mismatch rather than just tidying. `addElement()` was binding `kindInformations()["type"]` into `element_sub_type` while `populateElementTable()` bound `masterTypeToString()`. Every `element_sub_type` filter in `ElementQueryWidget` compares against `ElementData::masterTypeToString(...)` — commutator, coil, protection, plc — so unifying on that spelling makes live-added elements match the filters they previously could miss. `masterTypeToString()` returning empty for non-master types is harmless, since no filter looks for a non-master sub_type.

ispyisail:

This was the last untested surface in the stack: a dialog cannot be checked by loading and exporting a file. It has now been opened headlessly and read.

## The dialog, on `industrial.qet`

That is the project which returned **zero** rows before this work. Opened under Xvfb + openbox, captured from the live window:

```
671 conducteur(s) listé(s).

  Fil  | Composant 1 | Borne 1 | Composant 2 | Borne 2 | Folio
  308  | 4Q2         |         |             |         | 4
  309  | 4Q2         |         |             |         | 4
  401  | X1          |         | 4Q1         |         | 4
  401  | 4Q3         |         | X1          |         | 4
  404N | X1          |         | 4Q2         |         | 4
  421  | 4Q3         |         | 4Q4         |         | 4
  ...
```

That single screenshot confirms all three fixes in this slice at once:

1. **The count is right.** `QSqlQueryModel` fetches lazily, so `rowCount()` straight after `setQuery()` returned the first batch — 256 — not the size of the query. Measured with Qt's own QSQLITE driver: a 1000-row view reports 256 until drained, then 1000. The test project quoted in #628 has 280 conductors, so this was already showing 256 on our own data. The header now reads 671 because the model is drained first.
2. **No bogus exclusion warning.** `excludedConductorCount()` counted conductors whose terminals had no uuid — correct when that was why they were dropped. After `Terminal::stableUuid()` in #628 those conductors are *in* the table, so left alone this dialog would have announced **"671 conductors excluded" while listing all 671** — worse than the failure the count exists to prevent, because it undermines a list that is now correct. Both the count and its message now describe what actually remains: an endpoint attached to no element.
3. **The data is current.** The dialog refreshes the database before building its model.

Blank `Borne 1` / `Borne 2` is absent source data, not loss: `industrial.qet` stores the `"_"` placeholder on 1421 of its 1790 terminals.

## A correction to this PR's own description

An earlier revision said my impression that `--export-cables` produced blank output was wrong, and blamed unlabelled test projects. Measuring properly says otherwise: on the *same* project, `--export-cables` fills endpoint data on **0 of 671** rows while the database view fills **541 of 671**. The cause is that `WiringListExport` resolves endpoints from `element1`/`element2` attributes on the conductor XML, which older files never write (0 of 671 in `industrial.qet`).

That is still not an argument for a competing CSV export — `--export-cables` carries columns the view does not. It is the argument for this dialog: on the projects people actually have, the database path is the one with data in it.

## Reproducing it

`qet-gui-dialog.sh` in the `qelectrotech-docker` repo. Three things it encodes, each of which cost real time:

- Xvfb alone is not enough — without a window manager xdotool cannot activate a window and Qt menus never open. `openbox` is required.
- QET raises a **chain** of modal dialogs when opening a project (*Créer une copie de sauvegarde ?*, then *Fichier de restauration*), each swallowing input until dismissed. Dismissal has to be generic, not by name. This is also, I suspect, why `schema_indus.qet` blocks at zero CPU under any CLI flag — the case #661 addresses.
- `xwd -root` captures dialogs as solid black; `xwd -id <window>` captures the real contents. An empty-looking dialog is probably a capture artifact.

ispyisail:
@scorpio810 — this is slice 4 of the #503 stack; I have put the ask on #629 rather than repeating it here: https://github.com/qelectrotech/qelectrotech-source-mirror/pull/629#issuecomment-5613131105

Short version: both remaining slices are still clean against current master after 373 commits, the stack builds, and `qet-dbcheck --round-trip` passes on all 23 corpus projects with 0 failures and 0 warnings. This PR contains #629, so merging it takes both.

I have write access now but would rather have your go-ahead than merge my own database work unreviewed. Equally happy to wait, or to close if you would rather the direction changed.

scorpio810:
@ispyisail — thank you for the thorough work here, this stack is genuinely well documented and well tested. That said, I need to be upfront about pace: I'm currently supposed to be on vacation and resting, so review time on my end is limited right now, not a fixed schedule I can commit to.

Also, on the corpus: the example projects available to test against aren't representative of the large industrial projects some of our users run in production. Those can't be shared publicly for confidentiality reasons, so I can't just diff against the public examples and call it validated — I need to test this properly against that kind of scale and complexity before merging, specifically to make sure it doesn't break real users' projects.

I'll get to it as soon as I reasonably can. No need to close it in the meantime — just please don't read the delay as a sign something's wrong with the work.

scorpio810:
On top of that, I currently have too many PRs open and waiting on review at once, and I'm underwater trying to keep up. Would you be willing to close some of the related PRs in this stack for now and also so related with merged PR?

It would help me get through the backlog, and we can reopen or resubmit whichever ones are still relevant once I've caught up.

None of this reflects on the quality of the work — just where my bandwidth is right now.

ispyisail:
Done — I have closed **48** of my open PRs. You now have 2 from me instead of 50.

One thing I want to be straight about: you asked about this stack specifically, and these two are the ones I have kept open. That is the opposite of what you named, so it deserves an explanation rather than just doing it quietly.

The reason is focus. Rather than trim a few PRs and leave you with forty-odd, I would rather clear almost everything and work on one thing properly — and the wiring database is the piece I want to see through, because half of it is already merged (#625, #628) and leaving the other half unfinished is the worst of both worlds.

If you would rather these two went as well, say so and they are gone in a minute — the branches survive, so nothing is lost either way. Genuinely no attachment; the point was to get your queue down, and 50 → 0 is as good as 50 → 2 from my side.

For what it is worth, everything closed was closed on bandwidth, not on merit, and each one says so, so anything worth resurrecting is one click away.

scorpio810:
Thanks for this slice — careful work, and the reasoning in the comments and commit messages made it easy to review.

**Merge order.** This branch is stacked on #629 and contains its two commits (`6061c63`, `44ed01f`), so merging #630 merges #629 as well. Worth keeping in mind that #629 has not been reviewed on its own.

**What looks good**

- `excludedConductorCount()` mirrors the guard in `addConductor()` / `populateConductorTable()` exactly, so the count cannot disagree with the list.
- The view keeps every conductor: LEFT JOINs to `element_info` and `diagram`, no join to `element`. The comment explaining why is accurate.
- Draining the model with `fetchMore()` before reading `rowCount()` is the right fix for the 256-row prefetch.
- `bindElementValues()` / `bindElementInfoValues()`: for masters, `masterTypeToString()` produces the same strings `ElementQueryWidget` filters on (`coil`, `protection`, `commutator`, `plc`). For non-masters nothing reads `sub_type`, so the empty value is harmless. Moving the type mask into `element_nomenclature_view` keeps the nomenclature output unchanged.

** Minor**

- `ORDER BY diagram_position, wire_number` sorts wire numbers as text, so "10" comes before "9". Not a blocker, just worth knowing.
- `exportWiring()` follows the existing CLI exporters (`QTextStream`, plain `QFile`). On Qt6 the output is UTF-8, so encoding is fine. Once #830 is in, it could optionally reuse `BomExport::writeCsv()` to get a BOM, which Excel needs to detect UTF-8 when opening the file directly, and an atomic write.

PR #631, opened 2026-08-02
BOM: wire count per element (discussion #503, slice 5)

ispyisail:
>
> **The subquery now has indexes behind it.** `COUNT(DISTINCT c.uuid)` is correct, but the shape is a correlated scalar subquery evaluated once per element row, and there was no index on either conductor column it filters — `EXPLAIN QUERY PLAN` confirmed `CORRELATED SCALAR SUBQUERY` over a full scan. On a standalone SQLite harness at 2000 elements × 5000 conductors: **2134 ms unindexed, 10 ms indexed**, a 204× difference. This sits on `element_nomenclature_view`, which `updateDB()` rebuilds, so the cost landed on BOM export and every nomenclature refresh rather than on one dialog.
>
> The two indexes live in **#628** alongside the `conductor` table itself, since that is where the columns are declared and they carry foreign keys that attract lookups anyway. Only a comment lands here, tying the query to the thing that makes it affordable — the two live in different files and nothing else connects them.
>
> **The wire count is also now meaningful on legacy projects.** As first posted, a conductor reached the `conductor` table only if both terminals carried a catalog uuid, so `wire_count` read 0 for every element on `industrial.qet` and on 16 of the 20 example projects containing wiring. `Terminal::stableUuid()` in #628 fixes that at the root; all 20 measurable projects now contribute their full conductor count.

**Ready for review.** Slice 5 of #503 — the "strengthened BOM" half. Merge #625 → #628 → #629 → #630 first; the commit to review here is the last one, `Add a wire count per element to the nomenclature view` (2 files, +10/-1).

## What it adds

`element_nomenclature_view` gains a `wire_count` column, and the nomenclature query builder offers it as **Nombre de fils** next to the other non-info columns it already exposes (`position`, `folio`, `title`, `diagram_position`). This is the payoff for slices 1–4: the conductor connectivity is in the database, so a BOM can now carry "how many wires land on this part" without any new table.

```sql
(SELECT COUNT(DISTINCT c.uuid) FROM conductor c
  WHERE c.terminal1_element_uuid = e.uuid
     OR c.terminal2_element_uuid = e.uuid) AS wire_count
```

Written as a correlated subquery rather than a `LEFT JOIN` so the view's existing comma-join `FROM` clause is left exactly as it was — mixing a `LEFT JOIN` into that list is the kind of edit that changes results by accident.

## Why `COUNT(DISTINCT c.uuid)` and not a sum of endpoints

Because **a conductor with both ends on the same element is one wire touching it, not two** — and that case is common, not hypothetical:

| project | captured conductors | of which self-loops |
|---|---|---|
| `weneedpolonez-Polonez_MR89_wiring_diagram.qet` | 280 | **55 (20%)** |

I did not take that number on trust, since 20% looked high enough to suspect slice 2 was mis-populating `terminal*_element_uuid`. Parsing the project XML directly and counting conductors whose `element1` == `element2` gives **exactly 280 and 55** — the same figures the database produces, from a completely independent path. So the self-loops are real source data, and this doubles as independent confirmation that #628's conductor population is correct.

Summing the two endpoint columns instead would have silently overcounted all 55.

## Verified

**Invariant, all 22 example projects.** For every project, `SUM(wire_count)` over the whole `element` table must equal `2 × conductors − self_loops`. Checked by a temporary probe, removed before commit:

```
photovoltaique      conductors=22  selfloops=0   sum=44   expected=44   OK
schema_unifilaire…  conductors=33  selfloops=0   sum=66   expected=66   OK
tableau_domestique  conductors=46  selfloops=0   sum=92   expected=92   OK
Polonez             conductors=280 selfloops=55  sum=505  expected=505  OK
…18 more, all 0 conductors, all OK
```

Polonez is the one that actually exercises the `DISTINCT`: it is the only example with self-loops, and it is where the naive version would have returned 560 instead of 505.

**No change to existing output.** `--export-bom` across all 22 examples, before vs after: 21 byte-identical. `photovoltaique` has the same 64 rows with identical content as a multiset — four byte-identical rows (all with an empty `label`) sit in a different order. They tie on `ORDER BY label`, which has never defined an order among ties, so both orderings are valid outputs of the unchanged query. No row added, removed or altered.

**Cost.** Best of 3, `--export-bom`:

| project | before | after |
|---|---|---|
| `industrial.qet` (150 folios, 354 BOM rows) | 5.71 s | 5.51 s |
| `Polonez` (191 BOM rows, 280 conductors) | 1.30 s | 1.30 s |

Within noise — the subquery only runs when something queries the view, never on project load.

## The honest caveat

An element wired only by conductors whose terminals have no uuid reads **0**, because those conductors are absent from the `conductor` table by design (#628). `examples/industrial.qet` reads 0 for all 354 of its BOM rows for exactly this reason.

Worth knowing how narrow that gap actually is, because I had it wrong earlier in #503 and #628. I described it there as a property of the element catalog that a user cannot fix by re-saving. That was wrong in one important respect: **the shipped collection is essentially fully covered** — 5993 of the 5996 `.elmt` files that define terminals carry terminal uuids. The projects that come up empty do so because a `.qet` embeds its *own snapshot* of each element definition, and an old project's snapshot predates the uuid field:

| | embedded terminals | with uuid |
|---|---|---|
| `industrial.qet` | 1790 | **0** |
| `Polonez` | 1041 | 364 |
| shipped collection | 5996 | **5993** |

So a project drawn today gets full coverage, and this column is populated for it. Old projects stay at 0 until their embedded definitions are refreshed — and I could not find an existing action that does that refresh, which looks like the real follow-up work, separate from this PR.

## Scope note: this is the BOM half of slice 5 only

Slice 5 was scoped in #503 as "terminal plans / BOM strengthening". I deliberately did **not** build terminal plans on `wiring_list_view`, because QET already has a 29-file `sources/TerminalStrip/` subsystem — `TerminalStrip`, `PhysicalTerminal`, `RealTerminal`, bridges, a strip editor, a layout editor and a tree dock — with its own model that reaches conductors directly (`RealTerminal::conductor()`). Building a second terminal-plan path over the wiring view would compete with it, which is the same mistake I avoided in #630 by not writing a second CSV exporter.

What that subsystem does appear to lack is any export at all (no CSV/export anywhere under `sources/TerminalStrip/`). If terminal plans are wanted, that looks like the right place to start, built on the strip model rather than on this view — happy to take it on as its own piece of work if you agree.

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -155,6 +155,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/conductornumexport.cpp
   ${QET_DIR}/sources/wiringlistexport.h
   ${QET_DIR}/sources/wiringlistexport.cpp
+  ${QET_DIR}/sources/ui/wiringlistdialog.h
+  ${QET_DIR}/sources/ui/wiringlistdialog.cpp
   ${QET_DIR}/sources/conductornumexport.h
   ${QET_DIR}/sources/conductorprofile.cpp
   ${QET_DIR}/sources/conductorprofile.h
--- a/sources/cli_export.cpp
+++ b/sources/cli_export.cpp
@@ -70,6 +70,7 @@ const QHash<QString, QString> &exportFlags()
 		{"--export-cables", "cables"},
 		{"--export-wires", "wires"},
 		{"--export-bom", "bom"},
+		{"--export-wiring", "wiring"},
 		{"--export-nets", "nets"},
 		{"--export-links", "links"},
 		{"--info", "info"},
@@ -525,6 +526,54 @@ QHash<Element *, int> folioIndex(QETProject &project)
 	return folio;
 }
 
+/// From-to wiring list: one row per conductor, each endpoint resolved to its
+/// element label and terminal name.
+///
+/// Reads wiring_list_view out of the project database. --export-cables produces
+/// the same logical list from the document XML instead, and the two are meant
+/// to agree: running both and diffing them is a direct check that the database
+/// still describes the project, which is otherwise only observable through the
+/// GUI.
+int exportWiring(QETProject &project, const QString &output)
+{
+	// The project database is built lazily; force a (re)build before querying.
+	project.dataBase()->updateDB();
+
+	static const QStringList columns {
+		"wire_number", "from_element_label", "from_terminal",
+		"to_element_label", "to_terminal", "diagram_position", "conductor_uuid"
+	};
+
+	QSqlQuery query = project.dataBase()->newQuery(
+		"SELECT " % columns.join(", ") %
+		" FROM wiring_list_view ORDER BY diagram_position, wire_number");
+	if (!query.exec()) {
+		err << "Wiring list query failed: " << query.lastError().text() << "\n";
+		return 1;
+	}
+
+	QString csv = columns.join(";") % "\n";
+	int rows = 0;
+	while (query.next()) {
+		QStringList values;
+		for (int i = 0; i < columns.size(); ++i)
+			values << csvField(query.value(i).toString());
+		csv += values.join(";") % "\n";
+		++rows;
+	}
+
+	QFile file(output);
+	if (!file.open(QIODevice::WriteOnly | QIODevice::Text)) {
+		err << "Cannot open '" << output << "' for writing.\n";
+		return 1;
+	}
+	QTextStream fout(&file);
+	fout << csv;
+	file.close();
+	out << "Exported " << rows << " conductor(s) -> " << output << "\n";
+	return 0;
+}
+
 /// Electrical nets: groups of terminals joined into one potential.
 /// Walks QET's own potential graph, so each net is a connected component
 /// of terminals across all folios. The ground truth for connectivity.
@@ -823,6 +872,8 @@ int run(const QStringList &args)
 		return exportCsv(project, format, output);
 	if (format == "bom")
 		return exportBom(project, output);
+	if (format == "wiring")
+		return exportWiring(project, output);
 	if (format == "nets")
 		return exportNets(project, output);
 	if (format == "links")
--- a/sources/cli_export.h
+++ b/sources/cli_export.h
@@ -48,6 +48,7 @@ namespace CLIExport {
 		  qelectrotech --export-cables  <project.qet> <output.csv>
 		  qelectrotech --export-wires   <project.qet> <output.csv>
 		  qelectrotech --export-bom     <project.qet> <output.csv>
+		  qelectrotech --export-wiring  <project.qet> <output.csv>
 		  qelectrotech --export-nets    <project.qet> <output.json>
 		  qelectrotech --export-links   <project.qet> <output.csv>
 		  qelectrotech --info           <project.qet> [output.json]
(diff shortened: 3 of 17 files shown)


ispyisail:
Self-review, slice 5 of the #503 stack. Context in #625.

## The wire-count subquery has no index behind it

```sql
(SELECT COUNT(DISTINCT c.uuid) FROM conductor c
  WHERE c.terminal1_element_uuid = e.uuid
     OR c.terminal2_element_uuid = e.uuid) AS wire_count
```

The logic is right — `COUNT(DISTINCT c.uuid)` is the correct way to count a conductor with both ends on one element once, and the comment explaining that is accurate. The problem is the shape: a correlated scalar subquery evaluated once per element row, against a table with no index on either column it filters. There is no `CREATE INDEX` anywhere in this stack or on master, so both sides of that `OR` are full scans, and `EXPLAIN QUERY PLAN` confirms `CORRELATED SCALAR SUBQUERY`.

Synthetic measurement, 2000 elements × 5000 conductors:

```
no index   : 2134.5 ms
with index :   10.4 ms      (204x)
```

after adding

```sql
CREATE INDEX ... ON conductor(terminal1_element_uuid);
CREATE INDEX ... ON conductor(terminal2_element_uuid);
```

Numbers are from a standalone SQLite harness rather than from QET, so treat the absolute milliseconds as indicative — but the ratio is the engine's, not the harness's. This sits on `element_nomenclature_view`, which is rebuilt by `updateDB()`, so the cost lands on BOM export and on every nomenclature refresh, not just on opening one dialog. Two indexes seem cheap insurance for the first slice that puts a per-row subquery in that view.

Worth deciding whether the indexes belong here or in slice 2 with the table definitions — arguably slice 2, since that is where `conductor` is created and the columns already carry foreign keys that will attract lookups anyway.

ispyisail:

## `wire_count` is now non-zero on the projects that have wiring

As first posted, a conductor reached the `conductor` table only if both terminals carried a catalog uuid, so this column read **0 for every element** on `industrial.qet` and on 16 of the 20 example projects containing wiring. `Terminal::stableUuid()` in #628 fixes that at the root: all 20 measurable projects now contribute their full conductor count, so the column reports real numbers rather than zeros.

## The subquery has indexes behind it

`COUNT(DISTINCT c.uuid)` is the correct way to count — a conductor with both ends on one element is one wire touching it, not two — but the shape is a correlated scalar subquery evaluated once per element row, and there was no index on either conductor column it filters. `EXPLAIN QUERY PLAN` confirmed `CORRELATED SCALAR SUBQUERY` over a full scan.

Standalone SQLite harness, 2000 elements × 5000 conductors:

```
no index   : 2134.5 ms
with index :   10.4 ms      (204x)
```

The two indexes live in **#628** alongside the `conductor` table, since that is where the columns are declared and they already carry foreign keys that attract lookups. Only a comment lands here, tying the query to what makes it affordable — the two are in different files and nothing else connects them.

## End-to-end cost

The subquery sits on `element_nomenclature_view`, which `updateDB()` rebuilds, so the cost lands on BOM export and every nomenclature refresh rather than on one dialog. Measured against a real `master` build, best of three, load + `updateDB()` + BOM query:

| project | master | full stack | delta |
|---|---|---|---|
| `industrial.qet` (671 conductors, 50 folios) | 3.91 s | 4.11 s | **+0.20 s** |
| `m_000.qet` (457 conductors) | 3.01 s | 3.01 s | +0.00 |
| `weneedpolonez` (366 conductors) | 1.20 s | 1.21 s | +0.01 |

BOM contents themselves were checked against a pre-change baseline across the corpus by `qet-dbcheck.py`: **22 projects, 0 failures, 0 warnings**, with the nomenclature row count never exceeding the document's element count.

ispyisail:

PR #632, opened 2026-08-02
Fix Cyclique (modulo) parts defaulting to modulus 0 (follow-up to #593)

ispyisail:
Follow-up to #593, reported by @scorpio810 [here](https://github.com/qelectrotech/qelectrotech-source-mirror/pull/593#issuecomment-5152319870): a "Cyclique (modulo)" part counts upward forever instead of wrapping.

## What was wrong

Not the wrap/carry arithmetic — that part was fine. **The modulus was silently left at 0**, and 0 means "no cycle", so the part behaved exactly like a plain digit.

`setType()` defaulted the modulus to 8 inside the block that installs numeric behaviour. That block only runs when the *previous* type was non-numeric:

```cpp
if ( ((t==unit || … || t==wrap) && (type_==string || type_==folio || …)) || fnum )
{
    …
    if (t == wrap) ui->modulus_spinBox->setValue(8);   // ← skipped when type_ was numeric
}
```

A fresh part starts as `unit` ("Chiffre 1"). So the ordinary way to reach this feature — change the type of the part sitting in front of you — is exactly the path that skips the default. Going the long way round, via "Texte", sets it to 8 and works.

Driving the real widget through the same slot the combo box fires:

| path | modulus | result |
|---|---|---|
| `Chiffre 1` → `Cyclique (modulo)` | **0** | never wraps |
| `Chiffre 1` → `Texte` → `Cyclique (modulo)` | 8 | works |

That difference is the whole bug, and it explains why the feature tests fine whenever the context is built some other way.

## The fix

Move the default out of the numeric-behaviour block so it applies whatever the part was before, and fire it only when the current modulus is unusable so a deliberate value survives switching type away and back.

After, both paths give `modulus=8`.

## The arithmetic was already correct

Worth showing, since the report understandably read as "the carry logic is broken". Stepping the real `NumerotationContextCommands` with a carry target in front of the wrap part:

```
e. + unit(inc 0) + wrap(mod 7)
  e.00 e.01 e.02 e.03 e.04 e.05 e.06 e.10 e.11 … e.16 e.20 e.21 …

%IX + unit + . + wrap(mod 32)          ← the April 5000 case
  %IX0.0 %IX0.1 … %IX0.31 %IX1.0 %IX1.1 …
```

That is the requested pattern exactly: the wrap segment cycles and resets, the card digit ticks up.

One thing this makes explicit — **the wrap part needs a numeric part in front of it to carry into.** A wrap part on its own resets correctly but has nowhere to put the carry, so it just repeats `0…6, 0…6`. That is by design (`carry()` drops a carry with no target), but it is not discoverable from the UI, and it is a plausible second way to be disappointed by this feature. Happy to look at surfacing it if you think it is worth it.

## Compatibility

Saved configurations are untouched: a stored modulus loads and round-trips exactly as saved.

```
saved modulus 32 (PLC case)     -> spinbox 32, round-trips as 32   PRESERVED
saved modulus 7                 -> spinbox 7,  round-trips as 7    PRESERVED
saved modulus 0 (left by bug)   -> spinbox 0,  round-trips as 0    PRESERVED
```

**@scorpio810 — this is the part that affects you directly.** The last row is deliberate: I did not auto-rewrite stored 0s, because silently changing a saved numbering configuration seemed worse than leaving it alone. But it means **if you already saved a numbering config while this bug was present, that config still has modulus 0 and will still count upward after this fix.** Set the modulus once on that part and it will behave. Only new/edited parts get the default. Say the word if you would rather it repaired stored 0s on load and I will change it.

Built clean, no new warnings.

### DIFF
--- a/sources/autoNum/assignvariables.cpp
+++ b/sources/autoNum/assignvariables.cpp
@@ -39,6 +39,7 @@ namespace autonum
 	sequentialNumbers::sequentialNumbers(const sequentialNumbers &other)
 	{
 		unit          = other.unit;
+		wrap          = other.wrap;
 		unit_folio    = other.unit_folio;
 		ten           = other.ten;
 		ten_folio     = other.ten_folio;
@@ -57,6 +58,7 @@ namespace autonum
 			return (*this);
 
 		unit          = other.unit;
+		wrap          = other.wrap;
 		unit_folio    = other.unit_folio;
 		ten           = other.ten;
 		ten_folio     = other.ten_folio;
@@ -70,6 +72,7 @@ namespace autonum
 	bool sequentialNumbers::operator==(const sequentialNumbers &other) const
 	{
 		if (unit          == other.unit && \
+			wrap          == other.wrap && \
 			unit_folio    == other.unit_folio && \
 			ten           == other.ten && \
 			ten_folio     == other.ten_folio && \
@@ -107,6 +110,11 @@ namespace autonum
 						    document,
 						    "unit",
 						    unit.join(";")));
+		if (!wrap.isEmpty())
+			element.appendChild(QETXML::textToDomElement(
+						    document,
+						    "wrap",
+						    wrap.join(";")));
 		if (!unit_folio.isEmpty())
 			element.appendChild(QETXML::textToDomElement(
 						    document,
@@ -156,6 +164,11 @@ namespace autonum
 		from = element.firstChildElement("unit");
 		unit = from.text().split(";");
 
+			//Absent from files written before cyclic parts could be
+			//rendered; an empty list is the correct reading of that.
+		from = element.firstChildElement("wrap");
+		wrap = from.text().split(";");
+
 		from = element.firstChildElement("unitFolio");
 		unit_folio = from.text().split(";");
 
@@ -179,6 +192,7 @@ namespace autonum
 	void sequentialNumbers::clear()
 	{
 		unit.clear();
+		wrap.clear();
 		unit_folio.clear();
 		ten.clear();
 		ten_folio.clear();
@@ -434,13 +448,17 @@ namespace autonum
 						qMax(
 							qMax(m_seq_struct.hundred.size(),
 								 m_seq_struct.ten.size()),
-							m_seq_struct.alpha.size())
+							qMax(m_seq_struct.alpha.size(),
+								 m_seq_struct.wrap.size()))
 					);
 
 		for (int i=1; i<=max ; i++)
 		{
 			if (m_assigned_label.contains("%sequ_" + QString::number(i)) && m_seq_struct.unit.size() >= i) {
 				m_assigned_label.replace("%sequ_" + QString::number(i),m_seq_struct.unit.at(i-1));
+			}
+			if (m_assigned_label.contains("%seqw_" + QString::number(i)) && m_seq_struct.wrap.size() >= i) {
+				m_assigned_label.replace("%seqw_" + QString::number(i),m_seq_struct.wrap.at(i-1));
 			}
 			if (m_assigned_label.contains("%seqt_" + QString::number(i)) && m_seq_struct.ten.size() >= i) {
 				m_assigned_label.replace("%seqt_" + QString::number(i),m_seq_struct.ten.at(i-1));
@@ -553,6 +571,10 @@ namespace autonum
 			{
 				autonum::setSequentialToList(seqStruct.unit, context,"unit");
 			}
+			if (label.contains("%seqw_"))
+			{
+				autonum::setSequentialToList(seqStruct.wrap, context,"wrap");
+			}
 			if (label.contains("%sequf_"))
 			{
 				autonum::setSequentialToList(seqStruct.unit_folio, context,"unitfolio");
@@ -594,6 +616,7 @@ namespace autonum
 		QString value;
 		QString formula;
 		int count_unit = 0;
+		int count_wrap = 0;
 		int count_unitf = 0;
 		int count_ten = 0;
 		int count_tenf = 0;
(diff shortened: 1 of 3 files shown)


scorpio810:
Screenshot
These rules are always incorrect; one should count from 0 to 7 after 10 to 17, and so on, but this is not the case.

ispyisail:
@scorpio810 Thank you for the screenshot — that was the missing piece, and you are right that the rules are wrong. It is a bigger problem than the one this PR started out fixing, and I have pushed the rest of it here.

## What is actually wrong

**A Cyclique (modulo) part cannot be displayed at all.** The wrap-and-carry feature shipped without its rendering half: 68c2603 changed seven files, none of them `assignvariables.*`. So there is no `%seqw_` variable, no `wrap` list in `sequentialNumbers`, no branch in `setSequential()`, and none in `numerotationContextToFormula()`.

That last one is visible in your own screenshot: you have **two** parts configured, but the formula reads just `%sequ_1`. The Cyclique part contributed nothing to it, because there was no token for it to contribute.

**Invisible, but not inert.** The cyclic part still advances and still carries. So `Chiffre 1` in front of it gets +1 from the carry *on top of* its own increment, and the one digit that is displayed jumps by two. That is your missing 8. Reproduced exactly:

```
  n    unit (shown)   wrap (hidden)   label
  7    7              7               7
  8    9              0               9      <- 8 skipped
```

0 1 2 3 4 5 6 7 9 10 11 12 13 14 15 — the same sequence as your diagram, and the `16` left in your value field is where the counter lands after those 15 elements.

So the modulus-default bug this PR opened with was real, but secondary: even with the modulus set correctly, the wrapping digit could never appear.

## What I have added

`%seqw_N`, and the plumbing behind it: a `wrap` list on `sequentialNumbers` (copy/assign/compare/clear/toXml/fromXml), collection in `setSequential()`, substitution in `assignSequence()` including the loop bound so a context whose only counter is cyclic still resolves, and emission from `numerotationContextToFormula()` so adding a Cyclique part now puts a token in the formula.

Measured on your exact configuration, `Chiffre 1` + `Cyclique mod 8`:

```
formula generated now:  %sequ_1%seqw_1        (was: %sequ_1)

carry digit increment 1   00 11 22 33 44 55 66 77 90 101 112 ...
carry digit increment 0   00 01 02 03 04 05 06 07 10 11 ... 17 20 21 22 23
```

**The second line is the 0-7 / 10-17 / 20-27 you asked for.**

## One thing I would like your call on

The difference between those two lines is the **increment of the digit in front of the cyclic one**. For positional counting it has to be **0** — that digit should move only when the cyclic digit carries into it. Yours is 1 (the editor's default), which is why it also advances on its own and you get `00 11 22 33`.

I did not want to decide this for you. Options as I see them:

1. Leave it: the carry digit's increment is the user's to set, and it is now at least explicable.
2. Default the increment to 0 when a part sits directly in front of a Cyclique part.
3. Ignore a part's own increment when it is the carry target, so it only ever moves on carry.

(2) is the friendliest and least surprising to me, and (3) changes documented behaviour for anyone who deliberately wanted both. Tell me which you prefer and I will implement it.

## Compatibility

Old projects are unaffected: `<wrap>` is simply absent from files written before this, which `fromXml()` reads as an empty list, and those files have no cyclic parts to reference anyway. An older QElectroTech opening a newer file ignores the unknown child.

scorpio810:
Modulo is for I/O PLC autonum rules see attached projets: 
Screenshot
Screenshot-1

scorpio810:
I'd rather not lock this down yet.

My real cases span two different rack sizes:

- April 5000/2000 racks, 32-input/output cards: %IX0.0 .. %IX0.31, then %IX1.0. so the counter needs to wrap at 32 before the card digit advances.

- Schneider M340 (BMXP342020 + DDI6402K/DDO6402K): 64-point I/O cards, addressed in 4 blocks of 16 (e.g. I1.00-15, I1.16-31, I1.32-47, I1.48-63). so the counter needs to wrap at 64 before the card digit advances.


But different PLC projects draw their card layouts differently, and I don't want to guess a one-size-fits-all rule from just my own setups. I'd lean toward option 1 or 2 (leave it as user choice, or default to 0 but keep it adjustable) rather than option 3 (forcing it), so nothing gets locked out for a layout I haven't seen. Let's see what comes back from other users and real field usage before deciding between 1 and 2.

Thanks for digging into the missing rendering half too, that screenshot bug was exactly what was biting me.

Screenshot-2

ispyisail:
@scorpio810 Following up on your second layout — the M340 with 64-point cards addressed `I1.00-15 / I1.16-31 / I1.32-47 / I1.48-63`.

That one would **not** have worked even after the rendering fix in this PR. A cyclic part could only be drawn at its natural width, so you would have got `I1.0` where your drawing needs `I1.00`. And it could not be worked around: `Chiffre 01` pads but does not wrap, `Cyclique` wraps but does not pad.

The two layouts you gave genuinely conflict — April wants `%IX0.0` unpadded, M340 wants `I1.00` padded — so the width can't be inferred from the modulus or the type. It has to be yours to set.

So that is now a user-settable **format**, in **#634** (stacked on this PR): a run of zeros, the same convention a spreadsheet uses, where the length is the minimum number of digits.

```
Type                Valeur  Incr.  modulus  Format
Cyclique (modulo)     0       1      32     (vide)  ->  %IX0.0  %IX0.31  %IX1.0
Cyclique (modulo)     0       1      64       00    ->  I1.00   I1.63    I2.00
```

Measured on both of your layouts, through the rollover:

```
April, no mask    %IX0.29 %IX0.30 %IX0.31 %IX1.0 %IX1.1 %IX1.2
M340,  mask "00"  I1.00 I1.01 ... I1.62 I1.63 I2.00 I2.01
```

It applies to every numeric part type, not just cyclic ones, so `Chiffre 01` can be widened past two digits without needing yet another type in the dropdown.

**Nothing changes for existing projects**: an empty mask means the type's own natural width, so `Chiffre 1` is still `7`, `Chiffre 01` still `07`, `Chiffre 001` still `007`. Absent is the default, and absent changes nothing. It is stored the same way `modulus` was — a sixth field, written to XML only when set — so old files parse unchanged and an older QElectroTech ignores the attribute.

And to be clear on the other question: I have **not** touched the carry-digit increment. You asked to wait for field feedback before choosing between leaving it to the user and defaulting it to 0, and I agree that is the right order — this is only about how a part is displayed, not about when it advances.

PR #633, opened 2026-08-02
Never leave a collection folder without a name (replaces #622)

ispyisail:
**Correction (2026-08-02):** the "Windows path-encoding problem with accented characters" cited below as a live cause was already fixed, in #524 (merged 2026-06-29, before this branch existed) — @plc-user caught this on the PR thread. That PR switched `load_file()` to the wide-char pugixml overload, which is why the snippet below already shows `toStdWString()`. What follows is corrected to reflect that: this PR is about the remaining structural gap (any *other* reason `load_file()` can fail — genuinely missing or malformed files), not that specific trigger, which is no longer live. See the discussion on the PR for the full correction.

---

Replaces #622, which GitHub will not let me reopen — I force-pushed the rebase while it was closed, and that permanently blocks reopening. Same branch, same purpose, with the corrections described below. Discussion in #622 is worth reading; @plc-user's suggestion there is implemented here.

Fixes https://qelectrotech.org/bugtracker/view.php?id=332 — which is **still open** (`Status: new`, `Resolution: open`); #622 was closed on the understanding it had been resolved.

## The bug

`FileElementCollectionItem::localName()` sets a non-root folder's label only inside the success path of loading its `qet_directory`:

```cpp
if (docu.load_file(str.toStdWString().c_str()))
    if (QString(docu.document_element().name()) == "qet-directory")
        { … setText(nl.name()); }
// nothing set at all if either test fails
```

A fresh item's `text()` is null, so any failure — file missing or malformed — leaves the folder with a completely blank label. That is the reported symptom. (The third cause originally listed here, a Windows path-encoding problem with accented characters that @plc-user diagnosed on the tracker, was already fixed in #524 before this branch existed — see the correction above.)

## The fix

Resolve the name into a local and always fall back to the folder's own directory name, so the label is never empty whatever went wrong.

**The fallback is applied after `NamesList::name()`, not passed into it.** This matters and is the part I got wrong in #622. `name()` returns a caller-supplied fallback at `nameslist.cpp:274`, *before* it reaches its "first available translation" step at line 275:

```cpp
if (!map_names["en"].isEmpty()) return (map_names["en"]);
if (!fallback_name.isEmpty())   return (fallback_name);   // <- short-circuits
if (map_names.count())          return (map_names.begin().value());
```

So passing `m_path` in would replace a perfectly good name in another language with the raw directory name. A folder named only in French, viewed under an English locale, must keep showing `Accentué` — and with my original patch it stopped doing so.

## Telling the user the file is broken (@plc-user's suggestion)

Falling back silently cures the symptom but hides the cause: the user sees a plausible name and never learns there is a file to repair. A folder whose `qet_directory` could not be read now says so in its tooltip, naming the file, above the collection path that tooltip already carried.

Only a **file-level** failure is flagged. A readable `qet-directory` with no entry for the current language is not an error — `NamesList::name()` resolves that itself — so a folder named only in French gets its name and no warning. That distinction is what keeps the indicator meaningful rather than noise.

The flag is recorded in `localName()` and consumed in `setUpData()`, because `setUpData()` ends with an unconditional `setToolTip(collectionPath())` that runs *after* `localName()`. Setting the tooltip in `localName()` — which is what I tried first — is silently overwritten and does nothing at all.

## Measured

A fixture collection of four folders, driven through the real `FileElementCollectionItem`:

| folder | master today | this PR |
|---|---|---|
| named only in French | `Accentué` | `Accentué` — no warning |
| malformed `qet_directory` | **blank** | `malformed` + warning |
| no `qet_directory` at all | **blank** | `no_file` + warning |
| valid `qet_directory` | `Valid Folder` | `Valid Folder` — no warning |

Rebased onto current master: #622's branch predated #620 and its diff would have reverted the elements-panel dark-theme fix. Two files, +45/-3, builds clean.

## What this does not do

This does not fix the Windows accented-path trigger from bugtracker #332 — that was already fixed in #524, independently of this PR (see the correction at the top). What this does is close the remaining structural gap: any other reason `load_file()` can fail — a genuinely missing or malformed `qet_directory` — still leaves a folder blank on current master, and after this PR it doesn't. If the original #332 reporter can confirm their folders now show names, that closes the loop on #524, not on this PR specifically.

### DIFF
--- a/sources/ElementsCollection/fileelementcollectionitem.cpp
+++ b/sources/ElementsCollection/fileelementcollectionitem.cpp
@@ -430,13 +430,20 @@ void FileElementCollectionItem::setUpData()
  */
 void FileElementCollectionItem::setUpIcon()
 {
-		// Directories are cheap to (re-)decide: no early return for them.
-		// setUpData() -- which resolves m_qet_directory_unreadable via
-		// localName() -- runs asynchronously (QtConcurrent::map), so this can
-		// be called for a directory before that result is known; without
-		// this, the plain folder icon would get cached by the guard below
-		// and the warning badge would never appear for that item.
-	if (!isDir() && !icon().isNull())
+		// Must return unconditionally once an icon is set: setIcon() calls
+		// setData(), which emits dataChanged() regardless of whether the new
+		// icon differs from the old one (QIcon has no meaningful equality).
+		// QTreeView responds to dataChanged() by recomputing the row's size
+		// hint, which re-enters data() for this same index -- so without this
+		// guard, any repeated setIcon() here recurses until the stack
+		// overflows. Confirmed by crash report on PR #633.
+		//
+		// This item's m_qet_directory_unreadable is already final by the time
+		// this can run at all: ElementsCollectionModel only attaches itself
+		// to the tree view (making data() reachable) from loadingFinished(),
+		// which fires after the QtConcurrent::map over every item -- this one
+		// included -- has completed. So there is no race to work around here.
+	if (!icon().isNull())
 		return;
 
 	if (isCollectionRoot()) {
--- a/sources/ElementsCollection/fileelementcollectionitem.cpp
+++ b/sources/ElementsCollection/fileelementcollectionitem.cpp
@@ -22,7 +22,39 @@
 #include "../qeticons.h"
 #include "elementslocation.h"
 
+#include <QApplication>
 #include <QDir>
+#include <QPainter>
+#include <QPixmap>
+#include <QStyle>
+
+namespace {
+	/**
+		@return the folder icon overlaid with a small warning badge in the
+		bottom-right corner. Used for a directory whose qet_directory could
+		not be read (@see FileElementCollectionItem::m_qet_directory_unreadable),
+		so the problem is visible in the tree itself and not only on hover
+		via the tooltip. Built once: same folder icon, same badge, every time.
+	*/
+	const QIcon &unreadableFolderIcon()
+	{
+		static const QIcon icon = []() {
+			QPixmap pixmap = QET::Icons::Folder.pixmap(16, 16);
+			const QPixmap badge = QApplication::style()
+				->standardIcon(QStyle::SP_MessageBoxWarning)
+				.pixmap(9, 9);
+
+			QPainter painter(&pixmap);
+			painter.drawPixmap(pixmap.width() - badge.width(),
+					    pixmap.height() - badge.height(),
+					    badge);
+			painter.end();
+
+			return QIcon(pixmap);
+		}();
+		return icon;
+	}
+}
 
 /**
 	@brief FileElementCollectionItem::FileElementCollectionItem
@@ -398,7 +430,13 @@ void FileElementCollectionItem::setUpData()
  */
 void FileElementCollectionItem::setUpIcon()
 {
-	if (!icon().isNull())
+		// Directories are cheap to (re-)decide: no early return for them.
+		// setUpData() -- which resolves m_qet_directory_unreadable via
+		// localName() -- runs asynchronously (QtConcurrent::map), so this can
+		// be called for a directory before that result is known; without
+		// this, the plain folder icon would get cached by the guard below
+		// and the warning badge would never appear for that item.
+	if (!isDir() && !icon().isNull())
 		return;
 
 	if (isCollectionRoot()) {
@@ -417,7 +455,8 @@ void FileElementCollectionItem::setUpIcon()
 	else
 	{
 		if (isDir()) {
-			setIcon(QET::Icons::Folder);
+			setIcon(m_qet_directory_unreadable ? unreadableFolderIcon()
+							    : QET::Icons::Folder);
 		} else {
 			if (m_path.endsWith(".qetmak")) {
 				setIcon(QIcon());
--- a/sources/ElementsCollection/fileelementcollectionitem.cpp
+++ b/sources/ElementsCollection/fileelementcollectionitem.cpp
@@ -136,18 +136,41 @@ QString FileElementCollectionItem::localName()
 		}
 		else
 		{
+			// Fall back to the raw directory name (m_path) whenever the
+			// translated name can't be obtained -- qet_directory missing,
+			// unreadable (e.g. a Windows path-encoding issue with special
+			// characters, see bugtracker #332), malformed, or present but
+			// without a usable name entry -- rather than leaving the item
+			// blank.
+			QString display_name;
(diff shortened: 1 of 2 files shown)


sources/ElementsCollection/fileelementcollectionitem.cpp
```diff
@@ -136,18 +136,41 @@ QString FileElementCollectionItem::localName()
...
+						// Deliberately no fallback argument: a non-empty one
+						// is returned *before* NamesList::name() reaches its
+						// "first available translation" step, so passing
+						// m_path here would replace a perfectly good name in
+						// some other language with the raw directory name.
+						// The fallback belongs after the chain, not inside it.
+					display_name = nl.name();
 				}
 			}
+			setText(display_name.isEmpty() ? m_path : display_name);
```
plc-user, line 198: The additional hint-text is good to explain, what to do here, but my thought was to show the user direct here that something's wrong:
`setText(display_name.isEmpty() ? (QObject::tr("FixMe: ") % m_path) : display_name);`

plc-user:
The issue with missing directory names in win when special characters are used in the name seems to have already been fixed by a previous PR. Still, I think this PR is a good idea!

By the way:
You guys are setting a pace with these many pull requests that, at least for me, is hard to keep up with right now.

ispyisail:
@plc-user You're right, and thank you for catching it — I should have checked this before repeating the "Windows path-encoding problem with accented characters" framing in this PR's description.

**The specific trigger you diagnosed on bugtracker #332 was already fixed, in #524** (merged 2026-06-29, `sources/ElementsCollection/fileelementcollectionitem.cpp`): switching `pugi::xml_document::load_file()` from the narrow `toStdString()` overload to `toStdWString()`, so it calls `_wfopen` on Windows instead of `fopen_s` under the ANSI codepage. That's exactly the accented-path failure mode you were asking the reporter about on 2025-10-30. The tracker itself hasn't been updated to reflect that, which is presumably why the connection wasn't obvious from there.

I didn't catch this because #524 landed before I started working on #622/#633, so the code I was reading and quoting (`docu.load_file(str.toStdWString().c_str())`) already had the fix — I described *why* that call could still fail without checking whether the specific failure mode I was citing was still reachable. It wasn't. Sloppy attribution on my part; I've corrected the PR description accordingly.

**What #633 is actually for, then, is narrower than I originally framed it:** `localName()` still only calls `setText()` inside the success path of `load_file()`, so *any* remaining failure — a genuinely missing `qet_directory`, a malformed one, or any other reason `_wfopen`/`fopen` could fail in the future — still leaves the folder blank rather than falling back. #524 closed the one specific cause you diagnosed; #633 closes the structural gap underneath it, plus adds the tooltip you originally suggested so a real failure (whatever causes it) isn't silently hidden by the fallback. Complementary, not overlapping with #524, but no longer "the" fix for the accented-path case specifically since that's already resolved. Verified again just now: on current master (which already includes #524), a malformed or missing `qet_directory` still produces a blank label — that part of the bug is real and current.

Thanks for staying with this — you've now caught two real things on two different PRs today.

ispyisail:
> The issue with missing directory names in win when special characters are used in the name seems to have already been fixed by a previous PR. Still, I think this PR is a good idea!
> 
> By the way:
> You guys are setting a pace with these many pull requests that, at least for me, is hard to keep up with right now.

Real me, sorry

I just read the above text and its getting out of control. How do big projects handle this problem?

ispyisail:
> By the way: You guys are setting a pace with these many pull requests that, at least for me, is hard to keep up with right now.

On the plus side you guys have some cool features on the way.

plc-user:
Hi @ispyisail,
you didn't say anything about my wish to add a "FixMe: " to the display_name (see code-review above).
Do you add that fragment to your commit or should I add another commit after releasing this one?
I don’t mind if this appears as a commit under your name...!

ispyisail:
@plc-user Pushed as a follow-up commit: f6b448d

I didn't take the "FixMe: " prefix as-is because that string is `localName()`/`display_name` — the same value that shows up verbatim in the folder's dialog title (`tr("Propriété du dossier %1")`) and status-bar messages elsewhere in `elementscollectionwidget.cpp`. Baking a prefix into it there would leak into those too ("Propriété du dossier FixMe: Accentué").

What I pushed instead gets the same visibility without touching that value: `setUpIcon()` now overlays a small warning badge on the folder icon when `m_qet_directory_unreadable` is set, so it's visible directly in the tree, not just on tooltip hover. Same signal you were after, just via the icon rather than the name.

While in there I also fixed a latent race in `setUpIcon()`'s "already has an icon, skip" guard: `m_qet_directory_unreadable` is only known once the async `setUpData()`/`localName()` job finishes, so a directory painted before that completed would have had its plain folder icon cached forever. Fixed by not applying that guard to directories (cheap to redecide, unlike the SVG element icons the guard is actually protecting).

Let me know what you think of the icon approach.

plc-user:
With latest changes QET crashes with a segfault, when loading of elements-collection finishes and drawing of the elements-tree begins.
Am I possibly missing the icons?

ispyisail:
Thank you for the report — found it, and it was a real one.

The badge commit narrowed `FileElementCollectionItem::setUpIcon()`'s early return to skip directories, so the badge could pick up `m_qet_directory_unreadable` once it was resolved. That meant every directory now called `setIcon()` on *every* `data(Qt::DecorationRole)` query instead of just once. `QStandardItem::setIcon()` → `setData()` emits `dataChanged()` unconditionally (`QIcon` has no equality check to suppress it), `QTreeView` responds to `dataChanged()` by recomputing the row's size hint, and that re-enters `data()` for the same index — which calls `setIcon()` again. Unbounded mutual recursion, stack overflow, segfault, right as the tree starts drawing — matches what you saw exactly.

I isolated the pattern in a ~40-line standalone Qt program outside the QET tree to confirm the mechanism directly: it segfaults before the first paint even completes, with 100,000+ recursive frames on the stack.

Fixed in f83aa3f1b: restored the plain, unconditional "return once an icon is set" guard. The race it was trying to route around doesn't actually happen — `ElementsCollectionModel` only attaches itself to the tree view (the only way `data()` becomes reachable at all) from `loadingFinished()`, which fires after the background scan of *every* item, this one included, has already finished. So `m_qet_directory_unreadable` is always settled before `setUpIcon()` can run for the first time, and the badge still works correctly with the simple guard.

Pushed directly to this branch. Sorry for the regression, and thanks for catching it quickly.

plc-user:
Ah ... that's even better than a text that needs to be translated!
grafik
User can directly see that there is something "wrong" with the respective directory.

PR #637, opened 2026-08-02
Add "Export to SVG" to the element editor (discussion #605)

ispyisail:
Implements discussion #605: exporting a single element (`.elmt`) to a standalone SVG file, directly from the element editor.

## Problem

The diagram editor can already export a folio to SVG. The element editor — where a single symbol is drawn — has no export capability at all: no menu action, no dialog, nothing. Confirmed before starting by grepping its header for "export": nothing there.

## What it adds

`Fichier > Exporter en SVG` in `QETElementEditor`, next to the existing `Enregistrer dans un fichier` action.

## Deliberately differs from the discussion's proposed implementation

The discussion suggested sourcing this from `ElementPictureFactory`'s cached per-element `QPicture` (`m_pictures_H`) — the one that backs the elements-panel preview icons. Before building on it, I checked how that cache actually gets invalidated: **nothing does.** It's keyed by the element's on-disk uuid and nothing in the editor ever tells it to drop an entry on edit. So exporting from it would:

- silently produce **stale content** for any element already previewed once in the panel, after any edit, and
- produce **nothing at all** for a brand-new, never-saved element.

Neither is acceptable for a `File > Export` action a user expects to reflect what's currently on screen.

Instead this renders `ElementScene` directly — the same pattern `ExportDialog::generateSvg()` already uses for the diagram editor's own SVG export (`Diagram` → `QPicture` → `QSvgGenerator`-backed `QPainter`). No new drawing logic anywhere; only a new playback target, sized to the element's own content bounds via the existing `elementSceneGeometricRect()` helper.

## Hotspot cross excluded from the export

`ElementScene::drawForeground()` draws the red origin/hotspot cross on every `render()` call, unconditionally — an editing aid, not part of the element. Added a `hotspotVisible` flag on `ElementScene`, defaulting to `true` (the existing editing view is completely unaffected), turned off only for the duration of the export's `render()` call.

## Verified end-to-end via a real Xvfb session

Not just a build check:

1. Opened a real shipped element (`en_60617_05_06_04.elmt`, "Phototransistor"), exported it via the new menu action, and rendered the resulting SVG back to a bitmap with a small `QSvgRenderer`-based harness. Pixel-identical in shape to the element as shown in the editor. Confirmed the file is valid XML and contains no red/`#ff0000` stroke — the hotspot cross did not leak in.

2. **The case the whole "render live, not cached" decision was actually about:** opened the same element, drew a new line with the line tool, and exported again *without saving*. The new line is present in the exported SVG. `git status` on the source `.elmt` file after both exports shows it completely untouched — the export is read-only and reflects live, unsaved editor state, exactly the property a cache-based implementation would have gotten wrong.

Built clean, no new warnings. Two files, +105/-0.

### DIFF
--- a/sources/editor/elementscene.cpp
+++ b/sources/editor/elementscene.cpp
@@ -317,6 +317,10 @@ void ElementScene::contextMenuEvent(QGraphicsSceneContextMenuEvent *event)
 */
 void ElementScene::drawForeground(QPainter *p, const QRectF &)
 {
+	if (!m_hotspot_visible) {
+		return;
+	}
+
 	p -> save();
 
 	// desactive tout antialiasing, sauf pour le texte
--- a/sources/editor/elementscene.h
+++ b/sources/editor/elementscene.h
@@ -90,6 +90,8 @@ class ElementScene : public QGraphicsScene
 		    m_y_grid;
 
 		QPointer<CustomElementGraphicPart> m_single_selected_item;
+
+		bool m_hotspot_visible = true;
 	
 		// methods
 	public:
@@ -132,6 +134,14 @@ class ElementScene : public QGraphicsScene
 		QETElementEditor* editor() const;
 		void addItems(QVector<QGraphicsItem *> items);
 		void removeItems(QVector<QGraphicsItem *> items);
+
+			/// Whether drawForeground() draws the red hotspot cross. On by
+			/// default (the normal editing view); turned off around a
+			/// render() call meant to capture the element's actual content
+			/// only, e.g. exporting to SVG -- the cross is an editing aid,
+			/// not part of the drawn symbol.
+		void setHotspotVisible(bool visible) {m_hotspot_visible = visible;}
+		bool hotspotVisible() const {return m_hotspot_visible;}
 	
 	protected:
 		void mouseMoveEvent         (QGraphicsSceneMouseEvent *) override;
--- a/sources/editor/ui/qetelementeditor.cpp
+++ b/sources/editor/ui/qetelementeditor.cpp
@@ -1397,10 +1397,20 @@ bool QETElementEditor::on_m_save_as_file_action_triggered()
 */
 bool QETElementEditor::on_m_export_svg_action_triggered()
 {
+		//Suggest the element's own filename (without its .elmt extension) as
+		//the default export name, per plc-user's review on #637 -- the
+		//directory-only default below is unchanged for an element that has
+		//never been saved, since there is no filename to derive one from.
+	QString suggested_path = QETApp::customElementsDir();
+	if (!m_file_name.isEmpty()) {
+		QFileInfo file_info(m_file_name);
+		suggested_path = QDir(file_info.absolutePath()).filePath(file_info.completeBaseName());
+	}
+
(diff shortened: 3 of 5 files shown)


scorpio810:
https://github.com/plc-user/QET_ElementScaler/blob/master/README.md

https://qelectrotech.org/forum/viewtopic.php?id=2637
https://qelectrotech.org/forum/viewtopic.php?pid=19407#p19407
https://qelectrotech.org/forum/viewtopic.php?pid=19415#p19415


Online official/contrib collection viewer in HTML with element *elmt converted to *SVG by plc-user 's QET_ElementScaler https://github.com/plc-user/QET_ElementScaler

https://download.qelectrotech.org/qet/elements/index.html
https://download.qelectrotech.org/qet/elements_contrib/index.html

plc-user:
That's funny: You want me to evaluate a competitor to my own software, Laurent?  :hushed:

That shouldn't be a problem for me!   :grinning:

The graphic primitives are converted differently than QET_ElementScaler does. Accordingly, the file size is significantly different: In the attached example, it's a factor of 10 for the file created by QSvgGenerator.

The graphic primitives are rendered well, though the font sizes appear smaller in the QSvgGenerator version (see screenshots).

The absolute sizes of the respective primitives are about 33% larger than with QET_ElementScaler, while the fonts are the same size. This accounts for the visual size differences in the SVGs!

I can’t say whether there are any settings in QSvgGenerator for the graphic primitives or font sizes that would allow for adjustments, but the visual result isn’t so bad that we shouldn’t integrate this feature into QET.

The source code looks clean and well organized and does not produce any additional errors or warnings.
I can certainly imagine one improvement, though: The element filename could be included by default (without the extension) in the file dialog.

left: SVG from QET_ElementScaler
right: QET Element-Editor 
Bildschirmfoto__QET_ElementScaler

left: SVG from QSvgGenerator
right: QET Element-Editor 
Bildschirmfoto__QSvgGenerator

scorpio810:
@plc-user : Don't worry, QET_ElementScaler still has its place — we're just going for the "deluxe edition" 😉 Thanks for taking the time to test it and for the screenshots, really appreciate the thorough eval!


Thanks for the thorough writeup and the Xvfb verification, @ispyisail — really appreciate the rigor here, especially the "unsaved edits" test case.

Before merging: would you mind adding the small improvement plc-user suggested — pre-filling the file dialog with the element's filename (without the extension) as the default export name? Once that's in, this looks good to merge from my side.

ispyisail:
> we're just going for the "deluxe edition"

Dream big

ispyisail:
Just pushed the pre-fill improvement plc-user suggested — exporting an element that has already been saved now suggests its own filename (without the `.elmt` extension) as the default SVG name, in the element's own directory, instead of only defaulting to the custom elements directory with an empty name.

For an element that has never been saved (no filename to derive one from), the dialog still falls back to the custom elements directory as before.

Verified in a real GUI session (Xvfb): opening `filtre.elmt` and triggering "Exporter en SVG" now pre-fills the save dialog with "Look in: <element's directory>" and "File name: filtre".

@scorpio810 @plc-user let me know if this looks good to merge.

PR #638, opened 2026-08-02
Color management for conductor auto-numbering (discussion #606)

ispyisail:
Implements discussion #606: when conductor auto-numbering resolves a label text that matches a configured rule (e.g. "L1", "PE"), the conductor's color is set automatically — the same way its text already is — instead of every conductor needing recoloring by hand afterward.

## The single integration point

`ConductorAutoNumerotation::applyText(const QString& t)` is where every auto-numbered conductor's resolved label actually gets set. Checked before touching anything: it's called exactly once in the whole codebase, from `numerateNewConductor()`. So a lookup placed right next to the existing `cp.text = t` line reaches every auto-numbered conductor, applied identically to `m_conductor` and to every conductor in `conductor_list` — the same same-potential group whose text the existing code already keeps in sync.

`cp.color`, `cp.m_color_2` and `cp.m_bicolor` are exactly the three `ConductorProperties` fields `conductor.cpp:545` already reads to draw a conductor's line and its second stripe. Nothing new for the drawing code to learn.

## `ConductorColorRule` + `QETProject::conductorColorRules()`

A small class (`color`, `color_2`, `bicolor`) plus a project-level `QHash<QString, ConductorColorRule>`, persisted the same way `m_default_xref_properties` already is: an in-class default (`ConductorColorRule::defaultRules()` — one common phase/neutral/earth convention, not a claim of regulatory correctness for any specific country), read from a `<conductor_autonum_colors>` element alongside the existing `<conductors_autonums>`, and only replaced when that element is actually present — so a project saved before this feature keeps sensible built-in defaults rather than nothing.

## Configuration UI

A new **"Couleurs des conducteurs"** tab in `ProjectAutoNumConfigPage`, next to the existing "Conducteurs" tab. One row per rule: label text, a `KColorButton` for its color (the same picker class `conductorpropertieswidget.ui` already uses for conductor color), a bicolor checkbox, and a second `KColorButton` enabled only when bicolor is checked. Modelled on `SelectAutonumW`: applies via its own "Enregistrer" button, not gated by the surrounding dialog's OK — matching every sibling tab in that page (`applyProjectConf()` is deliberately empty; each tab applies itself).

## Verified

Built clean from a fresh configure, zero new warnings across all seven new/changed files.

Via a real Xvfb session:
- Config page renders with all five default rules in their real colors (L1 brown, L2 black, L3 grey, N blue, PE green + gold bicolor stripe).
- **Full persistence round trip through an actual project file**, not just the write path: saved a project, inspected the written XML directly (`<conductor_autonum_colors>` with five `<color_rule>` entries in the expected shape), hand-edited L1's color *in the file* to confirm the read path genuinely parses rather than reusing in-memory defaults, reopened the project, and confirmed the config page showed the hand-edited color — not the default — with everything else unchanged.
- Set up a numbering context resolving to the literal text "L1" and selected it as the diagram's active conductor auto-numbering scheme.

## What this could not verify

I could not get a live, on-canvas conductor auto-numbered end-to-end in this sandbox. Placing two connected elements requires drag-and-drop from the collection panel into the `QGraphicsView`, and this environment's synthetic X11 automation does not reliably trigger Qt's `QDrag`/XDND handshake — tried both a single-motion drag and a slow multi-step drag; no drag cursor or drop ever registered. That's a tooling limitation, not a code question: the one choke-point function this relies on was read in full and the change mirrors the existing, already-working text-assignment line for line, and every other part — the rule lookup's data source, the exact fields it writes, and their persistence — was independently verified above. Flagging the gap plainly rather than presenting a screenshot I don't have.

If anyone can try it against a real drag-and-drop session (or has a project file with two elements already wired), confirming the color actually paints on a live redraw would close that last gap.

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -155,6 +155,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/conductornumexport.cpp
   ${QET_DIR}/sources/wiringlistexport.h
   ${QET_DIR}/sources/wiringlistexport.cpp
+  ${QET_DIR}/sources/conductorcolorrule.cpp
+  ${QET_DIR}/sources/conductorcolorrule.h
   ${QET_DIR}/sources/conductornumexport.h
   ${QET_DIR}/sources/conductorprofile.cpp
   ${QET_DIR}/sources/conductorprofile.h
@@ -269,6 +271,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/autoNum/ui/autonumberingdockwidget.h
   ${QET_DIR}/sources/autoNum/ui/autonumberingmanagementw.cpp
   ${QET_DIR}/sources/autoNum/ui/autonumberingmanagementw.h
+  ${QET_DIR}/sources/autoNum/ui/conductorcolorrulesw.cpp
+  ${QET_DIR}/sources/autoNum/ui/conductorcolorrulesw.h
   ${QET_DIR}/sources/autoNum/ui/folioautonumbering.cpp
   ${QET_DIR}/sources/autoNum/ui/folioautonumbering.h
   ${QET_DIR}/sources/autoNum/ui/formulaautonumberingw.cpp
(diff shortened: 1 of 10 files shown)


ispyisail:

PR #639, opened 2026-08-02
Auto-break conductor

Kellermorph:
Adds an "Auto-break conductor" feature to QElectroTech. When enabled, placing a component on an existing conductor automatically breaks the conductor at the component's dock point and reconnects the broken ends to the new element's terminals — preserving bent conductor segments (L-shaped paths).

1. Toolbar button (Coupure automatique de conducteur(s)) next to the existing auto-conductor button, using the Conductor icon.
2. Algorithm in DiagramEventAddElement::addElement():
For each terminal of the new element, find the closest conductor within 5px whose path segment orientation matches the terminal orientation.
Split the conductor at the dock point (using Qet::orthogonalProjection).
Create a new conductor from the connect_to endpoint to the matched terminal.
Detect bent conductors (horizontal+vertical) and create a second conductor from the other endpoint to the nearest free terminal, preserving the original shape.
Track used terminals (used_terminals set) to distribute connections across all terminals of multi-pole elements.
3. Persistence:
Saved per-project in XML (auto_break_conductors attribute in <conductors_autonums>).
Also saved globally in QSettings so new projects inherit the last-used state.


sources/diagramevent/diagrameventaddelement.cpp
```diff
@@ -248,16 +253,171 @@ void DiagramEventAddElement::addElement()
...
+				Conductor *new_c2 = new Conductor(other, other_terminal);
+				new AddGraphicsObjectCommand(new_c2, m_diagram, QPointF(), undo_object);
+				ConductorAutoNumerotation can2(new_c2, m_diagram, undo_object);
+				can2.numerate();
+				if (m_diagram->freezeNewConductors() || m_diagram->project()->isFreezeNewConductors())
+					new_c2->setFreezeLabel(true);
+
+				broken_endpoints.insert(other);
+				used_terminals.insert(other_terminal);
+			}
```
ispyisail, line 477: Missing `break;` here (or another early exit from the enclosing `foreach (Conductor *c, m_diagram->conductors())` loop) once this match has been fully handled for terminal `t`.

Without it, if a second existing conductor's path also lies within 5px of the same dock point (e.g. the new component is placed at a crossing of two independent, unconnected wires — an ordinary layout), the loop keeps scanning and breaks/reconnects that conductor too, wiring it to the same terminal `t`. `Terminal::addConductor()` doesn't prevent this: it only rejects a duplicate link to the *same* other terminal, and explicitly allows fan-out to different terminals (for legitimate bus/junction points), so nothing downstream catches it. The result is two previously-separate nets silently bridged together through the new component's single terminal.

ispyisail (commented):
Reviewed by pulling the branch and building it against current master (clean, zero new warnings) rather than just reading the diff, since the concern below is a logic question, not something the compiler would catch.

## The core issue: nothing stops the same new terminal from being wired to two different original conductors

`addElement()` loops over the new element's terminals, and for each one scans every existing conductor looking for one whose path passes within 5px of that terminal's dock point. When it finds a match, it breaks that conductor and reconnects it to the new terminal — but there's no `break` (or any other early exit) out of the `foreach (Conductor *c, m_diagram->conductors())` loop afterward. If a *second* existing conductor's path also happens to lie within 5px of the same dock point, the loop keeps going and processes that one too, breaking it and wiring it to the same terminal a second time.

I checked whether the framework quietly prevents that. It doesn't: `Terminal::addConductor()` (`terminal.cpp:129`) only rejects a new conductor when the terminal is already linked to that *specific* other terminal — it explicitly allows fan-out to different terminals (needed for legitimate bus/junction points), so nothing downstream catches this. A terminal really can end up with two independent conductors created by two different original wires in one call.

That makes it more than a theoretical edge case: **placing a component at a point where two independent, unconnected wires happen to cross** — a completely ordinary layout in real schematics — would silently bridge both nets together through the new component's terminal. That's the exact scenario this feature exists for (dropping a part onto existing wiring), and the wrong result gives the user no visual reason to suspect anything happened.

Suggested fix: add a `break;` right after a match is fully handled (after line 406, `broken_endpoints`/`used_terminals` insertion for the "other" endpoint), so the terminal loop moves on to the next terminal instead of continuing to scan for further matches against the one it just resolved.

## Two smaller notes, not blockers

- The "preserve bent conductor" second connection (the `other_terminal` search) picks the new element's nearest free terminal by raw Euclidean distance only, with no check that its orientation matches the approach direction of `other_dock`. The primary `connect_to` selection does account for orientation via the `switch` above it; this second pick doesn't, so it could occasionally route the preserved segment into the "wrong side" of a terminal.
- The new toolbar action reuses `QET::Icons::Conductor` (a plain wire icon) for a checkable toggle that's conceptually closer to "auto-connect" than "draw a wire." Not confusing on its own since the icon isn't used elsewhere on a visible toolbar, but a bit generic for a feature toggle — `QET::Icons::Autoconnect` (already used by the sibling `m_auto_conductor` action) or a distinct icon might read more clearly at a glance.

Happy to be corrected if I'm missing a guard elsewhere that prevents the double-break case — but I couldn't find one, and the fix if needed is small.

Kellermorph:
Hi ispyisail,

Thanks for testing so thoroughly again.
I didn't notice any errors during my test. Because of your comment, I tested something else and ran into some issues there.
Thanks for being so observant.

I'll take another look at it and post an update.

Kellermorph:
Fix: Auto-break conductors from both sides when placing element on terminal strip

When placing a multi-pole element on existing conductors that all lead to the same terminal strip (e.g. 3 conductors from the left and 3 from the right), the auto-break feature now correctly breaks ALL conductors passing through each terminal dock point, not just the closest one.

Changes:

Break all conductors from the same circuit: When multiple conductors pass through a dock point, they are all broken and reconnected through the new element — but only if they share the same "other" endpoint (i.e. belong to the same circuit). This prevents silently bridging two independent nets when unrelated conductors happen to cross at a dock point.
Orientation-aware terminal matching: The "other" endpoint of a broken conductor is now connected to the nearest free terminal whose orientation matches the approach direction (e.g. a South terminal only accepts connections from below). This prevents routing a conductor into the wrong side of a terminal.
Tested with:

3-pole element between 3 left-side and 3 right-side conductors going to the same terminal strip — all 6 conductors correctly broken and reconnected
Builds clean, zero new warnings

ispyisail (commented):
Went through the latest "fix" commit (d850241f9) against the base "Auto-break conductor" commit, plus the surrounding infrastructure it leans on (`Terminal::orientation()`/`dockConductor()`, `Conductor`'s constructor, `DeleteQGraphicsItemCommand`, `QET::orthogonalProjection`).

**What the fix gets right:**
- Adding the `used_terminals.contains(t)` guard at the top of the outer terminal loop closes a real gap in the base commit — without it, a terminal already consumed as an "other" reconnection target for one pole could get reprocessed and double-connected while handling a later pole.
- I traced the North/South/East/West sign conventions in the new orientation-matching code against `Terminal::init()`'s docking-point offsets and they're internally consistent with the rest of the codebase.
- Grouping conductors by shared "other" endpoint before batch-breaking them is a sound way to avoid bridging two unrelated nets that happen to cross at the same point.

**Two things worth resolving before merge:**

1. **Pre-existing (in the base commit, not this fix), but a real bug:** the "is this dock point on the conductor" check only tests `it != QLineF::NoIntersection` from `QET::orthogonalProjection()`. That's true for both `BoundedIntersection` (the point falls within the actual drawn segment) *and* `UnboundedIntersection` (the point falls on the segment's infinite extension, off the physically drawn wire). So a new element's terminal dock point that's collinear with a wire but past its visible end can get falsely detected as "on" that conductor and spliced in, even though the wire never actually reached there. Worth checking `it == QLineF::BoundedIntersection` specifically (or clamping the projection parametrically to [0,1] along the segment) while this code is already being touched.

2. **In this fix commit:** `best` (the conductor whose group actually gets broken) is chosen by nearest `connect_to` distance across *all* candidates at a dock point, not by which group is larger. If three conductors cross the same point — two belonging to one net, one to another — and the lone outlier happens to have the nearest `connect_to`, the two same-net conductors get silently skipped this round rather than broken together. That's exactly the multi-conductor-per-point scenario this fix targets, so worth a look.

Non-code note: this branch is based on a commit from well before #620 merged (~180 commits behind current `master` as of this review), so it'll need a rebase regardless of the above.

Neither point is a five-alarm blocker — this commit is a clear improvement over the base — but #1 undermines the geometric premise the whole feature rests on, so I'd resolve it before merge.

Kellermorph:
Re Bug 1 (BoundedIntersection): This is already handled. QET::orthogonalProjection() is not a thin wrapper around QLineF::intersect() — it calls QET::lineContainsPoint() internally (qet.cpp:207), which checks both collinearity (unitVector()) and segment bounds (point_line.length() <= line.length()). So the function already returns false for projections that fall on the infinite extension but outside the actual drawn segment. No code change needed here.

Re Bug 2 (best by group size instead of distance): Fixed. The "best" group is now chosen by largest group size (most conductors sharing the same other endpoint), not by nearest connect_to distance. Matches are collected into a QMap<Terminal*, QList<ConductorMatch>> grouped by other endpoint, and the group with the most members wins. This ensures that when e.g. 2 same-net conductors and 1 independent conductor cross at a dock point, the 2-member group gets broken rather than the lone outlier.

ispyisail:
Fix for bug 1 from my earlier review — the unbounded-line intersection issue.

`QET::orthogonalProjection()` reports a hit for any point on a segment's *infinite extension*, not just the finite drawn segment (it only checks `QLineF::intersects() != NoIntersection`, which is also true for `UnboundedIntersection`). So a new element's terminal dock point that's collinear with a wire but past its actual drawn end was getting falsely detected as "on" that conductor and spliced in, even though the wire never physically reached there.

I don't have push access to this branch, so here's the fix as a patch — apply with `git am` against your current HEAD (`5ec49eedd`), or cherry-pick/adapt by hand, whichever's easier:

```diff
diff --git a/sources/diagramevent/diagrameventaddelement.cpp b/sources/diagramevent/diagrameventaddelement.cpp
index e8f0f8635..2f018e28b 100644
--- a/sources/diagramevent/diagrameventaddelement.cpp
+++ b/sources/diagramevent/diagrameventaddelement.cpp
@@ -31,6 +31,34 @@
 #include <QPainterPath>
 #include <limits>
 
+namespace {
+	/**
+		@brief distanceToSegment
+		@param point : point to measure from
+		@param segment : the finite segment (not the infinite line through it)
+		@return the distance from @a point to the closest point actually on
+		@a segment. Unlike QET::orthogonalProjection(), which reports a hit
+		for any point on the segment's infinite extension, this clamps the
+		projection to the segment itself: a point collinear with a wire but
+		past its actual drawn end is correctly reported as far away, not "on"
+		the wire.
+	*/
+	qreal distanceToSegment(const QPointF &point, const QLineF &segment)
+	{
+		const QPointF a = segment.p1();
+		const QPointF b = segment.p2();
+		const QPointF ab = b - a;
+		const qreal len2 = QPointF::dotProduct(ab, ab);
+
+		if (len2 <= 0.0)
+			return QLineF(point, a).length();
+
+		qreal t = QPointF::dotProduct(point - a, ab) / len2;
+		t = qBound(0.0, t, 1.0);
+		return QLineF(point, a + t * ab).length();
+	}
+}
+
 /**
 	@brief DiagramEventAddElement::DiagramEventAddElement
 	Defaut constructor
@@ -297,7 +325,9 @@ void DiagramEventAddElement::addElement()
 				    c->terminal2->parentElement() == element)
 					continue;
 
-					//Check if dock point lies on the conductor path.
+					//Check if dock point lies on the conductor path. Distance is
+					//measured to the segment itself (clamped), not the infinite
+					//line through it -- see distanceToSegment().
 				QPointF local_dock = c->mapFromScene(t_dock);
 				bool point_on_conductor = false;
 				QPainterPath path = c->path();
@@ -306,15 +336,10 @@ void DiagramEventAddElement::addElement()
 					const QPainterPath::Element &e1 = path.elementAt(i);
 					const QPainterPath::Element &e2 = path.elementAt(i + 1);
 					QLineF segment(QPointF(e1.x, e1.y), QPointF(e2.x, e2.y));
-					QPointF projection;
-					if (QET::orthogonalProjection(local_dock, segment, &projection))
+					if (distanceToSegment(local_dock, segment) < 5.0)
 					{
-						qreal dist = QLineF(local_dock, projection).length();
-						if (dist < 5.0)
-						{
-							point_on_conductor = true;
-							break;
-						}
+						point_on_conductor = true;
+						break;
 					}
 				}
```

Deliberately left `QET::orthogonalProjection()` itself alone — its other caller (`Conductor::movePointIntoPolygon`) has different semantics (picks the globally-nearest projection across every edge of a polygon), so changing its bounded/unbounded behavior wasn't in scope here.

Verified: full project builds and links clean with this change on top of `5ec49eedd`.

Kellermorph:
Thank you so much for your effort and help! Thank you! 
I've passed your message on to the AI so it can incorporate the patch

Kellermorph:
everything fine now?

scorpio810:
Merged, but didn't working here...

Kellermorph:
Hi Laurent,
what isn't working? There is a new button on the top. Did you activate it?

scorpio810:
Ooops , sorry Thomas..

Kellermorph:
so could u merge it again or how does this work now?

scorpio810:
https://github.com/qelectrotech/qelectrotech-source-mirror/commit/e8f80697f3cfe7278df49624073a8c9725f7d376

scorpio810:
Hi Thomas, was I saw yesterday 
1 copy an three power contacts element (an element already placed in the folio in the same conductors).
2 clic on break conductor icon and place this copied element-> conductors was not breaked.

ispyisail:
I reproduced this and tracked down why.

**Repro:** with "Coupure automatique de conducteur(s)" enabled, copy an already-placed/connected element (`Ctrl+C`), paste it (`Ctrl+V`), then drag the pasted copy so its terminal lands on an existing, unrelated conductor. The conductor stays a single unbroken wire — the new element just sits on top of it, never spliced in.

**Root cause:** the auto-break logic added here lives entirely inside `DiagramEventAddElement::addElement()` (`sources/diagramevent/diagrameventaddelement.cpp`), which only runs for the "drag a part in from the collection panel and drop it" flow (wired up in `diagramview.cpp` around the `new DiagramEventAddElement(...)` call). That's the only call site that checks `project()->autoBreakConductor()`.

Paste goes through a completely different path: `DiagramView::paste()` → `PasteDiagramCommand::redo()` (`sources/diagramcommands.cpp`), which just adds the copied items to the scene — it never calls into the break-detection code at all. Same story for repositioning an element that's already on the diagram (ordinary mouse-drag of an existing item never goes through `DiagramEventAddElement` either), so the feature currently only fires at the exact moment of a fresh drop from the parts panel.

**Suggested fix:** factor the per-terminal "find overlapping conductor and split it" logic out of `DiagramEventAddElement::addElement()` into a shared helper, and call it from `PasteDiagramCommand` too (and ideally from wherever ordinary element-move commits, so dragging an existing element onto a conductor behaves the same way). Happy to take a pass at this if useful — wanted to confirm the diagnosis first.

Kellermorph:
i will take a look to add this feature. This PR just include if u take a new Element with drag and drop

PR #640, opened 2026-08-02
Configuration export/import for switching QET config profiles (discussion #610)

ispyisail:
Implements discussion #610: an in-app way to save and load QET configuration profiles, replacing the manual OS-settings-file-renaming workaround Nuri (a translator working with multiple clients) currently uses.

## Why `--config-dir` doesn't already solve this

Checked before writing anything: `QETApp::configDir()` (overridable via `--config-dir`) is only consulted for a handful of specific paths — element text patterns, the custom stylesheet, directory creation at startup. It never touches `QSettings`, which uses Qt's own OS-native backend (an ini file on Linux, the registry on Windows, a plist on macOS) tied to the organization/application name set in `main.cpp`. Shortcuts, autosave interval, recent files, UI layout — the bulk of actual preferences — aren't affected by `--config-dir` at all. Nuri's manual file-renaming workaround is genuinely the only option that exists today.

## What it adds

Two entries under `Configuration`: **"Enregistrer la configuration sous..."** and **"Charger une configuration..."**.

Both go through `QSettings(path, QSettings::IniFormat)` for the saved/loaded file specifically, regardless of the live backend on that platform — a plain key/value copy loop (`allKeys()` + `value()`/`setValue()`), identical code on Windows, Linux and macOS. Export writes every key from the live settings to a chosen file. Import **clears** the live settings first, then copies the file's keys in.

Clearing rather than merging is deliberate: a key present in the current configuration but absent from the loaded file must not survive the switch, or loading a profile that never mentions a given key would silently keep whatever the previous profile set for it — the opposite of what "switch profiles" means.

## Why restart, not live reload

QET has no settings-changed signal wired through every consumer (shortcuts, autosave timer, docks, etc.), so faking a live reload risks parts of the UI silently running on stale settings — worse than the restart the original forum request already said was acceptable. Import reuses the exact same `closeEveryEditor()` call `quitQET()` already uses, so an unsaved project still prompts to save, and a cancelled prompt aborts the whole import before the settings are touched.

The restart starts a detached second instance of the same executable, then quits the current one. Checked `SingleApplication`'s construction in `main.cpp` first: it's built with `allowSecondary=true`, so a second instance starting before the first fully exits doesn't race a single-instance lock.

## UI placement

Both actions live in `QETMainWindow`, the shared base every top-level QET window already builds its `&Configuration` menu from — so they appear consistently in the diagram editor and the element editor with no per-window wiring. Uses `DocumentImport` (already defined, previously unused anywhere) paired with the existing `DocumentExport`.

## Verified end-to-end via a real Xvfb session, not just a build

Built clean, zero new warnings.

Exported the live configuration and confirmed a clean round trip: 102 keys on both sides, plain ini format.

Then the actual scenario the discussion describes: closed QET cleanly, hand-edited a live setting on disk (`zoom-out-beyond-of-folio`, flipped to `true`) to simulate drift since the export, relaunched, used "Charger une configuration..." to load the earlier export (which had that key as `false`), confirmed the restart-confirmation dialog appears with the expected message, clicked Yes, and confirmed:
- the old process actually exited and a **new** one started (different PID, no `--config-dir` argument — a genuinely fresh launch, not the same process reusing its window)
- the setting on disk read back as `false` again — reverted to exactly what the loaded profile contained, not left at the drifted value.

## Not in this PR (per the discussion's own scope)

Live reload without restart, automatic per-client profile detection, and cloud sync/sharing infrastructure are all deliberately out of scope — sharing a profile with someone is already "hand them the file."

### DIFF
--- a/sources/qetapp.cpp
+++ b/sources/qetapp.cpp
@@ -45,9 +45,12 @@
 #include <iostream>
 #define QUOTE(x) STRINGIFY(x)
 #define STRINGIFY(x) #x
+#include <QFileDialog>
 #include <QFontDatabase>
+#include <QProcess>
 #include <QProcessEnvironment>
 #include <QRegularExpression>
+#include <QSettings>
 #ifdef BUILD_WITHOUT_KF5
 #	include "ui/nokde/kautosavefile.h"
 #else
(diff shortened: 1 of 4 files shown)


ispyisail:

PR #641, opened 2026-08-02
Generate cabinet placement thumbnails from manufacturer/reference info (discussion #602)

ispyisail:
Implements the scope proposed in discussion #602: https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/602

## What this adds

A new context-menu action, **"Générer une vignette d'armoire"**, available whenever the current diagram selection includes at least one element with both `manufacturer` and `manufacturer_reference` filled in (`DiagramView::contextMenuActions()`).

Triggering it (`DiagramView::generateCabinetThumbnails()`):
- For each eligible selected element, synthesizes a minimal `ElementData::Thumbnail`-type `.elmt` definition: a plain labeled rectangle whose dynamic text is a `CompositeText` of `%{manufacturer}` / `%{manufacturer_reference}`, with those two values also baked into the definition's own `<elementInformations>` so the thumbnail displays correctly the instant it's placed.
- Files it into a dedicated **"Cabinet thumbnails"** folder under the project's embedded collection, created on demand via `XmlElementCollection::createDir()` + `addElementDefinition()` — the same primitives the collection-management UI itself uses, so the result shows up in the elements panel and is drag-and-droppable onto any diagram with no new UI plumbing.
- Dedups by `manufacturer + " " + manufacturer_reference` against what's already filed in that folder (checked via `elementsNames()`), so placing the same device on several schematic elements doesn't pile up redundant thumbnails.

## Out of scope (per the discussion)

The synthesized block's size is a fixed placeholder, not the device's true physical footprint — QET has no info field for physical width/height/depth today. Adding one so thumbnails could be generated at true scale is a natural follow-up, but separate scope from the generate-and-file workflow here.

## Testing

Built clean, no new warnings. Verified end-to-end in a real GUI session (Xvfb):
- Loaded a project with an element whose `elementInformations` had `manufacturer`/`manufacturer_reference` set.
- Right-clicked it; confirmed the new action appears (and only appears for elements with both fields filled — verified it's absent for elements without them).
- Triggered it, saved the project, and confirmed in the saved XML that a `Schneider Electric XB5-AA21.elmt` Thumbnail definition was correctly filed under a new `<category name="Cabinet thumbnails">`, with the manufacturer/reference baked into its `elementInformations` and the composite text holding the literal `%{manufacturer}`/`%{manufacturer_reference}` placeholders.
- Triggered the action a second time on the same element and re-saved; confirmed no duplicate was created (dedup works).

### DIFF
--- a/sources/diagramview.cpp
+++ b/sources/diagramview.cpp
@@ -38,7 +38,11 @@
 #include "ElementsCollection/xmlelementcollection.h"
 #include "NameList/nameslist.h"
 #include "elementdialog.h"
+#include "qetgraphicsitem/element.h"
+#include "qetinformation.h"
 #include <QDropEvent>
+#include <QSet>
+#include <QUuid>
 
 /**
 	Constructeur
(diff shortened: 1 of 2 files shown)


ispyisail:

PR #642, opened 2026-08-02
Add user-defined custom properties on elements (discussion #611)

ispyisail:
Implements the scope proposed in discussion #611: https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/611

## The gap this closes

`DiagramContext` (the free-form key/value store backing every element's info fields) already accepts and round-trips any key matching `^[a-z0-9-_]+$` generically via `toXml()`/`fromXml()`. But `ElementInfoWidget::buildInterface()` only ever builds one row per key from the fixed ~40 predefined `ELMT_*` constants — there was no way for a user to actually create a new key through the properties dialog, even though the storage layer already fully supported it.

## What this adds

- A new **"Ajouter une propriété personnalisée"** button at the bottom of the element Informations panel.
- Clicking it appends a `CustomElementInfoPartWidget` row — unlike the existing fixed `ElementInfoPartWidget` rows (each hardcoded to one predefined key), both the key name and the value are user-editable here.
- The typed key is validated live against the already-public `DiagramContext::isKeyAcceptable()` and the field gets a red border the moment it doesn't match, instead of the key silently failing to save.
- `ElementInfoWidget::updateUi()` now also scans the element's current informations for any key outside the predefined/special set and rebuilds a custom row for each — so a previously-added custom property re-displays correctly the next time that element is selected or the project is reopened.
- No changes needed to `DiagramContext::toXml()`/`fromXml()` — custom keys flow through the exact same generic path predefined keys already use.
- A remove button on each custom row deletes that property.

## Testing

Built clean, no new warnings. Verified end-to-end in a real GUI session (Xvfb) against `examples/741.qet`:
- Selected a resistor element, clicked "Ajouter une propriété personnalisée", typed key `puissance_max` / value `250W`, saved the project.
- Confirmed in the saved XML that `puissance_max` was written as a normal `<elementInformation>` entry alongside the predefined keys, with no special-casing needed.
- Fully restarted the app, reloaded the project, reselected the same element: the custom row reappeared pre-filled with `puissance_max` / `250W`.
- Edited the key to `Invalid Key!` (uppercase + space, violates the accepted key format) and confirmed the field immediately shows a red border.

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -683,6 +683,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/ui/dynamicelementtextitemeditor.h
   ${QET_DIR}/sources/ui/dynamicelementtextmodel.cpp
   ${QET_DIR}/sources/ui/dynamicelementtextmodel.h
+  ${QET_DIR}/sources/ui/customelementinfopartwidget.cpp
+  ${QET_DIR}/sources/ui/customelementinfopartwidget.h
   ${QET_DIR}/sources/ui/elementinfopartwidget.cpp
   ${QET_DIR}/sources/ui/elementinfopartwidget.h
   ${QET_DIR}/sources/ui/elementinfowidget.cpp
--- a/sources/ui/customelementinfopartwidget.cpp
+++ b/sources/ui/customelementinfopartwidget.cpp
@@ -1,0 +1,113 @@
+/*
+	Copyright 2006-2026 The QElectroTech Team
+	This file is part of QElectroTech.
+
+	QElectroTech is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 2 of the License, or
+	(at your option) any later version.
+
+	QElectroTech is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with QElectroTech.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#include "customelementinfopartwidget.h"
+
+#include "../diagramcontext.h"
+#include "../qeticons.h"
+
+#include <QGridLayout>
+#include <QLineEdit>
+#include <QToolButton>
+
+/**
+	@brief CustomElementInfoPartWidget::CustomElementInfoPartWidget
+	Constructor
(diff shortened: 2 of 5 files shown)


plc-user:
Is there a special reason, why the user-defined properties are only editable for elements already used in a diagram?
When I add custom properties to an element-file with a text-editor directly in XML these are also read by QET and are shown in the list as custom properties:
grafik
In my opinion it would make sense to edit additional properties (also) in element-editor.

plc-user:
When looking at the properties in elements already placed on a folio I find an edit-line with no text and no description:
grafik

When adding / editing a property that edit-line gets filled with a `false`:
grafik

Could this be a bug?

ispyisail:
@plc-user Thanks — both are good catches. Taking them in reverse order.

### The nameless row filling with `false` — yes, that's a bug

Traced it. `"exclude_from_bom"` is in `QETInformation::elementInfoKeys()`, and `buildInterface()` creates one `ElementInfoPartWidget` per key in that list, so the key gets a generic edit row *in addition to* its dedicated "Exclure de la nomenclature" check box. `translatedInfoKey()` has no case for it and falls through to `else return QString()`, so that row has no label — your first screenshot.

Then `currentInfo()` writes the key unconditionally:

```cpp
if (m_exclude_from_bom_cb) {
    info_.addValue(QStringLiteral("exclude_from_bom"), m_exclude_from_bom_cb->isChecked() ? "true" : "false");
}
```

so the next `updateUi()` refills that anonymous row with `false` — your second screenshot.

One thing worth being clear about: this isn't a regression from this PR. Both ingredients are already in master at `14ef848c4` (the merge's first parent) — the key was in `elementInfoKeys()` and the unconditional write was at `elementinfowidget.cpp:327`. What #642 changed was adding the custom-property rows underneath, which is what made the panel worth looking at closely.

Fix in #738. It drops the key from the list `buildInterface()` iterates, rather than from `elementInfoKeys()` itself — the latter would delete the `element_info` database column that `projectdatabase.cpp` builds from that list and that `elementquerywidget.cpp` reads, so it would break the BOM-exclusion feature. The check box stays the only way to set it and the stored value is unchanged.

### Custom properties in the element editor

Fair point, and I agree it would make sense — but I'd treat it as a separate feature rather than a gap in this one, because the two store different things.

What this PR added is per-*instance* properties: `DiagramContext` on the placed element, saved in the project file. The element editor edits the `.elmt` definition, so custom properties there would be per-*symbol* defaults, inherited by every future instance — a different code path (`ElementPropertiesEditorWidget`) and a different storage location.

Your observation that hand-written custom keys in a `.elmt` already show up in the list is the interesting part: it means the read path already inherits them, so what's missing is only the editing UI on the definition side, plus a decision about what should happen when a definition's default and an instance's value disagree. Happy to scope that separately if you think it's worth doing.

PR #643, opened 2026-08-02
Add adjustable background frame in the element editor (discussion #604)

ispyisail:
Implements the scope proposed in discussion #604: https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/604

## What this adds

A toggleable, visual-only background frame in the element editor, letting an element's drawing be proportioned directly against a realistic folio reference instead of a bare grid.

- Follows the exact precedent of the existing hotspot indicator (`ElementScene::drawForeground()`, `elementscene.cpp:318-333`): drawn in `ElementView::drawBackground()`, centered on the same origin, and never written to the saved `.elmt` XML.
- **"Afficher le cadre de fond"** toggles the frame on/off (menu + toolbar).
- **"Taille du cadre de fond..."** opens a small width/height dialog to resize it.
- Defaults to a representative folio's drawable area, derived from `BorderProperties::defaultProperties()` (`columns_count * columns_width` by `rows_count * rows_height` — 1020×640 out of the box), since element coordinates and diagram scene coordinates already share the same grid-unit space.
- Both the visibility and the last-used size persist via `QSettings`, the same way the existing grid-dot rendering prefs (`elementeditor/grid_pointsize_min`/`max`) already do — no reset between sessions.

## Testing

Built clean, no new warnings. Verified end-to-end in a real GUI session (Xvfb):
- Opened the element editor on an existing element, toggled the frame on, confirmed a dashed rectangle appears centered on the same origin as the hotspot crosshair, sized to the default 1020×640.
- Opened "Taille du cadre de fond...", changed it to 500×500, confirmed the drawn frame updates and stays centered on the origin.
- Fully quit and relaunched the element editor: confirmed the toggle was still on and the frame was still 500×500, restored from `QSettings` with no reconfiguration needed.
- Confirmed the element's `.elmt` file has zero diff and contains no trace of the frame — it's genuinely visual-only, like the hotspot marker.

### DIFF
--- a/sources/editor/elementscene.cpp
+++ b/sources/editor/elementscene.cpp
@@ -17,6 +17,7 @@
 */
 #include "elementscene.h"
 
+#include "../borderproperties.h"
 #include "../NameList/ui/namelistdialog.h"
 #include "../NameList/ui/namelistwidget.h"
 #include "../QPropertyUndoCommand/qpropertyundocommand.h"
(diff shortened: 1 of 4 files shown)


ispyisail:

PR #645, opened 2026-08-03
Cover auto-numbering counter changes with undo/redo (discussion #608)

ispyisail:
Implements the scope proposed in discussion #608: https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/608

## The bug

Placing an auto-numbered element or conductor advances a shared `NumerotationContext` counter (`QETProject::addConductorAutoNum()`/`addElementAutoNum()`) as a side effect that sits entirely outside the undo stack. Undo the placement and the visible label goes away, but the counter stays advanced — every undo of an auto-numbered placement silently burns a number, with no way to get it back except a manual reset (discussion #597).

## The fix

Adds `SetAutoNumContextCommand` (`sources/undocommand/`), a small `QUndoCommand` storing the old/new `NumerotationContext` and calling the matching `add*AutoNum()` setter on `undo()`/`redo()` — the same shape `QPropertyUndoCommand` already uses right next to it in `ConductorAutoNumerotation::applyText()`.

Wired into the exact call sites the discussion identified:
- `ConductorAutoNumerotation::newProperties()` (static) and `numerateNewConductor()` — both direct, unwrapped `addConductorAutoNum()` calls.
- `Element::setUpFormula()`, called from `DiagramEventAddElement::addElement()` when a new element is dropped onto a diagram. `setUpFormula()` now takes an optional parent `QUndoCommand *`; `addElement()` calls it *before* pushing its own `undo_object` (previously it ran after), so the counter change lands in the same undo macro as the element's own placement — **one Ctrl+Z reverts both together**, instead of leaving the counter adrift.

Deliberately **not** touched, matching the discussion's own scope:
- The project-properties config dialog's `add*AutoNum()` calls — those are deliberate edits to the numbering *rule* itself, not a side effect of placing something.
- `Diagram::loadElmtFolioSeq()`/`loadCndFolioSeq()` and the bulk folio-renumbering passes in `QETProject` — these run as load-time/bulk bookkeeping, not as part of any undoable user gesture, so wrapping them in undo commands would be actively wrong (there's no sensible "undo" for "the diagram was loaded").

## Testing

Built clean, no new warnings.

Verified in a real GUI session (Xvfb) that the reordered `setUpFormula()` call executes correctly with no crash or regression: repeatedly placed new elements onto a diagram via drag-and-drop from the collection panel (the exact code path through `DiagramEventAddElement::addElement()`), saved, and reloaded — all placements persisted correctly.

**One gap I want to be upfront about**: I wasn't able to get a reliable, reproducible interactive test of the *conductor* counter's undo/redo round-trip (draw a new wire → check counter advanced → Ctrl+Z → check counter restored) inside this sandbox — precise terminal-to-terminal mouse-drag wire-drawing didn't register reliably under Xvfb+xdotool's synthetic input in the time I spent on it, similar to the drag-and-drop limitation noted in PR #638's testing. I'm confident in the logic by code review — `SetAutoNumContextCommand` is a direct structural copy of the already-proven `QPropertyUndoCommand` pattern, and the `numerateNewConductor()`/static `newProperties()` changes are mechanical (replace a bare setter call with construction of the same command, parented or pushed exactly like `applyText()` already does two lines above each of them) — but I want to flag that the conductor path's undo/redo round-trip specifically wasn't exercised end-to-end in a live app the way the element path was. Happy to have a reviewer sanity-check this by hand, or point me at a more reliable way to script wire-drawing for next time.

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -761,6 +761,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/undocommand/movediagramcommand.h
   ${QET_DIR}/sources/undocommand/removediagramcommand.cpp
   ${QET_DIR}/sources/undocommand/removediagramcommand.h
+  ${QET_DIR}/sources/undocommand/setautonumcontextcommand.cpp
+  ${QET_DIR}/sources/undocommand/setautonumcontextcommand.h
   ${QET_DIR}/sources/undocommand/rotateselectioncommand.cpp
   ${QET_DIR}/sources/undocommand/rotateselectioncommand.h
   ${QET_DIR}/sources/undocommand/rotatetextscommand.cpp
--- a/sources/conductorautonumerotation.cpp
+++ b/sources/conductorautonumerotation.cpp
@@ -24,6 +24,7 @@
 #include "qet.h"
 #include "qetdiagrameditor.h"
 #include "ui/potentialselectordialog.h"
+#include "undocommand/setautonumcontextcommand.h"
 
 /**
 	@brief ConductorAutoNumerotation::ConductorAutoNumerotation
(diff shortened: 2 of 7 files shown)


scorpio810:
https://youtu.be/mkfUvRd55iE

PR #647, opened 2026-08-03
Add crash-time ring flush and diagnostics export UI (discussion #644, steps 4-5)

ispyisail:
Implements steps 4-5 of discussion #644: https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/644

**Stacked on #646** (steps 1-3, not yet merged) -- this branch is built on top of that one, so the diff below includes #646's commit too. Please review/merge #646 first; once it lands, this PR's diff will shrink to the single commit `Add crash-time ring flush and a diagnostics export UI`. Kept as its own PR rather than folded into #646, matching the discussion's own framing: step 4 is explicitly "the highest-risk piece ... lands last, behind its own switch."

## Step 4 — crash-time ring flush (`CrashHandler`)

Installs a handler for `SIGSEGV`/`SIGABRT`/`SIGBUS`/`SIGFPE`/`SIGILL` (POSIX) / `SetUnhandledExceptionFilter` (Windows) that flushes the in-memory ring to a fixed `crash_dump.log` before the process dies.

This required reworking `LogRing` (step 3, in #646) to be genuinely **lock-free**, not just mutex-protected: a signal handler that blocks on a lock the crashing thread (or another thread) already holds turns a clean crash into a hang — no ring dump *and* no core dump, worse than doing nothing. `append()` now claims a slot with a single atomic fetch-add; `dumpToFd()` reads the preallocated entries directly and writes them with `write(2)` only, looping on `EINTR`/short writes. Accepted tradeoff: at most one entry can be read torn if a crash lands mid-append into that exact slot — documented in `logring.h`; a seqlock to detect/retry wasn't judged worth the complexity for that window.

Other invariants implemented per the discussion:
- `sigaltstack` with a static 64 KiB buffer, `SA_ONSTACK` — a stack-overflow `SIGSEGV` has no usable stack for a handler without one.
- Nothing under the actual handler touches Qt, `QString`, or the allocator: the dump path and a small header (version/git/OS/Qt) are precomputed into fixed `char` buffers by `install()`, which runs once at startup in normal context.
- Atomic test-and-set so only the first crash writes a dump; a second concurrent/nested fault goes straight to restore-and-re-raise.
- After writing, the handler restores `SIG_DFL` and re-raises (POSIX) / returns `EXCEPTION_CONTINUE_SEARCH` (Windows), so the OS's own crash path — core dump, Windows Error Reporting — still runs. A handler that "fixed" the crash by swallowing the signal would destroy exactly the post-mortem evidence this whole design exists to preserve.

**Tested in this environment: POSIX/Linux only**, all five signals. Sent each directly to a running process and confirmed (a) `crash_dump.log` is written with the correct header and ring contents, mode 0600, and (b) the process still terminates via the signal with the kernel's own "core dumped" flag set (exit code 128+signal, confirmed for all five). **The Windows path is implemented per the discussion's guidance but is untested — no Windows build available in this sandbox.** Please sanity-check before relying on it.

## Step 5 — getting the data back out

- `QETApp::checkCrashDump()`, called from `checkBackupFiles()` only when there's no stale project file to recover this run (so the two prompts never both show, per the discussion), offers an unretrieved crash dump via `DiagnosticsReportDialog` and then deletes it regardless of the user's choice — offered exactly once.
- A new **"Aide > Enregistrer un rapport de diagnostic..."** action builds the same kind of report from the *current* session (header + this session's log file) for a manual "attach this to a bug report" flow, not tied to a crash.
- Both go through `QetLogger::redact()` before ever reaching the user: the one redaction implemented is a literal replace of the home directory with `~`, since an absolute path under it leaks the account name. The discussion's fancier "optionally redact project filenames too" isn't attempted — reliably telling a project path apart from arbitrary log text is a much fuzzier problem than a literal prefix match.
- `DiagnosticsReportDialog` shows the full (already-redacted) content before saving, per the discussion: "the user is about to attach this to a public tracker."

## Testing

Built clean, no new warnings.

Verified in a real GUI session (Xvfb):
- Sent `SIGSEGV` to a running instance, relaunched, confirmed the crash-report dialog appears with the correct header/ring content.
- Confirmed it does **not** reappear on a second relaunch (retrieval-marking works).
- Confirmed the manual "Save report" action produces a correctly-formatted report (live session header + current log) and saves it to a chosen path via the dialog's Save button.

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -116,10 +116,14 @@ set(QET_RES_FILES
 set(QET_SRC_FILES
   ${QET_DIR}/sources/cli_export.cpp
   ${QET_DIR}/sources/cli_export.h
+  ${QET_DIR}/sources/logging/crashhandler.cpp
+  ${QET_DIR}/sources/logging/crashhandler.h
   ${QET_DIR}/sources/logging/logring.cpp
   ${QET_DIR}/sources/logging/logring.h
   ${QET_DIR}/sources/logging/qetlogger.cpp
   ${QET_DIR}/sources/logging/qetlogger.h
+  ${QET_DIR}/sources/logging/ui/diagnosticsreportdialog.cpp
+  ${QET_DIR}/sources/logging/ui/diagnosticsreportdialog.h
   ${QET_DIR}/sources/pdf_links.cpp
   ${QET_DIR}/sources/pdf_links.h
   ${QET_DIR}/sources/import/edz/edzarchive.cpp
--- a/qelectrotech.pro
+++ b/qelectrotech.pro
@@ -174,7 +174,8 @@ HEADERS += $$files(sources/*.h) \
            $$files(sources/svg/*.h) \
            $$files(sources/import/edz/*.h) \
            $$files(sources/import/edz/lzma/*.h) \
-           $$files(sources/logging/*.h)
+           $$files(sources/logging/*.h) \
+           $$files(sources/logging/ui/*.h)
 
 SOURCES += $$files(sources/*.cpp) \
            $$files(sources/editor/*.cpp) \
@@ -221,7 +222,8 @@ SOURCES += $$files(sources/*.cpp) \
            $$files(sources/svg/*.cpp) \
            $$files(sources/import/edz/*.cpp) \
            $$files(sources/import/edz/lzma/*.c) \
-           $$files(sources/logging/*.cpp)
+           $$files(sources/logging/*.cpp) \
+           $$files(sources/logging/ui/*.cpp)
 
 
 # Needed for use promote QTreeWidget in terminalstripeditor.ui
--- a/sources/logging/crashhandler.cpp
+++ b/sources/logging/crashhandler.cpp
@@ -1,0 +1,171 @@
+/*
+	Copyright 2006-2026 The QElectroTech Team
+	This file is part of QElectroTech.
+
+	QElectroTech is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 2 of the License, or
+	(at your option) any later version.
+
+	QElectroTech is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with QElectroTech.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#include "crashhandler.h"
+
+#include "logring.h"
+#include "../qetversion.h"
+
+#include <QByteArray>
+#include <QSysInfo>
+#include <atomic>
+#include <cstring>
+
+#ifdef Q_OS_WIN
+#include <fcntl.h>
+#include <io.h>
+#include <share.h>
+#include <sys/stat.h>
+#include <windows.h>
+#else
+#include <cerrno>
+#include <csignal>
+#include <fcntl.h>
+#include <unistd.h>
+#endif
+
+namespace {
+
+// Everything the handler touches is preallocated here and filled in by
+// install() (normal context, runs once at startup) -- nothing under the
+// actual signal/exception path may allocate or touch QString/Qt.
+const LogRing *g_ring = nullptr;
+char g_dump_path[1024] = {};
+char g_header[1024] = {};
+int g_header_len = 0;
+
+// Guards against two threads crashing at once, or the handler itself
+// faulting while dumping: only the first crash writes a dump. See
+// crashhandler.h invariant 4.
+std::atomic<bool> g_already_dumped{false};
+
+#ifndef Q_OS_WIN
+
+// A stack-overflow SIGSEGV leaves no usable stack for a handler to run
+// on at all, hence the alternate signal stack (invariant: sized well
+// above any known SIGSTKSZ so this doesn't depend on
+// sysconf(_SC_SIGSTKSZ), which some libc versions require at runtime
+// rather than offering as a compile-time constant).
+char g_altstack[65536];
+
+const int kHandledSignals[] = {SIGSEGV, SIGABRT, SIGBUS, SIGFPE, SIGILL};
+
+void restoreDefaultAndReraise(int sig)
+{
+	struct sigaction sa {};
+	sa.sa_handler = SIG_DFL;
+	sigemptyset(&sa.sa_mask);
+	sa.sa_flags = 0;
+	sigaction(sig, &sa, nullptr);
+	raise(sig);
+}
+
+void signalHandler(int sig)
+{
+	if (g_already_dumped.exchange(true, std::memory_order_acq_rel)) {
+		// Not the first crash (concurrent fault on another thread, or
+		// this handler faulting while dumping): skip straight to
+		// restore-and-re-raise rather than risk a second, interleaved
+		// write to the same file.
+		restoreDefaultAndReraise(sig);
+		return;
+	}
+
+	// open/write/close are all on the POSIX async-signal-safe function
+	// list; nothing else is called here.
+	const int fd = ::open(g_dump_path, O_WRONLY | O_CREAT | O_TRUNC, 0600);
+	if (fd >= 0) {
+		if (g_header_len > 0) {
+			::write(fd, g_header, static_cast<size_t>(g_header_len));
+		}
+		if (g_ring) {
+			g_ring->dumpToFd(fd);
+		}
+		::close(fd);
+	}
+
+	restoreDefaultAndReraise(sig);
+}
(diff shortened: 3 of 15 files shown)


ispyisail:
Pushed English translations for this feature's UI text (c45ddcba1).

**What was happening:** the crash-recovery dialog and the manual "Save diagnostic report" dialog were written with French `tr()` source strings — correct, that's this codebase's convention throughout (checked: it's not unique to this file, it's how the whole app is written). But no English translations existed for these specific strings, because that only happens by running the `update_translations` (lupdate) developer target explicitly — it's not part of a normal build (CMakeLists.txt explains why: running lupdate on every build races with lrelease reading the same file). So `tr()` had nothing to fall back to but the raw French source, regardless of the user's own language setting.

**Fixed:** ran `update_translations`, then wrote reviewed English translations for the nine strings that are actually part of this feature — `DiagnosticsReportDialog`'s title/button/file-dialog/error text, and the two dialogs' body text in `QETApp::checkCrashDump()`/`showDiagnosticsReport()`. The other ~19 strings lupdate turned up are pre-existing gaps elsewhere in the app (auto-numbering widgets, the About dialog, etc.) — unrelated to this PR, left untouched. Other languages only got the new entries as empty placeholders, same as any lupdate run — filling those in is normal community-translation work, not something I attempted.

**On verification, being upfront about a limitation:** I confirmed the `.ts` entries are correct XML and confirmed the compiled `.qm` genuinely contains the new English text (Qt's `.qm` format is UTF-16LE, so a plain string search needs `-e l` to find it — a plain grep on the file looks empty even when the content is right, worth knowing if you check this yourself). I did not get a clean on-screen confirmation. My first attempt showed the dialog still in French despite no load errors — turned out to be `SingleApplication`'s instance-detection colliding with your own running `qet-testbuild` container over the shared host network (`network_mode: host` puts a native test binary and the container on the same network namespace, and `SingleApplication` doesn't appear to distinguish between them). I could see one of my launches reached your live container — it logged an "Elements collection reload" at the time, apparently harmless, no data affected, but a real unintended interaction with your running session. I stopped there rather than keep retrying against that risk. So: correct at the file and binary-content level, confirmed by inspection rather than by watching it render.

scorpio810:
Hello Shane,

Thanks for the follow-up and for being upfront about the verification.

Two things before we can move forward:

1. **Conflicts**: the branch needs a rebase on `master` (#646 has been merged since Aug 6, that's likely the source of the conflicts). Could you rebase `feature-eventloop-watchdog` (and the #647 branch too if it's diverged as well)?

2. **Translations (c45ddcb)**: appreciate the initiative, but could you drop the generated `.ts`/`.qm` files from this commit? I run `update_translations`/lupdate on my end as maintainer — that avoids bloating the PR with generated content and merge conflicts on those binary/XML files. The French source text in `tr()` is enough on your side, I'll handle the generation step.

3. **The network incident**: just to confirm — was the "qet-testbuild container" your own instance (e.g. from your qelectrotech-docker setup), not something on my end? Assuming that's the case, no action needed there, but wanted to double-check before treating it as harmless.

4. **qelectrotech.pro**: the CMake side is covered (you already added the two files to `qet_compilation_vars.cmake`), but the qmake `.pro` still needs an update. `sources/logging/*.h` and `sources/logging/*.cpp` are globbed so `crashhandler`/`eventloopwatchdog`/`logring`/`qetlogger` are picked up automatically — but `sources/logging/ui/` has no glob at all (unlike the other `ui/` subfolders, which each have one), so `diagnosticsreportdialog.{h,cpp}` won't build under qmake. Could you add:
```
$$files(sources/logging/ui/*.h)
$$files(sources/logging/ui/*.cpp)
```
to the HEADERS/SOURCES lists in `qelectrotech.pro`, alongside the other `ui/` subfolder globs? That way I don't have to push a follow-up fix after merging.


Thanks!
Laurent

ispyisail:
Working on it.

I was testing and i was having problems understanding .........then i realied it was in French........i don't understand French so i translated

I will fix

ispyisail:
Hi Laurent — all four done, thanks for the clear list.

**1. Rebase.** Both branches are rebased on current `master` (`ea5117b14`). #646's commit is now an ancestor rather than duplicated in the diff, which is where the conflicts were coming from. Both PRs report mergeable/clean again.

- #647 is now a single commit, 15 files.
- #665 (`feature-eventloop-watchdog`) is rebased on top of it and still stacked, 2 commits.

**2. Translations.** Dropped entirely — `c45ddcb` touched only `lang/*.ts` and `lang/*.qm` (58 generated files, ~25k lines), so removing the generated content removed the whole commit. #647 now has **zero** `.ts`/`.qm` files. The French `tr()` source strings are untouched in the `.cpp`, so `update_translations`/lupdate on your side has everything it needs. Agreed this is the better split — that commit was most of the PR's diff and would have conflicted constantly.

**3. The network incident.** Confirmed: that was entirely my own machine. It was my local `qelectrotech-docker` `qet-testbuild` container, and the cause was mundane — the compose service runs with `network_mode: host`, so it shares the host's network namespace, and `SingleApplication` keys its instance lock on a hash of the application name over an abstract Unix socket. A native test binary I launched with the same application name got adopted as a *secondary instance* of that container's QET and forwarded its arguments there instead of starting up. Nothing touched anything outside this machine, and nothing on your end was involved. I've since been isolating test runs by overriding the application name (and `HOME`/`XDG_CONFIG_HOME`) so it can't recur.

**4. `qelectrotech.pro`.** Added, exactly as you described:

```
           $$files(sources/logging/*.h) \
           $$files(sources/logging/ui/*.h)
```
```
           $$files(sources/logging/*.cpp) \
           $$files(sources/logging/ui/*.cpp)
```

You were right that this was the only gap — `sources/logging/*` already covers `crashhandler`/`logring`/`qetlogger`, and `eventloopwatchdog` in #665, but `sources/logging/ui/` had no glob of its own unlike the other `ui/` subfolders, so `diagnosticsreportdialog.{h,cpp}` would not have built under qmake.

I couldn't run a full `qmake` here (it stops earlier on `Unknown module(s) in QT: KWidgetsAddons KCoreAddons` — a missing optional KDE dep in my environment, unrelated to this change), so rather than assume, I evaluated the two globs on their own and confirmed they expand to include the new files:

```
sources/logging/crashhandler.h
sources/logging/logring.h
sources/logging/qetlogger.h
sources/logging/ui/diagnosticsreportdialog.h      <- new
```

Folded into the same commit that adds `sources/logging/ui/`, so the commit that introduces the directory also registers it in both build systems rather than leaving a follow-up for you.

CMake build re-verified after the rebase; `crashhandler`, `eventloopwatchdog`, `logring`, `qetlogger` and `ui/diagnosticsreportdialog` all compile and link.

Sorry about the `.ts`/`.qm` noise — I'll keep generated translation output out of future PRs.

scorpio810:
Hi Shane, thanks.

for qmake (Qt5) install debhelper (>= 11),
               libqt5svg5-dev,
               qt5-qmake,
               qtbase5-dev,
               libkf5widgetsaddons-dev,
               libkf5coreaddons-dev, 
               libsqlite3-dev,
               qtbase5-private-dev,
               pkgconf

scorpio810:
```
mkdir build && cd build
qmake ../qelectrotech.pro 
time make -j48
```

ispyisail:
That was the missing piece — thanks. `libkf5widgetsaddons-dev` and `libkf5coreaddons-dev` were the two I didn't have; installed them and ran your recipe verbatim. Both branches build clean under qmake now:

```
mkdir build && cd build
qmake ../qelectrotech.pro
time make -j24
```

- **#647**: `make exit=0`, no errors, 2m10s
- **#665** (stacked): `make exit=0`, no errors, 2m09s

So I can replace my earlier hand-wave with something actually checked.

**The `.pro` fix is confirmed necessary, and it was slightly worse than I described.** Comparing the generated `Makefile.Release` with and without the change:

| | object targets |
|---|---|
| without the globs | 496 |
| with them | 498 |

The two added are:

```
release/diagnosticsreportdialog.o
release/moc_diagnosticsreportdialog.o
```

The moc one is the part I'd missed. `DiagnosticsReportDialog` is a `QObject`, so without the glob it was neither compiled *nor* moc'd. Without the fix the only trace of it anywhere in the makefile is the header appearing in a dependency list of the translation unit that includes it — no compile step at all, so the link fails on undefined `DiagnosticsReportDialog::*`. Your diagnosis was exactly right.

Confirmed present in the linked qmake binary:

```
$ nm -C qelectrotech | grep DiagnosticsReportDialog | head -1
00000000005d5650 T DiagnosticsReportDialog::DiagnosticsReportDialog(QString const&, QString const&, QByteArray const&, QWidget*)
```

And on #665, `eventloopwatchdog.o` is picked up by the existing `sources/logging/*.cpp` glob with no `.pro` change needed, exactly as you said — all five objects present (`crashhandler`, `diagnosticsreportdialog`, `eventloopwatchdog`, `logring`, `qetlogger`), with `EventLoopWatchdog` symbols in the binary.

I'll build both ways before pushing from now on rather than assuming CMake coverage implies qmake coverage.

PR #648, opened 2026-08-03
IEC 81346 plant/location fallback and composite structure-id token (discussion #613)

ispyisail:
Implements discussion #613: https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/613

## What I found before implementing

I checked the discussion's claims against the current codebase, since it makes specific claims about what's wired up. One claim doesn't hold: the diagram's Installation (=) / Localisation (+) (`DIA_PLANT`/`DIA_LOCMACH`, `BorderTitleBlock::plant()`/`locmach()`) are **not** "never read anywhere else besides a tooltip" — they're already substituted via `%M`/`%LM` in element auto-numbering formulas (`assignvariables.cpp`), so a label formula like `=%M+%LM-%prefix%id` already works today.

The real, narrower gap: nothing connects the diagram's plant/location to an element's **own** `%{plant}`/`%{location}` composite-text tokens (a separate, independent field per element), and no composite-text token could reference the diagram's values or the element's prefix at all.

## What this PR does

`AssignVariables::replaceVariable()` (`sources/autoNum/assignvariables.cpp/.h`) gains an optional `const Element *elmt = nullptr` parameter:

- **Fallback**: `%{plant}` / `%{location}` resolve to the diagram's title block values when the element's own field is empty. Three-tier priority: manual element value → diagram value → empty.
- **`%{prefix}`**: exposes `Element::getPrefix()` to the composite-text system (previously only available as `%prefix` in the separate auto-numbering label-formula system, not in `%{...}` composite text).
- **`%{structure_id}`**: assembles `=plant+location-label`, omitting any segment whose value is empty. Deliberately uses the element's resolved *label* rather than `prefix+label` — in normal QET usage the auto-num formula already bakes the prefix into the label (e.g. "Q3"), so appending the prefix again would double it for the common case. Left the deeper multi-level/hierarchical case (`=E1.1+A2.3`) out of scope, matching the discussion's own stated scope.

All 9 call sites of `replaceVariable()` were checked individually. The 7 that place text on a diagram (`dynamicelementtextitem.cpp` x3, `dynamicelementtextmodel.cpp` x4) already had the `Element*` in scope and now pass it through. The 2 in `partdynamictextfield.cpp` run inside the *element editor* (composing a symbol, no diagram exists yet) and are intentionally left on the `nullptr` default.

`CompositeTextEditDialog`'s variable picker gets two new entries ("Préfixe (-)", "Désignation IEC 81346 (=+-)") so the tokens are discoverable, appended after the existing `elementInfoKeys()`-driven list since they're synthetic tokens computed by `replaceVariable()`, not real stored element info.

## Testing

Built clean (0 errors/warnings in the 5 changed files). Note: a *fresh* cmake configure in my sandbox mixes Qt5/Qt6 headers via the FetchContent-built KF5 addons (Qt6 dev packages are now also present in that environment) and fails to build the main target — confirmed this reproduces against unmodified `master` too, so it's a pre-existing environment issue, not something introduced here or that this PR needs to fix. Built against an already-Qt5-configured tree instead.

Verified live via Xvfb:
- Opened `examples/grafcet.qet`, set the folio's Installation/Localisation to `E1`/`A2`.
- Added a `%{structure_id}` composite text to an element with no plant/location of its own → rendered `=E1+A2` (label segment correctly omitted since empty for this element).
- Set the element's own Installation to `X9`, left Localisation empty → text updated live to `=X9+A2`: the element's override wins for Installation, the diagram fallback still applies to the untouched Localisation field.

### DIFF
--- a/sources/autoNum/assignvariables.cpp
+++ b/sources/autoNum/assignvariables.cpp
@@ -205,17 +205,52 @@ namespace autonum
 	/**
 		@brief AssignVariables::replaceVariable
 		Replace the variables in formula in form %{my-var}
-		to the corresponding value stored in dc
+		to the corresponding value stored in dc.
+		%{plant} and %{location} fall back to the containing diagram's
+		title block installation (=) / location (+) fields (IEC 81346)
+		when the element's own value is empty, so a project can set
+		these once at the folio level instead of on every element.
 		@param formula
 		@param dc
+		@param elmt the element the formula belongs to, if any;
+		used to resolve the diagram fallback above and the
+		%{prefix}/%{structure_id} tokens. May be null.
 		@return
 	*/
 	QString AssignVariables::replaceVariable(const QString &formula,
-						 const DiagramContext &dc)
+						 const DiagramContext &dc,
+						 const Element *elmt)
 	{
 		QString str = formula;
-		str.replace("%{label}", dc.value("label").toString());
-		str.replace("%{plant}", dc.value("plant").toString());
+
+		Diagram *diagram = elmt ? elmt->diagram() : nullptr;
+
+		QString plant_value = dc.value("plant").toString();
+		if (plant_value.isEmpty() && diagram)
+			plant_value = diagram->border_and_titleblock.plant();
+
+		QString location_value = dc.value("location").toString();
+		if (location_value.isEmpty() && diagram)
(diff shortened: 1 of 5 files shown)


ispyisail:

PR #650, opened 2026-08-03
Page-match abbreviation for %{structure_id} + %{structure_id_full} token (discussion #649)

ispyisail:
Implements the page-match abbreviation half of discussion #649: https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/649

**Stacked on #648** (this branch is built on top of that one, so the diff below includes its commit too — merge #648 first, this PR's diff then reduces to a single commit). The labeling-box exception from #649 is deliberately **not** part of this PR — that's a separate stacked follow-up on top of this one, kept independent so this smaller, self-contained piece can be reviewed and merged on its own.

## What this does

`%{structure_id}` (added in #648) now drops the `=`/`+` segments when they match the diagram's own title block plant/location — this is IEC 81346's own allowance for a reduced designation relative to a reference point, here the page. Because #648's fallback already makes "no override" resolve to the diagram's value, abbreviation becomes the automatic default for the common case (a device with no explicit plant/location of its own) with zero extra data entry. A device with an explicit, *different* plant/location still prints in full, so it's never mistaken for a local device.

`%{structure_id_full}` is a new token that always renders the complete, unabbreviated designation regardless of page context — for exports/BOM/nomenclature use where the canonical full designation is what's needed, independent of which page a device happens to be drawn on.

`CompositeTextEditDialog`'s variable picker gets the new token added alongside the existing one.

## Testing

Built clean (0 errors/warnings in the 2 changed files). Same pre-existing Qt5/Qt6 build-environment note as #648 applies here (this sandbox now has `qt6-base-dev` installed, which a *fresh* cmake configure mis-detects via the FetchContent-built KF5 addons) — worked around the same way, by reusing an already-Qt5-configured build tree.

Verified live via Xvfb against `examples/grafcet.qet`, folio Installation/Localisation set to `E1`/`A2`:
- Element with no plant/location of its own: `%{structure_id}` rendered **empty** (fully abbreviated — this grafcet element also has no label), `%{structure_id_full}` rendered `=E1+A2`.
- Element's own Installation set to `X9` (Localisation left empty): `%{structure_id}` rendered `=X9` (plant segment kept since it differs from the page; location segment still dropped since `A2` still matches the page), `%{structure_id_full}` rendered `=X9+A2`.

### DIFF
--- a/sources/autoNum/assignvariables.cpp
+++ b/sources/autoNum/assignvariables.cpp
@@ -210,11 +210,15 @@ namespace autonum
 		title block installation (=) / location (+) fields (IEC 81346)
 		when the element's own value is empty, so a project can set
 		these once at the folio level instead of on every element.
+		%{structure_id} is the resulting =plant+location-label designation,
+		reduced by dropping any =/+ segment that matches the diagram's own
+		title block (IEC 81346's allowance for a designation implied by
+		context); %{structure_id_full} is the same but never reduced.
 		@param formula
 		@param dc
 		@param elmt the element the formula belongs to, if any;
 		used to resolve the diagram fallback above and the
-		%{prefix}/%{structure_id} tokens. May be null.
+		%{prefix}/%{structure_id}/%{structure_id_full} tokens. May be null.
 		@return
 	*/
 	QString AssignVariables::replaceVariable(const QString &formula,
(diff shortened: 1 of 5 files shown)


ispyisail:
Part 2 of #649 is now up as #652 (the structure box itself), stacked on this one. This PR stays scoped to the page-match abbreviation.

ispyisail:

PR #652, opened 2026-08-03
Structure box: a labeling frame carrying its own IEC 81346 identity (discussion #649)

ispyisail:
Second half of discussion #649. **Stacked on #650** (page-match abbreviation), which is itself stacked on #648 — the diff shown here includes both until they merge.

## What it is

A **structure box** is a rectangle drawn on a folio purely for documentation. It carries its own plant (`=`) / location (`+`) / prefix (`-`) identity and has **no electrical meaning** — no terminals, no connectivity.

Elements visually inside it resolve `%{plant}` / `%{location}` / `%{structure_id}` against **the box** rather than the folio's title block. So a sub-assembly drawn on a shared page can carry its own designation without setting it on every device individually.

Resolution order is now: the element's own value → nearest enclosing box → the diagram's title block. Where boxes are nested, the **smallest (most specific)** one wins.

## Implementation notes

- Implemented as a `QetShapeItem` subclass fixed to `Rectangle`, so it reuses the existing resize-handle, move and selection behaviour unchanged.
- `DiagramEventAddShape` gained a `createShapeItem()` factory hook, so the new tool reuses all of its mouse handling instead of duplicating it.
- `buildStructureId()` was extracted from `replaceVariable()` so the box and the formula tokens assemble designations identically rather than by two parallel implementations.

## Three integration points a compile cannot catch

Worth calling out, because they're the ones a new `QGraphicsItem` type is easy to miss — all three were found by auditing dispatch sites rather than by the build:

1. **`DiagramContent`** switches on `item->type()`, and `StructureBoxItem` reports its own type rather than `QetShapeItem`'s — so boxes fell through into no bucket, leaving them **impossible to cut, copy, paste or delete**. Added to both the type switch and `hasDeletableItems()`.
2. **`PropertiesEditorFactory`** likewise had no case, so selecting a box left the selection-properties dock empty. It now returns the box's own editor — the useful thing to edit is its identity, not its pen and brush.
3. **`Diagram::toXml()`/`fromXml()`** persist boxes under a `structureBoxes` element, and they join `added_items` so paste positioning applies to them.

## Scope

Deliberately **excludes** discussion #617 (composite device grouping with real terminals and auto-wiring). That solves connectivity; this solves labeling. If #617 is ever built, its boundary could adopt the same reference-point mechanism — but they shouldn't block each other.

## Testing

Full Qt6 build, clean. **Also verified at runtime** on an isolated Xvfb display, driving the real GUI:

| # | Check | Result |
|---|---|---|
| 1 | Toolbar tool present, tooltip "Ajouter un cadre de repérage" | pass |
| 2 | Box draws at the intended size (click–move–click, like the other shape tools) | pass |
| 3 | Undo stack entry reads "Annuler Ajouter un cadre de repérage" | pass |
| 4 | Selection-properties dock shows Installation (=) / Localisation (+) / Préfixe (-) | pass — exercises the `PropertiesEditorFactory` case |
| 5 | Double-click opens the properties dialog | pass |
| 6 | Setting `E1` / `A2` / `K9` renders `=E1+A2-K9` on the box | pass — exercises `buildStructureId()` |
| 7 | Couper / Copier / Supprimer are enabled with a box selected | pass — exercises the `DiagramContent` fix |
| 8 | Supprimer actually removes the box | pass |
| 9 | Save writes `<structureBox location="A2" plant="E1" prefix="K9" …>` with pen, brush, geometry and z | pass |
| 10 | Reopening the saved file restores the box and its rendered identity | pass — full `toXml`/`fromXml` round-trip |

Checks 4, 7 and 8 are the ones that matter most, because they cover the three integration points above — those were written from reading dispatch sites, so they needed exercising rather than reasoning about.

### Still not covered

Being explicit about the gap rather than implying full coverage:

- **The core resolution path** — placing an element *inside* a box and confirming `%{structure_id}` resolves against the box rather than the title block. The plumbing either side of it is verified, but not that end-to-end behaviour.
- **Nested boxes** — that the smallest enclosing box wins.
- **Copy/paste round-trip** — the menu entries are enabled and delete works, but I didn't paste a box back.

Worth a reviewer's eye on those three, and on whether the draw-tool feel matches the other shape tools.

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -292,6 +292,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/diagramevent/diagrameventaddimage.h
   ${QET_DIR}/sources/diagramevent/diagrameventaddshape.cpp
   ${QET_DIR}/sources/diagramevent/diagrameventaddshape.h
+  ${QET_DIR}/sources/diagramevent/diagrameventaddstructurebox.cpp
+  ${QET_DIR}/sources/diagramevent/diagrameventaddstructurebox.h
   ${QET_DIR}/sources/diagramevent/diagrameventaddtext.cpp
   ${QET_DIR}/sources/diagramevent/diagrameventaddtext.h
   ${QET_DIR}/sources/diagramevent/diagrameventinterface.cpp
@@ -492,6 +494,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/qetgraphicsitem/qetgraphicsitem.h
   ${QET_DIR}/sources/qetgraphicsitem/qetshapeitem.cpp
   ${QET_DIR}/sources/qetgraphicsitem/qetshapeitem.h
+  ${QET_DIR}/sources/qetgraphicsitem/structureboxitem.cpp
+  ${QET_DIR}/sources/qetgraphicsitem/structureboxitem.h
   ${QET_DIR}/sources/qetgraphicsitem/qgraphicsitemutility.cpp
   ${QET_DIR}/sources/qetgraphicsitem/qgraphicsitemutility.h
   ${QET_DIR}/sources/qetgraphicsitem/reportelement.cpp
@@ -721,6 +725,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/ui/reportpropertiewidget.h
   ${QET_DIR}/sources/ui/shapegraphicsitempropertieswidget.cpp
   ${QET_DIR}/sources/ui/shapegraphicsitempropertieswidget.h
+  ${QET_DIR}/sources/ui/structureboxpropertieswidget.cpp
+  ${QET_DIR}/sources/ui/structureboxpropertieswidget.h
   ${QET_DIR}/sources/ui/thirdpartybinaryinstalldialog.cpp
   ${QET_DIR}/sources/ui/thirdpartybinaryinstalldialog.h
   ${QET_DIR}/sources/ui/titleblockpropertieswidget.cpp
--- a/sources/autoNum/assignvariables.cpp
+++ b/sources/autoNum/assignvariables.cpp
@@ -231,17 +231,52 @@ namespace autonum
 	/**
 		@brief AssignVariables::replaceVariable
 		Replace the variables in formula in form %{my-var}
-		to the corresponding value stored in dc
+		to the corresponding value stored in dc.
+		%{plant} and %{location} fall back to the containing diagram's
+		title block installation (=) / location (+) fields (IEC 81346)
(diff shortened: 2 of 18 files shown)


ispyisail:
Follow-up from trying to verify the core resolution path end-to-end at runtime. I couldn't complete that test, but the attempt turned up something worth recording — and it's a scope question rather than a defect.

## `%{structure_id}` and friends only work in dynamic texts, not label formulas

My first attempt set an element's **label formula** to `P=%{plant} L=%{location}`, placed the element inside a box, and reloaded. The saved result was:

```xml
<elementInformation name="formula" show="1">P=%{plant} L=%{location}</elementInformation>
<elementInformation name="label"   show="1">P=%{plant} L=%{location}</elementInformation>
```

— i.e. the tokens came through completely unsubstituted.

That isn't a bug in this PR. Tracing it:

- `Element::actualLabel()` → `AssignVariables::formulaToLabel()` → the `AssignVariables` constructor.
- `%{plant}` / `%{location}` / `%{structure_id}` / `%{structure_id_full}` are implemented in `AssignVariables::replaceVariable()`, which is a **separate** static function.
- `git grep replaceVariable` shows its only callers are `dynamicelementtextitem.cpp`, `dynamicelementtextmodel.cpp` and `partdynamictextfield.cpp` — all dynamic/composite text. Nothing in the label-formula path calls it.

So `enclosingStructureBox()` — and therefore this whole feature — is reachable **only** through a dynamic text in composite mode. That is consistent with #648, which added the token to `compositetexteditdialog.cpp` (the composite *text* editor), so I read it as intended scope rather than an oversight.

**Worth deciding explicitly**, though: a user who sees "Désignation IEC 81346 (=+-)" offered in the composite-text dropdown could reasonably expect `%{structure_id}` to work in a label formula too, and it silently passes through verbatim instead of erroring. Options are to document the limitation, or to route the label-formula path through `replaceVariable()` as well. I'd lean documenting for now, since changing the label path is a wider blast radius than this PR should carry.

## What I did not manage to verify

Being straight about it: I set up a controlled fixture (folio `FOLIO_P`/`FOLIO_L`, box `BOX_P`/`BOX_L`, element geometrically inside the box) intending to confirm the element resolves to the box's values rather than the folio's. I ran out of road on GUI automation — repeated `SingleApplication` instance-forwarding and display problems — before getting a clean reading through the composite-text path.

So **the box→element resolution itself is still only verified by code reading, not observed at runtime.** Everything in the results table in the description above stands; this specific assertion does not. Worth a reviewer with a real desktop spending two minutes on it: drop a dynamic text in composite mode with `%{structure_id}` on an element inside a box, and confirm it picks up the box.

ispyisail:

PR #654, opened 2026-08-03
Rotate crash-recovery backups instead of overwriting a single snapshot

ispyisail:
Closes/addresses discussion #598.

## Problem

`QETProject::writeBackup()` fires every 20 minutes and writes the
crash-recovery snapshot through a single `KAutoSaveFile`. If the
in-memory project is already bad when that tick lands, the one
recovery copy is overwritten with the bad state and there's nothing
left to roll back to (forum report: https://qelectrotech.org/forum/viewtopic.php?id=2477).

This is scoped to the crash-recovery path only, as agreed in the
discussion — autosave (the opt-in "Sauvegarde automatique des
projets" timer that writes straight to the real project file) is
untouched.

## Change

- `QETProject` now keeps `BackupGenerations` (3) rotating
  `KAutoSaveFile` snapshots in a `std::array`, written round-robin by
  `writeBackup()`. A bad write only clobbers the oldest generation,
  not the only one.
- `QETApp::checkBackupFiles()` groups the stale files it finds by the
  project they recover (there can now be up to 3 per crashed project)
  and shows a new `BackupRestoreDialog` letting the user pick which
  generation to reopen per project, defaulting to the most recent.
  Previously it would have reopened every stale file unconditionally,
  which — now that there are up to 3 per project — would have opened
  duplicate windows for the same crash.
- Added a Catch2 test proving `KAutoSaveFile::allStaleFiles()` already
  finds every generation for one managed file correctly; no changes
  were needed on the read/scan side, only on the write side and the
  restore UI.

## Not in scope (per discussion)

- Configurable generation count/interval.
- Full versioned save history.
- Any change to autosave's write-directly-to-the-real-file behavior.

## Test plan

- [x] New/existing Catch2 unit tests pass (`tests/catch`, both the
  `BUILD_WITHOUT_KF5` nokde `KAutoSaveFile` reimplementation and the
  default KF5-backed build)
- [x] Full app builds and links cleanly (`cmake --build`, default
  `BUILD_WITH_KF5=ON` config)
- [ ] Manual: crash QET mid-edit, confirm 3 rotating generations
  appear on relaunch and the dialog lets you pick one

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -677,6 +677,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/ui/diagramselection.h
   ${QET_DIR}/sources/ui/backupdialog.cpp
   ${QET_DIR}/sources/ui/backupdialog.h
+  ${QET_DIR}/sources/ui/backuprestoredialog.cpp
+  ${QET_DIR}/sources/ui/backuprestoredialog.h
   ${QET_DIR}/sources/ui/dialogwaiting.cpp
   ${QET_DIR}/sources/ui/dialogwaiting.h
   ${QET_DIR}/sources/ui/dynamicelementtextitemeditor.cpp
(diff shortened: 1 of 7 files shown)


ispyisail:

PR #655, opened 2026-08-03
Add configurable conductor end-cap style (square/round/flat)

ispyisail:
Builds discussion #616.

## Problem

Conductors (wires) are always drawn with a flush/square end — there's no way to control how a wire's stroke is rendered at its endpoint (square/flush, rounded, or projecting square past the point). `Conductor::conductor_pen` hardcodes `Qt::SquareCap` (`sources/qetgraphicsitem/conductor.cpp`) and `Conductor::paint()` never overrides it, even though every other pen attribute on the same `QPen` — line style, thickness, color — is already exposed as a user-configurable `ConductorProperties` field with a dialog control.

## Change

- Added a `Qt::PenCapStyle cap_style` field to `ConductorProperties`, alongside the existing `Qt::PenStyle style` field. Default is `Qt::SquareCap`, matching current behavior.
- Persisted through the same CSS-like `"style"` XML attribute / `QSettings` key that line style already uses (`readStyle`/`writeStyle` in `conductorproperties.cpp`) — e.g. `line-style: dashed; cap-style: round;`. A conductor using only defaults still produces an empty `writeStyle()`, so existing `.qet` files get no new attribute and round-trip unchanged.
- Added a combo box (Square/Round/Flat) next to the existing line-style combo in `ConductorPropertiesWidget`'s "Apparence" tab, wired the same way.
- `Conductor::paint()` now applies `m_properties.cap_style` onto the pen instead of leaving the hardcoded static value untouched.

## Why this shape

Same class, same dialog, same XML/QSettings round-trip pattern as line style — a small, idiomatic extension rather than new architecture. Only changes how the stroke terminates at both ends of the conductor line itself; doesn't touch terminal/junction rendering (junction dots are filled ellipses, unaffected by cap style) or the connection model.

## Not in scope (per discussion)

- Any marker (dot/tick/arrowhead) at the wire-terminal connection point — a different, unimplemented rendering concept.
- Dangling/unconnected conductor ends — conductors always connect exactly two terminals today; unchanged.

## Test plan

- [x] Full project builds and links clean (`cmake --build`, default `BUILD_WITH_KF5=ON` config), zero new warnings in the touched files
- [x] Verified by construction: default properties (`Qt::SolidLine` + `Qt::SquareCap`) produce an empty `writeStyle()`, so pre-existing `.qet` files are unaffected
- [ ] Manual: open the conductor properties dialog, switch cap style, confirm the wire endpoint changes visually and survives save/reload

### DIFF
--- a/sources/conductorproperties.cpp
+++ b/sources/conductorproperties.cpp
@@ -250,7 +250,8 @@ ConductorProperties::ConductorProperties() :
 	horiz_rotate_text(0),
 	m_show_text(true),
 	m_one_text_per_folio(false),
-	style(Qt::SolidLine)
+	style(Qt::SolidLine),
+	cap_style(Qt::SquareCap)
 {}
 
 /**
@@ -772,6 +773,7 @@ bool ConductorProperties::operator==(const ConductorProperties &other) const
 		other.m_color_2 == m_color_2 &&\
 		other.m_dash_size == m_dash_size &&\
 		other.style == style &&\
+		other.cap_style == cap_style &&\
 		other.text == text &&\
 		other.text_color == text_color &&\
 		other.m_formula == m_formula &&\
@@ -807,6 +809,7 @@ bool ConductorProperties::operator!=(const ConductorProperties &other) const{
 */
 void ConductorProperties::readStyle(const QString &style_string) {
 	style = Qt::SolidLine; // style par defaut
+	cap_style = Qt::SquareCap; // style d'extremite par defaut
 
 	if (style_string.isEmpty()) return;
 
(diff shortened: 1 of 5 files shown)


scorpio810:
Thanks for the PR, the code is clean and follows the existing style pattern well (XML/QSettings/dialog).

That said, I tested locally (changing the cap style in the properties dialog) and I don't see any visual difference on screen. Digging into the code:

cond_size defaults to 1 (conductorproperties.cpp), so at that line width the difference between SquareCap/RoundCap/FlatCap is sub-pixel — imperceptible without zooming in a lot.
Antialiasing is explicitly disabled in Conductor::paint(), which flattens the difference even further at this scale.

So I'm wondering about the practical value for end users: in typical QET usage (thin lines), the change is basically invisible. Would it be worth:

adding an example/screenshot with a thicker cond_size in the PR description, to actually show the intended effect,
and/or documenting in the combo box tooltip that the effect only becomes visible above a certain line width?

Otherwise the implementation logic looks solid to me.

PR #657, opened 2026-08-04
Consolidate QETDiagramEditor's ~60 QActions into a DiagramEditorActions pool

ispyisail:
Builds discussion #603.

## Problem

`QETDiagramEditor` declared roughly fifty `QAction`/`QActionGroup` members directly on itself, constructed inline across ~32 `new QAction(...)` sites in `setUpActions()`. A chunk of these had to be `public:` specifically so `DiagramView` — a separate, unrelated class — could reach into them (`qde->m_cut`, `qde->m_row_column_actions_group.actions()`, `qde->m_depth_action_group->actions()`, etc.) to build its own context menu. There was no single place to answer "what actions does this editor have," and the public visibility existed only for that one sibling class's benefit.

## Change

- New `DiagramEditorActions` class (`sources/diagrameditoractions.{h,cpp}`): owns every action that used to live on `QETDiagramEditor`. One instance is constructed per `QETDiagramEditor` window — **not** a singleton, since multiple editor windows can be open at once (tabbed or MDI), each needing its own independent action state tied to that window's current selection/diagram.
- `QETDiagramEditor::setUpActions()` is now a one-line delegating call (`m_actions = new DiagramEditorActions(this);`); the ~500-line body moved into the new class's constructor.
- `QETDiagramEditor::actions()` is a new public accessor. `DiagramView::contextMenuActions()` and `TerminalStripEditorWindow` (both already reaching into `QETDiagramEditor` — the latter as an existing `friend class`) now go through it instead of touching former-public members directly.
- `qetdiagrameditor.h` shrinks by the entire ~60-member declaration block, replaced by a single `DiagramEditorActions *m_actions` pointer.

## Scope decision: DiagramView's own 3 actions stay put

`DiagramView` owns three actions of its own (paste-here, multi-paste, create-template) that the discussion proposed folding into the same pool. I didn't: a `QETDiagramEditor` can have many `DiagramView` tabs open via MDI, and each currently owns an *independent* instance of those three — including per-view mouse-position state (`m_paste_here_pos`) that a single shared per-editor action can't correctly represent once more than one tab is open. These three were never part of the reach-through problem this PR fixes (they're already private to `DiagramView`), so folding them in would add real regression risk for a multi-tab paste-position bug, for no benefit toward the stated goal.

## Bug this refactor would otherwise have introduced

`QETDiagramEditor::~QETDiagramEditor()` was empty, so the new `m_actions` pointer was never deleted. Several `QActionGroup` members inside `DiagramEditorActions` are **value** members parented to the editor (`QActionGroup(editor)`). The original code relied on them being direct data members of `QETDiagramEditor` itself, so C++'s member-destruction order guaranteed they'd destruct — and self-unregister from Qt's parent/child list — *before* `QETDiagramEditor`'s own `QObject::~QObject()` ran. With them now living inside a separately-heap-allocated, never-destroyed `DiagramEditorActions` block, closing an editor window would eventually have made `QObject::~QObject()` call `delete` on subobjects that were never individually `new`'d — heap corruption, not just a leak. Fixed with an explicit `delete m_actions;` in the destructor (verified this is the sole destruction pathway: every `QETDiagramEditor` construction site in the codebase relies solely on `WA_DeleteOnClose`, set in `closeEvent()`, for cleanup).

## Test plan

- [x] Existing Catch2 test suite passes (doesn't cover `QETDiagramEditor`/`DiagramView` directly, general regression check only)
- [x] Launched the app headlessly under Xvfb: starts cleanly, fully builds all menus/toolbars/actions (exercising every line of the new constructor and every relocated call site), correctly shows the crash-recovery dialog
- [ ] Could not complete in this environment: a live click-through window close (this sandbox has no window manager, so X11 focus/`_NET_ACTIVE_WINDOW` requests fail) — the destructor fix is verified by direct reasoning against C++'s guaranteed destruction order instead, see above
- [ ] Manual: open several diagrams in separate tabs/windows, confirm toolbar/menu actions (undo/redo, cut/copy/paste, zoom, add-item tools, etc.) all still work and each window's state (e.g. auto-conductor checkbox) stays independent


ispyisail:
Merged current master in to resolve the conflicts from the ~60-QAction drift since this branch was cut (mainly `m_find`/`m_jump_to_element` from #775, plus `m_auto_break_conductor` and `m_rotate_group_selection`, which are now properly owned by `DiagramEditorActions`). Verified live: app launches clean, both rotate modes work correctly and distinctly, undo restores exactly, and the auto-break-conductor toolbar toggle works. Ready for review.

ispyisail:

PR #659, opened 2026-08-04
Preserve master/slave links when pasting or duplicating a folio

ispyisail:
Builds discussion #607.

## Problem

Cutting/copying a linked group of elements (a relay coil with its contacts, a PLC master with its slave I/O elements, etc.) — even the whole group together in one paste — currently drops the master/slave link entirely. The pasted copies come in as ordinary, independent elements with no link and no warning; the user has to notice and manually re-link them afterward.

## Traced end to end

- `Element::toXml()` writes each linked partner's uuid into a `<link_uuid>` entry — the clipboard XML genuinely contains the link information.
- `Element::fromXml()` reads that back — but only into a **deferred, unresolved** buffer, `tmp_uuids_link`. No linking happens at this point.
- The only code that ever resolves `tmp_uuids_link` into a real link is `Element::initLink(QETProject *prj)`, which builds an `ElementProvider` and resolves the pending uuids project-wide.
- `initLink()` is only ever called from `Diagram::refreshContents()`, which in turn is only called from two places: full project load, and macro-block insertion.
- `DiagramView::paste()` calls `Diagram::fromXml()` directly and never calls `refreshContents()` or `initLink()` anywhere in that path. Pasted elements' `tmp_uuids_link` gets populated correctly, but nothing ever resolves it — the link is silently dropped.

This is a different, narrower problem from the uuid-*collision* issue already fixed for whole-folio duplication (an element keeping its source uuid as the project database's primary key) — that's a separate bug on a separate code path.

## Fix

Added `Element::initLink(const QList<Element *> &candidates)` — resolves pending links against a caller-supplied candidate list instead of a project-wide `ElementProvider` search.

The scoping matters and is the subtle part: right after an XML round-trip and *before* `PasteDiagramCommand::redo()`'s existing `newUuid()` loop renews each pasted element's identity, a pasted element's `tmp_uuids_link` still holds its source's original partner uuid — which at that exact moment still equals the not-yet-renewed uuid of that partner's own pasted copy, if it was pasted too. Resolving only within the pasted batch (not the whole project) is what stops a linked pair being pasted together from matching against an original element left elsewhere in the project that happens to still carry that same soon-to-be-replaced uuid. If only one half of a linked group is pasted, its entry simply finds no match in the batch and is dropped — the same "leave it unlinked" outcome as today.

Wired into `PasteDiagramCommand::redo()`, right before the existing `newUuid()` loop, gated by the same `first_redo` flag so undo/redo cycles don't re-trigger either.

## Also fixed: the same bug on the folio-duplication path

Tracing this turned up the identical gap in `ElementsPanelWidget::duplicateDiagram()`: same XML round-trip, same per-element `newUuid()` renewal loop (added by an earlier commit to fix the separate uuid-collision issue), but nothing ever resolving `tmp_uuids_link` there either. Same root cause, same fix, reusing the new overload — flagging this clearly since it's beyond discussion #607's literal scope (cut/copy/paste), not silently bundled in.

## Test plan

- [x] Existing Catch2 test suite passes (general regression check; no existing coverage of linked-element paste specifically)
- [ ] Manual: link a master element to a slave, select both, copy, paste — confirm the pasted copies are linked to each other (not to the originals, and not silently dropped); paste only the master (partner not copied) — confirm it comes in unlinked, same as today
- [ ] Manual: duplicate a folio containing a linked master/slave pair — confirm the duplicated copies are linked to each other

### DIFF
--- a/sources/diagramcommands.cpp
+++ b/sources/diagramcommands.cpp
@@ -77,6 +77,20 @@ void PasteDiagramCommand::redo()
 
 		//this is the first paste, we do some actions for the new element
 		const QList <Element *> elmts_list = content.m_elements;
+
+			//Resolve master/slave links within the pasted batch before any
+			//uuid gets renewed below: each element's tmp_uuids_link still
+			//holds its source's original partner uuid at this point, which
+			//still equals the not-yet-renewed uuid of that partner's own
+			//pasted copy, if it was pasted too. Scoping the search to
+			//elmts_list (rather than the whole project) is what keeps this
+			//from matching an original element left elsewhere in the
+			//project that happens to share that same soon-to-be-replaced
+			//uuid. See Element::initLink(const QList<Element *> &).
+		for (Element *e : elmts_list) {
+			e->initLink(elmts_list);
+		}
+
 		for (Element *e : elmts_list)
 		{
 			//make new uuid, because old uuid are the uuid of the copied element
--- a/sources/elementspanelwidget.cpp
+++ b/sources/elementspanelwidget.cpp
@@ -642,16 +642,32 @@ void ElementsPanelWidget::duplicateDiagram()
 
 		new_diagram->fromXml(diagram_elmt, QPointF(0, 0), false, nullptr);
 
+		QList<Element *> new_elements;
 		for (QGraphicsItem *item : new_diagram->items()) {
 			if (Element *elmt = dynamic_cast<Element *>(item)) {
-				// The XML round-trip kept the source elements' uuids. Give the
-				// copies their own identity (as PasteDiagramCommand::redo()
-				// does for on-diagram paste): element.uuid is the PRIMARY KEY
-				// of the project database, so duplicates fail to insert and
-				// silently vanish from nomenclature/summary tables.
-				elmt->newUuid();
-				new_diagram->restoreText(elmt);
+				new_elements << elmt;
 			}
+		}
+
+			// Resolve master/slave links within the duplicated batch before
+			// any uuid gets renewed below, for the same reason
+			// PasteDiagramCommand::redo() does: each element's
+			// tmp_uuids_link still holds its source's original partner
+			// uuid at this point, matching the not-yet-renewed uuid of
+			// that partner's own copy in new_elements. See
+			// Element::initLink(const QList<Element *> &).
+		for (Element *elmt : new_elements) {
+			elmt->initLink(new_elements);
+		}
+
+		for (Element *elmt : new_elements) {
+			// The XML round-trip kept the source elements' uuids. Give the
+			// copies their own identity (as PasteDiagramCommand::redo()
+			// does for on-diagram paste): element.uuid is the PRIMARY KEY
+			// of the project database, so duplicates fail to insert and
+			// silently vanish from nomenclature/summary tables.
+			elmt->newUuid();
+			new_diagram->restoreText(elmt);
 		}
 	}
 
--- a/sources/elementspanelwidget.cpp
+++ b/sources/elementspanelwidget.cpp
@@ -26,7 +26,9 @@
 #include "titleblock/templatedeleter.h"
 #include <QFileInfo>
 #include <QMessageBox>
+#include <QSettings>
 #include "qetgraphicsitem/element.h"
+#include "qetgraphicsitem/conductor.h"
 
 /*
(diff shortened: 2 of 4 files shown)


ChuckNr11:
Hi ispyisail,
wenn ich es richtig verstanden habe möchtest Du beim Kopieren und Einfügen von querverweisrelevanten Elementen die Verknüpfung zwischen z.B. Spule und Kontakt erhalten. Ich denke das ist nicht der richtige Weg.
Die Verbindung zwischen einer Spule und einem Kontakt muss eindeutig sein, sie kann und darf nicht doppelt vorkommen. Irgendetwas an der Verbindung muss geändert werden, entweder eine andere Spule oder ein anderer Kontakt (z.B. Kontaktnummern ändern). Da QElectrotech nicht auf doppelte oder nicht eindeutige Verbindungen prüft, ist das Löschen eine gute Möglichkeit dafür zu sorgen, dass der Benutzer diese Verbindung korrekt wiederherstellt.
Wäre schön, wenn du das berücksichtigst
Danke Gruß Achim

with translater:

If I understand correctly, you want to preserve the link between—for example—a coil and a contact when copying and pasting elements involved in cross-references. I don't think that is the right approach.
The connection between a coil and a contact must be unique; it cannot and must not be duplicated. Something about the connection needs to be changed—either the coil or the contact (e.g., by changing contact numbers). Since QElectrotech does not check for duplicate or ambiguous connections, deleting the link is a good way to ensure the user re-establishes it correctly.
It would be great if you could take this into account.
Thanks, regards, Achim

scorpio810:
https://qelectrotech.github.io/qelectrotech-doc/_images/qet_configure_project.png
See: checkox m_save_label_paste "Ne pas conserver les labels des éléments lors des copier coller" generalconfigurationpage.ui

ispyisail:
Thanks — that is the right thing to point at, and it resolves most of @ChuckNr11's objection. Pushed `abb9438` acting on it.

## Why that preference matters here

`diagramcommands/erase-label-on-copy` is what makes preserving links safe on the paste path. At its default, `PasteDiagramCommand::redo()` (`diagramcommands.cpp:85-93`) clears label, formula, comment and location on every pasted element, so the copies arrive with **no designation at all** — there is nothing for two linked groups to be ambiguous about.

| setting | paste result | ambiguity |
|---|---|---|
| **true** (default) | label / formula / comment / location cleared | none |
| false | designations preserved | real — two identically designated groups |

## The gap it exposed

`ElementsPanelWidget::duplicateDiagram()` never consulted that preference. A duplicated folio always kept the source designations, and once this branch preserves the links too, that gives two identically designated linked groups — a genuinely ambiguous drawing rather than merely untidy.

`abb9438` makes duplication clear the same four fields paste clears, plus conductor texts, under the same preference. Anyone who deliberately turns it off keeps the old behaviour on both paths, which is the point of having it.

## On the original objection

@ChuckNr11's concern was right, but about labels rather than links. The stated mechanism does not apply:

- A duplicated link is not possible — `SlaveElement::linkToElement()` (`slaveelement.cpp:51-64`) calls `unlinkAllElements()` first, so a contact can only ever hold one master.
- The copies link **to each other**, not to the originals: `Element::initLink(const QList<Element*>&)` searches only the pasted batch.
- What could actually be ambiguous was the duplicate *designation*, and that is now handled on both paths.

Worth noting there is **no duplicate-designation validation anywhere in `sources/`** — an ambiguous drawing is possible today by other routes. That feels like it deserves its own issue rather than being smuggled in here.

## One thing this still does not solve

Masters and slaves usually live on **different folios**, and a selection cannot span folios while a paste lands on one. So a cross-referenced group spread across folios still cannot be copied and pasted as a unit — that would need a copy-everything-then-map-on-paste UI, which is a much larger feature and not something I am proposing here.

What this branch does is make the cases that *are* expressible today behave correctly: a linked group inside one folio, and a duplicated folio.

## Third interaction, flagged not fixed

`MultiPasteDialog` reads the same setting and, when auto-numbering is ticked, force-writes it to `false` (`multipastedialog.cpp:113`) so the auto-num pass has formulas to work against, restoring it at `:217`. With links preserved, a multi-paste with auto-num produces linked copies carrying the source labels which are then renumbered — arguably the ideal outcome, but I have not tested it and did not want to change that path silently.

(I did check the save/restore cannot strand the preference: the only `return` between those lines is inside a sort comparator.)

ispyisail:

PR #660, opened 2026-08-04
Add "rotate group" to actually rotate a selection as a whole

ispyisail:
Builds discussion #618.

## Problem

There's already a "rotate selection" action (Space), but tracing its actual math confirms it doesn't do what "rotating a group" usually means — it spins each item in place individually rather than rotating the group as a whole around a shared point.

`RotateSelectionCommand` does exactly one thing per selected item: `new QPropertyUndoCommand(item, "rotation", item->rotation(), item->rotation()+angle, this)`. `QGraphicsItem::setRotation()` rotates an item around its own local transform origin and leaves its scene `pos()` completely untouched. Select three elements arranged left-to-right in a row and rotate 90° — each one spins in place individually, but all three stay in their original row positions. The group doesn't become a vertical column.

## Change

- Added a `rotate_as_group` parameter to `RotateSelectionCommand` (default `false`, so the existing action and its one call site are fully unchanged). When set, it computes a shared pivot once up front — the bounding-box center of the whole selection — and queues a second, parallel `"pos"` `QPropertyUndoCommand` alongside the existing `"rotation"` one, rotating each item's position around that pivot by the same angle. Standard rotate-a-point-around-a-center math, verified to match `QGraphicsItem::setRotation()`'s clockwise-positive convention so a group rotation turns the same direction as each item's own spin.
- **Scoped the position change to `Element`/`IndependentTextItem`/`DiagramImageItem` only** — these are the only selectable types with scene-space `pos()`. `ConductorTextItem`, `DynamicElementTextItem`, and `ElementTextItemGroup` are all parent-relative children (confirmed by reading their constructors — `DynamicElementTextItem::setParentItem(parent_element)`, `ConductorTextItem`'s `DiagramTextItem(parent_conductor)`), so when their owning `Element` is also selected and gets its own `pos()` rotated, they're carried along for free by Qt's normal parent/child transform propagation — exactly what the existing "skip rotation if parent is also selected" guard for those three cases already assumes.
- Exposed as a new, separate action ("Pivoter le groupe", Shift+Space) next to the existing one, rather than changing Space's behavior — some workflows may already rely on the current per-item rotation.

## Test plan

- [x] Existing Catch2 test suite passes (general regression check; no existing coverage of selection rotation specifically)
- [ ] Manual: select 3+ elements arranged in a row, trigger "Pivoter le groupe" (Shift+Space), confirm the group becomes a column (not each element spinning in place); undo, confirm it cleanly reverts both rotation and position
- [ ] Manual: select an element together with one of its own dynamic text items, group-rotate, confirm the text item stays correctly attached (carried by the parent's transform, not independently repositioned)

### DIFF
--- a/sources/qetdiagrameditor.cpp
+++ b/sources/qetdiagrameditor.cpp
@@ -628,6 +628,7 @@ void QETDiagramEditor::setUpActions()
 		//Selections Actions (related to a selected item)
 	m_delete_selection     = m_selection_actions_group.addAction( QET::Icons::EditDelete,        tr("Supprimer")                 );
 	m_rotate_selection     = m_selection_actions_group.addAction( QET::Icons::TransformRotate,   tr("Pivoter")                   );
+	m_rotate_group_selection = m_selection_actions_group.addAction( QET::Icons::TransformRotate, tr("Pivoter le groupe")         );
 	m_rotate_texts         = m_selection_actions_group.addAction( QET::Icons::ObjectRotateRight, tr("Orienter les textes")       );
 	m_find_element         = m_selection_actions_group.addAction( QET::Icons::ZoomDraw,          tr("Retrouver dans le panel")   );
 	m_edit_selection       = m_selection_actions_group.addAction( QET::Icons::ElementEdit,       tr("Éditer l'item sélectionné") );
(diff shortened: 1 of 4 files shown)


ispyisail:
Pushed a follow-up commit fixing a real problem @plc-user spotted in [discussion #618](https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/618#discussioncomment-17924744): he hit the same thing rotating graphical primitives in the Element Editor, and warned that including the bounding box in the rotation calculation leaves parts at positions with many decimal places that can't be corrected from the UI afterwards.

It applied directly here. Reproduced on the diagram side: with a selection whose bounding box comes from font metrics and pen widths, the pivot was (132.67, 101.11), and an element sitting exactly on the grid at x=100 landed at **x=133.78** — permanently off-grid, with four successive 90° rotations failing to return it:

```
 90° -> (133.78000000000003, 68.43999999999998)
...
360° -> (99.99999999999999,  99.99999999999999)
```

Worth noting precisely which half does the damage: positions are saved via `QString::number()` (`%.6g`), so the floating-point noise is trimmed on write — but the 133.78 offset is not. The element really has moved off the grid, and no further rotation repairs it.

Two changes:

1. **Snap the pivot** with `Diagram::snapToGrid()`, which also respects the user's configured X/Y grid instead of assuming the 10 px default.
2. **Exact arithmetic for multiples of 90°** instead of `qCos`/`qSin`. Both rotate actions only ever pass right angles, and `qCos(90°)` is `6.12e-17`, not 0 — so the trig path was contributing error for no benefit. A quadrant is just an axis swap. Trig is kept as the fallback for any future arbitrary angle.

Verified against the real Qt types rather than on paper:

```
raw bbox centre QPointF(132.67,101.11) -> snapped pivot QPointF(130,100)
  90 deg -> QPointF(130,70)
 180 deg -> QPointF(160,100)
 270 deg -> QPointF(130,130)
 360 deg -> QPointF(100,100)
exact round-trip to (100,100): true
still on the 10 grid: true
45 deg (trig fallback): QPointF(108.787,78.7868)
```

Builds clean (CMake/Ninja Release, Qt 5.15.18).

plc-user:
In the meantime I tried your changes.
As you already said: Rotating four times by 90 degrees does not land at the starting-position. But I think we can live with that, because we can move the selected parts in the UI to the right position. Or user can press "Ctrl+z" to rotate back. I'll now merge this PR!

PR #661, opened 2026-08-04
Fix command-line tools hanging forever on a modal message box

ispyisail:
## The bug

`qelectrotech --resave examples/schema_indus.qet out.qet` never returns.

It isn't slow. Ten minutes of wall clock consumed **0.16 seconds of CPU** — it's blocked, not working. The GUI opens the very same project without complaint, so the file is fine and the fault is in the headless path.

A backtrace of the stuck process says exactly what's wrong:

```
main
  CLIExport::run
      QETProject::openFile(QFile*)
          QET::QetMessageBox::warning(...)
            QDialog::exec()          <- waits forever
```

`examples/schema_indus.qet` records `version="0.3"`, so loading it raises the *"partially compatible with your version"* warning in `readProjectXml()`. Interactively somebody presses **Open**; with no display nobody can, and `exec()` spins its event loop indefinitely.

The version warning is just the box an example file happens to trigger — **any** modal reachable during a load does this, so anyone scripting the CLI over a directory of older projects hangs on the first one.

## The fix

Fixed at the wrapper that all 52 call sites already funnel through, rather than at the one warning, so the whole class is closed rather than one symptom.

`QET::QetMessageBox` gains a non-interactive mode. When on, the four functions write the message to stderr and return an answer instead of constructing a dialog. `main.cpp` turns it on in the CLI branch, right beside the existing `QETProject::setBackupEnabled(false)` — which is there for the same underlying reason, that this path has no user and no event loop of its own.

The answer is chosen as:

1. the caller's `defaultButton`, when it named one and it's on offer;
2. otherwise the first "carry on" button offered — `Ok`, `Open`, `Yes`, `Save`, `Apply`…;
3. otherwise the first button set at all.

Both warnings in `readProjectXml()` offer `Open | Cancel` and abort the load on `Cancel`, so they resolve to `Open` and the project loads — which is what a batch invocation wants. The message text still reaches the user on stderr, where previously it was lost inside a dialog nobody could see.

**GUI behaviour is unchanged.** The flag defaults to `false` and is set in exactly one place — the command-line branch of `main()` — so no interactive path can reach it.

## Verification

- `schema_indus.qet`: **hangs indefinitely → resaves in 0.3s**, exit 0, valid output
- All **23** shipped example projects now complete a double `--resave`, with element / conductor / terminal counts and the full uuid set intact
- Builds clean, no new warnings; Catch2 suite passes

Found while building a save-determinism corpus over `examples/`, which is also how I know the 23-project sweep is clean — before this, one project could not be checked at all because the tooling hung on it.

## Test plan

- [x] The previously-hanging project resaves
- [x] All 23 examples double-resave with contents preserved
- [x] Builds clean, unit tests pass
- [ ] Manual: confirm message boxes still appear normally in the GUI (unchanged by construction — flag is CLI-only)

### DIFF
--- a/sources/main.cpp
+++ b/sources/main.cpp
@@ -19,6 +19,7 @@
 #include "machine_info.h"
 #include "qet.h"
 #include "qetapp.h"
+#include "qetmessagebox.h"
 #include "qetproject.h"
 #include "singleapplication.h"
 #include "utils/qetsettings.h"
@@ -249,6 +250,11 @@ QGuiApplication::setHighDpiScaleFactorRoundingPolicy(QetSettings::hdpiScaleFacto
 			// runs on a background thread referencing the project and races the
 			// process exit (intermittent segfault in QET::writeToFile).
 			QETProject::setBackupEnabled(false);
+			// Answer message boxes instead of showing them: opening a project
+			// saved by an older QElectroTech raises a warning from
+			// QETProject::readProjectXml(), and with nobody able to dismiss it
+			// QDialog::exec() would spin its event loop forever.
+			QET::QetMessageBox::setNonInteractive(true);
 			return CLIExport::run(export_app.arguments());
 		}
 	}
--- a/sources/qetmessagebox.cpp
+++ b/sources/qetmessagebox.cpp
@@ -17,6 +17,84 @@
 */
 #include "qetmessagebox.h"
 
+#include <QTextStream>
+
+namespace {
+	bool g_non_interactive = false;
+
+	/**
+		@brief autoAnswer
+		Report a message box on stderr and pick an answer, for use when there
+		is no user to click anything. @see QET::QetMessageBox::setNonInteractive
+		@param severity : short word naming the kind of box, for the log line
+		@param title
+		@param text
+		@param buttons : the buttons the caller offered
+		@param defaultButton : the caller's preferred answer, may be NoButton
+		@return the button to report as pressed
+	*/
+	QMessageBox::StandardButton autoAnswer(
+			const char *severity,
+			const QString &title,
+			const QString &text,
+			QMessageBox::StandardButtons buttons,
+			QMessageBox::StandardButton defaultButton)
+	{
+			//Honour the caller's own default when it named one.
+		if (defaultButton != QMessageBox::NoButton
+			&& (buttons & defaultButton)) {
+			QTextStream(stderr) << severity << ": " << title << " -- " << text
+								<< "\n(no display: answered with the caller's default button)\n";
+			return defaultButton;
+		}
+
+			//Otherwise prefer a "carry on" answer over one that cancels, so a
+			//batch run completes rather than silently doing nothing.
+		static const QMessageBox::StandardButton preference[] = {
+			QMessageBox::Ok,     QMessageBox::Open,   QMessageBox::Yes,
+			QMessageBox::Save,   QMessageBox::Apply,  QMessageBox::YesToAll,
+			QMessageBox::Retry,  QMessageBox::Ignore, QMessageBox::Close
+		};
+		for (auto candidate : preference) {
+			if (buttons & candidate) {
+				QTextStream(stderr) << severity << ": " << title << " -- " << text
+									<< "\n(no display: continuing)\n";
+				return candidate;
+			}
+		}
+
+			//Nothing affirmative on offer -- fall back to whatever is set.
+		for (int bit = QMessageBox::Ok; bit <= QMessageBox::RestoreDefaults; bit <<= 1) {
+			auto candidate = static_cast<QMessageBox::StandardButton>(bit);
+			if (buttons & candidate) {
+				QTextStream(stderr) << severity << ": " << title << " -- " << text
+									<< "\n(no display: answered automatically)\n";
+				return candidate;
+			}
+		}
+
+		QTextStream(stderr) << severity << ": " << title << " -- " << text
+							<< "\n(no display: no button offered)\n";
+		return QMessageBox::NoButton;
+	}
+}
+
+/**
+	@brief QET::QetMessageBox::setNonInteractive
+	@param non_interactive
+*/
+void QET::QetMessageBox::setNonInteractive(bool non_interactive) {
+	g_non_interactive = non_interactive;
+}
+
+/**
+	@brief QET::QetMessageBox::isNonInteractive
+	@return true when message boxes are answered without a user
+*/
+bool QET::QetMessageBox::isNonInteractive() {
+	return g_non_interactive;
+}
+
 /**
 	@see Documentation Qt pour QMessageBox::critical
 */
@@ -27,6 +105,9 @@ QMessageBox::StandardButton QET::QetMessageBox::critical (
 		QMessageBox::StandardButtons buttons,
 		QMessageBox::StandardButton defaultButton)
 {
+	if (g_non_interactive) {
+		return autoAnswer("Critical", title, text, buttons, defaultButton);
+	}
 #ifdef Q_OS_MACOS
 	QMessageBox message_box(
 				QMessageBox::Critical,
@@ -59,6 +140,9 @@ QMessageBox::StandardButton QET::QetMessageBox::information(
 		QMessageBox::StandardButtons buttons,
 		QMessageBox::StandardButton defaultButton)
 {
+	if (g_non_interactive) {
+		return autoAnswer("Information", title, text, buttons, defaultButton);
+	}
 #ifdef Q_OS_MACOS
 	QMessageBox message_box(
 				QMessageBox::Information,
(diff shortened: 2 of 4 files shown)


ispyisail:
Still current, and re-verified today against master `610001a84`.

`examples/schema_indus.qet` (version 0.3) hangs **every** headless verb, not just `--resave` — all eight time out at 60 s on an unpatched build, and all eight complete with this PR's approach applied:

| verb | without | with |
|---|---|---|
| `--info`, `--resave` | HUNG | ok |
| `--export-bom` / `-nets` / `-links` | HUNG | ok |
| `--export-pdf` / `-svg` / `-png` | HUNG | ok |

Controls: a normal project (`examples/741.qet`) round-trips with its conductor count unchanged, and the GUI still shows every dialog — only the CLI branch switches the mode.

Two notes for whenever this gets a look:

**One call site still bypasses the wrapper.** `QETProject::readProjectXml()`'s neighbourhood has a raw `QMessageBox::warning` on the element-load path (the "Système de contacts modifié" prompt, `qetproject.cpp`). It is reachable during a load, so it can still hang a headless run. Routing it through `QET::QetMessageBox::warning` closes the last hole — one line, happy to push it here if you'd like.

**This covers confirmation dialogs only, which I think is the right scope.** Dialogs that gather a *parameter* have no sensible unattended answer — suppressing them just yields "cancelled". Those need the parameter lifted into the API instead, which is what #752 does for `RotateTextsCommand`. There are very few of them, so the two approaches together close the class.

#737 fixes the same hang at the single `readProjectXml()` site; this PR supersedes it, and I'm happy to close #737 if you prefer this one.

ispyisail:
Pushed the one-line follow-up (`9285d12`): the "Système de contacts modifié" warning in `QETProject::addElement()`'s Erase branch was calling `QMessageBox::warning` directly, so it bypassed the wrapper and the new guard. It is reachable during a load, which is the path this PR exists to unblock, so it could still hang a headless run. That was the last one.

Interactive behaviour is unchanged; unattended it answers with the `Yes` the call site already passes as its default — the same "continue" the old code took.

Rebuilt and re-checked on this branch: `schema_indus.qet` completes on `--info`, `--resave` and `--export-bom`, with `(no display: continuing)` on stderr, and `741.qet` still round-trips with 67 conductors.

I've also closed #737 in favour of this PR.

ispyisail:
A concrete instance of this turned up while building an automated regression
sweep, which I think makes the case better than my original description did:
**a project shipped in this repository's own `examples/` directory cannot be
processed by any command-line verb.**

`examples/schema_indus.qet` declares `version="0.3"`, so it trips the
`m_project_qet_version <= versionZeroDotSix()` warning in
`QETProject::readProjectXml()` (`sources/qetproject.cpp:1494`). Under
`-platform offscreen` there is nobody to press Open or Cancel, so `exec()`
blocks forever:

```bash
timeout 30 qelectrotech -platform offscreen \
  --resave examples/schema_indus.qet /tmp/out.qet
# exit 124 -- hangs until killed. No output, no error.

timeout 30 qelectrotech -platform offscreen \
  --resave examples/741.qet /tmp/out.qet
# exit 0 in under a second (version 0.90)
```

Same binary (master `7307a59c1`), same flags — so it is the version gate, not
the file's size or content. It is the only pre-0.6 file in `examples/`.

Three things worth noting:

- **Every verb is affected**, not just `--resave`. The modal is in the shared
  load path, so `--info`, `--export-bom`, `--export-nets` and `--export-links`
  all hang on the same file.
- **It fails in the worst way for automation**: no error, no exit, no output.
  Only an external timeout ends it, so any caller without one hangs
  indefinitely.
- **Measured cost**: in a corpus sweep this single file burned **1202s**
  (5 verbs x 2 refs x 120s timeout) — about 20 minutes of a 23-minute run. I
  have since had to special-case it, which means that file is simply never
  tested.

The second modal in the same function (`:1476`, project saved by a *newer*
QElectroTech) is the same trap in the other direction — nothing in `examples/`
triggers it today, but any file from a future release will.

No rush intended — just adding evidence rather than pinging. The branch is
still `MERGEABLE` / `CLEAN` against current master at +119/-1, and I am happy to
rebase, split it smaller, or adjust the approach if any of it is unwelcome.

ispyisail:
Independently re-verified this today, tracing the exact hang with a real gdb backtrace rather than assuming: blocked in `QDialog::exec()`'s event loop via `QET::QetMessageBox::warning()`, called from `QETProject::readProjectXml()`'s version-compatibility check — `examples/schema_indus.qet` (v0.3) always triggers it on load, and headlessly nothing can ever click it, so it hangs forever (genuinely idle in `ppoll()`, not a busy-spin).

Built this branch directly and confirmed the fix: `--info` and `--resave` on `schema_indus.qet` both now complete cleanly (exit 0) instead of timing out, with the warning logged to stderr and auto-answered as documented. Also ran a full database/document consistency sweep (23 example projects) under this binary — 0 regressions.

Still `CLEAN`/`MERGEABLE` despite being a couple hundred commits behind current master, so no rebase needed on your end before review. This is the last several CLI-tooling scripts in [qelectrotech-docker](https://github.com/ispyisail/qelectrotech-docker) hardcode a skip for `schema_indus.qet` specifically because of this — happy to go remove those skip-lists once this lands.

ispyisail:
New evidence rather than a repeat of the above: this now blocks a gate, and I measured it on today's master.

While investigating a save-determinism failure I traced it back here. The determinism corpus check (`tests/determinism/check.py`, resave each example project twice and compare) **fails on current master solely because of this hang**, and passes with this PR applied.

Measured today against master `a583b3c43`, Qt 5.15.18, same binary, same corpus, only these two commits differing:

| | master | master + #661 |
|---|---|---|
| `--resave examples/schema_indus.qet` | hangs, killed at 100 s | **0.307 s**, exit 0, 163 796 byte output |
| I3 (meaning preserved) | 22/23, one ERROR | **23/23** |
| gate exit code | 1 (fail) | **0 (pass)** |

The hang signature matches the diagnosis already in this PR: `STAT=Sl`, `TIME=00:00:00`, `wchan=poll_schedule_timeout` — zero CPU, blocked in an event loop waiting for a click that cannot arrive under `QT_QPA_PLATFORM=offscreen`. Not a busy loop, not slowness: a 132 KB project that resaves in a third of a second once the modal is routed through the wrapper.

Two consequences beyond the one file:

- Any headless verb on an affected project hangs **forever** rather than failing, so CI or a scripted run has no way to recover except an external timeout. A gate that hangs is worse than a gate that fails, because it burns the whole job budget.
- It puts a floor under the determinism work: I1 currently sits at 5/23 with this applied, and those remaining failures cannot even be investigated on a project that will not resave.

Entirely your call on timing, and no rush from my side — but the PR is still `mergeable: clean` against master five weeks on, and it is now holding back a gate rather than only affecting one example file.

One practical note if you try to reproduce: master does not currently build with the default `QT_VERSION_MAJOR=5` (unrelated one-liner, #824), so I had to apply that first to get a binary at all.

ispyisail:
@scorpio810 this one needs to be pushed ASAP please

ispyisail:
@scorpio810 — merging this one myself rather than adding it to your queue, since you have asked for fewer PRs to review and this one removes a failure rather than adding a feature. Say the word and I will revert it.

Re-verified today against current master (`6d09041dc`), not against the branch's five-week-old base:

- merges **clean**, no rebase needed
- builds clean with **Qt6**, now the default after `a1c090781`
- `--resave examples/schema_indus.qet`: **0.307 s, exit 0**. On master without this, the same command never returns
- save-determinism gate: **exit 0**, I3 **23/23** (was 22/23 plus one ERROR), no regressions against baseline

The reason I am comfortable merging without review is that the blast radius is provably nil for GUI users: `setNonInteractive(true)` is set inside the `if (CLIExport::isExportRequest(raw_args))` block in `main.cpp`, which `return`s. A GUI run cannot reach it. It sits next to the existing `setBackupEnabled(false)`, which is the same CLI-only pattern already in that block.

Two things I noticed while verifying, both from the Qt6 flip rather than from this PR, in case they are news:

- **CUPS is now a hard build requirement.** A clean master configure fails with `Could NOT find Cups (missing: CUPS_LIBRARIES CUPS_INCLUDE_DIR)` until `libcups2-dev` is installed. I confirmed this reproduces on clean master before attributing it.
- **I1 (idempotent save) is 0/23 on the Qt6 build**, where the Qt5 build was giving 4/23 earlier today. Still "no regressions against baseline" because the baseline records 0, so the gate passes either way — but the Qt5 build was beating the baseline and the Qt6 one is not. That may be worth a look separately; it is not connected to this change, which only touches message boxes.

scorpio810:
>CUPS is now a hard build requirement. A clean master configure fails with Could NOT find Cups (missing: CUPS_LIBRARIES CUPS_INCLUDE_DIR) until libcups2-dev is installed. I confirmed this reproduces on clean master before attributing it.

Thanks @ispyisail to pointing this.

PR #662, opened 2026-08-04
Fix broken signal/slot relations

arummler:
Selecting an entry in the conductor/element/folio drop down boxes on the project auto-numbering properties page had no effect (when compiled with QT6): the related fields never updated. Root cause: currentIndexChanged has been removed in QT6.

### DIFF
--- a/sources/autoNum/ui/autonumberingdockwidget.cpp
+++ b/sources/autoNum/ui/autonumberingdockwidget.cpp
@@ -26,6 +26,7 @@
 #include "../numerotationcontext.h"
 #include "../numerotationcontextcommands.h"
 #include "ui_autonumberingdockwidget.h"
+#include "../../undocommand/changetitleblockcommand.h"
 
 #include <QComboBox>
 #include <QLineEdit>
@@ -371,7 +372,17 @@ void AutoNumberingDockWidget::on_m_folio_cb_activated(int) {
 		ip.folio = "%id/%total";
 		m_project->setDefaultTitleBlockProperties(ip);
 	}
-		emit(folioAutoNumChanged(current_autonum));
+
+	if (m_project_view && m_project_view->currentDiagram()) {
+		Diagram *diagram = m_project_view->currentDiagram()->diagram();
+		TitleBlockProperties old_properties = diagram->border_and_titleblock.exportTitleBlock();
+		TitleBlockProperties new_properties = old_properties;
+		new_properties.auto_page_num = ip.auto_page_num;
+		new_properties.folio = ip.folio;
+		if (new_properties != old_properties)
+			diagram->undoStack().push(new ChangeTitleBlockCommand(diagram, old_properties, new_properties));
+	}
+	emit(folioAutoNumChanged(current_autonum));
 	refreshRow(AutoNumCategory::Folio);
 }
 
--- a/sources/diagramview.cpp
+++ b/sources/diagramview.cpp
@@ -100,7 +100,7 @@ DiagramView::DiagramView(Diagram *diagram, QWidget *parent) :
 
 	connect(m_diagram, SIGNAL(showDiagram(Diagram*)), this, SIGNAL(showDiagram(Diagram*)));
 	connect(m_diagram, SIGNAL(sceneRectChanged(QRectF)), this, SLOT(adjustSceneRect()));
-	connect(&(m_diagram -> border_and_titleblock), SIGNAL(diagramTitleChanged(const QString &)), this, SLOT(updateWindowTitle()));
+	connect(&(m_diagram -> border_and_titleblock), &BorderTitleBlock::informationChanged, this, &DiagramView::updateWindowTitle);
 	connect(diagram, SIGNAL(findElementRequired(ElementsLocation)), this, SIGNAL(findElementRequired(ElementsLocation)));
 
 	QShortcut *edit_conductor_color_shortcut = new QShortcut(QKeySequence(Qt::Key_F2), this);
--- a/sources/exportdialog.cpp
+++ b/sources/exportdialog.cpp
@@ -140,12 +140,21 @@ QWidget *ExportDialog::initDiagramsListPart()
 	reset_mapper_     = new QSignalMapper(this);
 	clipboard_mapper_ = new QSignalMapper(this);
 	
-	connect(preview_mapper_,   SIGNAL(mapped(int)), this, SLOT(slot_previewDiagram(int)));
-	connect(width_mapper_,     SIGNAL(mapped(int)), this, SLOT(slot_correctHeight(int)));
-	connect(height_mapper_,    SIGNAL(mapped(int)), this, SLOT(slot_correctWidth(int)));
-	connect(ratio_mapper_,     SIGNAL(mapped(int)), this, SLOT(slot_keepRatioChanged(int)));
-	connect(reset_mapper_,     SIGNAL(mapped(int)), this, SLOT(slot_resetSize(int)));
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0) // TODO Qt6 only: remove, mappedInt() always available
+	connect(preview_mapper_, SIGNAL(mapped(int)), this, SLOT(slot_previewDiagram(int)));
+	connect(width_mapper_, SIGNAL(mapped(int)), this, SLOT(slot_correctHeight(int)));
+	connect(height_mapper_, SIGNAL(mapped(int)), this, SLOT(slot_correctWidth(int)));
+	connect(ratio_mapper_, SIGNAL(mapped(int)), this, SLOT(slot_keepRatioChanged(int)));
+	connect(reset_mapper_, SIGNAL(mapped(int)), this, SLOT(slot_resetSize(int)));
 	connect(clipboard_mapper_, SIGNAL(mapped(int)), this, SLOT(slot_exportToClipBoard(int)));
+#else
+	connect(preview_mapper_, &QSignalMapper::mappedInt, this, &ExportDialog::slot_previewDiagram);
+	connect(width_mapper_, &QSignalMapper::mappedInt, this, &ExportDialog::slot_correctHeight);
+	connect(height_mapper_, &QSignalMapper::mappedInt, this, &ExportDialog::slot_correctWidth);
+	connect(ratio_mapper_, &QSignalMapper::mappedInt, this, &ExportDialog::slot_keepRatioChanged);
+	connect(reset_mapper_, &QSignalMapper::mappedInt, this, &ExportDialog::slot_resetSize);
+	connect(clipboard_mapper_, &QSignalMapper::mappedInt, this, &ExportDialog::slot_exportToClipBoard);
+#endif
 	
 	diagrams_list_layout_ = new QGridLayout();
 	
--- a/sources/qetapp.cpp
+++ b/sources/qetapp.cpp
@@ -124,8 +124,11 @@ QETApp::QETApp() :
 	initSplashScreen();
 	initSystemTray();
 
-	connect(&signal_map, SIGNAL(mapped(QWidget *)),
-		this, SLOT(invertMainWindowVisibility(QWidget *)));
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0) // TODO Qt6 only: remove, mappedObject() always available
+	connect(&signal_map, SIGNAL(mapped(QWidget *)), this, SLOT(invertMainWindowVisibility(QWidget *)));
+#else
+	connect(&signal_map, &QSignalMapper::mappedObject, this, [this](QObject *object) { invertMainWindowVisibility(qobject_cast<QWidget *>(object)); });
+#endif
 	qApp->setQuitOnLastWindowClosed(false);
 	connect(qApp, &QApplication::lastWindowClosed,
 		this, &QETApp::checkRemainingWindows);
--- a/sources/qetdiagrameditor.cpp
+++ b/sources/qetdiagrameditor.cpp
@@ -107,7 +107,11 @@ QETDiagramEditor::QETDiagramEditor(const QStringList &files, QWidget *parent) :
 	m_workspace.setTabsClosable(true);
 
 		//Set the signal mapper
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0) // TODO Qt6 only: remove, mappedObject() always available
 	connect(&windowMapper, SIGNAL(mapped(QWidget *)), this, SLOT(activateWidget(QWidget *)));
+#else
+	connect(&windowMapper, &QSignalMapper::mappedObject, this, [this](QObject *object) { activateWidget(qobject_cast<QWidget *>(object)); });
+#endif
 
 	setWindowTitle(tr("QElectroTech", "window title"));
 	setWindowIcon(QET::Icons::QETLogo);
--- a/sources/qetproject.h
+++ b/sources/qetproject.h
@@ -244,7 +244,6 @@ class QETProject : public QObject
 			/// rebuild their rule lists; this one just says "re-read me".
 		void autoNumContextUpdated();
 		void folioAutoNumRemoved();
-		void folioAutoNumChanged(QString);
 		void defaultTitleBlockPropertiesChanged();
 		void conductorAutoNumChanged();
 
--- a/sources/titleblock/templatelogomanager.cpp
+++ b/sources/titleblock/templatelogomanager.cpp
@@ -217,7 +217,11 @@ QString TitleBlockTemplateLogoManager::confirmLogoName(const QString &initial_na
 			connect(replace_button, SIGNAL(clicked()), signal_mapper, SLOT(map()));
 			connect(rename_button,  SIGNAL(clicked()), signal_mapper, SLOT(map()));
 			connect(cancel_button,  SIGNAL(clicked()), signal_mapper, SLOT(map()));
+#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0) // TODO Qt6 only: remove, mappedInt() always available
 			connect(signal_mapper, SIGNAL(mapped(int)), rename_dialog, SLOT(done(int)));
+#else
+			connect(signal_mapper, &QSignalMapper::mappedInt, rename_dialog, &QDialog::done);
+#endif
 		}
 		rd_label -> setText(
 			QString(tr(
--- a/sources/ui/configpage/projectconfigpages.cpp
+++ b/sources/ui/configpage/projectconfigpages.cpp
@@ -396,17 +396,29 @@ void ProjectAutoNumConfigPage::buildConnections()
 		//Conductor Tab
 	connect(m_saw_conductor, &SelectAutonumW::applyPressed,  this, &ProjectAutoNumConfigPage::saveContextConductor);
 	connect(m_saw_conductor, &SelectAutonumW::removeClicked, this, &ProjectAutoNumConfigPage::removeContextConductor);
-	connect(m_saw_conductor->contextComboBox(), SIGNAL(currentIndexChanged(QString)), this, SLOT(updateContextConductor(QString)));
+#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0) // TODO Qt6 only: remove, textActivated() always available
+	connect(m_saw_conductor->contextComboBox(), SIGNAL(activated(QString)), this, SLOT(updateContextConductor(QString)));
+#else
+	connect(m_saw_conductor->contextComboBox(), &QComboBox::textActivated, this, &ProjectAutoNumConfigPage::updateContextConductor);
+#endif
 
 		//Element Tab
 	connect(m_saw_element, &SelectAutonumW::applyPressed,  this, &ProjectAutoNumConfigPage::saveContextElement);
 	connect(m_saw_element, &SelectAutonumW::removeClicked, this, &ProjectAutoNumConfigPage::removeContextElement);
-	connect(m_saw_element->contextComboBox(), SIGNAL(currentIndexChanged(QString)), this, SLOT(updateContextElement(QString)));
+#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0) // TODO Qt6 only: remove, textActivated() always available
+	connect(m_saw_element->contextComboBox(), SIGNAL(activated(QString)), this, SLOT(updateContextElement(QString)));
+#else
+	connect(m_saw_element->contextComboBox(), &QComboBox::textActivated, this, &ProjectAutoNumConfigPage::updateContextElement);
+#endif
 
 		//Folio Tab
 	connect(m_saw_folio, &SelectAutonumW::applyPressed,  this, &ProjectAutoNumConfigPage::saveContextFolio);
 	connect(m_saw_folio, &SelectAutonumW::removeClicked, this, &ProjectAutoNumConfigPage::removeContextFolio);
-	connect(m_saw_folio->contextComboBox(), SIGNAL(currentIndexChanged(QString)), this, SLOT(updateContextFolio(QString)));
+#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0) // TODO Qt6 only: remove, textActivated() always available
+	connect(m_saw_folio->contextComboBox(), SIGNAL(activated(QString)), this, SLOT(updateContextFolio(QString)));
+#else
+	connect(m_saw_folio->contextComboBox(), &QComboBox::textActivated, this, &ProjectAutoNumConfigPage::updateContextFolio);
+#endif
 
 		//	Auto Folio Numbering
 	connect (m_faw, SIGNAL (applyPressed()),				 this, SLOT (applyAutoNum()));
@@ -491,6 +503,10 @@ void ProjectAutoNumConfigPage::removeContextElement()
 		return;
 	m_project->removeElementAutoNum (m_saw_element->contextComboBox()->currentText());
 	m_saw_element->contextComboBox()->removeItem (m_saw_element->contextComboBox()->currentIndex());
+	// removeItem() removes the current selection programmatically but
+	// textActivated() does not react to (by design, see buildConnections()).
+	// Refresh the displayed pattern explicitly so it matches the new selection.
+	updateContextElement(m_saw_element->contextComboBox()->currentText());
 }
 
 /**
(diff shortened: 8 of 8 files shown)


ispyisail (changes requested):
Root cause is correctly identified: `currentIndexChanged(QString)` had its `QString` overload removed in Qt6, and an old-style `SIGNAL()/SLOT()` connect targeting a signal that no longer exists fails at runtime rather than at compile time — `connect()` just returns `false` and logs a warning, so the UI silently stops updating with no error anywhere obvious. That matches the report exactly, and I checked for other instances of the same pattern elsewhere in the codebase — these three are the only ones, so the fix is complete in scope.

Verified: builds clean, and I ran it live to confirm the dropdowns work again.

One behavioral difference worth flagging before merge, found by tracing where the new signal leads rather than just diffing the signatures.

`contextComboBox()` (`m_comboBox` in `selectautonumw.ui`) has `editable` set to `true`. `currentTextChanged` fires on every keystroke in an editable combo box, not just on an actual selection change the way the old `currentIndexChanged` did. Each firing calls `updateContextConductor`/`Element`/`Folio`, which calls `SelectAutonumW::setContext()` — and for any string that doesn't exactly match a saved context name, that method does:

```cpp
qDeleteAll(num_part_list_);
num_part_list_.clear();
...
on_add_button_clicked();  // rebuilds a single blank default part
```

So editing the name field character-by-character wipes the configured numbering parts back to one blank part on the very first keystroke — before Apply is ever clicked. I reproduced this directly: added a second numbering part to the Conductor tab (two "Chiffre 1" rows), clicked into the combo box's text field, typed a single character, and both parts collapsed to one blank default part immediately.

The realistic failure case: pick an existing multi-part context to use as a starting point for a similar one, then edit the name to create a variant — the first keystroke silently destroys the parts you meant to keep, before you get anywhere near Apply.

`QComboBox::textActivated(const QString &)` looks like a closer match to the original intent: it fires once a selection is actually committed (a dropdown click, or Enter on typed text), not on every intermediate keystroke, and it's Qt6-safe — it's the currently-recommended replacement for the now-deprecated `activated(QString)` overload, so it's the same generation of API as `currentTextChanged` rather than something older. Swapping the three connects to it should fix the reported bug without introducing the per-keystroke reset.

Nice, minimal diagnosis on a real bug — just want the replacement signal to land somewhere that doesn't trade one broken interaction for a quieter one.

arummler:
Yes, I agree. Either textActivated or alternatively currentIndexChanged(int) and translate with a lambda. Both would be better. But I was thinking about a renaming function. Personally I find this drop down box used for everything: creation, selection and potentially renaming a bit awkward. I would propose to add an add (and potentially rename) button next to the delete button and make the text in the drop down box non-editable.  But something to think of and not necessarily in this MR. I will do the minimal change.

arummler:
I applied the changes. I also scanned for more broken signal/slot connections and found a few. In general I would migrate alle the remaining SIGNAL/SLOT which are working to the modern style but that has to be done carefully. Most of the broken connections only appeared with Qt6 but one of them was there since 3 years.

ispyisail (commented):
## Overview

This repairs four independent signal/slot connections that fail silently under Qt6. Since string-based `SIGNAL`/`SLOT` connects resolve at *runtime*, these produced a `qWarning` at startup and then simply never fired — no compile error, no crash, just dead UI. That makes them easy to miss, so thanks for tracking these down.

The four root causes:

| Fix | Cause |
|---|---|
| Autonum combo boxes | `QComboBox::currentIndexChanged(const QString&)` removed in Qt6 |
| `QSignalMapper` connects (×4 sites) | `mapped()` overloads removed in Qt6 |
| `DiagramView` window title | Signal renamed in 27dcd5e (Oct 2023), connect never updated |
| Folio autonum not applying | Connect targeted a slot that doesn't exist |

Overall this looks sound and should land. I have one issue I'd suggest addressing before merge, one scope question, and some cosmetics.

## Verifying the "dead connection" claims

I checked these against the tree rather than assuming, and both hold up:

- **`slot_setAutoPageNum`** (`autonumberingdockwidget.cpp:113,152`) — grepping all of `sources/` finds this name *only* at the two connect/disconnect sites. There is no such slot on `BorderTitleBlock`. The connection has never worked in its current form, so removing it is right.
- **`diagramTitleChanged`** — removed from `BorderTitleBlock` by 27dcd5e, whose message explicitly says *"Remove signal diagramTitleChanged from BorderTitleBlock and use instead the signal informationChanged."* That commit updated `diagram.cpp` and `genericpanel.cpp` but missed `diagramview.cpp`, so the window title has been stale since October 2023. Nice catch.

## Findings

### 1. `importTitleBlock()` bypasses the undo stack

`sources/autoNum/ui/autonumberingdockwidget.cpp:340-342`

```cpp
if (m_project_view && m_project_view->currentDiagram()) {
    m_project_view->currentDiagram()->diagram()->border_and_titleblock.importTitleBlock(ip);
}
```

This mutates a diagram's title block — per-folio document state that gets serialized into the `.qet` — without going through the undo stack. There's a dedicated command for exactly this operation, and every other caller uses it:

- `sources/undocommand/changetitleblockcommand.cpp:48,55` wraps `importTitleBlock` for undo/redo
- called from `diagramview.cpp:273`, `searchandreplaceworker.cpp:82,308`, `diagrampropertiesdialog.cpp:101`

As written, Ctrl+Z after changing the folio autonum won't revert the title block.

In fairness, the surrounding code is already non-undoable (`setDefaultTitleBlockProperties`, `qetproject.cpp:554`) — but that only touches the *project default*, whereas this writes into a diagram. Something like:

```cpp
diagram->undoStack().push(new ChangeTitleBlockCommand(diagram, old_props, ip));
```

### 2. Only the current diagram is updated — scope question

The setting being changed is a project-wide default (`setDefaultTitleBlockProperties`), but the fix applies it to `currentDiagram()` only, so other open folios keep the old autonum until something else refreshes them.

Two questions:

- Should this iterate all diagrams in the project?
- If a folio has *customised* title block properties, does importing the project default clobber them? `importTitleBlock(ip)` replaces the whole `TitleBlockProperties` struct rather than just the autonum field, and `ip` is the project default read at line 338. Compare `qetproject.cpp:1369`, where this same call is used only for brand-new diagrams.

That second one is what I'd most want manually tested: set a custom title block on folio 2, then change the folio autonum while folio 2 is active.

### 3. `textActivated` narrows semantics vs. the original

`sources/ui/configpage/projectconfigpages.cpp`

The history here is informative — 1216855c used `&QComboBox::currentTextChanged` (the faithful Qt6 replacement, firing on programmatic *and* user changes), then fb514894 switched to `textActivated` (user interaction only). I'm assuming that was to stop the slot firing during `addItems()` population at `:365-373`, which is a fair reason.

The trade-off is that programmatic index changes no longer refresh the panel. Concretely: `removeContextElement` (`:493`) calls `removeItem()`, which moves the current index. Under Qt5's `currentIndexChanged` the displayed context followed along; with `textActivated` it won't, so the panel can show a deleted numbering's settings next to a different name in the combo.

Not a regression for Qt6 users, since nothing worked at all before — but a small one for Qt5. If population-time firing was the concern, `currentTextChanged` plus a `QSignalBlocker` around `addItems()` would preserve both behaviours. Your call, but it'd be worth a line in the PR description either way.

### 4. Preprocessor indentation doesn't match the codebase

The PR indents directives inside function bodies (`\t#if QT_VERSION >= ...`). All 72 existing `QT_VERSION_CHECK` uses in `sources/` put `#if` at column 0 — e.g. `qgimanager.cpp:77`, `conductorproperties.cpp:814`, `qetxml.cpp:280`.

Related: the established comment marker here is `// ### Qt 6: remove` (`bordertitleblock.cpp:59`, `qtextorientationwidget.cpp:258`, `titleblocktemplate.cpp:399`) rather than `// TODO Qt6 only: remove`. Existing guards also tend to be written `#if QT_VERSION < QT_VERSION_CHECK(...)` with the legacy path first; this PR inverts that. Functionally identical, just inconsistent.

### 5. Typo

`sources/titleblock/templatelogomanager.cpp:222` — `// TOD Qt6 only` → `TODO`.

### 6. Whitespace

`sources/autoNum/ui/autonumberingdockwidget.h:66-69` — removing the `signals:` block leaves two consecutive blank lines.

### 7. Optional drive-by

`sources/qetproject.h:244` declares `void folioAutoNumChanged(QString);` with no emitter and no connect anywhere in `sources/`. Dead before this PR, but it's the same signal name being cleaned up here, so it might be a natural inclusion.

## Things I checked that are correct

Noting these so nobody "corrects" them later:

- **`mappedObject` + `qobject_cast`, not `mappedWidget`.** Qt 5.15 added `mappedWidget(QWidget*)`, which looks like the tidier match for `mapped(QWidget*)` — but Qt6 moved `QSignalMapper` to QtCore and dropped it. `mappedObject` is the correct choice.
- **Version thresholds are accurate.** `mappedInt`/`mappedObject` → 5.15, `textActivated` → 5.14. The `#else` branches aren't dead code either: the project still carries pre-5.14 paths (`conductorproperties.cpp:814`, `richtexteditor.cpp:749`).
- **Lambdas pass `this` as the context object**, so they disconnect on destruction.
- **Null-safety of the casts holds.** Both mappers are only ever fed `QWidget*` (`qetdiagrameditor.cpp:2113`, `qetapp.cpp:2443`), and both targets tolerate `nullptr` anyway (`invertMainWindowVisibility:1724` guards via `qobject_cast`, `subWindowForWidget:1446` returns `nullptr`).
- **`exportdialog.cpp` clipboard connect is not duplicated** — it appears exactly once per branch.
- **Removing the `folioAutoNumChanged(QString)` signal doesn't break the same-named no-arg slot** (`autonumberingdockwidget.h:54`), which is still connected and used. Having both overloads was confusing, so this is an improvement.

## Suggested manual testing

No tests here, which is consistent with the rest of the project — these are GUI wirings, so unit tests aren't really available. That does raise the value of a manual pass, on **both Qt5 and Qt6**, since three of the four fixes touch version-guarded paths:

1. Qt6: change conductor/element/folio autonum in project properties → fields update
2. Qt5: confirm no regression from the `textActivated` narrowing (the `removeItem` case in #3)
3. Change a diagram title → window title updates
4. Folio with a customised title block + autonum change (finding #2)
5. Export dialog: preview, width/height, ratio, reset, clipboard
6. Tray-icon window toggle and the Windows menu (both mappers)

## Security / performance

Nothing of concern — no input handling, no allocation changes. `informationChanged` fires more often than the old `diagramTitleChanged` would have, but `updateWindowTitle()` is trivial and the old connect was dead anyway, so this is strictly a restoration.

---

Real bugs, correctly diagnosed, with the trickiest call (`mappedObject` over `mappedWidget`) made right. Only #1 seems worth changing before merge; #2 deserves an answer; the rest are cosmetic.

ispyisail (commented):
## Overview

This repairs four independent signal/slot connections that fail silently under Qt6. Since string-based `SIGNAL`/`SLOT` connects resolve at *runtime*, these produced a `qWarning` at startup and then simply never fired — no compile error, no crash, just dead UI. That makes them easy to miss, so thanks for tracking these down.

The four root causes:

| Fix | Cause |
|---|---|
| Autonum combo boxes | `QComboBox::currentIndexChanged(const QString&)` removed in Qt6 |
| `QSignalMapper` connects (×4 sites) | `mapped()` overloads removed in Qt6 |
| `DiagramView` window title | Signal renamed in 27dcd5e (Oct 2023), connect never updated |
| Folio autonum not applying | Connect targeted a slot that doesn't exist |

Overall this looks sound and should land. I have one issue I'd suggest addressing before merge, one scope question, and some cosmetics.

## Verifying the "dead connection" claims

I checked these against the tree rather than assuming, and both hold up:

- **`slot_setAutoPageNum`** (`autonumberingdockwidget.cpp:113,152`) — grepping all of `sources/` finds this name *only* at the two connect/disconnect sites. There is no such slot on `BorderTitleBlock`. The connection has never worked in its current form, so removing it is right.
- **`diagramTitleChanged`** — removed from `BorderTitleBlock` by 27dcd5e, whose message explicitly says *"Remove signal diagramTitleChanged from BorderTitleBlock and use instead the signal informationChanged."* That commit updated `diagram.cpp` and `genericpanel.cpp` but missed `diagramview.cpp`, so the window title has been stale since October 2023. Nice catch.

## Findings

### 1. `importTitleBlock()` bypasses the undo stack

`sources/autoNum/ui/autonumberingdockwidget.cpp:340-342`

```cpp
if (m_project_view && m_project_view->currentDiagram()) {
    m_project_view->currentDiagram()->diagram()->border_and_titleblock.importTitleBlock(ip);
}
```

This mutates a diagram's title block — per-folio document state that gets serialized into the `.qet` — without going through the undo stack. There's a dedicated command for exactly this operation, and every other caller uses it:

- `sources/undocommand/changetitleblockcommand.cpp:48,55` wraps `importTitleBlock` for undo/redo
- called from `diagramview.cpp:273`, `searchandreplaceworker.cpp:82,308`, `diagrampropertiesdialog.cpp:101`

As written, Ctrl+Z after changing the folio autonum won't revert the title block.

In fairness, the surrounding code is already non-undoable (`setDefaultTitleBlockProperties`, `qetproject.cpp:554`) — but that only touches the *project default*, whereas this writes into a diagram. Something like:

```cpp
diagram->undoStack().push(new ChangeTitleBlockCommand(diagram, old_props, ip));
```

### 2. Only the current diagram is updated — scope question

The setting being changed is a project-wide default (`setDefaultTitleBlockProperties`), but the fix applies it to `currentDiagram()` only, so other open folios keep the old autonum until something else refreshes them.

Two questions:

- Should this iterate all diagrams in the project?
- If a folio has *customised* title block properties, does importing the project default clobber them? `importTitleBlock(ip)` replaces the whole `TitleBlockProperties` struct rather than just the autonum field, and `ip` is the project default read at line 338. Compare `qetproject.cpp:1369`, where this same call is used only for brand-new diagrams.

That second one is what I'd most want manually tested: set a custom title block on folio 2, then change the folio autonum while folio 2 is active.

### 3. `textActivated` narrows semantics vs. the original

`sources/ui/configpage/projectconfigpages.cpp`

The history here is informative — 1216855c used `&QComboBox::currentTextChanged` (the faithful Qt6 replacement, firing on programmatic *and* user changes), then fb514894 switched to `textActivated` (user interaction only). I'm assuming that was to stop the slot firing during `addItems()` population at `:365-373`, which is a fair reason.

The trade-off is that programmatic index changes no longer refresh the panel. Concretely: `removeContextElement` (`:493`) calls `removeItem()`, which moves the current index. Under Qt5's `currentIndexChanged` the displayed context followed along; with `textActivated` it won't, so the panel can show a deleted numbering's settings next to a different name in the combo.

Not a regression for Qt6 users, since nothing worked at all before — but a small one for Qt5. If population-time firing was the concern, `currentTextChanged` plus a `QSignalBlocker` around `addItems()` would preserve both behaviours. Your call, but it'd be worth a line in the PR description either way.

### 4. Preprocessor indentation doesn't match the codebase

The PR indents directives inside function bodies (`\t#if QT_VERSION >= ...`). All 72 existing `QT_VERSION_CHECK` uses in `sources/` put `#if` at column 0 — e.g. `qgimanager.cpp:77`, `conductorproperties.cpp:814`, `qetxml.cpp:280`.

Related: the established comment marker here is `// ### Qt 6: remove` (`bordertitleblock.cpp:59`, `qtextorientationwidget.cpp:258`, `titleblocktemplate.cpp:399`) rather than `// TODO Qt6 only: remove`. Existing guards also tend to be written `#if QT_VERSION < QT_VERSION_CHECK(...)` with the legacy path first; this PR inverts that. Functionally identical, just inconsistent.

### 5. Typo

`sources/titleblock/templatelogomanager.cpp:222` — `// TOD Qt6 only` → `TODO`.

### 6. Whitespace

`sources/autoNum/ui/autonumberingdockwidget.h:66-69` — removing the `signals:` block leaves two consecutive blank lines.

### 7. Optional drive-by

`sources/qetproject.h:244` declares `void folioAutoNumChanged(QString);` with no emitter and no connect anywhere in `sources/`. Dead before this PR, but it's the same signal name being cleaned up here, so it might be a natural inclusion.

## Things I checked that are correct

Noting these so nobody "corrects" them later:

- **`mappedObject` + `qobject_cast`, not `mappedWidget`.** Qt 5.15 added `mappedWidget(QWidget*)`, which looks like the tidier match for `mapped(QWidget*)` — but Qt6 moved `QSignalMapper` to QtCore and dropped it. `mappedObject` is the correct choice.
- **Version thresholds are accurate.** `mappedInt`/`mappedObject` → 5.15, `textActivated` → 5.14. The `#else` branches aren't dead code either: the project still carries pre-5.14 paths (`conductorproperties.cpp:814`, `richtexteditor.cpp:749`).
- **Lambdas pass `this` as the context object**, so they disconnect on destruction.
- **Null-safety of the casts holds.** Both mappers are only ever fed `QWidget*` (`qetdiagrameditor.cpp:2113`, `qetapp.cpp:2443`), and both targets tolerate `nullptr` anyway (`invertMainWindowVisibility:1724` guards via `qobject_cast`, `subWindowForWidget:1446` returns `nullptr`).
- **`exportdialog.cpp` clipboard connect is not duplicated** — it appears exactly once per branch.
- **Removing the `folioAutoNumChanged(QString)` signal doesn't break the same-named no-arg slot** (`autonumberingdockwidget.h:54`), which is still connected and used. Having both overloads was confusing, so this is an improvement.

## Suggested manual testing

No tests here, which is consistent with the rest of the project — these are GUI wirings, so unit tests aren't really available. That does raise the value of a manual pass, on **both Qt5 and Qt6**, since three of the four fixes touch version-guarded paths:

1. Qt6: change conductor/element/folio autonum in project properties → fields update
2. Qt5: confirm no regression from the `textActivated` narrowing (the `removeItem` case in #3)
3. Change a diagram title → window title updates
4. Folio with a customised title block + autonum change (finding #2)
5. Export dialog: preview, width/height, ratio, reset, clipboard
6. Tray-icon window toggle and the Windows menu (both mappers)

## Security / performance

Nothing of concern — no input handling, no allocation changes. `informationChanged` fires more often than the old `diagramTitleChanged` would have, but `updateWindowTitle()` is trivial and the old connect was dead anyway, so this is strictly a restoration.

---

Real bugs, correctly diagnosed, with the trickiest call (`mappedObject` over `mappedWidget`) made right. Only #1 seems worth changing before merge; #2 deserves an answer; the rest are cosmetic.

ispyisail (changes requested):
Re-reviewed after the updates. The `textActivated` change is in and correct, and the extra fixes you found are genuinely good catches. One blocking issue, in `qetapp.cpp`.

## Blocking: the version guard in `qetapp.cpp` is inverted

```cpp
#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0) // TODO Qt6 only: remove, mappedObject() always available
	connect(&signal_map, SIGNAL(mapped(QWidget *)), this, SLOT(invertMainWindowVisibility(QWidget *)));
#else
	connect(&signal_map, &QSignalMapper::mappedObject, this, [this](QObject *object) { invertMainWindowVisibility(qobject_cast<QWidget *>(object)); });
#endif
```

Compare with the three other sites in this PR (`exportdialog.cpp`, `qetdiagrameditor.cpp`, `templatelogomanager.cpp`), which all correctly use `#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)` → old string-based, `#else` → new typed signal. This one has `>=`, so both arms end up on the wrong side:

**On Qt6** it takes the string-based branch — but `mapped()` was removed in Qt6 entirely. Qt6's `qsignalmapper.h` declares only:

```
void mappedInt(int);
void mappedString(const QString &);
void mappedObject(QObject *);
```

so the connect fails at runtime and the tray-icon show/hide toggle stays broken — the same class of bug this PR is fixing. Reproduced with a minimal Qt6 program doing exactly this connect:

```
QObject::connect: No such signal QSignalMapper::mapped(QWidget *)
old-style connect(mapped(QWidget*)) succeeded: false
```

**On Qt < 5.15** it takes the `#else` branch, which uses `&QSignalMapper::mappedObject` — added in 5.15, so that won't compile. Not an issue on 5.15/6, but it is inconsistent with the other three guards, which do handle it.

Fix is just the comparison direction:

```cpp
#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)
	connect(&signal_map, SIGNAL(mapped(QWidget *)), this, SLOT(invertMainWindowVisibility(QWidget *)));
#else
	connect(&signal_map, &QSignalMapper::mappedObject, this, [this](QObject *object) { invertMainWindowVisibility(qobject_cast<QWidget *>(object)); });
#endif
```

(Using `mappedObject` + `qobject_cast` rather than `mappedWidget` is the right call, incidentally — Qt5.15 has `mappedWidget`, but Qt6 dropped it along with `mapped`.)

## Verified good

- **`textActivated` guard** — correct direction, and 5.14 is the right boundary (`textActivated` was added in 5.14).
- **The other three `QSignalMapper` guards** — all correct.
- **`diagramview.cpp`** — confirmed real: `BorderTitleBlock` has no `diagramTitleChanged` signal at all (its signals are `borderChanged`, `displayChanged`, `titleBlockFolioChanged`, `needFolioData`, `needTitleBlockTemplate`, `informationChanged`), so master's connect was a silent runtime no-op. `informationChanged()` → `updateWindowTitle()` match on signature. Good find.
- **The folio rework** is better than the signal it replaces, for a reason beyond undo: the old `connect`/`disconnect` pair dereferenced `m_project_view->currentDiagram()->diagram()` with no null check, and bound to whichever diagram was current *at `setProject()` time* — so it also stopped applying once the user switched diagrams. Resolving the diagram at activation time behind an `if (m_project_view && m_project_view->currentDiagram())` fixes both. Copying `old_properties` and overriding only `auto_page_num`/`folio` is also what stops the clobbering.
- **Builds clean** — CMake/Ninja Release against Qt 5.15.18, no warnings from the changed files.

## Minor, non-blocking

1. `on_m_folio_cb_activated()` pushes a `ChangeTitleBlockCommand` unconditionally, and that command's constructor has no old==new short-circuit, so re-picking the entry that's already selected adds an empty step to the undo stack. `TitleBlockProperties` has `operator!=`, so a one-line guard covers it:

   ```cpp
   if (new_properties != old_properties)
       diagram->undoStack().push(new ChangeTitleBlockCommand(diagram, old_properties, new_properties));
   ```

2. In the `else` branch (empty selection) `ip.auto_page_num` keeps its previous value and is then copied onto the diagram. Harmless in practice since `folio` is `%id/%total` and `auto_page_num` isn't consulted, but it does leave a stale field behind.

Happy to re-check once `qetapp.cpp` is flipped. Nice work tracking down the three-year-old one.

arummler:
Thanks, I have fixed both points.

ispyisail (changes requested):
Both of my points are fixed correctly — thanks. Verified each:

- **`qetapp.cpp` guard** (3d2d4cd6b) — now `#if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)`, matching the other three sites, so Qt 6 takes the `mappedObject` branch. The blocking issue is resolved.
- **Empty undo entry** (2157a3dce) — `if (new_properties != old_properties)` at the call site.
- You also picked up the stale `auto_page_num` I'd mentioned as a side note, with `ip.setAutoPageNum(QString())` in the `else`. Nice.

Unfortunately the newer commit 3f8f1d4d2 introduces two new problems, so not quite an all-clear yet.

## 1. `removeContextConductor()` refreshes from the wrong combo box

```cpp
void ProjectAutoNumConfigPage::removeContextConductor()
{
	...
	m_saw_conductor->contextComboBox()->removeItem(...);
	updateContextConductor(m_saw_element->contextComboBox()->currentText());
	                       ^^^^^^^^^^^^
```

That reads the **element** tab's combo and feeds it to `updateContextConductor()`. `m_saw_conductor`, `m_saw_element` and `m_saw_folio` are three separate `SelectAutonumW` instances (lines 337/341/345), so after deleting a conductor numbering context the conductor pattern editor gets refreshed with whatever the Element tab happens to have selected.

The element and folio versions are both correct — only this one:

```
510:  updateContextElement  (m_saw_element  ->contextComboBox()->currentText());   OK
701:  updateContextConductor(m_saw_element  ->contextComboBox()->currentText());   <- m_saw_conductor
718:  updateContextFolio    (m_saw_folio    ->contextComboBox()->currentText());   OK
```

Worth flagging that this is the same *class* of bug the commit is fixing — showing a pattern editor that doesn't correspond to the selected context — so it would be easy to mistake the remaining symptom for the original one not being fully fixed.

## 2. Stray `#include <unistd.h>`

```cpp
+#include <unistd.h>
```

Nothing in the file uses it — no `sleep`/`usleep`/`read`/`close`/etc. — so it looks like a leftover from debugging. It's also POSIX-only and not available under MSVC, so it risks breaking the Windows build for no benefit.

The codebase's only other use of that header treats it exactly that way (`sources/import/edz/lzma/7zFile.c`):

```c
#ifdef _WIN32
  #include <io.h>
  ...
#else
  #include <unistd.h>
#endif
```

Simplest fix is to drop the line.

---

Everything else still builds clean here (CMake/Ninja Release, Qt 5.15.18). With those two sorted I'd be happy to approve.

arummler:
Yeah. 1) was a copy and paste artefact, made several in this MR... 2) was indeed left over from debugging. Fixed.

ispyisail (approved):
Both blocking points from my last round are fixed correctly in 8640a07f7, and it's the minimal change I'd expect:

- **`removeContextConductor()`** now reads `m_saw_conductor->contextComboBox()->currentText()` instead of `m_saw_element`'s. Checked it against the two sibling methods (`removeContextElement` at line 509 uses `m_saw_element`, `removeContextFolio` at line 717 uses `m_saw_folio`) — all three are now self-consistent, each refreshing from its own combo box rather than a neighbor's.
- **`#include <unistd.h>`** — removed, confirmed no remaining use in the file.

Nothing else changed in this commit (single file, 1 insertion / 2 deletions), so no new surface to review. Combined with everything verified in earlier rounds (Qt version guard direction, undo-stack wiring, `textActivated` vs. `currentTextChanged`, the stale `auto_page_num` fix), this looks complete. Approving.

arummler:
Merged master in order to resolve conflict.

scorpio810:
I 'm think #if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)  guards was no needed  with my packages:
My Ubuntu PPA is for:
Resolute (26.04) is version 5.15.18 
Ubuntu 24.04 LTS (Noble Numbat), the provided Qt 5 version is 5.15.13
Ubuntu 22.04 LTS (Jammy Jellyfish) includes Qt 5.15.3

BTW, we can drop Ubuntu 22.04.

Debian trixie (stable) Qt 5 version is 5.15.15
Debian (Unstable)Qt 5 version is  5.15.19

AppImage Qt 5 version is 5.15.12
Flatpak Qt 5 version is 5.15.18
AppImage arm64 Raspberry Pi 5 (Debian Trixie based Raspberry Pi OS)

macOS aarch64  Qt 5 version is 5.15.17 I 'm think
macOS x86_64 Qt 5 version is 5.15.16 I 'm think

Windows CI/CD Qt 5 version is  5.15.19

https://github.com/qelectrotech/qelectrotech-source-mirror/pull/698

PR #664, opened 2026-08-05
Fix element_info orphan row causing UNIQUE constraint errors on undo

ispyisail:
## What this fixes

`projectDataBase::removeElement()` only ran `DELETE FROM element WHERE uuid=:uuid`. It never touched `element_info`, even though every element also has a row there (`element_uuid` is its `PRIMARY KEY`, with a `FOREIGN KEY` back to `element.uuid` that isn't enforced by this connection — no `ON DELETE CASCADE` in effect). So deleting an element left its `element_info` row orphaned.

Re-adding an element with that same uuid later — undo of that same deletion, or a redo replaying it — goes through `Diagram::addItem()` → `projectDataBase::addElement()`, which `INSERT`s into both tables. The `element` insert succeeds (that row really was removed). The `element_info` insert hits the orphaned row's primary key and fails, silently: the error is logged and swallowed, so the element re-enters the scene with **no `element_info` row at all**, and nothing later re-syncs it. The element then stays silently absent from the database — and from anything that reads it, e.g. nomenclature/BOM views — until something else happens to touch it again.

## How I found it

Testing the combined build in [qelectrotech-docker](https://github.com/ispyisail/qelectrotech-docker), the app lagged during a normal session. Its log showed this repeating:

```
projectDataBase::addElement insert element info error :
QSqlError("19", "Unable to fetch row",
"UNIQUE constraint failed: element_info.element_uuid")
```

six times over a ~17-minute session — not a one-off. Notably it was *always* the `element_info` insert that failed and *never* the `element` insert (0 occurrences of "insert element error" in the same log), which is what pointed at asymmetric cleanup between the two tables rather than a general uuid-collision problem.

This is pre-existing on `master` — it predates and is independent of the wiring-database PR stack (#625/#628/#629/#630); those just also use this same table.

## The fix

`removeElement()` now also runs `DELETE FROM element_info WHERE element_uuid=:uuid`, so a removed element leaves no orphaned row behind for a later re-add to collide with.

## Verification

Built and ran the fix live: placed an element already present in a project, deleted it, undid the deletion. Before the fix, that sequence reliably reproduced the exact error above. After the fix, the same sequence produces nothing — the element reappears with no database error. Confirmed with screenshots at each step (selected → deleted → undone) against the actual compiled binary, not just from reading the code.

General regression check (`tests/catch`) still passes; it doesn't cover this file specifically, so the live reproduction above is the meaningful verification here.

### DIFF
--- a/sources/dataBase/projectdatabase.cpp
+++ b/sources/dataBase/projectdatabase.cpp
@@ -150,10 +150,30 @@ void projectDataBase::addElement(Element *element)
 */
 void projectDataBase::removeElement(Element *element)
 {
+	bool changed = false;
+
 	m_remove_element_query.bindValue(":uuid", element->uuid().toString());
-	if(!m_remove_element_query.exec()) {
-		qDebug() << "projectDataBase::removeElement remove error : " << m_remove_element_query.lastError();
+	if (m_remove_element_query.exec()) {
+		changed = true;
 	} else {
+		qDebug() << "projectDataBase::removeElement remove error : " << m_remove_element_query.lastError();
+	}
+
+		// element_info has no ON DELETE CASCADE (foreign keys aren't enforced
+		// by this connection), so it must be cleared explicitly here. Without
+		// this, the row is orphaned under the removed element's uuid, and
+		// re-adding an element with that same uuid later (undo of this same
+		// removal, or a redo replaying it) hits element_info's PRIMARY KEY
+		// constraint on element_uuid: the element re-add itself succeeds, but
+		// its element_info insert silently fails and is lost.
+	m_remove_element_info_query.bindValue(":uuid", element->uuid().toString());
+	if (m_remove_element_info_query.exec()) {
+		changed = true;
+	} else {
+		qDebug() << "projectDataBase::removeElement remove element_info error : " << m_remove_element_info_query.lastError();
+	}
+
+	if (changed) {
 		emit dataBaseUpdated();
 	}
 }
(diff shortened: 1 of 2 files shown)


ispyisail:
While testing something unrelated on current `master` I hit this error repeatedly in a session with **no deletions and no undo at all**, so I went looking for why. The trigger is broader than the title of this PR suggests, which I think strengthens the case for the fix.

### Ordinary element placement hits the same path

`DiagramEventAddElement::addElement()` deliberately removes and re-adds the *preview* element around the auto-conductor search, to stop it interfering with `Element::AlignedFreeTerminals()`:

```cpp
m_diagram->removeItem(m_element);   // sources/diagramevent/diagrameventaddelement.cpp:257
while (!element->AlignedFreeTerminals().isEmpty() && ... autoConductor()) { ... }
m_diagram->addItem(m_element);      // :272
```

`removeItem()` → `removeElement()` → `DELETE FROM element` only, orphaning the preview's `element_info` row. `addItem()` → `addElement()` → the `element` insert succeeds (that row really was freed) and the `element_info` insert then hits the orphan's PRIMARY KEY. So every commit click in place mode produces one failure.

### Measured

Fresh project on `master`, placing one element at a time, no deletes, no undo:

| action | cumulative `insert element info error` |
|---|---|
| drop (preview enters place mode) | 0 |
| 1st commit click | 1 |
| 2nd commit click | 2 |
| 3rd commit click | 3 |

Exactly one per placement. `insert element error` (the `element` table) stayed at **0** throughout — the same asymmetry that pointed at cleanup rather than a uuid collision in the first place.

### What this does and doesn't mean

To be precise about the impact: the failure on placement belongs to the *preview* element, not the real one. The real element is inserted once under a fresh uuid and both of its inserts succeed. So placement **leaks one permanent orphan row per add-element session** rather than corrupting user data — `element_info` accumulates rows keyed by uuids that no longer exist in `element`. They stay invisible to the summary query, which inner-joins `element`.

The data-correctness risk is still the undo/redo path already described in this PR, where a real element's `element_info` insert is the one that fails and its information goes stale.

What the placement path adds is that the defective code runs constantly during normal editing. Orphan rows are therefore guaranteed to be present in essentially every project, which makes the undo failure certain rather than dependent on an unlucky sequence. It also explains why this first showed up as felt "lag" in live use rather than only after an undo.

The fix in this PR covers both, since both come from the same missing `DELETE FROM element_info`. No change requested — just adding the evidence.

ispyisail:

PR #665, opened 2026-08-05
Add an event-loop responsiveness watchdog (discussion #644 follow-up)

ispyisail:
**Stacked on #647** (`feature-diagnostic-logging-crash`) — this diff includes that branch's commits until it merges; the only new files/changes are `sources/logging/eventloopwatchdog.{h,cpp}`, the two-line registration in `cmake/qet_compilation_vars.cmake`, and the construction + `start()` call in `main.cpp`.

## Why

[`QetLogger`](https://github.com/qelectrotech/qelectrotech-source-mirror/pull/646) (discussion #644, steps 1-3) only captures what an explicit `qDebug()`/`qInfo()`/`qWarning()` call already decided to report. Most of a session — painting, dragging, a slow synchronous operation — produces no log output at all, so a silent multi-second gap in the log can't be told apart from the user simply not doing anything.

That gap wasn't theoretical — a user testing a combined build reported the app "lagged," and diagnosing it meant inferring stalls from timestamp gaps between unrelated log lines, which is exactly the kind of thing that can't distinguish a real freeze from normal idle time. (The actual cause that time turned out to be a database bug, #664 — but confirming *whether and when* a stall genuinely happened took more effort than it should have.)

## What this adds

`EventLoopWatchdog`: a `QTimer::PreciseTimer` repeating tick every 50ms, measuring the *actual* elapsed time since the previous tick via `QElapsedTimer` (monotonic — unaffected by system clock/NTP adjustments, unlike `QDateTime`). Qt does not queue up missed fires for a normal repeating timer, so if the main thread blocks for 600ms, the timer fires once as soon as the loop frees up, with ~600ms measured since the last tick — that gap *is* the stall, measured at its source rather than inferred after the fact.

Only logs when a tick is late by more than 200ms (`kStallThresholdMs`), through the existing `qWarning()` path — so it reuses `QetLogger`'s file/ring/rotation with no new plumbing, and a healthy session produces zero output from this class, matching `QetLogger`'s own bounded-log design philosophy (see its class comment). Same `QET_WATCHDOG_DISABLE=1` escape-hatch convention as `QetLogger`'s `QET_LOG_DISABLE=1`.

**Deliberately not included:** attributing a stall to what caused it. This tells you a stall happened and how long — pairing that timestamp with gdb attached to a running session (as used for the CLI hang, #661) is still how you get from "it stalled" to a root cause. A watchdog and a profiler are different tools.

## Verification

Checked against the compiled binary, not just read: temporarily injected a `QThread::msleep(600)` via a one-shot `QTimer` 2s after startup, confirmed the exact expected warning through the real `qWarning()`/`QetLogger` path:

```
EventLoopWatchdog: main thread stalled for 620 ms (expected a tick every 50 ms)
```

then removed the test hook entirely and reconfirmed a normal run produces no output from this class at all. General regression suite (`tests/catch`) still passes.

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -116,10 +116,14 @@ set(QET_RES_FILES
 set(QET_SRC_FILES
   ${QET_DIR}/sources/cli_export.cpp
   ${QET_DIR}/sources/cli_export.h
+  ${QET_DIR}/sources/logging/crashhandler.cpp
+  ${QET_DIR}/sources/logging/crashhandler.h
   ${QET_DIR}/sources/logging/logring.cpp
   ${QET_DIR}/sources/logging/logring.h
   ${QET_DIR}/sources/logging/qetlogger.cpp
   ${QET_DIR}/sources/logging/qetlogger.h
+  ${QET_DIR}/sources/logging/ui/diagnosticsreportdialog.cpp
+  ${QET_DIR}/sources/logging/ui/diagnosticsreportdialog.h
   ${QET_DIR}/sources/pdf_links.cpp
   ${QET_DIR}/sources/pdf_links.h
   ${QET_DIR}/sources/import/edz/edzarchive.cpp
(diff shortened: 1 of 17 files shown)


ispyisail:
Rebased on current `master` alongside #647 (see [that reply](https://github.com/qelectrotech/qelectrotech-source-mirror/pull/647#issuecomment-5203907049) for the full detail). This branch is still stacked on #647 and is mergeable/clean again.

No `.pro` change was needed here — `eventloopwatchdog.{h,cpp}` sit directly in `sources/logging/`, which is already globbed. The `sources/logging/ui/` globs that were missing are added in #647.

CMake build re-verified after the rebase.

PR #668, opened 2026-08-06
Show a live IEC 81346-2 placeholder label when nothing else is configured

ispyisail:
## This PR has been reworked — please re-read

The original version of this PR invented a new `designation-letter` attribute on `qet_directory`, written without knowing QET already has a working, wired-in mechanism for exactly this: `qet_labels.xml` + `autonum::elementPrefixForLocation()`, populating `Element::m_prefix` at construction and already surfaced via the `%prefix` label-formula variable. That system isn't broken — verified live on current master, a coil with formula `%prefix` correctly renders its `qet_labels.xml` prefix — it's just under-populated, and the original version of this PR duplicated it instead of building on it. Full comparison (27 agree / 45 disagree against the curated data) is in the review on [qelectrotech-elements#69](https://github.com/qelectrotech/qelectrotech-elements/pull/69), which is closed in favor of contributing to `qet_labels.xml` directly instead.

## What this version actually does

When an element has no numbering formula and no manually-set label, `DynamicElementTextItem` shows `Element::getPrefix() + "?"` instead of nothing — the same prefix `%prefix` already resolves, just visible *before* a formula is configured rather than only after. No new data source, no new attribute, no new file format.

- Net diff: ~30 lines in `dynamicelementtextitem.cpp` (a small helper, used at the two existing label-rendering call sites), plus a 6-line reorder in `element.cpp`'s constructor (see below).
- Deliberately **not** part of `Element::actualLabel()`: that value gets written back into the element's stored information at five call sites in `element.cpp`, and a placeholder saved as if it were a real label would corrupt project data, not just look wrong. The placeholder lives at display time only.

## A real bug found while verifying this live

`Element`'s constructor called `buildFromXml()` (which constructs this element's `DynamicElementTextItem` children and triggers their first render) *before* `setPrefix()`. So every element's first paint saw an empty prefix, with nothing ever triggering a repaint to pick up the real value afterward. The existing `%prefix`-in-a-formula path never hit this, since formulas are evaluated on demand, long after construction — only a placeholder computed at first-render time exposes it. Fixed by moving `setPrefix()` before `buildFromXml()` — 6 lines, no behavior change for anything that isn't this new placeholder.

## Test plan

Live-verified (Xvfb, real `10_electric` + `qet_labels.xml` collection installed):
- A fresh, unconfigured coil from `310_relays_contactors_contacts/01_coils` shows **`K?`**.
- Setting its label formula to `%prefix` replaces it with **`K`** (placeholder gone, no `?`).
- Saving and reopening the project: unconfigured coils still show `K?`, and `elementInformations` stays empty for them — confirmed both by direct XML inspection of the saved file and by the live Informations panel showing empty Label/Formule fields after reload.
- A coil with `formula="%prefix"` round-trips correctly: `elementInformations` contains that real formula (not a placeholder), and it renders `K`.

### DIFF
--- a/sources/qetgraphicsitem/dynamicelementtextitem.cpp
+++ b/sources/qetgraphicsitem/dynamicelementtextitem.cpp
@@ -32,6 +32,30 @@
 #include <QDomElement>
 #include <QGraphicsSceneMouseEvent>
 
+namespace {
+	/**
+		@brief displayLabelFor
+		Purely presentational: when @element has no numbering formula and
+		no manually-set label yet, show its qet_labels.xml classification
+		prefix (Element::getPrefix(), already resolved at construction time
+		via autonum::elementPrefixForLocation()) as a "K?"-style placeholder,
+		so the folio reads as intentionally-unnumbered rather than blank.
+		Deliberately NOT part of Element::actualLabel(): that value gets
+		written back into the element's stored information in several
+		places (e.g. Element::setElementInformations()), and a placeholder
+		saved as if it were a real label would be a correctness bug, not a
+		display nicety.
+	*/
+	QString displayLabelFor(Element *element)
+	{
+		const QString label = element->actualLabel();
+		if (!label.isEmpty())
+			return label;
+		const QString prefix = element->getPrefix();
+		return prefix.isEmpty() ? QString() : prefix + QStringLiteral("?");
+	}
+}
+
 /**
 	@brief DynamicElementTextItem::DynamicElementTextItem
 	Constructor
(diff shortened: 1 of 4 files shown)


scorpio810:
@ispyisail : see https://github.com/qelectrotech/qelectrotech-elements/blob/main/10_electric/qet_labels.xml and search on forum, sorry for my delay I don't have the time or the health for it at the moment.

ispyisail:
Force-pushed a rework of this PR — it now reads the existing `qet_labels.xml`/`%prefix` mechanism instead of inventing a new attribute, and fixes a real element-construction ordering bug found along the way. See the updated PR description above for the full explanation; happy to answer questions about the change.

ispyisail:
Rebase-checked against current master: no conflicts, builds clean. Ready for review whenever you have a moment.

ispyisail:

PR #672, opened 2026-08-06
Fix element prefix lookup: relocated collections, non-10_electric trees, unbounded array

### LINKED ISSUE #671: Element prefix lookup (qet_labels.xml / %prefix): breaks on relocated collections, ignores non-10_electric trees, unbounded array
`autonum::elementPrefixForLocation()` (`sources/autoNum/assignvariables.cpp`) resolves an element's IEC 81346 designation prefix from `qet_labels.xml`, surfaced as the `%prefix` label-formula variable. Investigating why prefixes are widely reported as unreliable turned up several distinct defects. Filing them together since they're all in that one function; @scorpio810 asked for bug reports on this in [forum #2178](https://qelectrotech.org/forum/viewtopic.php?id=2178).

Everything below is reproduced against the shipped collection on current `master`, except where noted as inspection-only.

## 1. Relocating the common collection silently kills every prefix

Probably the one users actually hit. `QETApp::commonElementsDir()` returns the configured path **verbatim**, without appending a trailing `/` — unlike its sibling `customElementsDir()`, which explicitly normalises:

```cpp
if(!m_custom_element_dir.endsWith("/")) { m_custom_element_dir.append("/"); }
```

The lookup then concatenates naively (`assignvariables.cpp:788`):

```cpp
QString filepath = QETApp::commonElementsDir().append(qet_labels);   // "10_electric/qet_labels.xml"
```

If the configured path has no trailing slash, that yields `<path>10_electric/qet_labels.xml` — file not found, empty prefix, no error logged anywhere.

Default installs are immune because both build systems bake in a path that ends in `/` (`cmake/paths_compilation_installation.cmake`, `qelectrotech.pro`), which is why this doesn't show up in development. The `#ifndef QET_COMMON_COLLECTION_PATH` fallback at `qetapp.cpp:622` (`applicationDirPath() + "/elements"`) has no trailing slash either.

**Reproduced**: set `elements-collections/common-collection-path` to a valid collection path without a trailing slash, restart, place a coil from `10_allpole/310_relays_contactors_contacts/01_coils` (which has `<prefix>K</prefix>`), save, inspect the `.qet`:

| common-collection-path | saved `prefix` attribute |
|---|---|
| default (trailing `/`) | `prefix="K"` |
| relocated (no trailing `/`) | `prefix=""` |

Same binary, same collection content — only the path differs. This looks like the cause behind [#2651](https://qelectrotech.org/forum/viewtopic.php?id=2651) (relocating the file for a shared team library) and the report in [#2178](https://qelectrotech.org/forum/viewtopic.php?id=2178) that it breaks when things are "not in the default directory".

## 2. Common-collection trees other than `10_electric` read the wrong file

The branch condition is a literal folder-name comparison:

```cpp
if (current_location.fileName() != "10_electric") { /* read the CUSTOM collection */ }
else                                              { /* read the COMMON collection */ }
```

The common collection ships five top-level trees — `10_electric`, `20_logic`, `30_hydraulic`, `50_pneumatic`, `60_energy` — and only `10_electric` contains a `qet_labels.xml`. So an element from any of the other four fails the test, takes the custom branch, and goes looking in the **user's** collection.

**Reproduced** by adding a second common-collection tree with its own labels file (`99_test/qet_labels.xml` declaring `<category name="widgets"><prefix>Z</prefix></category>`) and placing an element from `99_test/widgets`: `prefix=""` on current master, where it should resolve `Z`.

This also means there is currently no way to add prefixes for `20_logic` etc. at all.

## 3. Fixed-size array with an unbounded index

```cpp
QString path[10];
int i = -1;
while (...) { i++; path[i] = current_location.fileName(); ... }   // nothing bounds i
```

The deepest category in the shipped collection needs 9 of those 10 slots (8 directory levels + the element). A custom collection can nest arbitrarily deep, and overflowing writes `QString` objects past the end of a stack array.

*Inspection + depth measurement only — I did not build a deep enough custom collection to force the overflow.*

## 4. The company collection is never considered

`QETApp::companyElementsDir()` is a third collection; it isn't handled, and falls into the "custom" branch by default.

## 5. The XML matching ignores nesting

The scan is a flat token walk matching any element whose `name` attribute equals the next path segment, with **no check that the match is a child of the previous match**. It produces correct results on the shipped file only because depth-first document order happens to line up with the hierarchy.

Relatedly, the documented inheritance rule ("If a directory does not have a prefix the element will assign its parent directory prefix and so on") is implemented by scanning *forward* with `readNextStartElement()`/`skipCurrentElement()`, which can only ever move deeper or sideways, never back up to a parent. It gives the right answer only because the file format mandates that `<prefix>` appear *after* child `<category>` elements — a constraint documented in `qet_labels.xml`'s own header comment that exists purely to accommodate this parser.

## 6. Minor

- `file.isReadable();` at lines 748 and 790 discards its return value and does nothing.
- The common and custom branches are ~40 lines of copy-paste differing only in the file path.

## Underlying design note

Which collection an element originally came from is **not recoverable** after import: `XmlElementCollection::addElement()` builds the embedded path as `"import/" % location.collectionPath(false)`, and `collectionPath(false)` strips the `common://` / `custom://` / `company://` protocol. That's why the code resorts to guessing from a hardcoded folder name. Any fix has to either try each collection in turn, or preserve the origin at import time.

---

I have a fix for 1–6 and will open a PR shortly, but wanted the analysis on record separately in case you'd rather take a different approach — particularly on whether `commonElementsDir()` should guarantee a trailing separator (only one caller in the tree concatenates onto it directly, and it's the buggy line above).

ispyisail:
Fixes #671.

`autonum::elementPrefixForLocation()` resolves an element's IEC 81346 designation prefix from `qet_labels.xml` (surfaced as the `%prefix` label-formula variable). The analysis in #671 found six defects in it; this rewrites the function rather than patching around them, because several dissolve entirely once the parsing is done properly. Net: 115 lines → ~50, plus a ~45-line helper.

## What changes

1. **Relocated common collections work again.** `commonElementsDir()` returns the configured path verbatim with no trailing separator, and the old code did `commonElementsDir().append("10_electric/qet_labels.xml")` — producing a mangled path and a silent empty prefix. Now joined with `QDir::filePath()`.
2. **Each common-collection top-level tree gets its own candidate**, so `20_logic`, `30_hydraulic` etc. can carry a `qet_labels.xml` instead of being sent to the user's collection by the hardcoded `!= "10_electric"` test.
3. **`QString path[10]` with an unbounded `i++` replaced by a `QStringList`.**
4. **The company collection is now considered** alongside common and custom.
5. **Nesting-aware parsing.** The old flat token scan matched any element whose `name` attribute equalled the next path segment with no parent/child check, and implemented "inherit the parent's prefix" by scanning *forward* — which only works because the file format requires `<prefix>` after child `<category>` elements. Replaced with a `QDomDocument` walk that descends by actual nesting and remembers the nearest ancestor prefix, so ordering within a category no longer matters. Existing files keep working.
6. Removed two no-op `file.isReadable();` statements and ~40 lines of copy-paste between the two collection branches.

## Behaviour change worth flagging

An element in a directory the labels file doesn't mention now inherits its **nearest listed ancestor's** prefix instead of getting none. That's what the format documents ("If a directory does not have a prefix the element will assign its parent directory prefix and so on") — the old forward-scan couldn't implement it. Practically it means more elements get a sensible prefix than before, but it is a change.

## On collection identity

Which collection an element came from isn't recoverable post-import (`addElement()` builds the embedded path from `collectionPath(false)`, which strips the protocol), so candidates are tried in turn and the first that describes the element's path wins. Custom files written against the common-collection layout are still accepted, so existing setups don't break.

If you'd rather fix this at the root by having `commonElementsDir()` guarantee a trailing separator, that's a smaller change — only one caller in the tree concatenates onto it directly, and it's the buggy line. I went with the local fix to keep the blast radius small; happy to switch.

## Test plan

Verified against the real collection with A/B runs of the same binary before and after, reading the `prefix` attribute out of the saved `.qet`:

| case | before | after |
|---|---|---|
| relocated common collection (no trailing `/`) | `prefix=""` | `prefix="K"` |
| default install | `prefix="K"` | `prefix="K"` (no regression) |
| second common tree with its own `qet_labels.xml` | `prefix=""` | `prefix="Z"` |
| inheritance: element under `90_terminal_strips_diagram` (no prefix of its own) | `prefix="X"` | `prefix="X"` |

The coil used is from `10_allpole/310_relays_contactors_contacts/01_coils`, which declares `<prefix>K</prefix>`.

Not reproduced, fixed by construction: the `path[10]` overflow (would need a custom collection nested >9 deep) — see #671 item 3.

### DIFF
--- a/sources/autoNum/assignvariables.cpp
+++ b/sources/autoNum/assignvariables.cpp
@@ -24,6 +24,9 @@
 #include "../qetgraphicsitem/element.h"
 #include "../qetxml.h"
 #include "../qetproject.h"
+#include <QDir>
+#include <QDomDocument>
+#include <QFile>
 #include <QStringList>
 #include <QVariant>
 #include <utility>
@@ -704,6 +707,54 @@ namespace autonum
 	}
 
 	/**
+		@brief labelsPrefixFor
+		Look up @a dir_segments (outermost category first) in the
+		qet_labels.xml at @a file_path.
+		A category without a <prefix> of its own inherits the nearest
+		ancestor that has one, as documented in that file.
+		@return the prefix that applies, or a null QString if the file
+			cannot be read or nothing along the path carries one.
+	*/
+	static QString labelsPrefixFor(
+			const QString &file_path, const QStringList &dir_segments)
+	{
+		QFile file(file_path);
+		if (!file.open(QIODevice::ReadOnly | QIODevice::Text))
+			return QString();
+
+		QDomDocument document;
+		if (!document.setContent(&file))
+			return QString();
+
+		QDomElement node = document.documentElement();
+		if (node.isNull())
+			return QString();
+
+		QString prefix;
+		for (const QString &segment : dir_segments)
+		{
+			QDomElement child =
+					node.firstChildElement(QStringLiteral("category"));
+			while (!child.isNull()
+				   && child.attribute(QStringLiteral("name")) != segment) {
+				child = child.nextSiblingElement(
+							QStringLiteral("category"));
+			}
+				//Deepest category this file describes: keep whatever an
+				//ancestor already provided rather than losing it.
+			if (child.isNull())
+				break;
+
+			node = child;
+			const QString own = node.firstChildElement(
+						QStringLiteral("prefix")).text().trimmed();
+			if (!own.isEmpty())
+				prefix = own;
+		}
+		return prefix;
+	}
+
+	/**
 		@brief elementPrefixForLocation
 		@param location
 		@return the prefix for an element represented by location,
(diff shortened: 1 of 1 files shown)


ispyisail:
Rebase-checked against current master: no conflicts, builds clean. Ready for review whenever you have a moment.

ispyisail:

ispyisail:
Closing this out for good — #686 (@IBSYSLevi's PR) covers the same ground and is further along in review, so it's the one going in.

The two most load-bearing fixes from this PR's analysis are now on #686 as well:

- the unbounded `QStringList` replacing `QString path[10]` (#671 item 3 — a deep custom collection could overflow a fixed-size stack array)
- the `QDir::filePath()` fix for the relocated-collection path join (#671 item 1 — `commonElementsDir() + "10_electric/qet_labels.xml"` silently mangles into one word when the configured path has no trailing separator; likely the most-reported cause of "prefixes just don't work")

Along with those, review on #686 turned up a real bug in the refactored lookup itself — an empty `<prefix/>` override not stopping the search — and I found one more while fixing that: the token-matching loop advances twice per matched element, which only produces the right result on a pretty-printed file; a minified `qet_labels.xml` silently returns no prefix at all. Both are fixed there too, verified against the real shipped `qet_labels.xml` in a standalone harness.

What #672 doesn't carry over: the `QDomDocument`-based nesting-aware rewrite (#671 item 5) and per-tree common-collection candidates (#671 item 2) aren't in #686. Both are real, still-open gaps — I'll file them separately if they turn out to matter in practice, rather than block #686 on a bigger rewrite.

ispyisail:
Filed the two remaining items as promised: #837, stacked on #686.

PR #673, opened 2026-08-06
show terminalnames in export

Kellermorph:
Terminal names/labels are now always rendered in PDF, PNG, and other exports, even when the "Dessiner les bornes" (Draw terminals) checkbox is not active.

Previously, Diagram::setDrawTerminals(false) called Terminal::setVisible(false), which hid the entire terminal QGraphicsObject — including the name label. This happened because both the terminal visual (red line + docking point) and the name are drawn inside Terminal::paint(), and Qt skips paint() entirely for invisible items.

The fix changes setDrawTerminals() to only set a flag and trigger a repaint, instead of hiding terminals. Terminal::paint() now checks diagram()->drawTerminals() to decide whether to draw the red line and docking point, but always renders the name/label regardless of the flag.

### DIFF
--- a/sources/diagram.cpp
+++ b/sources/diagram.cpp
@@ -2397,9 +2397,10 @@ QPointF Diagram::snapToGrid(const QPointF &p)
 	\~French true pour afficher les bornes, false sinon
 */
 void Diagram::setDrawTerminals(bool dt) {
+	draw_terminals_ = dt;
 	foreach(QGraphicsItem *qgi, items()) {
 		if (Terminal *t = qgraphicsitem_cast<Terminal *>(qgi)) {
-			t -> setVisible(dt);
+			t -> update();
 		}
 	}
 }
--- a/sources/diagram.cpp
+++ b/sources/diagram.cpp
@@ -67,6 +67,7 @@ Diagram::Diagram(QETProject *project) :
 	m_project                (project),
 	use_border_              (true),
 	draw_terminals_          (true),
+	draw_terminal_names_     (true),
 	draw_colored_conductors_ (true),
 	m_event_interface        (nullptr),
 	m_freeze_new_elements    (false),
@@ -2319,6 +2320,7 @@ ExportProperties Diagram::applyProperties(
 	old_properties.draw_border = border_and_titleblock.borderIsDisplayed();
 	old_properties.draw_titleblock = border_and_titleblock.titleBlockIsDisplayed();
 	old_properties.draw_terminals          = drawTerminals();
+	old_properties.draw_terminal_names     = drawTerminalNames();
 	old_properties.draw_colored_conductors = drawColoredConductors();
 	old_properties.exported_area = useBorder() ? QET::BorderArea
 						   : QET::ElementsArea;
@@ -2327,6 +2329,7 @@ ExportProperties Diagram::applyProperties(
 	// applique les nouvelles options de rendu
 	setUseBorder             (new_properties.exported_area == QET::BorderArea);
 	setDrawTerminals         (new_properties.draw_terminals);
+	setDrawTerminalNames     (new_properties.draw_terminal_names);
 	setDrawColoredConductors (new_properties.draw_colored_conductors);
 	setDisplayGrid           (new_properties.draw_grid);
 	setDisplayGuides         (new_properties.draw_guides);
@@ -2398,6 +2401,20 @@ QPointF Diagram::snapToGrid(const QPointF &p)
 */
 void Diagram::setDrawTerminals(bool dt) {
 	draw_terminals_ = dt;
+	foreach(QGraphicsItem *qgi, items()) {
+		if (Terminal *t = qgraphicsitem_cast<Terminal *>(qgi)) {
+			t -> update();
+		}
+	}
+}
+
+/**
+	@brief Diagram::setDrawTerminalNames
+	Defines whether or not to display the terminal names/labels
+	@param dt : true to display the terminal names, false otherwise
+*/
+void Diagram::setDrawTerminalNames(bool dt) {
+	draw_terminal_names_ = dt;
 	foreach(QGraphicsItem *qgi, items()) {
 		if (Terminal *t = qgraphicsitem_cast<Terminal *>(qgi)) {
 			t -> update();
--- a/sources/diagram.h
+++ b/sources/diagram.h
@@ -126,8 +126,9 @@ class Diagram : public QGraphicsScene
 		bool use_border_;
 		bool draw_guides_;
 		QList<Diagram::Guide> m_guides_list;
-		bool draw_terminals_;
-		bool draw_colored_conductors_;
+	bool draw_terminals_;
+	bool draw_terminal_names_;
+	bool draw_colored_conductors_;
 
 		QString m_conductors_autonum_name;
 		DiagramEventInterface *m_event_interface;
@@ -226,6 +227,8 @@ class Diagram : public QGraphicsScene
 	
 		bool drawTerminals() const;
 		void setDrawTerminals(bool);
+		bool drawTerminalNames() const;
+		void setDrawTerminalNames(bool);
 		bool drawColoredConductors() const;
 		void setDrawColoredConductors(bool);
 	
@@ -424,6 +427,15 @@ inline QGIManager &Diagram::qgiManager() {
 inline bool Diagram::drawTerminals() const
 {
 	return(draw_terminals_);
+}
+
+/**
+	@brief Diagram::drawTerminalNames
+	@return true if terminal names are rendered, false otherwise
+*/
+inline bool Diagram::drawTerminalNames() const
+{
+	return(draw_terminal_names_);
 }
 
 /**
--- a/sources/diagram.h
+++ b/sources/diagram.h
@@ -126,9 +126,9 @@ class Diagram : public QGraphicsScene
 		bool use_border_;
 		bool draw_guides_;
 		QList<Diagram::Guide> m_guides_list;
-	bool draw_terminals_;
-	bool draw_terminal_names_;
-	bool draw_colored_conductors_;
+		bool draw_terminals_;
+		bool draw_terminal_names_;
+		bool draw_colored_conductors_;
 
 		QString m_conductors_autonum_name;
 		DiagramEventInterface *m_event_interface;
--- a/sources/exportproperties.cpp
+++ b/sources/exportproperties.cpp
@@ -34,7 +34,7 @@ ExportProperties::ExportProperties() :
 	draw_border(true),
 	draw_titleblock(true),
 	draw_terminals(false),
-	draw_terminal_names(true),
+	draw_terminal_names(false),
 	draw_bg_transparent(false),
 	draw_colored_conductors(true),
 	exported_area(QET::BorderArea)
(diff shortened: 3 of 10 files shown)


ispyisail (approved):
Reviewed and built (CMake/Ninja Release, Qt 5.15.18) — compiles clean. The approach looks right to me, and I think it quietly fixes a second bug beyond the one you describe. A few things I checked, and one caveat worth knowing about.

## It also fixes `drawTerminals()` being permanently stale

Worth calling out because it isn't in the description. On `master`, `draw_terminals_` is written **only** in the constructor initialiser (`draw_terminals_(true)` — diagram.cpp:69) and nowhere else; `setDrawTerminals()` only ever called `setVisible()`. So `Diagram::drawTerminals()` always returned `true`, no matter what.

That matters for `applyProperties()` (diagram.cpp:2321), which is how export/print saves and restores render options:

```cpp
old_properties.draw_terminals = drawTerminals();   // always true on master
...
setDrawTerminals(new_properties.draw_terminals);
```

so the "restore afterwards" step always restored terminals to *on* rather than to whatever they actually were. Adding `draw_terminals_ = dt;` makes the accessor honest and the round-trip correct. Might be worth a line in the commit message, since someone bisecting a terminal-visibility change later will want to know.

- **Label clipping** — no issue: `Terminal::boundingRect()` already unites `m_br` with the label rect (including the `-2/+2` padding), so painting the name on a terminal that no longer draws its marker stays inside the reported bounds.
- **Nothing else depended on terminals being hidden** — `setVisible` on a `Terminal` appears in exactly one place in the tree, the line this PR replaces, and `setDrawTerminals()` has exactly one caller (`applyProperties`). So the visible→always-painted switch has no other reader to upset.
- **Export area is unchanged.** This was my main worry: with terminals no longer hidden, would they start contributing to `itemsBoundingRect()` and enlarge the exported region when "elements area" is used (`toPaintDevice()`, diagram.cpp:674)? Checked it rather than assumed — `QGraphicsScene::itemsBoundingRect()` includes hidden items anyway:

  ```
  both visible:       QRectF(-0.5,-0.5 111x111)
  second item hidden: QRectF(-0.5,-0.5 111x111)
  ```

  So no change to the exported area either way.
- **Semantics** — gating the marker on `drawTerminals()` while leaving the name on the terminal's own `show_name` flag (default `false`, set per-element from XML) reads as the right split: "Draw terminals" is about the red marker, `show_name` is about the label. A user who deliberately authored `show_name="true"` presumably wants it in the output.

## One caveat: "always rendered" isn't quite always

`Terminal::paint()` still starts with

```cpp
if (options && options->levelOfDetailFromTransform(painter->worldTransform()) < 0.5)
    return;
```

which sits *above* the label block, so names are still dropped entirely below 0.5 level-of-detail. Exports go through `toPaintDevice(pix, width, height, ...)`, so an export scaled to under ~50% of the diagram's natural size will still lose the names.

Pre-existing, not something this PR introduces, and arguably fine since the text would be unreadable at that scale — but it does mean the PR description's "always rendered in PDF, PNG and other exports" has an asterisk. If you wanted names to genuinely always survive, the LOD early-return would need to move below the label block (or gain a `show_name` exemption). I'd be inclined to leave it and just not promise "always" — your call.

## Minor

`setDrawTerminals()` now calls `t->update()` on every terminal unconditionally, including when the flag hasn't actually changed. Harmless (it's only called from the export path), but an early `if (draw_terminals_ == dt) return;` would make it a no-op in that case.

Nothing blocking from me — happy to see this go in.

plc-user:
> Terminal names/labels are now always rendered in PDF, PNG, and other exports, even when the "Dessiner les bornes" (Draw terminals) checkbox is not active.

Maybe the user unchecked this box intentionally?
I'd prefer an additional checkbox to decide whether to add terminal-labels to pdf-export or not.

Kellermorph:
Thanks for the thorough review 

Changes made
New checkbox "Dessiner les noms des bornes" (Draw terminal names)

Added a dedicated checkbox to independently control terminal name/label visibility in exports, separate from the existing "Dessiner les bornes" (Draw terminals) checkbox which controls the red terminal marker.

ExportProperties: Added draw_terminal_names bool (default: true), with serialization via toSettings/fromSettings
ExportPropertiesWidget: Added checkbox "Dessiner les noms des bornes" in the export dialog (PNG/SVG/PDF export)
ProjectPrintWindow: Added checkbox in the print dialog (PDF print)
Diagram: Added draw_terminal_names_ member, drawTerminalNames() getter, setDrawTerminalNames() setter, wired through applyProperties() for save/restore round-trip
Terminal::paint(): Label drawing now checks diagram()->drawTerminalNames() in addition to m_show_name
This means users can now:

Show terminal markers only (check "Draw terminals", uncheck "Draw terminal names")
Show terminal names only (uncheck "Draw terminals", check "Draw terminal names")
Show both or neither

ispyisail (approved):
Re-reviewed after the checkbox commit. Builds clean (CMake/Ninja Release, Qt 5.15.18). @plc-user's point was the right one to raise and this is the right answer to it — splitting the marker and the label into two independent options is better than either "always" or "never".

The wiring is complete and mirrors the existing `draw_terminals` path exactly, which is what I'd want to see here:

- member + constructor initialiser, getter/setter
- **both** directions in `applyProperties()` — saved into `old_properties` *and* applied from `new_properties`. Worth noting because that save/apply symmetry is precisely what was broken for `draw_terminals` before this PR, so it would have been easy to reproduce the same bug in the new field.
- `toSettings`/`fromSettings` with matching defaults, and `defaultPrintProperties()` picks it up for free since it routes through `fromSettings`
- both dialogs (export widget + print window), plus the `on_m_draw_terminal_names_cb_clicked()` preview-refresh slot following the existing auto-connect naming
- the null-diagram guard `(!diagram() || diagram()->drawTerminalNames())` matches the one used for the marker

Nothing blocking. Three notes, one of which is a decision rather than a defect.

## 1. Accidental indentation change in `diagram.h`

The new member came in at one tab instead of two, and pulled two pre-existing lines down with it:

```
		QList<Diagram::Guide> m_guides_list;
	bool draw_terminals_;             <- was two tabs
	bool draw_terminal_names_;
	bool draw_colored_conductors_;    <- was two tabs
		QString m_conductors_autonum_name;
```

`draw_terminals_` and `draw_colored_conductors_` didn't need to change at all. Worth restoring so the block stays aligned and the diff only shows the real addition.

## 2. The default still changes existing output — worth being deliberate about

`draw_terminal_names` defaults to `true` (constructor and `fromSettings`), while `draw_terminals` defaults to `false`. So for a user who never opens the new checkbox, the shipping combination is "no markers, but names".

Concretely: someone who authored terminals with `show_name="true"` and exports today gets no names (the old code hid the whole item); after this they will. That is probably what they want — `show_name` defaults to `false` per terminal and has to be set deliberately in the element, so the affected population is exactly the people who asked for names in the first place.

But it is still a change in default output, which is the flavour of thing @plc-user was flagging. Defaulting to `false` would make it purely opt-in and preserve every existing export byte-for-byte. I don't have a strong view — just worth being an explicit choice rather than a side effect.

## 3. Minor

`applyProperties()` now calls `setDrawTerminals()` and `setDrawTerminalNames()` back to back, and each walks every item in the scene calling `update()`. Two full passes where one would do; irrelevant for correctness, and only on the export path, but easy to collapse if you're touching it anyway.

## Still true from my earlier pass

`Terminal::paint()` keeps its early return above the label block:

```cpp
if (options && options->levelOfDetailFromTransform(painter->worldTransform()) < 0.5)
    return;
```

so names still vanish below 0.5 level-of-detail regardless of the new checkbox — an export scaled under ~50% won't show them even with the box ticked. Pre-existing and arguably fine (unreadable at that size), but now that there's a checkbox explicitly promising terminal names, the gap between the promise and the behaviour is a little more visible than it was. Fine to leave; just don't be surprised by the bug report.

Approving — the whitespace nit aside, this is a clean, complete implementation.

Kellermorph:
Whtiespace is fixed

plc-user:
Das sieht doch schon schön aus!
An die lokalisierten Texte zumindest auf Deutsch denkst Du noch? 
@ispyisail hat die gute Idee, die Standard-Einstellung auf "false" zu setzen.

English:
That's looking pretty good already!
Are you still thinking about the localized texts at least in German?
@ispyisail had the good idea to set the default setting to “false.”

Kellermorph:
Natürlich denke ich noch da dran.

Ich warte jetzt ab, was es in diesem Thema gibt
https://github.com/qelectrotech/qelectrotech-source-mirror/pull/639
Wenn das alles fertig ist, mache ich in einem Schwung die ganzen deutschen Texte fertig.

Aus meiner Sicht ist die Standardeinstellung auf False verkehrt.
Wenn jemand die neue Funktion nutzt, dass die Anschlusstexte vom Master auf den Slave übertragen werden, will er diese auch im Export sehen. Nutzt jemand diese Funktion nicht, ist es auch egal wenn diese Funktion auf True steht, da sich am Export dann nichts ändert.
So muss man immer daran denken, die Checkbox zu aktivieren.

Englisch:

Of course, I’m still thinking about it.

I’m waiting to see what comes up in this thread:
https://github.com/qelectrotech/qelectrotech-source-mirror/pull/639
Once all that’s done, I’ll finish up all the German texts in one go.

In my opinion, the default setting of “False” is wrong.
If someone uses the new feature that transfers the connection texts from the master to the slave, they’ll want to see them in the export as well. If someone doesn’t use this feature, it doesn’t matter if the setting is set to “True,” since the export won’t change either way.
So you always have to remember to check the box.

Translated with DeepL.com (free version)

plc-user:
> (...) So muss man immer daran denken, die Checkbox zu aktivieren.

Das wird doch hoffentlich nicht nötig sein: 
Die Status anderen CheckBoxen werden doch auch in den Settings gespeichert ... dieser nicht?

ispyisail:
@plc-user is right, checked the diff to be sure rather than assume: `draw_terminal_names` round-trips through `QSettings` exactly the same way `draw_terminals` (and every other export checkbox) already does —

```cpp
// ExportProperties::toSettings()
settings.setValue(prefix % "drawterminals", draw_terminals);
settings.setValue(prefix % "drawterminalnames", draw_terminal_names);   // symmetric

// ExportProperties::fromSettings()
draw_terminals = settings.value(prefix % "drawterminals", false).toBool();
draw_terminal_names = settings.value(prefix % "drawterminalnames", true).toBool();  // symmetric
```

So the *default* only ever matters the very first time someone opens the export dialog on a fresh install with no saved settings. After that, whatever they last checked persists automatically for every export after — the same as every other checkbox in that dialog already behaves. "You always have to remember to activate it" isn't actually a consequence of the default being `false`; nothing about this mechanism requires re-checking it each time.

That doesn't settle what the *first-ever* default should be — @Kellermorph's point about #639's master→slave terminal-name feature making names more valuable in exports is a separate, legitimate consideration, and reasonable people could land either way on it. But it does mean that specific objection to `false` isn't accurate, so it shouldn't be what the decision hinges on.

Kellermorph:
> > (...) So muss man immer daran denken, die Checkbox zu aktivieren.
> 
> Das wird doch hoffentlich nicht nötig sein: Die Status anderen CheckBoxen werden doch auch in den Settings gespeichert ... dieser nicht?

Du hast recht. Da war ich gestern zu übereifrig. Es wird gespeichert. 
Jetzt ist es auf standard fault gesetzt. Danke euch beiden.

English
You're right. I was a little too eager yesterday. It's been saved.
Now it's set to “standard fault.” Thanks to both of you.

Kellermorph:
if this one is merged too, i do the translations

scorpio810:
> if this one is merged too, i do the translations

Done, Thomas.
https://github.com/qelectrotech/qelectrotech-source-mirror/commit/d2b2a2eadbb8699bc2541794ebd0da9d0802bac4

PR #679, opened 2026-08-07
Fixing minimum width calculation for title block

arummler:
When double clicking on "A4_1" template to open it the program crashed with:
Fatal: ASSERT: "!std::isnan(value)" in file /usr/include/x86_64-linux-gnu/qt6/QtCore/qnumeric.h, line 501 (/usr/include/x86_64-linux-gnu/qt6/QtCore/qnumeric.h:501, )

That was only the case for this template as it uses only columns with relative sizes causing the minimum width calculation to go bonkers. Introduced a guard to avoid division by zero. Trying again in the newest published version it does not happen anymore although nothing touched the template nor this particular place -- not clear why. In any case guarding it cannot hurt as a preventive measure.

### DIFF
--- a/sources/titleblock/templateview.cpp
+++ b/sources/titleblock/templateview.cpp
@@ -999,20 +999,34 @@ void TitleBlockTemplateView::updateDisplayedMinMaxWidth()
 	int max_width = tbtemplate_ -> maximumWidth();
 
 	QString min_max_width_sentence;
-	if (max_width != -1) {
+	if (min_width != -1 && max_width != -1) {
 		min_max_width_sentence = QString(
 			tr(
 				"Longueur minimale : %1px\nLongueur maximale : %2px\n",
 				"tooltip showing the minimum and/or maximum width of the edited template"
 			)
 		).arg(min_width).arg(max_width);
-	} else {
+	} else if (min_width != -1) {
 		min_max_width_sentence = QString(
 			tr(
 				"Longueur minimale : %1px\n",
 				"tooltip showing the minimum width of the edited template"
 			)
 		).arg(min_width);
+	} else if (max_width != -1) {
+		min_max_width_sentence = QString(
+			tr(
+				"Longueur maximale : %1px\n",
+				"tooltip showing the maximum width of the edited template"
+			)
+		).arg(max_width);
+	} else {
+		min_max_width_sentence = QString(
+			tr(
+				"Longueur non contrainte.\n",
+				"tooltip shown when the edited template has neither a minimum nor a maximum width constraint"
+			)
+		);
 	}
 
 	// the tooltip may also display the split label for readability purpose
--- a/sources/titleblock/templateview.cpp
+++ b/sources/titleblock/templateview.cpp
@@ -999,34 +999,47 @@ void TitleBlockTemplateView::updateDisplayedMinMaxWidth()
 	int max_width = tbtemplate_ -> maximumWidth();
 
 	QString min_max_width_sentence;
-	if (min_width != -1 && max_width != -1) {
-		min_max_width_sentence = QString(
-			tr(
-				"Longueur minimale : %1px\nLongueur maximale : %2px\n",
-				"tooltip showing the minimum and/or maximum width of the edited template"
-			)
-		).arg(min_width).arg(max_width);
-	} else if (min_width != -1) {
-		min_max_width_sentence = QString(
-			tr(
-				"Longueur minimale : %1px\n",
-				"tooltip showing the minimum width of the edited template"
-			)
-		).arg(min_width);
-	} else if (max_width != -1) {
-		min_max_width_sentence = QString(
-			tr(
-				"Longueur maximale : %1px\n",
-				"tooltip showing the maximum width of the edited template"
-			)
-		).arg(max_width);
-	} else {
-		min_max_width_sentence = QString(
-			tr(
-				"Longueur non contrainte.\n",
-				"tooltip shown when the edited template has neither a minimum nor a maximum width constraint"
-			)
-		);
+
+	switch (static_cast<TitleBlockTemplate::WidthConstraintCase>(min_width)) {
+		case TitleBlockTemplate::WidthConstraintCase::RelativeWidthExceeds100Percent:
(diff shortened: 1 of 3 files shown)


ispyisail (approved):
Reviewed and built (CMake/Ninja Release, Qt 5.15.18) — compiles clean, no warnings on the changed file.

I tracked down the actual repro rather than just trusting the description. `titleblocks/A4_1.titleblock` has this `cols` attribute:

```
cols="t6%;t4%;t2%;t1%;t7%;t3%;t4%;t3%;t4%;t22%;t18%;t4%;t17%;t5%;"
```

All fourteen columns are `t`-prefixed (`RelativeToTotalLength`, per the regex in `parseColumns()`), and they sum to exactly 100. So `columnTypeTotal(QET::Absolute)` is 0 and `columnTypeTotal(QET::RelativeToTotalLength)` is 100, which makes the old denominator `(100.0 - 100) / 100.0 == 0.0`, and the division becomes `0 / 0.0` — an actual `0.0/0.0`, which is `NaN` under IEEE 754, not `+inf`. That lines up exactly with the reported `!std::isnan(value)` assert; the crash isn't generic division-by-zero, it's specifically the "zero absolute columns *and* relative columns summing to exactly 100" case. Worth knowing because it explains why this was so template-specific rather than a broader crash — I'd guess it's why it stopped reproducing when you tried again after nothing touched the template: `qRound`'s NaN assert is new in Qt 6, so it likely depends on which Qt build you launched against, not the template.

I also checked the other branch the guard now has to handle, since the fix folds two different situations (`RTT == 100` and `RTT > 100`) into one `denominator <= 0.0` check:

- `RTT == 100, abs_total > 0`: re-deriving the inequality by hand, this really is "no finite width satisfies it" — the derivation in the comment above (`TOT >= sum(ABS)/((100-sum(REL))/100)`) forces `TOT` to infinity here, so `INT_MAX` is the mathematically honest answer, not an arbitrary sentinel.
- `RTT > 100`: the denominator goes negative, and I worked through what dividing by a negative denominator does to the original inequality (`TOT >= sum(ABS)/D` becomes `TOT <= sum(ABS)/D` once you divide by a negative number and flip the sign). For `abs_total == 0` that reduces to `TOT <= 0`, i.e. `TOT == 0` is the only self-consistent solution — matching the `0` this fix returns. For `abs_total > 0` it reduces to "no non-negative `TOT` satisfies this," same as the `RTT == 100` case — matching `INT_MAX`. So the two-way ternary isn't a simplification that loses a case, it's actually correct for both branches of `denominator <= 0.0`, not just the one in the bug report.

One thing I checked because it looked like a possible regression: for `RTT > 100` the *old* code didn't crash — `abs_total` (int, promoted to double) divided by a negative denominator is a finite negative (or `-0.0`) double, and `qRound` of a finite value is fine. So pre-fix, a malformed `RTT > 100` template silently returned a *negative* "minimum width." Post-fix it returns `0` or `INT_MAX` depending on `abs_total`. Both are visible only in a tooltip (`TitleBlockTemplateView::updateDisplayedMinMaxWidth()`, the only caller of `minimumWidth()`), so this is a display-only change and an improvement either way — a negative minimum was already nonsensical, and now it's at least self-consistent with the derivation.

No other callers of `minimumWidth()` to worry about — grep turns up exactly the one, in the tooltip builder.

Minor, non-blocking: in the `abs_total > 0` branch, the tooltip will render `Longueur minimale : 2147483647px`. That's only reachable for a template whose relative columns already overflow the page, so it's a cosmetic rough edge on an already-broken template rather than something a real template will ever show — not asking for a change, just flagging it in case a friendlier fallback (e.g. reusing `-1` the way `maximumWidth()` does for "no constraint," with `updateDisplayedMinMaxWidth()` adjusted to skip the line) is easy enough to be worth doing while you're in here.

Approving — the guard is correct, it's derived from the same math as the comment above it rather than guessed, and it fixes the exact template that was crashing.

arummler:
Fixed and there was another undiscovered bug affecting the tool tip of the large horizontal bar in the template editor.

ispyisail (approved):
Reviewed 99151b9c0 — this is exactly what I'd suggested (reusing `maximumWidth()`'s `-1` = "no constraint" convention). Reconfirmed `minimumWidth()` still has exactly one caller (`templateview.cpp:998`) and traced the new four-way branch in `updateDisplayedMinMaxWidth()`; it's correct and covers the common case well (a template made only of relative columns, the `abs_total == 0` branch of the old fix).

One edge case worth flagging, non-blocking: the `-1` sentinel now also covers RTT-columns-summing->=100% *with* absolute columns present (`abs_total > 0`) — the branch your div-by-zero fix used to report as `INT_MAX`. That's not "no constraint," it's "no finite width satisfies this template": its own relative columns already claim >=100%, so any absolute-width columns on top can never fit. That state is reachable through the GUI today — the per-column spinbox caps each `RelativeToTotalLength` column at 100 (`dimensionwidget.cpp:184`), but there's no check on the *sum* across columns, so e.g. two RTT columns at 60% each plus one Absolute column gets you there. Since `maximumWidth()` is also `-1` whenever any non-Absolute column exists, this now surfaces as "Longueur non contrainte." — telling the user their template has no width constraints at all, when it's closer to the opposite (over-constrained and unsatisfiable). The old `INT_MAX` display was ugly but at least signaled something was wrong; this trades that for a friendly message that hides it.

Doesn't need to block on this — it only affects an already-broken template state, no crash risk, and the mainline cases are unambiguously improved. If it's worth closing later, distinguishing "unbounded" (`abs_total == 0`) from "impossible" (`abs_total > 0`) with a separate signal would do it, but that's a call on how much this corner deserves attention.

Small nit while in there: `return -1;` at `titleblocktemplate.cpp:985` lost a tab of indentation relative to the surrounding block (sits level with the `if` instead of nested inside it) — cosmetic only.

Approving — the crash fix stands, and this is a genuine improvement over `INT_MAX` for the case that matters day to day.

arummler:
Okay, it might be overkill but taking the opportunity this commit should cover all the cases. Also adding proper doxygen documentation for the two methods (another big effort for the whole project). Translations missing to be done later when merged. Code embedded French to be checked by a native speaker, that's what the translator delivered.

PR #682, opened 2026-08-07
Reject non-finite element positions and illegal XML control bytes

ispyisail:
Two independent crash/hang bugs, found by an adversarial mutation harness I built this session specifically to stress-test `--resave` against corrupted real project files and check results with gdb rather than guess.

## 1. A single NaN/Inf coordinate hangs `--resave` forever, not crashes

`x="nan"` on an `<element>` (or `<terminal>`) survives XML parsing — `QString::toDouble(&ok)` returns `ok=true` for `"nan"`/`"inf"`/`"-inf"`, which I verified directly rather than assumed:

```
nan  -> ok=true  value=nan  isfinite=false
inf  -> ok=true  value=inf  isfinite=false
```

The non-finite position then propagates into `Conductor::shape()` during diagram load. gdb on the hung process shows it stuck at 100% CPU — not a blocked wait, confirmed via `/proc/<pid>/stat`: 2.98 of 3.00 wall-clock seconds were user CPU time:

```
QPainterPathStroker::createStroke()
<- Conductor::shape()
<- QGraphicsItem::collidesWithItem()
<- Conductor::calculateTextItemPosition()
<- Diagram::addItem() <- Diagram::fromXml()
```

**Fix:** `QET::attributeIsAReal()` (`sources/qet.cpp`) — the shared "is this XML attribute a usable real number" helper already used at 26 call sites across the codebase — now also rejects non-finite values. `Element::valideXml()` and `Terminal::valideXml()` had their own duplicate, equally-buggy `toDouble(&conv_ok)` parsing (same gap: `conv_ok` is also `true` for `"nan"`), so I switched both to use the shared, now-correct helper instead of maintaining a second copy of the same check.

## 2. A single corrupted byte segfaults inside Qt's own XML parser

`QETProject::openFile()` hands the raw file straight to `QDomDocument::setContent(QIODevice*)`, which does not fail cleanly on a control character XML doesn't permit (the spec's `Char` production allows only tab/LF/CR among bytes below `0x20`) — it segfaults inside `libQt5Xml` itself. gdb shows the crash never reaches any QET code at all:

```
QDomDocument::setContent(QIODevice*, ...)
<- QETProject::openFile() <- QETProject::QETProject()
```

This is a class of input, not one specific byte — confirmed with three different corrupted files, **identical crash address in gdb across all three**:
- an embedded NUL (`0x00`) in XML whitespace
- an `0x0E` in XML whitespace (found by the harness's own mutation sweep, after the NUL fix, on a completely different seed file)
- an `0x19` inside base64-encoded embedded logo data

**Fix:** a new `QET::isWellFormedXmlByteStream()` (`sources/qet.cpp`) — read the raw bytes first and refuse to hand a stream containing an illegal control character to the parser at all, rather than try to recover from a crash inside a library this code doesn't control.

## Verification

- Both fixes verified against the original repro files: clean rejection (`XmlParsingFailed`, exit 1) or clean resave instead of hang/crash.
- Full-corpus regression: all 23 files in `qet-fix/examples/` still resave cleanly — the byte-stream check does not false-positive on real QET output.
- A 150-iteration adversarial sweep with all 8 mutation strategies active against the fixed build found **zero crashes**.
- Built clean, CMake/Ninja Release, Qt 5.15.18, mold-linked.

## Deliberately not fixed here

Found during validation; listed rather than silently folded in or silently left undocumented.

**`QDomDocument::setContent()` is called the same way at roughly twenty other sites** and likely shares the same crash — none of them are covered by this fix:

`sources/cli_export.cpp`, `sources/titleblocktemplate.cpp`, `sources/qetxml.cpp`, `sources/diagramview.cpp`, `sources/diagramevent/diagrameventaddmacro.cpp`, `sources/import/edz/edzpart.cpp`, `sources/elementtextpattern.cpp`, `sources/ElementsCollection/elementslocation.cpp`, `sources/ElementsCollection/elementstreeview.cpp`, `sources/titleblock/templateview.cpp`, `sources/editor/elementview.cpp`, `sources/editor/ui/qetelementeditor.cpp`. Element loading, macro/title-block loading, EDZ import, and clipboard paste are all plausibly reachable.

**A `DynamicElementTextItem`'s position accepts and persists a non-finite value.** Confirmed: `y="-inf"` survives two consecutive resaves without crashing — lower severity than the two bugs above, since it doesn't reach `Conductor::shape()` the same way an element position does, so it's a silent data-integrity gap rather than a hang. Different parsing path from the one this PR touches.

I'm happy to follow up on either of these if there's interest, but wanted this PR to stay scoped to the two bugs that are actually reproduced and fixed.

### DIFF
--- a/sources/qet.cpp
+++ b/sources/qet.cpp
@@ -19,6 +19,7 @@
 #include "qeticons.h"
 #include "shortcutmanager.h"
 
+#include <cmath>
 #include <limits>
 #include <QGraphicsSceneContextMenuEvent>
 #include <QAction>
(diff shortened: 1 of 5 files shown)


arummler:
Additional manual testing certainly does not hurt but that likely should be fixed in Qt6? At least according to https://doc.qt.io/qt-6/xml-changes-qt6.html
"control characters"?

plc-user:
We’ve had a few cases where someone asked on the forum why a project file wouldn’t load. In most cases, this happened when (for whatever reason!) the file contained a “nan” as the numerical value for a position or typos in XML-tags. 
I then suggested using QET_ElementScaler to check for a corrupted project file: 
There, I use a function from pugixml that flags corrupt XML formatting or similar issues.

In light of @arummler’s comment, I’m wondering what the current status is with the Qt6 port. 
Perhaps we should focus on that first before implementing the many ideas from the discussions and PRs into the “old” version...
And perhaps that will also resolve the “nan” issue without any additional work.

arummler:
Just to stress that I do not want to claim that the problem is solved by QT6. I was just curious why a xml parsing library (as built into Qt) would not be able to deal with malformed input. That seems quite basic for a parser. The manual seems to indicate that it should be capable to do it but should be tested (and also whether it covers the same scope of issues). 

Regarding Qt6: considering that there were several(?) attempts to migrate to Qt6 which failed as they were trying to go for a complete migration on a separate branch, running out of time and probably diverging too much, I would rather try to continue fixing until Qt6 is considered "ready" on the master. (IMHO in general the better strategy; diverging branches are always painful.) And indeed to try to go for "ready" as soon as possible and then to abandon Qt5 as soon as possible to avoid splitting the effort. What is missing: I guess some signal/slot, KF5/KF6 macro, some depreciated stuff. There are ca. 3 MRs open to deal with issues and I have another one in work. It might be good if we could label the MRs in order to prioritize.

In light of that a compromise for this MR - still not having checked the exact Qt6 XML capabilities - could be to merge but mark it with "TODO might be removed when going for Qt6 only" macro. In any case the comment mentions Qt5 and at least that should be edited when dropping Qt5.

ispyisail:

PR #683, opened 2026-08-07
Add --test-ops: headless scripted editing for automated regression testing

ispyisail:
Not an end-user feature. Adds a new CLI verb so the load-edit-save cycle QET's undo commands drive can be exercised headlessly, for automated regression testing.

## Why

I've been building a fuzzing/regression harness this session (a Python simulator, in a separate repo, driving `--resave`). It found and I fixed two real crash/hang bugs in the XML load path this way — but most of the *other* interesting bugs in this codebase turned out not to be reachable through the existing CLI at all, because `--resave`/`--export-*`/`--info` only ever load and re-serialize. They can't delete an element, rotate a selection, or undo anything — and several real bugs (PR #664, PR #660) live specifically in what happens *during* those operations, not in load/save.

## What it adds

```
qelectrotech --test-ops <project.qet> <ops.json> <output.qet>
```

`ops.json` is a small, fixed vocabulary applied to the first diagram, through the exact same code path the GUI uses:

```json
[
  {"op": "select", "uuids": ["{...}"]},
  {"op": "delete"},
  {"op": "undo"},
  {"op": "redo"},
  {"op": "rotate", "angle": 90}
]
```

- `select` resolves uuids against the diagram's elements and selects them (unresolved uuids warn on stderr, not fatal)
- `delete` pushes `DeleteQGraphicsItemCommand` on the current selection — the same command the GUI's Delete action pushes
- `rotate` pushes `RotateSelectionCommand` — same as the GUI's plain "Pivoter"
- `undo`/`redo` step the diagram's `QUndoStack`

On completion it saves and prints one JSON line to stdout:
```json
{"ops_applied": 3, "element_count": 65, "element_info_count": 65}
```
The last two are `SELECT COUNT(*)` against the in-memory project database's `element` and `element_info` tables — directly useful for catching the PR #664 bug class (an orphan `element_info` row), independent of which operation caused it.

## Verified against a real project, not just written and hoped

Ran each op against `qet-fix/examples/741.qet`:

- **select + delete**: `element_count` drops 65→64, `element_info_count` stays 65 — reproduces PR #664 on this (unpatched) branch immediately.
- **select + delete + undo**: element restored, and stderr shows exactly the error PR #664's title names: `UNIQUE constraint failed: element_info.element_uuid`.
- **select (2 elements) + rotate**: verified with the docker repo's `simulator/canon.py` that both elements' positions and orientation changed correctly and consistently.
- **rotate + undo + redo**: round-trips to the same rotated state.
- **unknown op name**: fails cleanly, exit 2, no crash.
- **uuid that doesn't resolve**: warns on stderr, treated as an empty selection, doesn't crash.
- **`"as_group": true`**: rejected with a clear message rather than silently rotating per-item — see note below.
- Existing CLI surface (`--resave`, `--info`, `--export-*`) re-verified unaffected by the corpus-wide determinism/regression fixtures already in the docker repo.

### Cross-check against the actual PR #664 fix

To make sure this isn't just detecting *a* difference but the *right* one, I also built this against `fix-element-info-orphan-row` (your open PR #664) and ran the identical select+delete+undo sequence: `element_count` and `element_info_count` both land on 65, and the `UNIQUE constraint failed` line is gone from stderr. Same tool, same ops, opposite (correct) result depending on whether the fix is present.

## Scope notes

- **`"as_group"` rotation is intentionally not supported.** `RotateSelectionCommand` only gained that mode in PR #660, which isn't merged yet. Rather than silently falling back to per-item rotation if someone passes `as_group: true`, it's rejected outright with a message naming why — so a caller relying on it fails loudly instead of getting a quietly different result.
- I kept the vocabulary to exactly what two known, real bugs needed (delete/undo/rotate/select), not a general scripting language. There's a separate, still-undecided `SCRIPTING-RFC.md` drafted in the docker repo for broader programmatic editing — this is deliberately much smaller than that, and I don't want this PR to be read as pre-empting that decision.


ispyisail:
Note: this is stacked on #682 (not yet merged) — that's why the diff currently shows `sources/qet.cpp`/`qet.h`/`qetgraphicsitem/element.cpp`/`terminal.cpp`/`qetproject.cpp` in addition to the two files this PR actually touches (`cli_export.cpp`/`cli_export.h`). Once #682 merges, this diff will shrink to just those two.

Happy to rebase directly on `master` instead if you'd rather review it standalone — just say the word. I stacked it because `--test-ops` was built and tested using the fixed binary throughout, so it's guaranteed to work on top of #682's changes; I haven't separately verified it against plain master without that PR.

ispyisail:
Rebased against current master (this branch predated the #625/#628 conductor/terminal table stack) and added `conductor_count`/`conductor_distinct`/`terminal_count` to `--test-ops`'s JSON summary, alongside the existing element counts. This was needed to wire a delete/undo/redo regression check for the database layer (qelectrotech-docker's `qet-db-liveops.py`) into a CI gate — without it there was no way to verify that deleting an element's conductors and undoing survives with the original uuids intact, which is the main thing worth checking on this code path. Verified against the full example corpus (22 projects, 0 failures) with both the new fields and the existing element-count behavior unchanged. No force-push — pushed as a fast-forward, so history is unchanged.

ispyisail:

PR #690, opened 2026-08-08
Draft/preview: SVG icons for the concepts no icon theme provides (88 icons)

ispyisail:
**Draft — this is for looking at, not for merging.** Opened so the icons can be built and seen in the running app rather than judged from a folder of files. Discussion: #689.

**Browsable gallery** (each icon beside the current one, plus true 16/22/32 px renders):
https://claude.ai/code/artifact/bef8f7fd-d9aa-4608-9f9d-ef7ffbca1a1d

## What this is

88 SVG icons for the QET-specific concepts that no general icon theme contains — conductors, terminals, terminal strips, folios, cross-references, terminal orientation marks, conductor end styles, element master/slave links, title blocks.

Matching all 183 icon names against the current upstream Breeze tree (7,107 names) gives **93 hits and 90 misses**. The misses are precisely the electrical concepts, and no upstream will ever carry them. #687 covers vendoring the 93 that already exist upstream; this covers the remainder.

## Why drawn, rather than traced or generated

- **Autotracing** the existing 16–22 px bitmaps would turn their anti-aliasing into permanent bézier wobble — a scalable icon that still looks bad.
- **Image generators** emit raster, so it would need tracing anyway, and they do not hold stroke weight or optical size across a set.

These are simple technical symbols: a terminal is a rectangle and a stub, a conductor is a polyline. That is path data, not illustration.

## The spec was measured, not invented

`#4d4d4d` is the dominant stroke colour across QET's current icons. The terminal red/blue split came from sampling `ico/22x22/terminal.png` (80% red, 20% blue).

Semantics are preserved — red = phase, blue = neutral, green = earth is electrical convention, not decoration — while the values are harmonised onto the Breeze palette so these sit correctly beside the 93 stock icons they share a toolbar with. A mixed toolbar was assembled to check this, and the two sets are hard to tell apart.

Coherence is structural rather than a matter of care: one `badge_new()` serves every "add" icon, the four stacking-order icons and the five conductor end styles are each generated by a single function, and the four orientation marks are one drawing rotated.

## Incidental fix

`ConductorSettings`, `DiagramAdd` and `DiagramDelete` each carried **different artwork at 16 px and 22 px** — `conductor-reset` vs `conductor2`, `folio-new` vs `diagram_add`, `folio-delete` vs `diagram_del`. Collapsing each to one scalable file resolves that.

Also `simplifyrichtext` was loaded by a direct `QIcon(":/ico/32x32/…")` in `richtexteditor.cpp` rather than through the registry; that call site is updated too.

## Verified

- Builds clean, CMake/Ninja Release, Qt 5.15 — 820/820, links.
- Runs under Xvfb with **no resource or icon-engine warnings**; the new glyphs render in the real toolbar.
- All 84 wired resource paths resolve to files on disk.
- `ico/copyright` gains a stanza for `ico/scalable/*` recording these as GPL-2+ and **not** derived from Breeze artwork.

## Deliberately not done

- **Superseded PNGs are left in place.** The change stays additive and trivially revertible; deleting them is follow-up work.
- **The 93 stock Breeze icons are untouched** — that is #687's job, and it should not be entangled with this.
- **`project-delete.svg` was dropped** from the branch: nothing in the tree references that asset.

## Honest limits

**The eight icons previously flagged as needing another pass have been redrawn** — `autoconnect`, `item-move`, `export-csv`, `run-dxf`, `folder-only-this` and the three properties icons. None carries a flag now. The properties mark in particular was tried twice as a corner badge before being moved inside the shape: at the ~12&nbsp;px a badge gets, detail collapses to a smudge whatever sits behind it.

**Optical sizing was audited by measurement, not by eye.** Rendering every icon and measuring its real ink extents found ink running 10–22 wide and 2–20 tall, 30 icons off-centre, and 18 breaching a 1&nbsp;px safe margin — three clipped at the canvas edge. One cause was invisible in the source: `stroke-linejoin="miter"` on the stacking-order parallelograms threw a spike past the nominal geometry. All four counts are now zero.

**The existing stacking-order icons are arguably cleaner than these replacements.** The gain there is HiDPI rendering, not aesthetics.

Most importantly: **this is one contributor's judgement on legibility, not a user's on whether the metaphors are right.** That is the part that needs other eyes, and the reason this is a draft rather than a merge request. If the direction is unwanted, saying so now costs nothing — the generator is a script, and the set can be regenerated to a different spec.

https://claude.ai/code/session_01Fq6v5jnifdpcwCoPocrp6e


ispyisail:

PR #691, opened 2026-08-08
Dialogs remember their size/position; drop a hardcoded light-mode color

ispyisail:
Small UI-polish PR from a broader windows/controls audit (main window + ~35 dialogs), not a full "modernization" — most of the app held up well (HiDPI already handled properly, no forced style/palette, 21/23 dialogs already use `QDialogButtonBox` correctly). Two concrete, low-risk findings survived that audit:

## 1. Dialogs don't remember their size or position

`QETElementEditor`, `QETDiagramEditor` and `QETTemplateEditor` already persist geometry via `QSettings`. None of the ~35 modal `QDialog` subclasses did — resize "Search and Replace" or "Diagram properties" to see more of its content, and it's back to the default size the next time you open it, every session.

Added `QET::trackDialogGeometry()` (`sources/qet.h`/`qet.cpp`), one call per dialog constructor, following the exact `restoreGeometry()`/`saveGeometry()` pattern already used by the three editors above — keyed by class name by default. `PropertiesEditorDialog` (a single class templated over several unrelated wrapped editors — conductor properties, element properties, etc.) gets an explicit key per wrapped editor type, so they don't all fight over one saved size.

Deliberately not touched: `BackupDialog`, which `setFixedSize()`s itself — a small fixed Yes/No confirmation prompt has no meaningful size to remember.

**Verification**: full Release build (504/504, no new warnings). Also verified the restore/save round-trip directly against `qet.cpp` in a small standalone Qt program outside the main build — saved a 600×400 geometry, then confirmed a fresh dialog defaulting to 200×150 picked up the saved 600×400 after the tracking call, in the same order the real fixes use (explicit default `resize()`/`setMinimumSize()` first, tracking call after, so a restored size correctly wins over a hardcoded default). Three of the ~35 constructors needed care rather than a mechanical insertion: `RichTextEditorDialog` and `AddLinkDialog` are hand-built with no `setupUi()` anchor (and share a file with an unrelated class, `RichTextEditor`, that does have one — easy to insert in the wrong constructor by grepping alone), `ReplaceFolioDialog` similarly shares a file with `ReplaceFolioWidget`, and `DiagramPropertiesDialog` calls `exec()` on itself from inside its own constructor, so the tracking call has to land *before* that call rather than at the end.

## 2. Hardcoded light-gray background breaks under a dark palette

`masterpropertieswidget.cpp` set an inline `background-color: #f0f0f0` on the disabled PLC-IO type combo box. Qt already renders disabled widgets correctly without help; the hardcoded light color just shows as a bright box in an otherwise dark dialog under a dark OS theme. Dropped the stylesheet.

---

A third finding from the same audit — a permanently-`setHidden(true)` button in `ReplaceConductorDialog` that looked like dead code from the `.cpp` alone — turned out to be a deliberate mechanism once I checked the `.ui`'s own Designer-level signal/slot connections: several controls are wired to call `click()` on it (which doesn't require visibility) purely to share one `updatePreview()` trigger point. Dropped that one from this PR since it isn't actually a bug.

### DIFF
--- a/sources/ElementsCollection/ui/renamedialog.cpp
+++ b/sources/ElementsCollection/ui/renamedialog.cpp
@@ -20,12 +20,14 @@
 #include <QDate>
 #include "ui_renamedialog.h"
 
+#include "../../qet.h"
 RenameDialog::RenameDialog(QString path, QWidget *parent) :
 	QDialog(parent),
 	ui(new Ui::RenameDialog),
 	m_path(std::move(path))
 {
 	ui->setupUi(this);
+	QET::trackDialogGeometry(this);
 	m_name = m_path.split("/").last();
 	if (m_name.endsWith(".elmt")) m_name.remove(".elmt");
 	ui->m_label->setText(tr("L'élément « %1 » existe déjà. Que souhaitez-vous faire ?").arg(m_path));
--- a/sources/NameList/ui/namelistdialog.cpp
+++ b/sources/NameList/ui/namelistdialog.cpp
@@ -19,6 +19,7 @@
 #include "ui_namelistdialog.h"
 #include "namelistwidget.h"
 
+#include "../../qet.h"
 #include <QPushButton>
 #include <QMessageBox>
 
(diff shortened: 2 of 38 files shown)


ispyisail:

PR #692, opened 2026-08-08
Draft: fast element insertion — all four phases from #676

ispyisail:
**Draft** — the questions in #676 are still open, and two of them change what this should look like. Opened so the behaviour can be tried rather than argued from a description.

Builds all four phases from #676. Three commits, each self-contained and useful alone; **the first can be cherry-picked on its own** if only Phase 1 is wanted.

## Phase 1 — reach the placement mode without a drag (`716ced5`)

`DiagramEventAddElement` is already a good interactive placement mode — ghost on the grid, left click drops, Space rotates, the element stays loaded, commits through `AddGraphicsObjectCommand` so undo behaves. It had **one caller in the whole tree**, inside `DiagramView::handleElementDrop()`, so the only way to reach it was to finish a drag.

- `DiagramView::startElementPlacement()` split out of `handleElementDrop()`; `defaultPlacementPos()` uses the cursor when it is over the view, the centre of the visible area otherwise.
- `ElementsCollectionWidget` emits `insertElementRequested()` on double click or Enter. It deliberately does **not** reach for an ancestor editor the way the status-bar hint does — the picker has no editor ancestor, so the host connects the signal.
- `QETDiagramEditor` places it and remembers it, which gives **"Insérer le dernier élément"** on `A`.

`A` rather than `Space`: Space already rotates the pending element *inside* placement mode (`diagrameventaddelement.cpp:174`) and is taken three more times in this editor.

**Answering question 1 in #676**: double click inserts by default, with a preference to restore the editor (`elementscollection/double-click-inserts`), worded in Configuration as an opt-out so the default is the unchecked box. Editing stays on the context menu where it already was.

## Phase 2 — flat ranked results (`16409b7`)

The tree search hides non-matching rows, so hits stay scattered. Searching `diode` against the shipped collection leaves you looking at five levels of expanded tree with the first visible hit being *Avalanche diode bidirectional*.

Same matches, shown as a ranked flat list. No new index — this ranks what `match()` already returns against `Qt::UserRole+1`.

```
exact name       1000     name contains   600 - match position
starts with       800     info field only 300
                        minus the name length
```

`diode` now returns: **Diode**, diode-tube, Diodes Module Pilot Wire, Photodiode.

## Phases 3 & 4 — cursor picker and folder palette (`a7cd955`)

`Insert` opens a picker where the mouse is. Empty field shows a quick palette as an icon grid; type to search everything; Enter places; Esc closes.

**The model is not hoisted to a `QETApp` singleton.** #676's Phase 3 proposes that, but **question 2 asks permission for it and nobody has answered**, and it touches loading code several people have worked on. It also turns out not to be needed: the picker builds no model at all, it asks the dock's widget via the new `ElementsCollectionWidget::rankedSearch()` — extracted from Phase 2 so both share one implementation. No second model, no doubled startup, and the two cannot disagree on ranking. If the singleton is wanted later, nothing here blocks it.

**The palette is a folder** — subfolders are categories, contents are entries. Set-up is dragging into it, ordering is the `01_`/`02_` convention already used by the shipped collection, sharing is `companyElementsDir()`. Only the path is in `QSettings`, defaulting to the custom collection. No schema, parser, merge rules or "reset to defaults".

Also fixed: both new actions did nothing silently with no folio open — now enabled from `slot_updateActions()` like every other placement action.

## Verified live

Under Xvfb against `examples/741.qet` with the shipped collection, not just compiled:

- Search `diode` → double-click a result → ghost follows the cursor grid-snapped; **no element editor opens**
- `Insert` over the folio → picker at the cursor → type `relay` → ranked list, best hit preselected
- `Insert` with an empty field → palette icon grid with thumbnails → `Enter` → chosen element in placement mode

Builds clean, CMake/Ninja Release, Qt 5.15, no new warnings in the touched files.

## Still open

**Question 3 in #676 is unanswered and Phase 4 rests on it** — the folder-as-palette design assumes people already curate folders in their custom collection. If that is not how they actually work, the palette grid is the part to reconsider; Phases 1–3 do not depend on it.

The `Insert` and `A` defaults are `ShortcutManager` registrations, so they are defaults rather than commitments and appear in the Shortcuts preference page.

Discussion: #676

https://claude.ai/code/session_01Fq6v5jnifdpcwCoPocrp6e

### DIFF
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -411,6 +411,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/ElementsCollection/elementcollectionitem.h
   ${QET_DIR}/sources/ElementsCollection/elementscollectionmodel.cpp
   ${QET_DIR}/sources/ElementsCollection/elementscollectionmodel.h
+  ${QET_DIR}/sources/ElementsCollection/elementpickerpopup.cpp
+  ${QET_DIR}/sources/ElementsCollection/elementpickerpopup.h
   ${QET_DIR}/sources/ElementsCollection/elementscollectionwidget.cpp
   ${QET_DIR}/sources/ElementsCollection/elementscollectionwidget.h
   ${QET_DIR}/sources/ElementsCollection/elementslocation.cpp
--- a/sources/ElementsCollection/elementpickerpopup.cpp
+++ b/sources/ElementsCollection/elementpickerpopup.cpp
@@ -1,0 +1,323 @@
+/*
+	Copyright 2006-2026 The QElectroTech Team
+	This file is part of QElectroTech.
+
+	QElectroTech is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 2 of the License, or
+	(at your option) any later version.
+
+	QElectroTech is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with QElectroTech. If not, see <http://www.gnu.org/licenses/>.
+*/
+#include "elementpickerpopup.h"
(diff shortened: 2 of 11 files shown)


ispyisail:

PR #693, opened 2026-08-08
Fix crash changing dynamic text color and confirming with Enter (bugtracker #323)

ispyisail:
Fixes https://qelectrotech.org/bugtracker/view.php?id=323 — "Crash on change label color". Reported crash: changing a dynamic text's color and confirming with **Enter** segfaults; confirming with the mouse on the dialog's own OK button does not. Reported on Windows 11 and Debian, with `QObject::installEventFilter(): Cannot filter events for objects in a different thread` immediately before the crash.

## Root cause

`DynamicTextItemDelegate::createEditor()`'s `color` case constructs a `QColorDialog` and returns it directly as the item view's editor widget for that cell:

```cpp
case DynamicElementTextModel::color:
{
    QColorDialog *cd = new QColorDialog(index.data(Qt::EditRole).value<QColor>(), parent);
    cd->setObjectName("color_dialog");
    return cd;
}
```

Every other case in this function returns a small inline widget appropriate for an item-view editor (`QSpinBox`, `QComboBox`, ...). The `font` case right above it needs a full picker too, and does it the safe way — call the blocking static `QFontDialog::getFont()`, then return a plain placeholder `QWidget` with the result stashed on it:

```cpp
case DynamicElementTextModel::font:
{
    bool ok;
    QFont font = QFontDialog::getFont(&ok, index.data(Qt::UserRole+2).value<QFont>(), parent);
    QWidget *w = new QWidget(parent);
    if (ok) { w->setFont(font); w->setProperty("ok", ok); }
    w->setObjectName("font_dialog");
    return w;
}
```

`color` didn't follow that pattern. `"color_dialog"` isn't one of the objectNames this delegate's own `eventFilter()` special-cases (the "commit on every keystroke" hack list, or the combo-box `FocusIn` list), so Enter on it falls through to the base `QStyledItemDelegate::eventFilter()`, which treats whatever `createEditor()` returned as an ordinary small editor to commit-and-destroy on Enter — racing the `QColorDialog`'s own internal OK-button accept/close path, which on Windows can hand off to the native color picker. Clicking OK with the mouse doesn't go through the same view-level key-press handling, which is consistent with the report that only Enter crashes.

## Verification

Structural: built a standalone Qt program that constructs a bare `QStandardItemModel` + `QTreeView` with the real `DynamicTextItemDelegate` and drives it through `QAbstractItemView::edit()` exactly as the real view does. Confirmed the *old* code's `QColorDialog` editor is a **child widget of the view's viewport**, not a top-level window (a search of `QApplication::topLevelWidgets()` never found it; only a search of the viewport's children did) — i.e. concretely not being used as a dialog is meant to be used, independent of whether this Linux/Xvfb environment reproduces the exact native-dialog-thread segfault from the report.

End-to-end with the fix: same harness, but now the `QColorDialog` **is** found among top-level widgets (proper popup behavior). Clicked its real OK button, confirmed the picked color landed on the placeholder widget's stashed properties, sent the editor a synthetic Enter keypress (the exact trigger from the bug report), and confirmed the final value committed to the model matches the picked color. No crash, clean exit.

Also did a full Release build (333/333, CMake/Ninja, Qt 5.15.18) — no new warnings.

### DIFF
--- a/sources/ui/dynamicelementtextmodel.cpp
+++ b/sources/ui/dynamicelementtextmodel.cpp
@@ -36,6 +36,7 @@
 #include <QHash>
 #include <QModelIndex>
 #include <QStandardItem>
+#include <QTimer>
 #include <QUndoCommand>
 
 static int src_txt_row   = 0;
@@ -1595,6 +1596,15 @@ DynamicTextItemDelegate::DynamicTextItemDelegate(QObject *parent) :
 	QStyledItemDelegate(parent)
 {}
 
+void DynamicTextItemDelegate::commitAndCloseDeferred(QWidget *editor) const
+{
+	auto *self = const_cast<DynamicTextItemDelegate *>(this);
+	QTimer::singleShot(0, self, [self, editor]() {
+		emit self->commitData(editor);
+		emit self->closeEditor(editor);
+	});
+}
+
 QWidget *DynamicTextItemDelegate::createEditor(
 		QWidget *parent,
 		const QStyleOptionViewItem &option,
@@ -1682,6 +1692,7 @@ QWidget *DynamicTextItemDelegate::createEditor(
 				w->setProperty("ok", ok);
 			}
 			w->setObjectName("font_dialog");
+			commitAndCloseDeferred(w);
 			return w;
 		}
 		case DynamicElementTextModel::color:
@@ -1716,6 +1727,7 @@ QWidget *DynamicTextItemDelegate::createEditor(
 				w->setProperty("ok", true);
 			}
 			w->setObjectName("color_dialog");
+			commitAndCloseDeferred(w);
 			return w;
 		}
 		case DynamicElementTextModel::pos:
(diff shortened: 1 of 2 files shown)


plc-user:
The crash has been averted!
But here's something I've noticed:
The text in the “Color” field and the text color itself aren't updated until you exit the Properties list and click in the diagram. Previously, the color and text were updated immediately when clicking "Ok". We should restore that behavior; otherwise, users will have questions about that in the forum...

ispyisail:
@plc-user Good catch, thank you — that's a real regression from the crash fix, not intended behavior.

**Cause**: the crashing code returned a *live* `QColorDialog` as the editor; clicking its OK button hid it, and hiding the active editor happens to trip Qt's own focus-lost commit path — so the value applied immediately, but only as an accidental side effect of the same mechanism that crashed on Enter. The fix replaced that with resolving the dialog synchronously and returning an inert placeholder widget instead, which is correct for the crash but has no focus to lose, so nothing told the view to read the value back until some unrelated click happened to trigger it.

**Fix pushed**: the editor now explicitly tells the view to commit and close itself right after the dialog resolves, instead of waiting on that accidental trigger. Applied to both the color and font pickers, since both share the same shape and thus the same gap — font just hadn't been reported yet.

Verified with the same test harness I used for the crash fix, this time specifically checking it commits *without* any further interaction (just the OK click), and reconfirmed the crash fix itself still holds. Should update immediately again now.

PR #698, opened 2026-08-08
Migrate string based signal/slot to method pointer

arummler:
This is the preferred way nowadays and will avoid silent disconnects as it is checked during compile time. This is a quite critical MR as I am touching basically all parts. I would propose to merge it after #662 which was the starting point. I will then rebase or merge and then it should be ready.


arummler:
Merged master. I would propose to handle first #679 as it also touches autonum.

plc-user:
As with many GUI toolkits, signal handling and the associated calls in Qt are changed fairly regularly with new versions.
When changes are made to these calls, we need to ensure that they work in both Qt5 and Qt6 if we want to incorporate them into master branch. This should not require any adjustments when we make the transition to Qt6. 
You’ve probably already tested this, @arummler, right?

I haven’t tried these changes yet, but everything looks plausible and reasonable to me.

scorpio810:
⚠️ Heads up: this PR would revert a recently merged feature

I checked the diff with Ai on sources/undocommand/rotateselectioncommand.cpp/.h, and this branch drops:

The rotate_as_group parameter on RotateSelectionCommand's constructor
The whole addGroupPositionUndo() method (shared-pivot group rotation, with exact right-angle math and grid snapping)
That functionality comes from #660 ("Add 'rotate group' to actually rotate a selection as a whole") plus its two grid-snapping follow-up fixes, which are already on master but aren't present on this branch's head commit (c7ed322). The "Merge branch 'master' into master-modernize-signal-slot" commit in this PR's history happened before #660 was merged, so this branch never picked it up.

If merged as-is, this would silently regress group rotation and the grid-alignment fix (reported by plc-user, discussion #618) — unrelated to the signal/slot modernization itself.

Could you rebase (or re-merge) onto current master and re-check this file? Since your migration hasn't touched rotateselectioncommand.cpp yet, there shouldn't be much to reconcile, but worth confirming no conflicts with the rotate_as_group logic.

Rest of the diff (signal/slot migration across the other 52 files) looks solid — thanks for the thorough work on this.

scorpio810:
Found two more regressions while reviewing sources/elementspanelwidget.cpp — both connect() calls were dropped entirely during the migration rather than converted, not moved elsewhere in the file:

elements_panel's customContextMenuRequested(const QPoint &) → handleContextMenu(const QPoint &) is gone. handleContextMenu() is still defined (line 506 on this branch) but nothing calls it anymore — right-click context menu on the elements panel is broken.
elements_panel's requestForTitleBlockTemplate(const TitleBlockTemplateLocation &) → QETApp::instance()->openTitleBlockTemplate(...) is gone too — double-clicking a title block template in the panel no longer opens it. (The two remaining openTitleBlockTemplate() calls elsewhere in the file are a different code path and don't cover this.)
Original (master), for reference:



```
connect(elements_panel, SIGNAL(customContextMenuRequested(const QPoint &)), this, SLOT(handleContextMenu(const QPoint &)));
connect(
    elements_panel,
    SIGNAL(requestForTitleBlockTemplate(const TitleBlockTemplateLocation &)),
    QETApp::instance(),
    SLOT(openTitleBlockTemplate(const TitleBlockTemplateLocation &))
);
```

Could you re-add these two, converted to the pointer-to-member syntax? Everything else I checked in this file and in qetdiagrameditor.cpp (overload resolution via qOverload<> for activateProject, closeProject, editProjectProperties, showError, diagramAdded) looks correct.

scorpio810:
One more regression, in sources/exportpropertieswidget.cpp: the draw_terminal_names checkbox's stateChanged/checkStateChanged connect to optionChanged() is missing from both branches of the #if QT_VERSION block. The checkbox itself is still built and still read/written in applyTo()/setExportProperties(), but toggling it no longer emits optionChanged() like the other six checkboxes in the same group (draw_grid, draw_border, draw_titleblock, draw_terminals, draw_bg_transparent, draw_colored_conductors) — so the export preview won't refresh when this option changes.

Could you add it back, e.g.:



`connect(draw_terminal_names, &QCheckBox::stateChanged, this, &ExportPropertiesWidget::optionChanged);`
(and the checkStateChanged equivalent in the #else branch)

arummler:
Sorry, got lost somewhere during the repeated three way merges. Hope we did not miss anything else. What I deliberately did not touch and which still uses the old style for connects is the richtext folder. From what I see this was copied at some time from the QTcreator code. Instead of doing it manually I would probably try to copy the current version at some point (for qt6) and see whether it needs any customization. At the first glance the current qtcreator code looks very similar.

scorpio810:
Please ask us when you think this is ready to merge, big thanks for your works.

arummler:
I have once again run a diff between the two branches. It builds both with Qt6 (my primary development target) and Qt5. A project of mine - quite large - reads without issues. Trying a couple operations also does not show anything strange and I do not see any signal/slot message in the commant line. I would say we can dare to merge it if no one else (or their favorite AI) wants to go through it.

PR #700, opened 2026-08-08
Follow up Auto-break conductors

Kellermorph:
Extracted the auto-break conductor logic into a shared helper function autoBreakConductors() in new files autobreakconductor.h/.cpp, called from three flows:

Drag from panel (diagrameventaddelement.cpp)
Paste (diagramcommands.cpp — PasteDiagramCommand)
Move (elementsmover.cpp — ElementsMover::endMovement)
Key improvements over the previous approach:

distanceToSegment() replaces QET::orthogonalProjection() for bounded point-on-segment detection (reviewer patch)
Conductors are grouped by shared far-end endpoint; only the largest group is broken to prevent bridging independent nets
Orientation-based selection of connect_to and other_terminal
Best group chosen by distance, not group size
Bug fixed: On first paste, autoBreakConductors() created DeleteQGraphicsItemCommand and AddGraphicsObjectCommand as children of m_break_cmd, but m_break_cmd->redo() was never called — so the break commands were never executed. Added the missing m_break_cmd->redo() call after creation.


ispyisail (changes requested):
Good follow-up — this is exactly the generalization I'd suggested after tracking down why paste didn't trigger the break (#639's thread), and the extraction is clean: the moved block in `autobreakconductor.cpp` is byte-for-byte identical to what was deleted from `diagrameventaddelement.cpp`, so the already-reviewed core algorithm (bounded `distanceToSegment()`, grouping by shared far-end endpoint, orientation-aware matching) carries over with no regression risk from the refactor itself.

One real concern with the two new call sites, though.

## Multiple elements in one batch can double-process the same conductor

Both new integrations loop over multiple elements and call `autoBreakConductors()` once per element, deferring execution:

```cpp
// PasteDiagramCommand::redo()
for (Element *e : content.m_elements) {
    autoBreakConductors(diagram, e, m_break_cmd);
}
...
m_break_cmd->redo();   // only now do the children actually run

// ElementsMover::endMovement()
for (Element *e : m_moved_content.m_elements) {
    autoBreakConductors(m_diagram, e, undo_object);
}
...
m_diagram->undoStack().push(undo_object);   // redo() cascades here
```

`autoBreakConductors()`'s `conductors_handled`/`used_terminals` bookkeeping is local to a single call — nothing carries over between elements in the same loop. That's fine as long as the diagram's conductor list is up to date between calls, but it isn't: `DeleteQGraphicsItemCommand`'s constructor just stores the content (checked `deleteqgraphicsitemcommand.cpp` — no scene mutation happens there), and the actual removal only happens on `redo()`, which for both these call sites only fires once, after the whole loop has already finished deciding what to break for every element.

So if two elements in the same paste or the same multi-element drag both have a terminal landing on the *same* existing conductor, the first call to `autoBreakConductors()` picks that conductor for element A (via `diagram->conductors()`, which still lists it — nothing's actually gone yet) and queues its deletion; the second call, processing element B, sees the exact same still-present conductor in `diagram->conductors()` and independently queues *another* deletion of it. Two `DeleteQGraphicsItemCommand`s end up targeting the same `Conductor*`, and when the parent's `redo()` cascades to both, the second one operates on an already-deleted item.

This is new exposure specifically from generalizing to paste/move: the original drag-from-panel call site only ever handled one element at a time, so the same conductor could never be claimed twice within a single call. Paste and multi-select move both routinely involve several elements at once, so I'd consider this reachable rather than theoretical — e.g. pasting two components side by side onto an existing bus bar, where both land within 5px of the same segment. I traced this from the code rather than reproducing it live (positioning terminals precisely via GUI automation to land within the 5px threshold felt too fiddly to get right blind), so flagging it as a well-reasoned concern rather than a confirmed crash — worth double-checking directly since a double-delete on a `QGraphicsItem` is the kind of thing that segfaults rather than misbehaving quietly.

Two directions that would fix it:
- Thread a shared `conductors_handled`/`used_terminals` state across the whole batch (one pass over all elements' terminals before committing any break), instead of one fully-independent call per element, or
- Call `redo()` on each element's break commands before moving to the next element, so `diagram->conductors()` is genuinely current by the time the next element is considered.

## Things I checked and are fine

- **The paste redo/undo fix itself** (the stated purpose — `m_break_cmd->redo()` was never called) is correct. `QUndoCommand`'s base `redo()`/`undo()` cascades to children in/reverse insertion order, which is the standard "bare `QUndoCommand()` as a macro parent" pattern already used elsewhere in this codebase, so calling `m_break_cmd->redo()` right after populating it is the right fix. The `undo()`-before-`removeItem()` ordering (so terminals are still on the scene when the break's own undo runs) is correct, and the destructor's `delete m_break_cmd` is safe against the earlier `delete; = nullptr;` path (no double-free).
- **`ElementsMover::endMovement()`'s discarded return value** (`broken_endpoints` isn't captured, unlike the drag-drop call site) looked like a possible duplicate-connection risk against the auto-connect pass right after it, but it isn't: `AlignedFreeTerminals()` — used by that auto-connect logic — only offers up terminals with no attached conductor, and `Conductor`'s constructor (`conductor.cpp:96-97`) calls `terminal->addConductor()` synchronously, before any undo-command `redo()` runs. So by the time `AlignedFreeTerminals()` evaluates, the just-broken terminals are already correctly excluded, independent of the discarded return value. `PasteDiagramCommand` doesn't have an auto-connect pass at all, so the same discard there has no consumer either way.

Kellermorph:
Done.
Tested with 2 x 3 Lines object.
Copy + Paste and all lines get broken und reconnected

PR #701, opened 2026-08-08
Make folio-referencing arrows clickable and show link state

ispyisail:
## What this adds

Coming/going arrows (`previous_report`/`next_report` elements, used for wire references between folios) already carried an in-memory link to their counterpart on another folio, but there was no way to *act* on that link and no visual cue that it existed.

- **Double-click a linked arrow** to jump straight to the linked folio, with the counterpart selected and centered.
- **Double-click an unlinked arrow** still opens the properties dialog, exactly as before — that dialog's "Folio referencing" tab is the only way to create the link in the first place, so this path can't be given up.
- **A dangling link** (the linked element was deleted after linking) shows a small popup instead of silently doing nothing.
- **Linked arrows get a persistent blue halo** — the same radial-gradient highlight `Element::drawHighlight()` already draws transiently on hover for linked elements elsewhere in QET, now shown continuously on report arrows whenever they resolve to a real counterpart, so an active link reads visually the way a hyperlink does.

## Implementation

- `ReportElement::mouseDoubleClickEvent()` — new override. Falls back to the inherited `QetGraphicsItem::mouseDoubleClickEvent()` (properties dialog) when `isFree()`; otherwise resolves `connected_elements.first()`, calls `Diagram::showMe()` to switch folio tabs, then selects and centers the target.
- `ReportElement::paint()` — new override. Calls the now-`protected` `Element::drawHighlight()` before the normal paint when `!isFree()`.
- `Element::drawHighlight()` moved from `private` to `protected` so `ReportElement` can reuse it instead of duplicating the gradient-drawing code.

## Testing

Verified live under Xvfb with a hand-built two-folio project containing a linked coming/going arrow pair:
- Halo renders on both linked arrows.
- Double-clicking the coming arrow switches to the linked folio and selects/centers the going arrow.
- Double-clicking a still-unlinked arrow opens the properties dialog unchanged (checked twice, no regression from the new override).

Build: CMake/Ninja Release, Qt 5.15, clean incremental rebuild, no new warnings.

### DIFF
--- a/sources/qetgraphicsitem/element.h
+++ b/sources/qetgraphicsitem/element.h
@@ -217,11 +217,11 @@ class Element : public QetGraphicsItem
 	protected:
 		void drawAxes(QPainter *, const QStyleOptionGraphicsItem *);
 		void setSize(int, int);
-
-	private:
 		void drawHighlight(
 				QPainter *,
 				const QStyleOptionGraphicsItem *);
+
+	private:
 		bool buildFromXml(const QDomElement &, int * = nullptr);
 		bool parseElement(const QDomElement &dom);
 		bool parseInput(const QDomElement &dom_element);
--- a/sources/qetgraphicsitem/reportelement.cpp
+++ b/sources/qetgraphicsitem/reportelement.cpp
@@ -24,6 +24,10 @@
 #include "../qetproject.h"
 #include "dynamicelementtextitem.h"
 
+#include <QGraphicsSceneMouseEvent>
+#include <QGraphicsView>
+#include <QMessageBox>
+
 ReportElement::ReportElement(const ElementsLocation &location, const QString& link_type,QGraphicsItem *qgi, int *state) :
 	Element(location, qgi, state,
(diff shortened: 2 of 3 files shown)


scorpio810:
Screenshot

Thanks for the PR Shane. However, double-click navigation to the linked folio already exists in QET, so that part doesn't add anything new.

That leaves the persistent blue halo (drawHighlight() called continuously in paint()), and I have two concerns there:

Visually: on a diagram with many folio references, having a halo lit up permanently on each one risks cluttering the drawing rather than helping readability. Hover was already enough to signal that a link exists.
More problematic: this halo is drawn inside paint(), so it also shows up in PDF export as well as PNG/SVG exports. This is an editing-state indicator that has no place on a final document meant to be printed or shared.

Since the navigation already existed and the halo is problematic as-is, I think it's better to close this PR rather than iterate on it. If you'd still like a visual link indicator, it would need to be redesigned from scratch, restricted to interactive display only (hover, or excluded at export/print time).

ispyisail:
fair enough, it was actaully broken anyway

PR #702, opened 2026-08-09
Add auto-numbering for folio-reference (report) arrow labels

ispyisail:
## Background

Follows on from #701 (clickable folio-reference arrows). Discussion there moved to: manually typing an ID for every coming/going arrow is tedious, but an earlier scheme that baked the *destination* folio number into the ID itself had a real bug — the number went stale whenever pages were reordered, because it was resolved once at creation time instead of computed live.

The design that came out of that discussion: keep the ID itself a boring, stable, auto-incrementing counter (with manual override, since labels are already editable), and move the *disambiguating* information (which folio, which device) into the picker dialog instead, where it's recomputed live every time the list is built and so can never go stale.

## What this adds

**A fourth auto-numbering category, "Renvois"**, alongside the existing Conducteur/Elément/Folio ones — same `NumerotationContext` engine, same UI pattern, not a new bespoke mechanism:
- `QETProject` gets a `report_autonum` hash (named rules + a "current" one), mirroring `conductorAutoNum`/`elementAutoNum` exactly, with its own `<report_autonums>` XML block.
- `Element::setUpFormula()` — the existing placement-time auto-label hook (previously an unconditional no-op for report arrows) — now sources its formula/context from the report hash for report-type elements. A newly-placed coming/going arrow gets the next number automatically.
- A "Renvois" tab in the project's "Numérotation auto" page, and a "Renvoi" row in the auto-numbering dock widget — both direct copy-adapts of the existing Element tab/row.
- The part-type list for this category deliberately **excludes** the folio-number and element-location tokens available to conductors/elements — baking the destination folio into the ID is exactly the bug this is meant to avoid.
- Manual override needs no new UI: the label is an ordinary dynamic text, already editable by double-click like any other element's label.

**A new column in the link picker**: `LinkSingleElementWidget`'s report-arrow candidate list (the dialog you use to link a coming arrow to its matching going arrow) gets a new first column, "Élément connecté" — the label of whatever device sits on the far end of that candidate's own wire, computed fresh every time the list builds (terminal → conductor → other terminal → `parentElement()` → `actualLabel()`). Same wire-tag/page-number/page-title columns that were already there stay as-is.

**Préfixe/Suffixe fields on the Renvois tab**: two plain text fields sit above the generic +/- part editor so a user can get e.g. "R1-A, R2-A..." without needing to understand multi-part numerotation contexts. They're backed by ordinary leading/trailing `"string"` context parts — the formula composition and XML persistence already handled these unchanged, so this is a pure UI addition (`SelectAutonumW::toNumContext()`/`setContext()` now wrap/unwrap the surrounding string parts around the counter's own rows).

## Testing

Verified live under Xvfb:
- Configured a Renvois rule via the project's Numérotation auto page; confirmed the dock widget's Renvoi row shows the live value/increment/next-value preview.
- Placed two report arrows from the collection panel; both were auto-labeled (1, then 2), counter advancing correctly each time, matching the dock's live preview throughout.
- The picker column addition was verified by build success + code review — it's the identical `Terminal`/`Conductor`/`Element` API sequence already used two lines above it in the same function for the existing wire-tag column — rather than an additional screenshot; double-click-triggered dialogs became unreliable in the automation harness later in the session, confirmed independent of this change (checked against an unrelated, unmodified menu action).
- Préfixe/Suffixe: typed "R"/"-A" into the new fields, Applied, and confirmed the saved project XML contains the expected `<part type="string" value="R".../><part type="unit".../><part type="string" value="-A".../>` sequence and composite `formula="R%sequ_1-A"`. Closed and reopened the project in a fresh process and confirmed the fields repopulate correctly from the saved rule, with the counter's own part editor showing only its own row (no phantom prefix/suffix row).

**Fix: report labels went blank after saving and reopening the project.** While testing print output for this feature, found that the "%f-%l%c"-style live destination text a report arrow shows for its linked partner — a pre-existing QET mechanism, not something this PR adds — silently stopped working the moment a project was saved and reopened, which defeats the entire point for anyone relying on it for printed drawings. Root cause was two separate, pre-existing bugs, unrelated to this PR's other changes:
- `QETProject::refresh()` — which re-establishes each element's dynamic-text connections on load — was `private` and never actually called from anywhere, including project open, despite its own doc comment saying that's exactly what it's for.
- `DynamicElementTextItem::refreshLabelConnection()` called `updateReportFormulaConnection()`, which re-applies the formula's signal connections but never refreshes `m_other_report` from `linkedElements()`. For a text item that just loaded from a file, `m_other_report` was still null, so every connection became a no-op and the label stayed blank permanently — reordering, deleting, or moving the partner did nothing either, since the very connections meant to react to those changes were never established in the first place.

Fixed by wiring `refresh()` into `openAndAddProject()` and having `refreshLabelConnection()` call `reportChanged()` (the function that actually re-reads `linkedElements()`) instead. Also found `diagramAdded` was missing from the set of signals `DynamicElementTextItem` listens to for `%f`/`%id` (only `projectDiagramsOrderChanged` and `diagramRemoved` were wired) — meaning inserting a new page wouldn't live-update an already-placed arrow's destination text even within the same session. Added it, and fixed the matching `disconnect()` to be symmetric (`diagramRemoved` was connected but never disconnected — a pre-existing leak).

Verified live: placed two arrows from the standard collection, linked them via the picker, saved, reopened in a fresh process, and confirmed both now show live `"<folio>-<position>"` text computed from their partner's actual position (e.g. `"2-C7"`) — previously both were blank after reopening, confirmed via debug tracing of `m_dynamic_text_list`/`m_other_report` before diagnosing the two root causes above.

## Testing

Verified live under Xvfb:
- Configured a Renvois rule via the project's Numérotation auto page; confirmed the dock widget's Renvoi row shows the live value/increment/next-value preview.
- Placed two report arrows from the collection panel; both were auto-labeled (1, then 2), counter advancing correctly each time, matching the dock's live preview throughout.
- The picker column addition was verified by build success + code review — it's the identical `Terminal`/`Conductor`/`Element` API sequence already used two lines above it in the same function for the existing wire-tag column — rather than an additional screenshot; double-click-triggered dialogs became unreliable in the automation harness later in the session, confirmed independent of this change (checked against an unrelated, unmodified menu action).
- Préfixe/Suffixe: typed "R"/"-A" into the new fields, Applied, and confirmed the saved project XML contains the expected `<part type="string" value="R".../><part type="unit".../><part type="string" value="-A".../>` sequence and composite `formula="R%sequ_1-A"`. Closed and reopened the project in a fresh process and confirmed the fields repopulate correctly from the saved rule, with the counter's own part editor showing only its own row (no phantom prefix/suffix row).
- Report-label reload fix: see verification described above (placed, linked, saved, reopened, confirmed live "2-C7"/"1-D8"-style text on both ends).

Build: CMake/Ninja Release, Qt 5.15, clean, no new warnings.

### DIFF
--- a/sources/autoNum/ui/autonumberingdockwidget.cpp
+++ b/sources/autoNum/ui/autonumberingdockwidget.cpp
@@ -64,12 +64,15 @@ void AutoNumberingDockWidget::clear()
 	ui->m_conductor_cb->clear();
 	ui->m_element_cb->clear();
 	ui->m_folio_cb->clear();
+	ui->m_report_cb->clear();
 	ui->m_conductor_value_le->clear();
 	ui->m_element_value_le->clear();
 	ui->m_folio_value_le->clear();
+	ui->m_report_value_le->clear();
 	ui->m_conductor_next_le->clear();
 	ui->m_element_next_le->clear();
 	ui->m_folio_next_le->clear();
+	ui->m_report_next_le->clear();
 }
 
 /**
@@ -88,6 +91,9 @@ AutoNumberingDockWidget::Row AutoNumberingDockWidget::rowFor(AutoNumCategory cat
 		case AutoNumCategory::Folio:
 			return {ui->m_folio_cb, ui->m_folio_value_le,
 				ui->m_folio_increase_sb, ui->m_folio_next_le};
+		case AutoNumCategory::Report:
+			return {ui->m_report_cb, ui->m_report_value_le,
+				ui->m_report_increase_sb, ui->m_report_next_le};
 	}
 	return {nullptr, nullptr, nullptr, nullptr};
 }
@@ -127,7 +133,13 @@ void AutoNumberingDockWidget::setProject(QETProject *project,
 			    this,SLOT(elementAutoNumChanged()));
 		disconnect (m_project,SIGNAL(elementAutoNumAdded(QString)),
 			    this,SLOT(elementAutoNumChanged()));
-	
+
+			//Report Signals
+		disconnect (m_project,SIGNAL(reportAutoNumRemoved(QString)),
+			    this,SLOT(reportAutoNumChanged()));
+		disconnect (m_project,SIGNAL(reportAutoNumAdded(QString)),
+			    this,SLOT(reportAutoNumChanged()));
+
 			//Folio Signals
 		disconnect (m_project,SIGNAL(folioAutoNumRemoved()),
 			    this,SLOT(folioAutoNumChanged()));
@@ -166,6 +178,12 @@ void AutoNumberingDockWidget::setProject(QETProject *project,
 		 this,SLOT(elementAutoNumChanged()));
 	connect (m_project,SIGNAL(elementAutoNumAdded(QString)),
 		 this,SLOT(elementAutoNumChanged()));
+
+		//Report Signals
+	connect (m_project,SIGNAL(reportAutoNumRemoved(QString)),
+		 this,SLOT(reportAutoNumChanged()));
+	connect (m_project,SIGNAL(reportAutoNumAdded(QString)),
+		 this,SLOT(reportAutoNumChanged()));
 
 		//Folio Signals
 	connect (m_project,SIGNAL(folioAutoNumRemoved()),
@@ -225,11 +243,20 @@ void AutoNumberingDockWidget::setContext()
 		{ ui->m_folio_cb -> addItem(str);}
 	}
 
+	//Report Combobox
+	ui->m_report_cb->addItem("");
+	QList <QString> keys_report = m_project->reportAutoNum().keys();
+	if (!keys_report.isEmpty()) {
+		foreach (QString str, keys_report)
+		{ ui->m_report_cb -> addItem(str);}
+	}
+
 		//The combo boxes have just been repopulated, so the value fields next
 		//to them are showing whatever the previous project left there.
 	refreshRow(AutoNumCategory::Conductor);
 	refreshRow(AutoNumCategory::Element);
 	refreshRow(AutoNumCategory::Folio);
+	refreshRow(AutoNumCategory::Report);
 
 	this->setActive();
 }
@@ -266,6 +293,11 @@ void AutoNumberingDockWidget::setActive()
 		QString active_element_autonum = m_project->elementCurrentAutoNum();
 		int el_index = ui->m_element_cb->findText(active_element_autonum);
 		ui->m_element_cb->setCurrentIndex(el_index);
+
+			//Report
+		QString active_report_autonum = m_project->reportCurrentAutoNum();
+		int report_index = ui->m_report_cb->findText(active_report_autonum);
+		ui->m_report_cb->setCurrentIndex(report_index);
 
 			//Folio
 		if (m_project->defaultTitleBlockProperties().folio == "%autonum") {
@@ -338,6 +370,33 @@ void AutoNumberingDockWidget::on_m_element_cb_activated(int)
 }
 
 /**
+	@brief AutoNumberingDockWidget::reportAutoNumChanged
+	Add new or remove report (folio-reference arrow) auto num from combobox
+*/
+void AutoNumberingDockWidget::reportAutoNumChanged()
+{
+	ui->m_report_cb->clear();
+
+	//Report Combobox
+	ui->m_report_cb->addItem("");
+	QList <QString> keys_report = m_project->reportAutoNum().keys();
+	if (!keys_report.isEmpty()) {
+		foreach (QString str, keys_report) {ui->m_report_cb -> addItem(str);}
+	}
+	setActive();
+}
+
+/**
+	@brief AutoNumberingDockWidget::on_m_report_cb_activated
+	Set new report (folio-reference arrow) AutoNum
+*/
+void AutoNumberingDockWidget::on_m_report_cb_activated(int)
+{
+	m_project->setCurrentReportAutoNum(ui->m_report_cb->currentText());
+	refreshRow(AutoNumCategory::Report);
+}
+
+/**
 	@brief AutoNumberingDockWidget::folioAutoNumChanged
 	Add new or remove folio auto num from combobox
 */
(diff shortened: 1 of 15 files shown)


ispyisail:
This is broken, don't merge

scorpio810:
https://qelectrotech.org/forum/viewtopic.php?id=2063
https://qelectrotech.org/forum/viewtopic.php?pid=15275#p15275
https://qelectrotech.org/forum/viewtopic.php?pid=22605#p22605
https://qelectrotech.org/forum/viewtopic.php?pid=21850#p21850
https://qelectrotech.org/forum/viewtopic.php?pid=16153#p16153

scorpio810:
Thanks for the work on this, and for flagging it yourself before we merged — this matches something that keeps coming back on the forum for years (see the links I posted above).

The underlying issue is that auto-numbering evaluates the formula once, at creation time, while a folio report/cross-reference needs to be re-evaluated live every time a page is opened, moved, or inserted — that's why the label goes blank or picks up the wrong number after save/reopen. Rather than patching the NumerotationContext engine further, here's the approach I use that avoids the problem entirely, no code changes needed:

In the project settings (or general config), for any report/cross-reference formula:

Use %F (folio label) instead of %f (folio position/number). %F stays stable when folios are moved, renamed, or inserted — %f doesn't, since it depends on the folio's physical position in the project.
In each folio's titleblock, use %id alone instead of %id/%total. Otherwise cross-references show something like (folio_label/total_folios)-(line)-(column) instead of a clean reference.

To apply this to an existing project: use Ctrl+F (find/replace), go to the "Folio" section, and replace %id/%total with %id across the whole project in one pass.

With this, you can move folios around, rename them, or insert a new one in the middle — without breaking report links or XRefs.

> I work on old project at my work, I need to move some folio so I rename it to special folio label :9-1, 9-2, 9-3 to insert after folio 9 and before folio 10 I didn't use position but folio label for my Xref.

ispyisail:
Yes. that is why i flagged the problem. 

If i can solve these problems will you accept a merge?

scorpio810:
I don’t think it’s worth wasting your time on this PR...

ispyisail:
## Post-mortem: what breaks when auto-numbering meets folio reports, and what a working version would need

Recording this for whoever picks this up next, so the investigation doesn't have to be repeated. @scorpio810's diagnosis above is correct — below is the specific mechanism, and an honest assessment of what doing it properly would cost.

### First, a correction to my own commit on this branch

Commit `e0a5f97` ("Fix live report-link (Renvoi) destination text not updating") **should not be trusted — I'd suggest ignoring or reverting it.** I re-tested it properly and the bug it claims to fix does not exist:

* Stock QET already re-displays the folio reference correctly after save/reopen. I verified by reverting the change, rebuilding, and reloading the same project — the reference renders fine without it.
* The blank text I originally saw was an artifact of a hand-written test file with empty `<dynamic_texts/>`. `Element::fromXml()` deletes the definition's text items and rebuilds them from that block, so those arrows had no text items at all — nothing to display. I then "confirmed the fix" against a properly *saved* file, which would have worked either way.
* The commit message also claims `QETProject::refresh()` is never called. That's wrong — `readProjectXml()` calls it right after loading the diagrams.

Two smaller things in it may still be worth keeping, on much narrower grounds: `diagramRemoved` is connected in `setConnectionForReportFormula()` but never disconnected in the matching remove function; and `diagramAdded` isn't listened to at all — harmless for "add folio" (it appends, shifting nobody), but it does leave a stale reference after **undoing a folio deletion**, since `RemoveDiagramCommand::undo()` re-inserts mid-list via `addDiagram(d, pos)`.

### The actual conflict

For a folio report, QET redefines `label` to mean **"where my linked partner is"**, not "this element's own name". That redefinition is hard-coded in at least three places:

* `DynamicElementTextItem::updateReportText()` — evaluates the project's `<report label>` formula against the *linked* element
* `DynamicElementTextItem::reportReplacedCompositeText()` — replaces `%{label}` with that same cross-reference
* `DynamicElementTextItem::zoomToLinkedElement()` — treats `info_name == "label"` as the click-to-navigate target

Auto-numbering is built on the opposite assumption: `Element::setUpFormula()` writes the element's *own* identity into `elementInformations["label"]`.

Same key, two incompatible meanings — and the stock arrow elements have exactly one dynamic text (`info_name=label`) to hold them. There is no ordering fix, because in the auto-numbering use case both meanings are wanted at the same time.

### How that surfaces as a bug

`DynamicElementTextItem::elementInfoChanged()` unconditionally does `setPlainText(element->actualLabel())` when `m_info_name == "label"`, with no `AllReport` guard. It is connected to `Element::elementInfoChange`, because `elementUseForInfo()` returns the arrow itself for `NextReport` / `PreviousReport`.

So `elementInfoChanged()` and `updateReportText()` both write the same text item, with no arbitration — whichever signal fires last wins.

**This is not new to this PR.** `Element::setElementInformations()` also emits `elementInfoChange` and is reachable from `ChangeElementInformationCommand`. On master a report arrow has no formula, so `actualLabel()` returns empty and the reference gets **blanked** — a plausible mechanism behind the "report label disappears" reports in the forum threads linked above. What this PR changed is that `setUpFormula()` now runs for reports, so the clobbering value is an auto-number instead of an empty string: same defect, just louder.

A narrow guard in `elementInfoChanged()` (skip when `linkType() & AllReport` and `m_info_name == "label"`) looks like the correct root-cause fix for that family, and is independent of auto-numbering. I'm not proposing it here because it needs testing I haven't done.

Worth contrasting with master/slave, which solves the same "my text displays another element's data" problem cleanly: `elementUseForInfo()` returns the **master**, so the normal element-info path remains the single writer. Reports added a second parallel writer instead.

### What a working implementation would need

The C++ pattern already exists; the cost is mostly outside it.

1. **A distinct info key for the arrow's own number — don't reuse `label`.** There's already a precedent in the same file: `conductorPropertiesChanged()` dispatches on `m_info_name` (`conductor_function`, `conductor_color`, …), so each subsystem owns its own text item. Something like `report_id` fits that convention, and the two writers stop contending.

2. **A second `<dynamic_text>` on the arrow elements**, bound to that key. This is the expensive part:
   * `01coming_arrow.elmt` / `02going_arrow.elmt` live in the separate **qelectrotech-elements** repo — a cross-repo change.
   * Projects embed their own copy of the element definition under `<collection>`, so **existing projects would not gain the new text**. That needs a migration story.
   * User-made custom arrow elements wouldn't have it either.

3. **Deciding whether it's wanted at all.** The existing reference is already bidirectional, and arguably better for printed drawings: in a two-folio test the going arrow reads `2-C7` while its partner reads `1-D8` — each end says which page *and where on it*. A matching serial number carries strictly less information. Printed navigation was the original motivation for this PR, and stock behaviour already covers it.

### Status

Given the above, I'm not pursuing the feature further and don't think this PR should be merged in its current form. @scorpio810's `%F` + `%id` guidance stands on its own as the practical answer — `%f` legitimately tracks physical folio position, which is exactly what you don't want in a reference someone reads off paper.

Thanks for taking the time to look at it.

PR #704, opened 2026-08-09
Autonumbering: renumbering of elements

arummler:
For the moment only for elements. Similar to a python script offered in the forum which worked directly on the qet file. It works but draft for the moment as there might be additional features. Also would like to extend to conductors and folios.

Some adjacent things that might be useful:
- changing a naming schema should propagate it through the project; does not make sense to have "orphaned" formulas assigned; or at least not without a clear warning (might be useful if already build although that likely means someone messed up)
- deleting a naming schema while assigned to any object should be either impossible or at least also only possible after a clear warning

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ doc/*
 !doc/QElectroTech.qch
 QElectroTech.tag
 !doc/doc-utils
+ui_*.h
--- a/cmake/qet_compilation_vars.cmake
+++ b/cmake/qet_compilation_vars.cmake
@@ -277,6 +277,8 @@ set(QET_SRC_FILES
   ${QET_DIR}/sources/autoNum/numerotationcontextcommands.h
   ${QET_DIR}/sources/autoNum/numerotationcontext.cpp
   ${QET_DIR}/sources/autoNum/numerotationcontext.h
+  ${QET_DIR}/sources/autoNum/renumberelementscommand.cpp
+  ${QET_DIR}/sources/autoNum/renumberelementscommand.h
   ${QET_DIR}/sources/autoNum/ui/autonumberingdockwidget.cpp
   ${QET_DIR}/sources/autoNum/ui/autonumberingdockwidget.h
   ${QET_DIR}/sources/autoNum/ui/autonumberingmanagementw.cpp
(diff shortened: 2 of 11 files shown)


ispyisail:
Built this against **Qt6** since you asked — and since this branch predates the Qt6 default flip (`a1c090781`) by a month, it seemed worth confirming rather than assuming.

## Verified

Current master, Qt 6.10.2, gcc 15.2.0, Ubuntu 26.04:

- **262 commits behind master, still merges clean** — no rebase needed
- **builds clean: 490/491, exit 0**, and no warnings mentioning any of the new files
- binary runs (`0.200.1-dev`) and `--export-bom examples/industrial.qet` succeeds

So nothing in the Qt6 transition has broken it while it sat.

## What reads well

**It is properly undoable.** `RenumberElementsCommand final : public QUndoCommand` with `undo()`/`redo()` is exactly the right shape for a bulk mutation, and it is the thing the forum Python script could never offer — that one rewrote the `.qet` from outside, with no way back. Doing it inside the undo stack is a real improvement over the approach it replaces, not just a port of it.

**The CMake wiring is complete** — all four new sources are registered in `qet_compilation_vars.cmake`, so the build picks them up without anything extra.

## One small thing

`.gitignore` gains an unrelated line:

```diff
 !doc/doc-utils
+ui_*.h
```

Ignoring generated `ui_*.h` is sensible on its own, but it has nothing to do with renumbering, and unrelated changes riding along in a feature branch make a later bisect more confusing than it needs to be. Worth splitting out, or at least mentioning in the description.

## On the two design points you raised

> changing a naming schema should propagate it through the project… deleting a naming schema while assigned to any object should be either impossible or at least also only possible after a clear warning

Both look right to me, and I would go slightly further: they read less like adjacent nice-to-haves and more like **prerequisites**. Renumbering makes the existing inconsistency easier to reach — the more you renumber, the more chances there are to end up pointing at a schema that has since changed or gone. An orphaned formula is worse after this feature than before it, because the feature encourages exactly the operation that creates them.

If it helps decide the ordering: the "cannot delete a schema that is in use" guard is the cheaper half and stands on its own regardless of whether renumbering lands.

Happy to test any follow-up the same way. Nothing here blocks it as a draft — the mechanics work.

PR #705, opened 2026-08-10
Continued signal/slot migration

arummler:
Some of the newly guarded (Qt5 path) signal/slot connects could be still converted to method pointers. Mostly for completeness as they will go away anyhow soon.

Remaining: richtext editor (separate MR as that still requires to decide whether to re-import or carefully modernize).


scorpio810:
Please remove guards for 
#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0) not packageed here on upstream, so and i didn't think mainstream Linux distritions package it with Qt5.14 now...

Same for #if QT_VERSION < QT_VERSION_CHECK(5, 15, 0)  

I drop old Ubuntu Jammy 22.04 PPA, so minimal Qt5 version support is now #if QT_VERSION < QT_VERSION_CHECK(5, 15, 12).

scorpio810:
At the moment I don’t have any spare time to rework my packaging scripts; otherwise, they’d all be set to `cmake KF6=ON` and `Qt6`. 
Many packagings scripts is based on DigiKam C.Gilles scripts like Appimages, MacPorts on my mac mini M2 8 G RAM need  ~24 hours to compile all frameworks... and I think stopping KVM/Qemu Intel X86_64  macOS.
No need for windows, a CI/CD is now working on Github,
 It saves me from having to spend two hours cross-compiling MXE (500W  24C/48Threads) on my big workstation, which is really handy.

arummler:
Would you be interested mid-term in CI (and potentially CD) on each MR/merge and also for the other OS?

scorpio810:
> Would you be interested mid-term in CI (and potentially CD) on each MR/merge and also for the other OS?

Yes if the GLS bandiwch cost me 0€ ?

scorpio810:
Not an each merge, but pushed by QET's admin, like Windows CI/CD.
Some merges aren’t particularly useful for our users; others require the code to be checked thoroughly – we don’t want to break your lovely projects, let alone introduce a virus or a backdoor.

arummler:
For open source repos (at least until Microsoft runs out of money or motivation) github actions are free. The usual approach is to run on each push to an MR and also to add badges in the README.md which reflect whether it compiled and checks succeeded during the merge. True that there are security concerns if one uses any kind of github action secret (password, etc. in a hidden variable). That has to be configured carefully. Alternatives are that it runs only for pre-approved commiters or that it has to confirmed by a repo admin.

scorpio810:
I’d be delighted not to have to deal with the tedious task of creating packages for the ‘nightly’ builds of Debian stable and unstable on the amd64 architecture, and no longer have to worry about a flapak repository, nor spend hours recompiling an amd64 QEMU/KVM AppImage into QEMU/KVM or arm64 on my Raspberry Pi 5. Ultimately, all these scripts don’t take much time once the build environment is set up and automated by scripts, but it did require extensive testing and a great deal of time.
And I’m really glad to be able to avoid using the MXE toolchain to cross-compile Windows packages.

The problem is that, until now, all my scripts have been based on Qmake, and the transition to CMake is going to require a huge amount of work all over again.

scorpio810:
https://github.com/qelectrotech/qelectrotech-source-mirror/tree/master/build-aux/flatpak
https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/misc/MacQetDeploy.sh
https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/misc/MacQetDeploy_arm64.sh
https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/packaging_script_AppImage.sh
https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/packaging_script_AppImage_aarch64.sh
https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/packaging_script_Debian_Ubuntu_Windows.sh
https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/packaging_script_Flatpak.sh

scorpio810:
Nice cleanup, consistent with #698. One thing worth checking before merge:

In sources/ui/configpage/projectconfigpages.cpp, the three conversions inside the #if QT_VERSION < QT_VERSION_CHECK(5, 14, 0) blocks use:


```
connect(m_saw_conductor->contextComboBox(), &QComboBox::activated, this, &ProjectAutoNumConfigPage::updateContextConductor);
QComboBox::activated is overloaded in Qt5 (activated(int) and activated(const QString&)), so taking its address without qOverload<QString>() is ambiguous and should fail to compile — unlike the QSignalMapper::mapped conversions elsewhere in this same patch, which correctly use qOverload<...>().
```

Two ways to resolve:

Wrap it: qOverload<QString>(&QComboBox::activated)
Or, simpler and in line with what's already being discussed in this thread — just drop the < 5.14.0 / < 5.15.0 guards entirely (per the note above about no longer packaging for those Qt5 versions) and keep only the textActivated branch.
Everything else (exportdialog.cpp, qetapp.cpp, qetdiagrameditor.cpp, templatelogomanager.cpp) looks correct — qOverload is applied properly there and mirrors the existing #else (Qt6) code path.

arummler:
Done.

PR #709, opened 2026-08-10
Qt6 migration beyond signal/slot

arummler:
Based on the stale ispyisail/qt6-cmake branch and related MR #505 :
- same options for KF6 as for KF5
   - vendored in replacment
   - system KF6
   - download and build KF6
- small update of translation related cmake 
- postponed for another MR: writing qm files into build directory instead of source directory which is better for out-of-source builds; install still works; cmake workflow always rebuilds qm files from ts files: I propose to remove the qm files if there is no benefit to keep the binaries in the repo

Remaining:
- the GuiPrivate target handling as in #505 
- probably several small deprectiation fixes as in #505 or in separate MR
- there are even more depreciations but probably not in this MR; nothing stopping Qt6 builds


scorpio810:
Please don't add in your PR translations files.

arummler:
Okay, I will the separate the qm file question out. Moving the path change and deletion of qm files into a separate MR for discussion.

scorpio810:
Thanks for this, the version-agnostic BUILD_WITH_KF/BUILD_KF rename and the KF6 fetch path are a nice cleanup — much easier to reason about than three separate variable sets.

Two things I'd like to see fixed before merging:

1. `cmake/define_definitions.cmake` still references the old variable name:
```cmake
   if(BUILD_WITH_KF)
     message("KF_GIT_TAG               :" ${KF5_GIT_TAG})
```
   `KF5_GIT_TAG` no longer exists (renamed to `KF_GIT_TAG` in fetch_kdeaddons.cmake), so this will just print an empty value. Should be `${KF_GIT_TAG}`.

2. In `cmake/fetch_kdeaddons.cmake`, `set(BUILD_SHARED_LIBS OFF)` is a global CMake variable set with no save/restore around it, and it's included before `add_subdirectory(tests)` in the top-level CMakeLists. As written it will leak into everything configured afterward, not just the fetched KF targets. Could you save the previous value before the FetchContent block and restore it right after, e.g.:
```cmake
   set(_qet_prev_build_shared_libs ${BUILD_SHARED_LIBS})
   set(BUILD_SHARED_LIBS OFF)
   ...
   set(BUILD_SHARED_LIBS ${_qet_prev_build_shared_libs})
```

Minor nit, non-blocking: the `option(BUILD_KF "Build KF5 libraries, use system ones otherwise" YES)` description text is still KF5-specific, while `BUILD_WITH_KF` right above it got a proper version-agnostic comment. Worth aligning for consistency.

Aside from that, happy with the direction — the BUILD_DESIGNERPLUGIN/KCOREADDONS_USE_QML/BUILD_QCH OFF additions to keep Qt6Qml out of the fetched build are a good call.

scorpio810:
Separate from the review comments above — since the PR description lists "the GuiPrivate target handling as in #505" as remaining work, flagging some context on that from the Windows CI side that might save some time.

`.github/workflows/windows-build.yml` currently has a discrepancy between the Qt5 and Qt6 jobs: the Qt5 job passes `-DBUILD_TESTING=OFF`, but this project gates its tests via `PACKAGE_TESTS` (CMakeLists.txt), not `BUILD_TESTING` — so that flag is a no-op left over from when the `qt5-base-private` MSYS2 package was briefly added and removed on the same day, and `BUILD_TESTING` got flipped back to OFF the next morning without anyone realizing it wasn't wired to anything. The Qt6 job correctly uses `-DPACKAGE_TESTS=OFF`.

More relevant to the GuiPrivate work itself: neither job currently pulls in a `*-base-private` MSYS2 package (`qt5-base-private` or `qt6-base-private`) on master. The need for `Qt::GuiPrivate` comes from `QPdfEngine::drawHyperlink` (internal PDF hyperlinks), documented in `cmake/qet_compilation_vars.cmake`. On Qt5 there's no separate `Qt5GuiPrivate` package — the target is created implicitly alongside `Gui` — but on Qt6, `GuiPrivate` has to be requested explicitly as a component, which also means the Qt6 Windows CI job will need the corresponding `qt6-base-private` MSYS2 package added once this lands, or the Qt6 build will fail to find the private headers.

Worth keeping in mind when #505's GuiPrivate handling gets pulled into this branch, so the CI package list gets updated at the same time rather than in a follow-up fix.

arummler:
Okay, all done. I cannot test it but looking on msys2.org I have the impression there is no -private separation and it should be in the base package. The rest I would separate into another MR as it touches more files.

arummler:
Ready to merge. Afterwards please #705

scorpio810:
@arummler , thanks you but not free time to me to review it this day... maybe this week-end, I don't know?

Issue #307, opened 2024-09-24
Cannot open qelectrotech.app on macOS Sequoia 15.0


MacValley:
I downloaded qelectrotech-0.100.0-r8146-arm64.zip which contains the file qelectrotech.app.
When opening the file I get: “qelectrotech.app” is damaged and can’t be opened. You should move it to the Bin.
I have a MacBook Pro M2 with chip Apple M2 Pro, running the latest OS, macOS Sequoia 15.0.

Anyone experiencing the same problem?

Spokplacenta:
Same error

MacValley:
Today I downloaded qelectrotech-0.100.0-r8152-arm64.zip, and this was successfully opened. So whatever the difference
 is between r8152 and r8146, it now works. Great! I will close the issue that I opened last week.

Spokplacenta:
I just tested the r8152 version, I still can't open it
![image](https://github.com/user-attachments/assets/cfd63792-15f8-45de-b63c-ad1fb9262feb)

nightscape:
Same here on a MacBook Air M1 on MacOs Sequoia. It says
![image](https://github.com/user-attachments/assets/cbc7ceef-d745-469a-bcc6-c5e514636e35)
for both 8152 and 8166.

gieljnssns:
I've tried with every version I can download on https://download.qelectrotech.org/qet/builds/MAC_OS_X/arm64/
Always the same error
<img width="274" alt="Scherm­afbeelding 2024-11-08 om 11 13 36" src="https://github.com/user-attachments/assets/9847daa5-255c-41e2-8e84-5bcc884b97af">

gieljnssns:
If you do this, you can get it working
![Scherm­afbeelding 2024-11-08 om 11 29 16](https://github.com/user-attachments/assets/be480ead-7dfa-46f6-9566-26d9369a2ce4)

PR #710, opened 2026-08-10
Fix bugtracker #307: QET file type description bad encoding

ispyisail:
## Bug

https://qelectrotech.org/bugtracker/view.php?id=307

Windows file-type descriptions for `.qet`/`.elmt` files show up mojibake-corrupted in Explorer/regedit — "Schéma QET" renders as "SchÃ©ma QET" and "Élément QET" as "Ã‰lÃ©ment QET".

## Root cause

`build-aux/windows/QET64.nsi` writes these descriptions via `WriteRegStr` with the correct accented French text. The `.nsi` source was saved as UTF-8 **without a byte-order mark** and had no explicit `Unicode true` declaration. NSIS's encoding auto-detection for scripts without a BOM can fall back to the system ANSI codepage for short, mostly-ASCII literals like these — the two UTF-8 bytes of `é` get read as two separate CP1252 characters (`Ã` + `©`), producing the classic double-encoded mojibake once written into the (Unicode) registry value.

Git history shows this was hit before: commit `9fceb6f00` restored the correctly-encoded accented strings, which is exactly what triggered the mojibake reports. Commit `e9e2ea5b0` ("Try to fix bug #307") then worked around the *symptom* by replacing the strings with plain-ASCII English ("Diagram QET" / "Element QET") — avoiding the corruption, but silently dropping the original French wording rather than fixing the underlying encoding problem. That workaround is the current state on `master`.

## Fix

Per NSIS's own documented guidance for scripts containing non-ASCII literals:
- Save `QET64.nsi` as UTF-8 **with a BOM**.
- Add an explicit `Unicode true` declaration (this is already NSIS 3's default, but stating it makes the encoding contract explicit rather than relying on an implicit default that could change).
- Restore the original `"Schéma QET"` / `"Élément QET"` strings.

## Verification

- Compiled with `makensis` (NSIS 3.10, installed locally via `apt`). The compiler log changed from silently defaulting to ANSI-ambiguous parsing to explicitly reporting:
  ```
  Processing script file: "QET64.nsi" (UTF8)
  ```
  and the build completes cleanly, producing an `x86-unicode` installer.
- No encoding-related warnings. The only warnings present are `7010: ... no files found` for payload directories (`elements/`, `lang/`, `fonts/`, etc.) that aren't populated in this stripped source checkout — those are supplied by the real CI build and are unrelated to this change.
- I was **not** able to verify the installed registry values actually render correctly in a real Windows File Explorer/regedit, since no Windows or Wine environment is available in this sandbox. Confidence rests on NSIS's documented BOM-based source-encoding detection, and on the compiler's own encoding-mode log line reflecting the fix (previously ambiguous/ANSI-defaultable, now explicitly UTF-8).

### DIFF
--- a/build-aux/windows/QET64.nsi
+++ b/build-aux/windows/QET64.nsi
@@ -1,4 +1,4 @@
-; this file is part of installer for QElectroTech
+﻿; this file is part of installer for QElectroTech
 ; Copyright (C)2015 QElectroTech Team <scorpio@qelectrotech.org>
 ;
 ; This program is free software; you can redistribute it and/or
@@ -34,6 +34,13 @@
 ;--------------------------------
 ; NSIS 3 requires SetCompressor BEFORE any Section or Function
 SetCompressor /FINAL /SOLID lzma
+
+; This script is saved as UTF-8 with a BOM. Without an explicit BOM, NSIS
+; falls back to interpreting non-ASCII literals (e.g. the accented registry
+; strings below) using the system ANSI codepage instead of UTF-8, which
+; double-encodes them into mojibake (bug #307). Unicode is NSIS 3's default,
+; but is declared here explicitly so this doesn't silently regress.
+Unicode true
 
 ;--------------------------------
 ; Includes
@@ -315,7 +322,7 @@ Section ""
     WriteRegStr   HKEY_CLASSES_ROOT "Applications\qelectrotech.exe\shell\open\command" "" \
         '"$final_qet_exe" "%1"'
     WriteRegStr   HKEY_CLASSES_ROOT ".qet"                                              "" "qet_diagram_file"
-    WriteRegStr   HKEY_CLASSES_ROOT "qet_diagram_file"                                  "" "Diagram QET"
+    WriteRegStr   HKEY_CLASSES_ROOT "qet_diagram_file"                                  "" "Schéma QET"
     WriteRegDWORD HKEY_CLASSES_ROOT "qet_diagram_file"                                  "EditFlags"    0x00000000
     WriteRegDWORD HKEY_CLASSES_ROOT "qet_diagram_file"                                  "BrowserFlags" 0x00000008
     WriteRegStr   HKEY_CLASSES_ROOT "qet_diagram_file\DefaultIcon"                      "" "$final_project_ico"
@@ -323,7 +330,7 @@ Section ""
 
     ; File associations - .elmt
     WriteRegStr   HKEY_CLASSES_ROOT ".elmt"                                             "" "qet_element_file"
-    WriteRegStr   HKEY_CLASSES_ROOT "qet_element_file"                                  "" "Element QET"
+    WriteRegStr   HKEY_CLASSES_ROOT "qet_element_file"                                  "" "Élément QET"
     WriteRegDWORD HKEY_CLASSES_ROOT "qet_element_file"                                  "EditFlags"    0x00000000
     WriteRegDWORD HKEY_CLASSES_ROOT "qet_element_file"                                  "BrowserFlags" 0x00000008
     WriteRegStr   HKEY_CLASSES_ROOT "qet_element_file\DefaultIcon"                      "" "$final_element_ico"


ispyisail:
This same encoding root cause (missing UTF-8 BOM in `QET64.nsi`, corrupting non-ASCII literals when compiled by NSIS) also matches bugtracker #211 ("Charset-Issue on current installer (0.8-RC)" — German umlauts in "Zeichnungsköpfe" rendering corrupted in the installer). Flagging as a likely duplicate of the issue this PR fixes.

scorpio810:
Hi @ispyisail,

Thanks a lot for taking the time to dig into this — the analysis of the NSIS BOM/encoding behavior is genuinely useful and well explained.

That said, I think this one has been overtaken by events a bit. Since bug #307 was reported back in 2024-05-23, `QET64.nsi` has been completely rewritten and the installer moved from NSIS 2.x to NSIS 3.x. On top of that, the "Schéma QET" / "Élément QET" strings aren't hardcoded French literals anymore — they're now driven by variables. So the specific scenario this PR fixes (a hardcoded accented literal getting misread by NSIS's ANSI fallback) doesn't really apply to the current script anymore, and restoring hardcoded French text would go a bit against that change.

Would you mind rebasing against current `master` and seeing whether the issue still shows up there? If the strings are now handled through variables, there's a good chance the original bug simply doesn't apply anymore — which would be good news. Might be worth a quick look at #211 too, since you flagged it as a likely duplicate.

I'll close this for now, but please don't hesitate to reopen if you find it still reproduces against the current file — and thanks again for the thorough writeup, it's appreciated.

ispyisail:
You're right to close this, and I can now say so with a measurement rather than an opinion — I went and tested it instead of arguing from the NSIS docs, and the test refutes my own root-cause claim.

## The premise of this PR does not survive NSIS 3.x

I rebuilt the exact structure with `makensis 3.10`: a main `.nsi` with **no BOM and no `Unicode true`**, `!include`-ing a UTF-8 `.nsh` with **no BOM**, containing German umlauts.

```
Processing script file: "a_nobom.nsi" (UTF8)
Processed 1 file, writing output (x86-unicode):
```

| test | result |
|---|---|
| no BOM anywhere (= the current tree) | detected as **UTF8**, unicode installer |
| identical script, BOM added | installer is **byte-identical** — same md5 |
| include actually encoded CP1252 | **hard compile error**, not silent mojibake |

Two things follow. Adding the BOM is a **no-op** on this toolchain, so the fix in this PR would have changed nothing. And NSIS 3.x does not silently fall back to ANSI — feed it genuinely non-UTF-8 bytes and the build fails loudly. The silent-corruption mechanism I described was real under NSIS 2.x, and `145810f33` ("Update Windows installer NSIS to 3.X version") is what actually killed it. You named that migration in your comment; it is the decisive half of your answer.

## Retracting my #211 comment

I flagged bugtracker #211 (German umlauts) as a likely duplicate of this. That was wrong, and for the same reason. The umlauts are still there, unBOM'd, in `lang_extra.nsh`:

```nsi
LangString Refrigeration ${LANG_GERMAN} "Kältetechnik"
LangString wrongArch     ${LANG_GERMAN} "Dieses Programm läuft ausschließlich unter Windows 64 Bit."
```

but by the test above they compile correctly, so #211 is resolved by the toolchain move as well. Both can be closed on that basis.

## One correction, only because it changes what the follow-up should be

> the "Schéma QET" / "Élément QET" strings aren't hardcoded French literals anymore — they're now driven by variables

They are still hardcoded literals on current `master` — what changed is that they are now English:

```nsi
QET64.nsi:318   WriteRegStr HKEY_CLASSES_ROOT "qet_diagram_file" "" "Diagram QET"
QET64.nsi:325   WriteRegStr HKEY_CLASSES_ROOT "qet_element_file" "" "Element QET"
```

That is commit `e9e2ea5b0` ("Try to fix bug #307"), which sidestepped the encoding problem by removing the accents rather than by fixing the encoding. And there is no `LangString` for either description anywhere in `lang_extra*.nsh` — unlike, say, `Titleblocks`, which carries about thirty translations.

## What I think is actually left

The encoding bug is gone. The thing #307 was reported *about* is not: on a French Windows the file type still reads "Diagram QET" rather than "Schéma QET", and on a German one it reads the same English.

The clean fix is the one your comment describes — drive them through variables:

```nsi
LangString FileTypeDiagram ${LANG_ENGLISH} "QET Diagram"
LangString FileTypeDiagram ${LANG_FRENCH}  "Schéma QET"
LangString FileTypeElement ${LANG_FRENCH}  "Élément QET"
...
WriteRegStr HKEY_CLASSES_ROOT "qet_diagram_file" "" "$(FileTypeDiagram)"
```

That puts them where every other user-visible installer string already lives, restores the accented text without hardcoding French into `QET64.nsi`, and — per the test above — is safe unBOM'd on NSIS 3.x. Happy to open that as a small separate PR if you want it; otherwise I'm content to leave this closed.

Thanks for taking the time to explain the close rather than just clicking the button — checking it properly is what turned up the measurement, which is a more useful outcome than the patch would have been.

Issue #209, opened 2022-02-05
Project properties Window: Half page hidden


raul-theion:
Hello, 
I've just realized, that at certain resolutions at least, half of the Project properties are hidden due to the ability of that window to resize.
(Once you realize you can resize it, you figure out you've missed half of the properties).
Probably this can easily be fixed just using the Qt property "setMinimumWidth()"

Cheers!

ispyisail:
Looking at this now — the cause is almost certainly the `WindowModal` setting that was on the Project Properties dialog until recently.

**What was happening:** `ConfigDialog` was opened with `Qt::WindowModal`, which makes Qt constrain the dialog to appear *within* the bounds of its parent MDI subwindow (the project view). If that subwindow was resized to something small, the dialog was clipped to fit inside it — half the content invisible, no way to scroll out of the parent's bounds.

**Current code:** The dialog already has a `QScrollArea` wrapping the content and sets its maximum size to 94 % of the available screen area, so on its own it handles small screens correctly. The missing piece was the modality.

**Fix already in flight:** PR #528 changes the modality to `Qt::ApplicationModal`, which makes the dialog a true top-level window not constrained to the MDI subwindow. That should resolve the clipping reported here as a side-effect of the primary SIGSEGV fix.

@scorpio810 — if PR #528 lands, would you be happy to close this one as fixed by it? Happy to add a dedicated test step to #528 confirming the dialog appears full-size regardless of the parent subwindow dimensions if that would help.

ispyisail:
Resolved by PR #528 (merged 2026-06-22). Switching the Project Properties dialog to `Qt::ApplicationModal` makes it a true top-level window no longer constrained to the MDI subwindow bounds, so the half-hidden content issue is gone.

PR #711, opened 2026-08-10
Fix bugtracker #306: crash when restoring backup files on startup

ispyisail:
## Bug

https://qelectrotech.org/bugtracker/view.php?id=306

On startup, QET offers to restore backup/autosave files left over from a previous session. Clicking **Cancel** works fine. Clicking **OK** to actually restore the listed files crashes the application. The same files open fine via File > Open or the OS file explorer. The reporter also noted the restore list keeps growing across sessions.

Related issue #209 (fixed in 2021) was about restore filenames becoming too long due to URL-encoding of the full path — a different, already-fixed problem. This is a distinct crash in the restore/reopen path itself.

## Root cause

`QETDiagramEditor::openBackupFiles()` (`sources/qetdiagrameditor.cpp`):

```cpp
QETProject *project = new QETProject(file, this);
if (project->state() != QETProject::Ok)
{
    if (project -> state() != QETProject::FileOpenDiscard) { ... warning ... }
    delete project;
    DialogWaiting::dropInstance();
}
addProject(project);   // <-- runs unconditionally, even after delete above
DialogWaiting::dropInstance();
```

When a backup project fails to reach `ProjectState::Ok`, `project` is deleted — but there's no `continue`/`else`, so `addProject(project)` still runs on the now-dangling pointer. `addProject()` immediately dereferences it (`new ProjectView(project)`, etc.), producing a use-after-free.

This matches the reported behavior exactly:
- **Cancel** never calls `openBackupFiles()` — it just deletes the stale markers directly (`QETApp::checkBackupFiles()`), so it's unaffected.
- **OK** calls `openBackupFiles()`, which crashes on this UAF whenever *any* listed backup fails to open cleanly.
- Because the crash happens mid-loop, cleanup for that (and any later) backup file in the batch never completes — plausibly explaining why the restore list kept growing across sessions.

## Fix

Add the missing `continue` so a failed project is skipped instead of being passed use-after-free into `addProject()`.

## Verification

- Clean rebuild: only the intended object file recompiled, linked successfully with no errors or warnings.
- I attempted a live repro under Xvfb by crafting a malformed stale-file marker to force `ProjectState != Ok` and clicking OK. However, this local build links against real KDE Frameworks (`BUILD_WITH_KF5=ON`, confirmed via `CMakeCache.txt`) rather than the in-tree `sources/ui/nokde/kautosavefile.cpp` reimplementation I initially targeted — real `KF5::KAutoSaveFile` uses a different stale-marker directory/naming scheme (`~/.local/share/stalefiles/<app>/`) that I wasn't able to reverse-engineer accurately enough in the time available to produce a matching malformed marker.
- Confidence in the fix instead rests on the code itself: this is an unambiguous, textbook use-after-free — the exact object is deleted on the line immediately above the missing `continue` — with a single-line, side-effect-free fix.

### DIFF
--- a/sources/qetdiagrameditor.cpp
+++ b/sources/qetdiagrameditor.cpp
@@ -2101,6 +2101,7 @@ void QETDiagramEditor::openBackupFiles(QList<KAutoSaveFile *> backup_files)
 			}
 			delete project;
 			DialogWaiting::dropInstance();
+			continue;
 		}
 		addProject(project);
 		DialogWaiting::dropInstance();


ispyisail:
Also resolves bugtracker #240 ("Au démarrage, apparition d'un message de restauration de fichier... crash") — identical symptom pattern: clicking OK on the restore-files dialog crashes, closing via the window's X button works fine, and the dialog kept recurring at every startup (matching the explanation above: the crash interrupts the per-file cleanup loop mid-batch, so the stale marker for that file never gets removed and keeps reappearing next launch).

ispyisail:

ispyisail:
Reopening this one. It was closed in the batch that cleared @scorpio810's review backlog, not on merit, and it is a one-line fix for a crash — so it seemed worth bringing back on its own rather than leaving it sitting.

Re-verified against current `master`, and two things turned up that make the case stronger than the original description:

**The sibling function shows this is an oversight, not a design choice.** `QETDiagramEditor::openProject()` (line ~1257) has the identical structure — same `state() != Ok` check, same `FileOpenDiscard` special case, same `delete project` — and correctly ends with `return(false);`. `openBackupFiles()` is that same code adapted into a loop, where the early exit was dropped. The `continue` restores the pattern that already exists in the same file.

**The trigger is more ordinary than "a corrupt backup file".** `FileOpenDiscard` is set when the user clicks **Cancel** on the version-compatibility dialog (`qetproject.cpp:1512` and `:1536`). So the sequence is:

1. a backup exists from a project saved by a newer QET version
2. the user clicks OK to restore it
3. the version warning appears
4. **the user clicks Cancel**
5. `state() == FileOpenDiscard` → `delete project` → `addProject(project)` → use-after-free

`addProject()` dereferences immediately, via `QETApp::registerProject(project)` and `new ProjectView(project)`. A user declining a dialog crashes the application. That also fits the reporter's note that the restore list kept growing — the crash aborts the loop before the remaining backups are cleaned up.

Incidentally the `continue` also fixes a double `DialogWaiting::dropInstance()` on that path.

The change cannot affect the success path; it only adds an early exit to a branch that currently ends in undefined behaviour.

PR #247, opened 2023-01-14
German translations (including binary file)

plc-user:
The binary file was created with qtlinguist Version 6.4.2 from Debian-unstable-Repository.
On my Debian unstable the binary version of the translation-file works with this QET-version:
QElectroTech V 0.100-DEV+86487b392b0a58100c957a06d79751d3
Compilation: GCC 12.2.0
Built with Qt 5.15.8 - x86_64 - Date : Jan 14 2023 : 07:24:47

PR #718, opened 2026-08-11
Fix bugtracker #247: XRef slave reference hidden with dark themes

ispyisail:
## Bug

https://qelectrotech.org/bugtracker/view.php?id=247

With a dark theme/stylesheet applied to QElectroTech, the slave cross-reference text (e.g. `_(1-D3)_`, shown next to a slave element pointing back to its master) renders invisible (white-on-white or similar) — while other diagram text stays visible. The issue also affects generated PDFs.

## Root cause

`DynamicElementTextItem`'s slave cross-reference sub-item (`m_slave_Xref_item`) hardcodes `Qt::black` in three places (`sources/qetgraphicsitem/dynamicelementtextitem.cpp`): on creation, on hover-leave, and when restoring color after text editing.

The *parent* text item's color is a real, user-configurable property — `color()`/`setColor()`, persisted in the diagram XML, exposed in the text editor's color picker — so users can already work around dark themes for ordinary diagram text by explicitly setting a lighter color. But the slave-Xref sub-item never used it, so this one specific piece of text had no escape hatch at all, regardless of what the user tried.

## Fix

Replace the three `Qt::black` hardcodes with `color()` (the parent `DynamicElementTextItem`'s own configured color, inherited from `DiagramTextItem`). This doesn't change default appearance — `color()` defaults to black, same as before — but lets the slave-Xref text finally follow whatever color the user sets on the parent text field.

## Verification

- **Not verified**: a live visual confirmation of the slave-Xref text picking up a non-default color. Reproducing this requires constructing a master/slave-linked element pair with composite text containing `%{label}` in an actual multi-folio project, which was out of scope for the time available. Confidence rests on this being a direct, mechanical substitution of an existing, already-used accessor (`color()`) for a hardcoded constant, applied identically to the exact three call sites that previously hardcoded `Qt::black` for this item, with no other logic changed.


scorpio810:
BTW, if you could add XREfs and folio report on DXF export : https://qelectrotech.org/forum/viewtopic.php?id=2481

ispyisail:
Had a look at the DXF export for the forum thread. Two findings:

**Folio reports already export.** Before changing anything I exported `examples/industrial.qet` to DXF and counted: the 317 folio/position strings present are the folio-report texts. They are `DynamicElementTextItem`s and the walker already collects them. If something specific is still missing there, a sample project would help.

**Cross-references were genuinely missing, and the slave half is now fixed** in #740. The slave label — the `(6-G15)` pointing back to the master — is a plain `QGraphicsTextItem` child of a `DynamicElementTextItem`, so it matched none of the casts in `generateDxf()` and was dropped. Comparing against the PDF export, which renders the whole scene:

| | before | after | PDF |
|---|---|---|---|
| slave xrefs `(n-Xn)` | 0 | **41** | 41 |
| folio/position strings | 317 | 358 | 403 |

The remaining 45 are the master-side cross-reference table drawn by `CrossRefItem`. That one paints itself with custom `QPainter` code in three modes (`drawAsCross`, `drawAsContacts`, `drawAsPlcTable`) including contact symbols, so a faithful DXF version is a much bigger job — probably better done as a DXF-writing paint device than per-item. Say the word if you want that pursued.

(#740 is separate from this PR, which is still just the dark-theme colour fix for bugtracker #247.)

ispyisail:

ispyisail:
Reopening. Closed in the batch that cleared @scorpio810's review backlog rather than on merit, and re-reviewed against current `master` before bringing it back.

**Merged with master**, which had restructured `updateXref()` since this was written (different nesting, `xrp` hoisted out of the branch). Rather than hand-resolving the conflict markers I took master's version of the file and re-applied the three substitutions onto it, so the result sits on the current structure.

**The change covers exactly the right call sites.** There are three places that hardcode `Qt::black` for `m_slave_Xref_item` — in `mousePressEvent()`, on `GraphicsSceneHoverLeave`, and where the item is constructed in `updateXref()` — and all three are changed. The `Qt::blue` at the hover-*enter* case is the deliberate highlight and is left alone.

**Default appearance is unchanged.** `DiagramTextItem::color()` returns `defaultTextColor()`, which is black unless the user has set something, so this only takes effect for a text field whose colour was explicitly configured.

One spot is slightly more correct than before: in `mousePressEvent()` the line immediately above is `setDefaultTextColor(m_user_color)`, so `color()` now returns the colour just restored — where the old code reset the cross-reference to black regardless.

**Known limitation, pre-existing and not made worse here:** `DiagramTextItem::setColor()` does not propagate to `m_slave_Xref_item`, so changing the colour live in the editor will not refresh the cross-reference until `updateXref()` or a hover-leave next fires. The load path is fine — `fromXml()` calls `setColor()` before `updateXref()` runs — which is what matters for the reported dark-theme and PDF-export cases.

**Verified:** Qt 6.10.2 / KF 6.10.0, build clean (0 errors), `ctest` 4/4.

**Not verified:** the visual result, interactively. Confirming it needs a master/slave pair with composite text in a multi-folio project under a dark stylesheet, which is not reachable headlessly. The argument rests on the substitution being mechanical and on `color()` defaulting to the same black it replaces — both checkable from the diff.

PR #719, opened 2026-08-11
Fix bugtracker #245: bare %name custom variables not detected in title blocks

ispyisail:
## Bug

https://qelectrotech.org/bugtracker/view.php?id=245

When a title block template cell contains a bare `%name` custom-variable placeholder (as opposed to the braced `%{name}` form), it isn't detected when the template is selected in Project Properties — the "Custom" tab doesn't get auto-populated with an entry for it. This fails both for a single bare variable with a trailing space (`%name2 `) and for two bare variables in the same cell (`%name2 %name3`). Manually adding the variable in Project Properties works and renders correctly.

## Root cause

`TitleBlockTemplate::listOfVariables()` (`sources/titleblocktemplate.cpp`) — which scans a template's cells to auto-populate the Custom tab (a feature recently added by another contributor, see `TitleBlockPropertiesWidget::addTemplateVariables()`) — only matched the braced `%{name}` form:

```cpp
static const QRegularExpression rx(QStringLiteral("%\\{([^}]+)\\}"));
```

The bare `%name` form is a fully legitimate, already-supported substitution syntax — `TitleBlockTemplate::interpreteVariables()` already replaces both forms when rendering — but `listOfVariables()` never matched it at all. Both reported symptoms are really the same underlying gap: any cell using only the bare form produces zero detected variables, not "first only" or "trailing space breaks it" specifically.

## Fix

Extend the regex to also match a bare `%name` as the longest run of identifier characters immediately after `%`, via a second alternative/capture group. This naturally stops at whitespace, so `%name2 ` and `%name2 %name3` are both now correctly detected — no change to existing braced-form handling, and the existing `globalMatch()` loop already correctly finds multiple matches per cell.

## Verification

- Wrote a standalone test of the regex/extraction logic covering exactly the reported repro cases — `%name2`, `%name2 ` (trailing space), `%name2 %name3` (two bare variables), `%{name2}` (braced, unaffected), a braced+bare mix, plain text with no variables, and two built-in-style names — all extracted correctly with no regression to the previously-working braced form.
- **Not verified**: the actual Project Properties "Custom" tab UI auto-populating live end-to-end, since exercising the full Xvfb GUI flow was out of scope given the extraction logic itself was already precisely verified in isolation.


ispyisail:

ispyisail:

The thing worth checking is what the bare-name branch of the regex newly picks up, since over-matching would clutter the Custom tab. Scanning the ten shipped title blocks, the bare `%name` forms present are:

`aprouv`, `author`, `client`, `control`, `date`, `date1`, `date2`, `dessin`, `filename`, `folio`, `label1`…`label5`, `model`, `objet`, `projet`, `rev`, `revision`, `title`

Those are real custom variables that were previously invisible to `listOfVariables()`. The standard fields among them (`author`, `date`, `title`, `filename`, `folio`) are already dropped by the `reserved` set in `TitleBlockPropertiesWidget::addTemplateVariables()`, so they do not reach the Custom tab. Note `[A-Za-z0-9_-]+` includes the hyphen, so a bare `%folio-id` still matches the reserved `folio-id` rather than truncating to `folio`.

The capture selection is right in the edge cases too: for a braced match group 1 is set, for a bare match group 1 is empty and group 2 set, and a literal `%{}` leaves both empty so the existing `!name.isEmpty()` guard drops it.

**Not verified:** the Custom tab rendering the newly detected variables in the GUI.

Issue #238, opened 2022-12-06
Question about ARM target in future release


george-andrei:
Hello,

I just discovered this awesome project, and first I would like to say a big thank you to the community around this project. Looks amazing! 😃 

As I am a user of Linux on ARM, I am wondering if there is any plan to compile the binaries for ARM, or if someone already experimented around this. If not, I am gladly open to test for you. 
In my opinion, ARM is getting more and more popular on desktop. First Apple with their M1s and now Microsoft is pushing it as well, so I see only benefits. 
To be honest, I don't know how much work the implementation of ARM target would mean for this project, so don't get me wrong, I am just curious if this has been already discussed or not. 

Cheers!

scorpio810:
Like Debian for QET stable version, in my PPA you have ARM ports for the 0.9-dev
https://launchpad.net/~scorpio/+archive/ubuntu/qelectrotech-dev/+packages

Architecture tag: armhf: ARM ARMv7 Hard Float 
Architecture tag: arm64: ARM ARMv8

george-andrei:
Hi @scorpio810,

Thanks a lot, that worked like a charm! 
Because use a PineBook Pro with Manjaro on it, I had to convert the .deb into a .pkg with `debtap`, but after this step, I installed the package with not problem:

![image](https://user-images.githubusercontent.com/12943454/208714765-ce98f9e8-73da-46f9-980e-cb8c8e1a7ee7.png)

With the holydays approaching, I will not have the time to further test this, but will surely do in 2023 😉 

(Note for the bots: I use Manjaro ARM with kernel 6.0.12 and QET V0.90-DEV)

scorpio810:
Hi @george-andrei,

thanks for your feedback,
You could share QET logs?

scorpio810:
![208714765-ce98f9e8-73da-46f9-980e-cb8c8e1a7ee7](https://user-images.githubusercontent.com/19955399/208990518-45a7bdaa-149b-44f2-88b1-e72b56dcd2fa.png)
Your version is very old, try latest builds

scorpio810:
Hi @george-andrei,
I would like to buy the new mac mini M2 to make universal bundles for macOS (arch:amd64 & arm64):

https://qelectrotech.org/forum/viewtopic.php?id=2393

> Hello,
> 
> I just discovered this awesome project, and first I would like to say a big thank you to the community around this project. Looks amazing! smiley
> 
[...]

PR #720, opened 2026-08-11
Fix bugtracker #238: table of contents pages in wrong order

ispyisail:
## Bug

https://qelectrotech.org/bugtracker/view.php?id=238

The summary/table of contents report displays pages in the wrong order.

## Root cause

`SummaryQueryWidget::queryStr()` (`sources/dataBase/ui/summaryquerywidget.cpp`) built the SQL `ORDER BY` clause purely from whichever columns the user chose to *display* in the report, in the order they picked them. "Position" (`pos`, the folio's actual position in the project) is just one of many optional display columns, defaulting to not selected — so a report was left completely unordered by folio position unless the user happened to both add "Position" as a displayed column **and** place it first among their chosen columns.

This exactly matches the forum diagnosis linked from the bug report ([viewtopic.php?pid=13910](https://qelectrotech.org/forum/viewtopic.php?pid=13910#p13910)): *"When you set the position to column 1, the order works"* — because the `ORDER BY` clause mirrored the display-column order, so `pos` had to be both present and first to get correct sorting, which nothing in the UI suggested or required.

## Fix

Always sort by `pos` (a genuine column of `project_summary_view` — see `projectDataBase::createSummaryView()`) as the primary sort key, regardless of which columns the user chose to display or in what order — with the user's chosen display columns still applied as secondary sort keys for ties, same as before. The `SELECT` column list (what's actually shown) is left untouched.

## Verification

- Wrote a standalone test of the query-building logic covering: display columns not including `pos`, `pos` chosen first, `pos` chosen in the middle, `pos` chosen alone, and no columns chosen — all produced correct SQL with `pos` as the primary `ORDER BY` term, proper comma placement, and no duplicate `pos` term when the user had already selected it themselves.
- **Not verified**: an actual generated table-of-contents report's visual row order in the running application — constructing a multi-folio project with a custom summary query configuration was out of scope for the time available, given the SQL-generation logic itself was already precisely verified in isolation, and `pos` was confirmed to genuinely exist in the underlying view.

### DIFF
--- a/sources/dataBase/ui/summaryquerywidget.cpp
+++ b/sources/dataBase/ui/summaryquerywidget.cpp
@@ -63,7 +63,6 @@ QString SummaryQueryWidget::queryStr() const
 	QStringList keys = selectedKeys();
 
 	QString select ="SELECT ";
-	QString order_by = " ORDER BY ";
 
 	QString column;
 	bool first = true;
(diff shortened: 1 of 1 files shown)


ispyisail:

PR #721, opened 2026-08-11
Enable Information tab in element editor for Slave and Terminal basetypes

### LINKED ISSUE #663: [Feature Request] basetype: Terminal Block "enable" the information menu in element editor
In my opinion generally for basetype: Slave, Terminal Block (and others), also would be nice to have the Information tab available form Element editor.
Specially for the terminal blocks, when I work with TBs I just want to get out from the elements list eg. the "Phoenix Contact PTTB 2,5" and there it will be filled out from the element source the full property list.
Also make this sense when working with "buildable" contactors, (eg: Siemens) when the options highly exceeds the auxiliary 1-4, and every Contact (Slave) element has different part number than the Coil.
Even while hiding the Master's linked fields.

ispyisail:
Closes #663.

## Problem

The "Informations" tab in the element editor's properties dialog was only made visible for `Simple` and `Master` basetypes:

```cpp
ui->tabWidget->setTabVisible(1,
    (type_ == ElementData::Simple || type_ == ElementData::Master));
```

`Slave` and `Terminal` elements had no way to attach manufacturer/reference/part-number metadata directly in the element editor, even though `ElementData::m_informations` already stores and round-trips this data identically for every basetype - the tab was just never shown for these types.

`updateTree()` was already partially expecting this: it already enables the tree for `Terminal`, and `populateTree()` already injects PLC-specific info rows for `Slave` elements with `m_slave_type == PLCSlave` - but none of that was reachable because the tab itself was hidden.

## Fix

- Extend the `setTabVisible` condition to include `Slave` and `Terminal`.
- Flip `updateTree()`'s `Slave` case from `setDisabled` to `setEnabled` so the tree is actually editable once visible (matching `Terminal`'s existing behavior).

## Use case (from the issue)

> Specially for the terminal blocks, when I work with TBs I just want to get out from the elements list eg. the "Phoenix Contact PTTB 2,5" and there it will be filled out from the element source the full property list. Also make this sense when working with "buildable" contactors, (eg: Siemens) when the options highly exceeds the auxiliary 1-4, and every Contact (Slave) element has different part number than the Coil.

## Known follow-up (not addressed here)

The issue also raises hiding "the Master's linked fields" on Slave elements - i.e. filtering which info keys are shown/editable on a Slave to avoid confusion with fields that conceptually belong to its Master. That's a UX/semantics decision (which keys, if any, should be filtered per basetype) rather than a technical blocker, and there's precedent for basetype-aware filtering in the existing PLC-Slave code path. Left out of this PR to keep it minimal; happy to follow up if maintainers want that filtering.

## Testing

Built cleanly (incremental build, only the touched file recompiled). Interactive GUI verification of the element editor dialog was not performed in this sandbox - not verified live.


ispyisail:
Merged current master in to resolve the conflict. The clash was only in the `setTabVisible` block: master dropped the `#if QT_VERSION >= 5.15` guard around it (5.15 is now the floor), while this branch had extended the condition inside that guard. Resolved by keeping the extended condition without the guard. The `updateTree()` Slave `setDisabled`→`setEnabled` hunk applied cleanly. Builds clean against master (Qt5).

ispyisail:

ispyisail:
Reopening, with one extra commit — reviewing this before merging turned up a defect in the original change.

## The problem with the PR as it stood

Making the Informations tab visible for `Slave` was only half the change. `ElementScene::toXml()` decides which base types persist their info:

```cpp
if (type_ == ElementData::Simple ||
    type_ == ElementData::Master ||
    type_ == ElementData::Terminal ||
    type_ == ElementData::Thumbnail)
{
    // write <elementInformations>
}
```

`Slave` is absent, and that is the only place in the tree that writes the block. So the editor would have shown an editable tab on a slave element, accepted whatever was typed into it, and **dropped it silently on save** — worse than the tab being hidden, because it invites the data entry and then discards it.

The shipped collection matches the condition exactly:

| base type | elements | carrying `<elementInformations>` |
|---|---|---|
| **slave** | 75 | **0** |
| terminal | 70 | 41 |

`Terminal` was never affected — it is in that list already, and `updateTree()` already called `setEnabled(true)` for it, so the tab being hidden was a plain inconsistency.

## The added commit

`Persist elementInformations for Slave elements too` adds `ElementData::Slave` to that condition, with a comment recording why it is there.

Safe in both directions: `ElementData::fromXml()` reads `<elementInformations>` unconditionally with no base-type check, so existing slave elements are unaffected and newly written ones load back correctly. It also makes `populateTree()`'s PLC-slave branch reachable for the first time — those five PLC rows live in `m_informations`, so they could not have been saved either.

## Verified

Qt 6.10.2 / KF 6.10.0: build clean (0 errors), `ctest` 4/4 passing, merges cleanly with current `master`.

Not verified: the editor dialog itself, interactively. `ElementScene::toXml()` sits in the GUI editor and the test suite only covers header-only helpers, so the save path could not be exercised headlessly. The reasoning above rests on the code and on the corpus evidence, both of which are checkable from the diff.

PR #507, opened 2026-06-16
snap: migrate to core24, kde-neon-6 (Qt6), cmake build

ispyisail:
## Problem

Issue #373: the snap crashes on Ubuntu 24.04 Wayland/X11 hosts:

```
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized.
```

The snap was built against `core22` (Ubuntu 22.04). The xcb Qt platform plugin links against 22.04 xcb libraries. When the snap runs on a 24.04 host, `dlopen()` fails because the host ABI does not match — producing the misleading "found but could not load" error.

## Solution

Migrate the snap from `core22` + Qt5 to `core24` + Qt6, rebuilding all Qt libraries and platform plugins against the Ubuntu 24.04 ABI.

### snapcraft.yaml changes
- `base: core22` → `base: core24`
- `extensions: [kde-neon]` → `extensions: [kde-neon-6]` (Qt6 + KF6 content snap)
- `architectures:` → `platforms:` (required for core24 in snapcraft 9.x)
- Remove manual `kde-sdk-setup` rsync part — `kde-neon-6` extension provides the SDK automatically
- Build system: `qmake` → `cmake` (qmake is unsupported in KDE Neon 6 extension)
  - `-DBUILD_WITH_KF=ON -DBUILD_KF=OFF`: use system KF6 from SDK content snap
  - `-DPACKAGE_TESTS=OFF`: skip Catch2/GoogleTest FetchContent in snap build
- `kf5-5-110-qt-5-15-11-core22` cleanup snap → `kf6-core24`
- `python3.10` → `python3.12` (core24 default Python)
- Pin `qet-tb-generator` source to `source-tag: v1.31` (issue #202)
- Stage `libxcb-cursor0` — required by Qt's xcb platform plugin since Qt 5.15.x / Qt 6.x; the `kf6-core24` content snap does not bundle it

### cmake changes (from this PR)
- `cmake/build: make Qt::GuiPrivate (PDF hyperlinks) optional` — the `kde-qt6-core24-sdk` snap does not ship Qt6 private headers; guarded behind a `QET_PDF_PRIVATE_HEADERS` compile flag so builds without private headers skip link-injection silently

## Commit structure

This PR sits on top of **PR #505** (cmake Qt6/KF6 support + C++ compatibility fixes) and adds only three commits of its own:

| Commit | Scope |
|--------|-------|
[...]

PR #722, opened 2026-08-11
CI/CD overhaul

arummler:
Started as a follow-up of the discussion with @scorpio810 in #705 
- implemented very basic linux, mac and freebsd workflow
- (fails partially, partially due to missing #709  )
- windows left out for the moment as working implementation exists (needs careful extension to on-pull trigger)


scorpio810:
You forgot   for Qt5 qtbase5-dev,
               libkf5widgetsaddons-dev,
               libkf5coreaddons-dev, 
               libsqlite3-dev,
               qtbase5-private-dev,
               pkgconf

arummler:
Thanks, yes it is only a sketch. But MacOS compiles already through on my fork (with the migration PRs included; partially this one fails as master does not have the KF5/6 and GuiPrivate fixes yet),

arummler:
On my repo copy the compilation already runs through. That will be visible here only after the code fixes are done on master. When I have time I will look into the packaging step. There I might have some questions regarding standard procedures and design decisions.

arummler:
@scorpio810 Just to avoid duplication: I have on another branch currently all OS running through and also a Mac bundle gets build. As several tries were necessary I will squash the commit later and transfer into this branch.

arummler:
Current status: 
- Basic on-pull check for linux-gcc, macos, freebsd, windows-msvc. 
- MacOS DMG package creation and upload as an artifact. Currently on-pull for testing. Signing should be supported but not tested yet as I was not able to generate a compatible certificate on Linux to shim it.
Proposed:
- separate existing CI into build and package step so that the build can run on-pull
- add a workflow step which collects all artefacts if successful and uploads them to the github page
- further packages: AppImage (does it have to be a RPi image? possible but could use also oldest Ubuntu), Snap (with migration to core24 as in #507 ), Flatpack, any need to test deb/rpm packaging on the CI?

arummler:
@scorpio810 Regarding the version: I would propose to introduce a VERSION file and to pick it up automatically in the CMakeLists.txt PROJECT and qetversion.cpp using standard CMake mechanisms. Mostly to avoid having them in two places and also to avoid having to regex multiple times during packaging. I can do that in this branch if you agree or later.

scorpio810:
I try to update my Qemu/KVM to Ubuntu 22.04 and use https://invent.kde.org/graphics/digikam/-/tree/master/project/bundles/appimage?ref_type=heads,

Maybe if you have an Appimage builder's?

scorpio810:
BTW, for macOS, have you tried DMG, export database work, and export PDF with links is ok?
If I want to upgrade Digikam script for macOS with MacPorts  for Qt6 this take 24 hours to compile it on my Apple mini M2 8G   RAM

arummler:
Regarding the DMG for MacOs:
It gets built but unfortunately I do not have a Mac so I cannot test it. It gets currently uploaded as a couple of artefacts so if you could test it, it would be valuable. The whole MaacOs build takes <10 minutes if a github runner is available. I have currently set BUILD_WITH_KF=OFF as I do not see an advantage but if you prefer some to compile it with the KF libraries I can try. You had previously a minimum MacOs version of 10.3 which is really old. I am taking for implification the Qt from homebrew which drives the minimum MacOs version. In the moment it is 14.0. In case you would also want to serve people with really old computers one would need to add a Qt compilation step. That's possible - I did it already for another project - but takes multiple hours and one has definitely to cache it to avoid waiting each time so long for the result.

arummler:
Regarding the AppImage:
If I understand it correctly this digikam script basically compiles the whole KDE library which takes that long? Why do you need that? Why not the vendored-in fallback or just the selected couple of libraries for whicch we have already the mechanism built-in?

I am setting up currently on a Ubuntu 22.04 x86_64 and arm image. I do not have really experience with AppImage but according to the documentation there is not really a presecribed base system; just a "the older the more compatible". Does it have an advantage to do it on a RPi? I can load an appropriate image in CI and use e.g. docker if needed. In any case the avantage would be that you do not have to do it manually anymore and it will be much faster....

scorpio810:
> Does it have an advantage to do it on a RPi?


No, compiling an AppImage takes more than 10 minutes on RPI5 8G + SSD PCIEX; I could have emulated a Debian arm64 on my Mac mini M2 and build it in one or two minutes – I just did it for fun.

arummler:
Okay, two AppImage files get built successfully. I tried to start it locally and it seems to work. One could still bundle with the standard plugins. But I did only the Qt6 version considering you announced the quite imminent phase-out of Qt5 support.

scorpio810:
@arummler  great, thanks you, I not tried yet but finished to push my flatpak-builder on Qt 6.11 pfffff
https://qelectrotech.org/forum/viewtopic.php?pid=23174#p23174

scorpio810:
@arummler  I saw you use qmake, please go to cmake and use KF=ON

scorpio810:
Hum Jammy Qt6 version is very old 
QElectroTech V 0.200.1-dev
Compilation : GCC 11.4.0
Built with Qt 6.2.4 - x86_64 - Date : Aug 16 2026 : 12:01:59
Git Revision : 5a3cae3127ee11d58c3b9173409a38236575a69a 
Run with Qt 6.2.4 using 48 thread(s)
CPU : model name : AMD Ryzen Threadripper 3960X 24-Core Processor 
RAM Total : 64 GB
RAM Available : 58 GB

scorpio810:
PDF links works, but export database no. maybe forgot libsqlite3-dev?

arummler:
> @arummler I saw you use qmake, please go to cmake and use KF=ON

That refers to AppImage? Yes, it uses linuxdeploy with linuxdeploy-plugin-qt which currently still requires qmake internally to find some paths. I saw on their issue tracker that a couple people asked, e.g https://github.com/linuxdeploy/linuxdeploy-plugin-qt/issues/156 But no action yet. I could try alternatively not using linxudeploy, e.g. https://www.qt.io/blog/deploying-to-linux-with-cmake

I will try to build the KF libraries.

arummler:
> Hum Jammy Qt6 version is very old
> QElectroTech V 0.200.1-dev
> Compilation : GCC 11.4.0
> Built with Qt 6.2.4 - x86_64 - Date : Aug 16 2026 : 12:01:59
> Git Revision : [5a3cae3](https://github.com/qelectrotech/qelectrotech-source-mirror/commit/5a3cae3127ee11d58c3b9173409a38236575a69a)
> Run with Qt 6.2.4 using 48 thread(s)
> CPU : model name : AMD Ryzen Threadripper 3960X 24-Core Processor
> RAM Total : 64 GB
> RAM Available : 58 GB

I would be happy to move to Noble. The only advise regarding the AppImagge base system was "oldest LTS" so I did that. Alternative would be not to use the system Qt but build it myself. Possible but a bit annoying.

scorpio810:
As for me, the QEMU/KVM 20.04 AppImage VM was successfully upgraded to 22.04, but the compilation of Digikam’s Qt6 frameworks failed...

scorpio810:
> I would be happy to move to Noble. The only advise regarding the AppImagge base system was "oldest LTS" so I did that. Alternative would be not to use the system Qt but build it myself. Possible but a bit annoying.

Yes I'm think it's a good idea, for PPA deb (I’ve dropeed Jammy (Qt62.4) and Noble (Qt6.4.2)...keep resolute (Qt6.10.12))

https://github.com/marketplace/actions/install-qt

```
 name: Qt6 Ubuntu CI

on:
  push:
    branches: [ main ]
  pull_request:
    branches: [ main ]

jobs:
  build:
    runs-on: ubuntu-24.04

    steps:
      # 1. Récupération du code source
      - name: Checkout Source Code
        uses: actions/checkout@v4

      # 2. Installation automatique de Qt6 (Ex: Qt 6.8 LTS)
      - name: Install Qt6
        uses: jurplel/install-qt-action@v4
        with:
          version: '6.8.*'              # Installe la dernière révision stable de Qt 6.8
          host: 'linux'                 # Spécifie la plateforme hôte
          target: 'desktop'             # Cible classique
          install-deps: 'true'          # Installe les dépendances système requises (libxcb, etc.)

      # 3. Configuration et compilation du projet avec CMake
      - name: Configure and Build
        run: |
          cmake -B build -S . -DCMAKE_BUILD_TYPE=Release
          cmake --build build --parallel $(nproc)

```

arummler:
> As for me, the QEMU/KVM 20.04 AppImage VM was successfully upgraded to 22.04, but the compilation of Digikam’s Qt6 frameworks failed...

I know that you are following what they did for Digikam. But that is a fully KDE integrated program and hence needs all the KDE libs which QElectrotech is not and does not need beyond what is necessary for autosave, and a couple of widgets. But beyond KDE there is Qt itself...

For some strange reason they basically compile all dependencies themselves and do not rely at all on system provided libs (beyond the bare minimum of the kernel). Not sure why; to have very high control of what is being shipped? In that case I understand why it takes 24h on your little SBC. If you think that this is the way to go and want to do it on a CI runner instead it should be cached.

scorpio810:
I’ve spent a lot of time today on the Qt6 AppImage VM, and also on the Flatpak Builder, which is working fine, so I’m very interested in your AppImage CI/CD.

I can't try macOS x86_64 because my VM was based on macOS Monterey and I can't bypass flag on bundle.
I don't want to try to update Digikam script to Qt6 on this VM, maybe also not try it on mac mini if you solve database export with sqlite3 but see -> MacOS Homebrew environment (old)
* Note : we can't use Qt 5 sqlite3 plug-in without compile Qt 5
https://qelectrotech.org/wiki_new/doc/macosx


I need to try Apple silicon bundle.

scorpio810:
@arummler fyi I tried macOS Qt6 aarch64 works weel, database export, PDF links OK but you forgot elements dir tiitleblocks, lang, etc

arummler:
> @arummler fyi I tried macOS Qt6 aarch64 works weel, database export, PDF links OK but you forgot elements dir tiitleblocks, lang, etc

Hmm, the files are included in the bundle, I had checked.  But I believe there is a bug in the search path used by QElectrotech to find the directories.

arummler:
I tried to fix it and looking at the embedded strings/paths it should be right for DMGs now. But resolveConfiguredDataPath() is still not super straightforward and might not work properly (or only by accident) in other package formats. I think the path handling could profit from a clean up. Afaik there is no lookup for "examples". The user has to find them on their own.

I update the AppImage base system to 24.04. Can still add a custom Qt. 

Next thing should be to separate out on-pull from package in your existing windows workflow. And to add one job which collects all the artefacts and uploads them.

scorpio810:
I have the same problem, on Debian packaging, Flatpak, binary no find element common, lang, etc. so the issue in the code or in cmake files.

```
[06:38:05] laurent@debian:~/paquet_qet/debian$ cat qelectrotech.install 
usr/bin/qelectrotech
usr/share/icons/hicolor/16x16
usr/share/icons/hicolor/22x22
usr/share/icons/hicolor/32x32
usr/share/icons/hicolor/48x48
usr/share/icons/hicolor/64x64
usr/share/icons/hicolor/128x128
usr/share/icons/hicolor/256x256
usr/share/qelectrotech/lang
usr/share/applications/org.qelectrotech.qelectrotech.desktop
usr/share/mime/packages/qelectrotech.xml
usr/share/appdata
usr/share/doc/qelectrotech/LICENSE
usr/share/doc/qelectrotech/ELEMENTS.LICENSE
usr/share/doc/qelectrotech/CREDIT
usr/share/doc/qelectrotech/ChangeLog
debian/icons/*  usr/share/pixmaps/
man/files/nl_BE/man1/* usr/share/man/nl_BE/man1/
man/files/cs/man1/* usr/share/man/cs/man1/
man/files/ca/man1/* usr/share/man/ca/man1/
man/files/el/man1/* usr/share/man/el/man1/
man/files/es/man1/* usr/share/man/es/man1/
man/files/fr/man1/* usr/share/man/fr/man1/
man/files/it/man1/* usr/share/man/it/man1/
man/files/nl/man1/* usr/share/man/nl/man1/
man/files/pl/man1/* usr/share/man/pl/man1/
man/files/pt/man1/* usr/share/man/pt/man1/
man/files/ru/man1/* usr/share/man/ru/man1/
man/files/fr.UTF-8/man1/* usr/share/man/fr.UTF-8/man1/
man/files/fr.ISO8859-1/man1/* usr/share/man/fr.ISO8859-1/man1/
man/files/man1/* usr/share/man/man1/
```

scorpio810:
extract debian data.tar.xz

```
06:41:07] laurent@debian:~/data/qelectrotech_0.200.1.r9347-1_amd64/usr$ tree
.
├── bin
│   └── qelectrotech
└── share
    ├── appdata
    │   └── qelectrotech.appdata.xml
    ├── applications
    │   └── org.qelectrotech.qelectrotech.desktop
    ├── doc
    │   └── qelectrotech
    │       ├── changelog.Debian.gz
    │       ├── changelog.gz
    │       ├── copyright
    │       ├── CREDIT.gz
    │       ├── ELEMENTS.LICENSE.gz
    │       ├── LICENSE.gz
    │       └── README
    ├── icons
    │   └── hicolor
    │       ├── 128x128
    │       │   ├── apps
    │       │   │   └── qelectrotech.png
    │       │   └── mimetypes
    │       │       ├── application-x-qet-element.png
    │       │       ├── application-x-qet-project.png
    │       │       └── application-x-qet-titleblock.png
    │       ├── 16x16
    │       │   ├── apps
    │       │   │   └── qelectrotech.png
    │       │   └── mimetypes
    │       │       ├── application-x-qet-element.png
    │       │       ├── application-x-qet-project.png
    │       │       └── application-x-qet-titleblock.png
    │       ├── 22x22
    │       │   ├── apps
    │       │   │   └── qelectrotech.png
    │       │   └── mimetypes
    │       │       ├── application-x-qet-element.png
    │       │       ├── application-x-qet-project.png
    │       │       └── application-x-qet-titleblock.png
    │       ├── 256x256
    │       │   ├── apps
    │       │   │   └── qelectrotech.png
    │       │   └── mimetypes
    │       │       ├── application-x-qet-element.png
    │       │       ├── application-x-qet-project.png
    │       │       └── application-x-qet-titleblock.png
    │       ├── 32x32
    │       │   ├── apps
    │       │   │   └── qelectrotech.png
    │       │   └── mimetypes
    │       │       ├── application-x-qet-element.png
    │       │       ├── application-x-qet-project.png
    │       │       └── application-x-qet-titleblock.png
    │       ├── 48x48
    │       │   ├── apps
    │       │   │   └── qelectrotech.png
    │       │   └── mimetypes
    │       │       ├── application-x-qet-element.png
    │       │       ├── application-x-qet-project.png
    │       │       └── application-x-qet-titleblock.png
    │       └── 64x64
    │           ├── apps
    │           │   └── qelectrotech.png
    │           └── mimetypes
    │               ├── application-x-qet-element.png
    │               ├── application-x-qet-project.png
    │               └── application-x-qet-titleblock.png
    ├── lintian
    │   └── overrides
    │       └── qelectrotech
    ├── man
    │   ├── ca
    │   │   └── man1
    │   │       └── qelectrotech.1.gz
    │   ├── cs
    │   │   └── man1
    │   │       └── qelectrotech.1.gz
    │   ├── el
    │   │   └── man1
    │   │       └── qelectrotech.1.gz
    │   ├── es
    │   │   └── man1
    │   │       └── qelectrotech.1.gz
    │   ├── fr
    │   │   └── man1
    │   │       └── qelectrotech.1.gz
    │   ├── fr.ISO8859-1
    │   │   └── man1
    │   │       └── qelectrotech.1.gz
    │   ├── fr.UTF-8
    │   │   └── man1
    │   │       └── qelectrotech.1.gz
    │   ├── it
    │   │   └── man1
    │   │       └── qelectrotech.1.gz
    │   ├── man1
    │   │   └── qelectrotech.1.gz
    │   ├── nl
    │   │   └── man1
    │   │       └── qelectrotech.1.gz
    │   ├── nl_BE
    │   │   └── man1
    │   │       └── qelectrotech.1.gz
    │   ├── pl
    │   │   └── man1
    │   │       └── qelectrotech.1.gz
    │   ├── pt
    │   │   └── man1
    │   │       └── qelectrotech.1.gz
    │   └── ru
    │       └── man1
    │           └── qelectrotech.1.gz
    ├── mime
    │   └── packages
    │       └── qelectrotech.xml
    ├── pixmaps
    │   ├── qelectrotech-16.xpm
    │   └── qelectrotech.xpm
    └── qelectrotech
        └── lang
            ├── qet_ar.qm
            ├── qet_ca.qm
            ├── qet_cs.qm
            ├── qet_da.qm
            ├── qet_de.qm
            ├── qet_el.qm
            ├── qet_en.qm
            ├── qet_es.qm
            ├── qet_fi.qm
            ├── qet_fr.qm
            ├── qet_hr.qm
            ├── qet_hu.qm
            ├── qet_it.qm
            ├── qet_ja.qm
            ├── qet_mn.qm
            ├── qet_nb.qm
            ├── qet_nl_BE.qm
            ├── qet_nl.qm
            ├── qet_no.qm
            ├── qet_pl.qm
            ├── qet_pt_BR.qm
            ├── qet_pt.qm
            ├── qet_ro.qm
            ├── qet_rs.qm
            ├── qet_ru.qm
            ├── qet_sk.qm
            ├── qet_sl.qm
            ├── qet_sr.qm
            ├── qet_sv.qm
            ├── qet_tr.qm
            ├── qet_uk.qm
            └── qet_zh.qm

65 directories, 88 files

```

arummler:
With my fix it should be fine for DMG. You could check your binary for the embedded strings, e.g. :
strings qelectrotech | grep -i "resources/elements\|resources/lang\|usr/local"

I am getting:
../Resources/elements/
../Resources/lang/

scorpio810:
https://qelectrotech.org/forum/viewtopic.php?pid=23199#p23199

arummler:
I am adapting the workflow to upload all currently implemented artefacts to "nightly".
image
The versioning could still profit from sime simplification.

scorpio810:
https://qelectrotech.org/forum/viewtopic.php?pid=23203#p23203

Kellermorph wrote:
I've already noticed 5 errors.

- Full slave elements are no longer filtered out of the selection properties.
- Connection transfer no longer works between PLC master and PLC slave.
- When I copy and paste a terminal, the existing wires are connected to the new terminal.
- The text editor no longer works correctly for free text.
- The configured layout of the windows for numbering, component collection, etc., is not saved and is missing the next time the program is opened.

scorpio810:
I havent not time yet for your CI/CD , maybe see this week end.

arummler:
> Kellermorph wrote: I've already noticed 5 errors.
> 
> * Full slave elements are no longer filtered out of the selection properties.
> * Connection transfer no longer works between PLC master and PLC slave.
> * When I copy and paste a terminal, the existing wires are connected to the new terminal.
> * The text editor no longer works correctly for free text.
> * The configured layout of the windows for numbering, component collection, etc., is not saved and is missing the next time the program is opened.

At least for the first four points I do not see any way how it could be related to packaging. There were many functional changes merged meanwhile so I would guess stuff broke there. A proper fully automatic release pipeline will at least allow that people can test faster even if not able to compile themselves while reducing your workload.

arummler:
> I havent not time yet for your CI/CD , maybe see this week end.

No problem, this here is also not yet finished. The only functional changes in this branch touch the search path as discussed above. I would just avoid to diverge too much in the packaging parts. Up to now I tried to keep it on top of master by merging or rebasing regularly.

arummler:
@Kellermorph @scorpio810 And the CI finds its first issue: QPdfDocument::pagePointSize was only introduced with Qt6.4. The PDF/Image import from PR #772 needs there a guard.

scorpio810:
Hi @arummler,

Since you mentioned separating build/package in the existing Windows workflow is next on your list — could you fold in dropping the Qt5 track while you're in there? Qt5 is end-of-life for QET now (announced on the forum), so `windows-build.yml` / `windows-msi.yml` shouldn't need to keep building it anymore. Job 1 (`build-windows`, the Qt5 track) in `windows-build.yml` can go entirely; the Qt6 job becomes the only Windows build.

A few things worth knowing before you touch it, since they're not fully settled on my end yet either:

1. **Version numbering** — the Qt6 job currently hardcodes the package version as `0.200.1` (deliberately not parsed from `qetversion.cpp`, to keep it visually distinct from the real Qt5 version and because Qt-version self-detection inside the binary proved unreliable — see commit e1aa65f reverting 608ca984). Once Qt6 is the only track, we still need to decide: keep `0.200.1` fixed, or go back to parsing the real version from `qetversion.cpp` (safe now since `QetVersion::currentVersion()` no longer branches on Qt version). Not decided yet — happy to discuss.

2. **MSI signing** — `windows-msi.yml` currently only signs the `qt5` matrix flavor via SignPath (guarded by your fork-safety check, `fix/msi-signing-fork-guard`, d3f60c88 — thanks for that). If Qt5 goes away, that condition needs to flip so the remaining (Qt6-based) MSI is the one that gets signed. Otherwise we'd end up shipping an unsigned MSI.

3. **"Qt6 experimental" labeling** — once it's the only track, the "EXPERIMENTAL" labeling should go too: the `label_suffix: "-qt6-EXPERIMENTAL"` in the MSI matrix, the warning block in `generate-page.py` (Qt6 section on the nightly download page), and the "Qt6 builds are experimental" paragraph in the nightly release body in `windows-build.yml`.

Let me know if you'd rather I take a pass at `windows-build.yml`/`windows-msi.yml` myself and you focus on the rest of the overhaul (Linux/macOS/FreeBSD/AppImage/Snap) — whatever avoids us duplicating work. Happy to sync on the version/signing decisions above whenever suits you.

Thanks for all the work on the CI/CD overhaul, it's moving fast!

scorpio810:
Hi @arummler,

On the CI/CD triggering strategy: I'd like to move to a model where the packaging/release workflows are triggered manually by an admin, run sequentially, and stop the chain on first failure rather than all firing independently.

Concretely: `windows-build.yml` runs first (admin-triggered). If it passes, `windows-msi.yml` fires — same as today, via `workflow_run`, no change there. Then the other packaging workflows (macOS DMG, AppImage, Snap, Flatpak, etc.) run one after another. If any one fails, the remaining ones in the chain don't run.

Reasoning: with everything you're adding (Linux, macOS, FreeBSD, Windows-MSVC, AppImage, Snap, Flatpak...), running all of it on every push/PR risks hitting the Free plan's 20-concurrent-job cap once a few PRs are active at once, and burns CI time on packaging steps that don't need to run per-commit anyway.

One nuance I'd suggest, though: let's keep the lightweight "on-pull" compile checks (Linux especially — fast, cheap) running automatically on every push, so contributors still get quick feedback that their PR compiles. Reserve the manual sequential chain for the full packaging/release pipeline (Windows build→MSI, macOS DMG, AppImage, Snap, Flatpak) — that one doesn't need to run more often than before a nightly or a release. That's basically how `windows-build.yml` already works today (monthly cron + manual dispatch, no push trigger), so it'd extend the same logic to the new packaging workflows rather than changing the on-pull compile checks you already built.

Let me know what you think — happy to hop on the PR thread to hash out the trigger config once you've had a look.

scorpio810:
Hi @arummler,

Looked at `build-aux/macos/README.md` — solid work overall, especially the dynamic min-macOS detection from Qt's load commands and the honesty about the signing path not being confirmed end-to-end yet.

On that signing path specifically: I compared it against the original macOS deploy script we've used for manual releases, `misc/MacQetDeploy_arm64.sh` (https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/misc/MacQetDeploy_arm64.sh), and it does two things your README doesn't mention — worth checking `package_dmg.sh` handles both, since they're exactly the kind of thing that passes locally but fails at real notarization:

1. **Explicit per-component signing, in order**, rather than relying on `macdeployqt` alone or a single `codesign --deep` on the bundle (Apple explicitly discourages `--deep` for this). The script signs, separately: all `.dylib` files in `Frameworks/`, then `.framework` bundles, then plugins in `PlugIns/`, then any dylibs in `MacOS/`, then the main executable, then the bundle itself last — each with `--timestamp --options=runtime`.

2. **Re-signing everything a second time after the DMG copy.** The script's comment explains why: `hdiutil` can invalidate Sealed Resources while copying into the writable (UDRW) DMG, so every component gets re-signed again once the DMG is mounted, before converting to the final compressed UDZO. Skipping this step means the signature can look fine on the built bundle but be invalid inside the DMG users actually download — something that likely wouldn't surface until notarization or a `spctl` check on a real machine.

The original script also notarizes and staples the `.app` and the final DMG as two separate steps, not just once at the end — might be worth confirming `package_dmg.sh` does the same.

Given the README already flags the signing path as unconfirmed with a real cert, this seemed worth raising now rather than after a first real signing attempt. Happy to hop on the PR thread if useful.

arummler:
> Hi @arummler,
> 
> Since you mentioned separating build/package in the existing Windows workflow is next on your list — could you fold in dropping the Qt5 track while you're in there? Qt5 is end-of-life for QET now (announced on the forum), so `windows-build.yml` / `windows-msi.yml` shouldn't need to keep building it anymore. Job 1 (`build-windows`, the Qt5 track) in `windows-build.yml` can go entirely; the Qt6 job becomes the only Windows build.

I already separated them in the sense that there is only one artefact upload job for all packaging jobs. And the exe and msi jobs are consecutive. I can explain more about the structure in the bottom. If the decision is to retire now the Qt5 build completely I am happy to comply - I had to jump through some loops to get it working. For me it makes sense to focus on the Qt6 build and fix anything still not working. 

> 
> A few things worth knowing before you touch it, since they're not fully settled on my end yet either:
> 
> 1. **Version numbering** — the Qt6 job currently hardcodes the package version as `0.200.1` (deliberately not parsed from `qetversion.cpp`, to keep it visually distinct from the real Qt5 version and because Qt-version self-detection inside the binary proved unreliable — see commit [e1aa65f](https://github.com/qelectrotech/qelectrotech-source-mirror/commit/e1aa65f1ee8772195676d3b439c9b7dd172a0001) reverting [608ca98](https://github.com/qelectrotech/qelectrotech-source-mirror/commit/608ca984a401b08feb3374ab77ce21914c040a69)). Once Qt6 is the only track, we still need to decide: keep `0.200.1` fixed, or go back to parsing the real version from `qetversion.cpp` (safe now since `QetVersion::currentVersion()` no longer branches on Qt version). Not decided yet — happy to discuss.

In my branch here I got already rid of the fixed version as it complicated things and did not really make sense looking at SemVer. We can discuss the tag/release workflow in the bottom but what I already did is to work towards having the version number only in one place. In the moment I introdiced two files: QET_VERSION and QET_RELEASE. These are fed through CMake mechanisms both into the cpp as well as the CMake project version. Advantages:
- only one place to change the version minimizes the risk of mismatch
- parsing becomes much easier and less error prone


> 2. **MSI signing** — `windows-msi.yml` currently only signs the `qt5` matrix flavor via SignPath (guarded by your fork-safety check, `fix/msi-signing-fork-guard`, [d3f60c8](https://github.com/qelectrotech/qelectrotech-source-mirror/commit/d3f60c881449914423298190a8cd138726b8cd75) — thanks for that). If Qt5 goes away, that condition needs to flip so the remaining (Qt6-based) MSI is the one that gets signed. Otherwise we'd end up shipping an unsigned MSI.

Hmm, I though I had already added the signing to the Qt6 branch. I will check again and make sure.

> 3. **"Qt6 experimental" labeling** — once it's the only track, the "EXPERIMENTAL" labeling should go too: the `label_suffix: "-qt6-EXPERIMENTAL"` in the MSI matrix, the warning block in `generate-page.py` (Qt6 section on the nightly download page), and the "Qt6 builds are experimental" paragraph in the nightly release body in `windows-build.yml`.

I had already removed the qt6 suffix and added rather a qt5 suffix. If - as you proposed above - it is time to retire Qt5, I will clean out all the remains

> 
> Let me know if you'd rather I take a pass at `windows-build.yml`/`windows-msi.yml` myself and you focus on the rest of the overhaul (Linux/macOS/FreeBSD/AppImage/Snap) — whatever avoids us duplicating work. Happy to sync on the version/signing decisions above whenever suits you.

I would propose that I will do the Qt5 removal of the Windows branch. And we should agree on the release workflow. Then I would say it is time to merge and you can test the signing (which was difficult for me to test; only trying unsigned releases) and re-work that path as needed. 

> 
> Thanks for all the work on the CI/CD overhaul, it's moving fast!

arummler:
Hi @scorpio810 ,

>  
> On the CI/CD triggering strategy: I'd like to move to a model where the packaging/release workflows are triggered manually by an admin, run sequentially, and stop the chain on first failure rather than all firing independently.
> 
> Concretely: `windows-build.yml` runs first (admin-triggered). If it passes, `windows-msi.yml` fires — same as today, via `workflow_run`, no change there. Then the other packaging workflows (macOS DMG, AppImage, Snap, Flatpak, etc.) run one after another. If any one fails, the remaining ones in the chain don't run.
> 
> Reasoning: with everything you're adding (Linux, macOS, FreeBSD, Windows-MSVC, AppImage, Snap, Flatpak...), running all of it on every push/PR risks hitting the Free plan's 20-concurrent-job cap once a few PRs are active at once, and burns CI time on packaging steps that don't need to run per-commit anyway.
> 
> One nuance I'd suggest, though: let's keep the lightweight "on-pull" compile checks (Linux especially — fast, cheap) running automatically on every push, so contributors still get quick feedback that their PR compiles. Reserve the manual sequential chain for the full packaging/release pipeline (Windows build→MSI, macOS DMG, AppImage, Snap, Flatpak) — that one doesn't need to run more often than before a nightly or a release. That's basically how `windows-build.yml` already works today (monthly cron + manual dispatch, no push trigger), so it'd extend the same logic to the new packaging workflows rather than changing the on-pull compile checks you already built.
> 
> Let me know what you think — happy to hop on the PR thread to hash out the trigger config once you've had a look.

that's roughly how also intended to do it. In the moment I still have a on-pull trigger on the build-package pipeline but that's primarily for quick testing of the pipeline. I agree that one should not run uselessly but in the moment there are quite few limitations. E.g. the 20 concurennen you mentioned refer afaik to the "free" tier of private repos. Public repos like ours are not really among any of the tiers mentioned and mostly limited on certain runners. But I agree that what is mentioned on https://docs.github.com/en/actions/reference/limits is formulated in a misleading way.

1) basic builds on-pull / merge
Those are basically finished after I have now removed the Qt5 branches:
- linux-build.yml
- macos-build.yml
- windows-build-mingw.yml
- windows-build-msvc.yml
- freebsd-build.yml
What can be done later is to add a testing step. And you could generate the usual "badges" and add them to the README.md to make the status visible. Triggers yet to be checked but they should run a) on a merge request and b) when the merge is done.

2) packaging (nightly/develop and release)
image

I have one workflow file package.yml whose job it is to: check whether it is a develop/nightly or a release, call the package files linux-package-appimage.yml  linux-package-flatpack.yml  linux-package-snap.yml  macos-package-dmg.yml  windows-package-exe.yml  windows-package-msi.yml, upload their artefacts either  to a) the nightly/development release/tag replacing the old one and generating and uploading a new overview page for all of them (not only Windows) OR b) upload to  the release and add the Changelog.

For the versioning the major change is that I would like to have it in one place. In the moment I have a QET_VERSION and QET_RELEASE file. One possible workflow I had first in mind is:
- commit QET_VERSION, QET_RELEASE and Changelog (whichever of the three files is the correct one)
- create a tag/release on the github GUI
- the workflow runs through, checks that the version and release makes sense, edits the release text and attaches the binaries, changes the QET_RELEASE back to "development" in form of an automatic commit. If necessary you can still edit the release text manually afterwards.

But this whole thing seems to be too much back and forth. Some alternatives are:
- QET_VERSION, QET_RELEASE and ChangeLog stay but one triggers by creating a tag only. Difference is that Release is fully automatically created.
- The same but instead the workflow also creates the tag and release if you trigger it manually either from the GUI or from a small script using the GITHUB API.
- Or a different approach: you create the tag/release using the GUI, that one is the authorative information, the workflow commits automatically the tag, release (and potentially even the Changelog) to the repo, then runs, attaches the binaries to the existing release.

Which one does make most sense for you?

Regarding development/nightly: I have called it in the moment development and run it immediately after a merge. But I guess a nightly would be sufficient as well.

Future possible extensions: 
- automatic snap upload (or do you use PPA to build it)
- automatic upload / trigger of external build (where applicable) of any other distribution channel: AppImage, FlatPack, Chocolately, Debian/Ubuntu Deb, etc. From what I saw some externally managed OS repos are lagging behind or lost their maintainer? gentoo?
- upload to the qelectrotech.org  web page (but that one probably needs to be re-worked first completly, might be best/easiest just to point to the releases on github)

arummler:
Now there are two main workflows. The packaging is triggered on merge (could be still changed to nightly if too heavy, e.g. if there are multiple merges consecutively) and on tag. I am now calling the doxygen generation in order to have only one main trigger point. That helps with the issue that piblishing new gh-pages overwrites them. That requires that both the doxygen doc as well as the "development/nightly" page are uploaded at the same time. In the moment there is no caching of any dependencies as the workflows are quite quick. That can be still fine tuned. 

What needs action (setting up variables) and testing is the signing in the different paths. And naturally if there are further extensions desirable. Please let me know how you would like to proceed.

scorpio810:
## macOS: KF6 dependency notes for arummler (Homebrew build)

I spent some time this week digging into the KF6-on-Homebrew situation for the macOS build and wanted to share what I found, so this isn't blocking or urgent — just a pointer for when you get to it.

### Starting point

I based my testing on Gilles Caulier's (digiKam maintainer) Qt6/KF6 compilation setup for Homebrew, rather than the MacPorts framework — trying to see if it could be a faster/simpler path than MacPorts for getting KF6 on macOS.

### What I ran into

- Some KF6 modules in Gilles' Homebrew tap/build failed to compile. After checking, none of the failing modules are actually needed by QET, so they can just be skipped/excluded rather than fixed.
- I also revisited the old QET wiki instructions for a KF5 Homebrew tap (`kde-mac/kde`, https://invent.kde.org/packaging/homebrew-kde.git). That same tap is still alive today and has been extended to include KF6 formulas alongside the old KF5 ones (mirror: https://github.com/KDE/homebrew-kde).
- That tap is currently **in the process of migrating most frameworks into `homebrew/core`**, which could explain some of the build conflicts — formulas in the tap may clash with versions already provided by `homebrew/core`.
- The old wiki tap command uses `--force-auto-update`, which **no longer exists** — it was removed from Homebrew in March 2024 (4.2.13). The correct command today is just:
  ```
  brew tap kde-mac/kde https://invent.kde.org/packaging/homebrew-kde.git
  ```
- The wiki also references a `do-caveats.sh` helper script (`"$(brew --repo kde-mac/kde)/tools/do-caveats.sh"`), which didn't work for me either — worth checking whether it still exists at that path once the tap is tapped correctly (without the obsolete flag).

### Relevant note

Historically (KF5 era), QET only ever needed two modules from that tap: `kf5-kcoreaddons` and `kf5-kwidgetsaddons`. The KF6 equivalents are presumably `kf6-kcoreaddons` and `kf6-kwidgetsaddons` — worth targeting only those instead of trying to build the whole tap.

Happy to share more details if useful, but wanted to get this pointer written down while it's fresh.

scorpio810:
## macOS: KF6 dependency notes for arummler (Homebrew build, based on Gilles Caulier's digiKam scripts)

I spent some time this week testing the KF6-on-Homebrew route for the macOS build and wanted to share what I found — not blocking or urgent, just a pointer for when you get to it.

### Starting point

I based my testing on Gilles Caulier's (digiKam maintainer) Qt6/KF6 Homebrew bundle scripts (`01-build-homebrew.sh` + `02-build-extralibs.sh`), as an alternative to the MacPorts framework — trying to see if it's a faster/simpler path for getting KF6 on macOS.

### `01-build-homebrew.sh` — takes hours

This script installs Qt6 and all of digiKam's other Homebrew dependencies. On my Mac Mini M2, it takes **hours** to complete — much heavier than `02-build-extralibs.sh` below. Worth keeping in mind for CI: even with caching, a from-scratch run of this step alone is a significant time cost.

### `02-build-extralibs.sh` — builds the KF6 components themselves

This one builds the actual KF6 modules on top of the Homebrew Qt6 from step 1. Out of the box, several modules failed to compile for me. None of them are needed by QET, so I commented them out:

- `karchive`, `ki18n`, `kconfigwidgets`, `kiconthemes`, `kservice`, `kxmlgui`, `kbookmarks`, `kimageformats`, `sonnet`, `ktextwidgets`, `libksane`, `kcalendarcore`, `breeze`, `opencv`

With those excluded, the remaining set built cleanly, no errors, in about 5 minutes on my Mac Mini M2:

- `extra-cmake-modules`, `kconfig`, `breeze-icons`, `kcoreaddons`, `kwindowsystem`, `solid`, `threadweaver`, `kdbusaddons`, `kcrash`, `kcodecs`, `kauth`, `kguiaddons`, `kwidgetsaddons`, `kitemviews`, `kcompletion`

That build time is a lot better than MacPorts (which takes ~24h), so this looks like a promising direction for CI — worth checking whether QET actually needs all of the modules above, or an even smaller subset.

### Note on the old KF5 Homebrew tap

For the record: the old QET wiki also references a KF5 Homebrew tap (`kde-mac/kde`, https://invent.kde.org/packaging/homebrew-kde.git). I checked it — it's still alive but **only has KF5 formulas** (`kf5-*`), nothing KF6. Not a usable path for a KF6 build; Gilles' scripts above remain the only concrete KF6-on-Homebrew route I found.

Happy to share more details if useful.

cgilles:
Hi. The Homebrew script as a port of official Macports ones, that i wrote. The Homebrew script have been ported by Michael Miller. I tried to use these on my macbook pro M1, without success. The target Homebrew hierarchies is completely different than macports and to bundles all files it's a big puzzle. I still use Macports which run fine since more than 9 months. Never try to compile whole install from scratch to quickly after a main release of macOS and Xcode, developers needs time to adjust the ports.

cgilles:
The failed compilation of KF6 need to backport my patches used with Macports. Again, I maintain Macports, not Homebrew. Considering to use Macports is more safe...

cgilles:
>As for me, the QEMU/KVM 20.04 AppImage VM was successfully upgraded to 22.04, but the >compilation of Digikam’s Qt6 frameworks failed...

The KF6/Qt6 must be compiled under XUbuntu 22.04, not 20.04 (Qt6 do not support it, only Qt5). 

Tip: When i rebuild a main version of the AppImage, i reinstall all from scratch (system + recompilation o all one by one).

scorpio810:
> As for me, the QEMU/KVM 20.04 AppImage VM was successfully upgraded to 22.04, but the >compilation of Digikam’s Qt6 frameworks failed...

>The KF6/Qt6 must be compiled under XUbuntu 22.04, not 20.04 (Qt6 do not support it, only Qt5).

>Tip: When i rebuild a main version of the AppImage, i reinstall all from scratch (system + recompilation o all one by one).


I wanted to avoid having to start all over again from scratch and get the VM up and running again, set up SSH, remote, etc, and so on.

arummler:
> > As for me, the QEMU/KVM 20.04 AppImage VM was successfully upgraded to 22.04, but the >compilation of Digikam’s Qt6 frameworks failed...
> 
> > The KF6/Qt6 must be compiled under XUbuntu 22.04, not 20.04 (Qt6 do not support it, only Qt5).
> 
> > Tip: When i rebuild a main version of the AppImage, i reinstall all from scratch (system + recompilation o all one by one).
> 
> I wanted to avoid having to start all over again from scratch and get the VM up and running again, set up SSH, remote, etc, and so on.

Sorry for the likely naive question but what is the reason that you compile your own frameworks - both Qt and KDE - for the AppImage builds? To be in precise control of the versions? Usually I compiled for other projects Qt only if I either needed a specific version which was not distributed or attempted a static build. And why each time from scratch?

As far as I understood from documentation there is not really a mandatory or even recommended base system. It just mentions something along the line "the oldest possible LTS" in order to achieve maximum backward compatibility with host systems. So I went the easy way, used Ubuntu (22.04), installed the packages from the standard repo and deployed them. It is possible to build your own versions in the CI/CD, using its own workflow, caching or using docker, package registry, etc. if really needed. You stressed that you were using XUbuntu: to stay rather on X11 than Wayland? How do you verify your AppImage compatibility? Is there even such a thing?

Correction: similar to the Mac build I did not use the KDE libs but the in-built replacement. Afaik it should be 1:1 so not sure it is necessary to go to the trouble to bundle the KDE libs just for 2-3 widgets?

scorpio810:
>Correction: similar to the Mac build I did not use the KDE libs but the in-built replacement. Afaik it should be 1:1 so not sure it is necessary to go to the trouble to bundle the KDE libs just for 2-3 widgets?


https://github.com/search?q=repo%3Aqelectrotech%2Fqelectrotech-source-mirror+KAutoSaveFile&type=pullrequests

https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/598

>Proposed scope
Add rotation to the crash-recovery path specifically (not autosave, which intentionally writes the real file on an opt-in schedule):

>Keep a small, fixed number of rotating snapshots (e.g. 3) instead of one KAutoSaveFile.
On each writeBackup() tick, rotate existing generations before writing the new one.
Extend checkBackupFiles()'s stale-file scan/prompt to list all generations for a project, letting the user pick which to restore (most recent by default).
No new settings UI initially — a fixed small default covers the reported case.
Why this shape
Autosave intentionally commits to the real file on a user-chosen schedule; adding generations there is a different feature (versioned saves) with different UX questions. The crash-recovery file is already synthetic and out of the user's sight, so extending it from 1 to 3 generations is a contained change to writeBackup()/checkBackupFiles() without touching autosave semantics.

cgilles:
Qt and KDE need tune to be optimized. KDE especially do not compile everywhere fine (depending of the module pulled on the workflow. MacOS for ex is a good example of the poor support platform from KDE.

You need to take a care also to the binary compatibility. I seen for ex the KF6.29 switch to more recent C++ standard which break the compatibility with older macOS | XCode version. This is not fine for end user to always switch to recent macOS.

Another point about the KDE framework is the way to add everywhere a dependency to support the Kirigami | QtQuick | Qml. We don't care about that in digiKam as we don't use and never use it at all. Qt/C++ do the stuff perfectly and it's the robust foundation. Also the support for the phone/tablets is not planned in digiKam.

We still maintain the Qt5 version of the build system for one only reason : the macOS Intel, for older macOS release. But the day are number, probably next year we will drop the Qt5 support definitively in digiKam.

About your static libs question: the response is no. Qt/KDE still compiled as shared libs. 

AppImage compatibility : handmaid + users feedback. Using an older XUbuntu allows to run Appimage almost everywhere without problem. It's a gain. The only point that remain in time is the openssh compatibility. For this point Qt is compiled with a static openssh as well.

About X11 | Wayland debate : Using X11 allows to work everywhere, as Wayland X11 layer is great. No need a pure Wayland build for the moment, but probably later in few years when all users system will be switched to Wayland.

scorpio810:
See: https://github.com/qelectrotech/qelectrotech-source-mirror/pull/799

arummler:
> See: #799

Sorry, but did you mean by chance another PR/issue or is the "scroll bug" analyzed and rectified in that PR somehow related to the packaging/QT6 version/KF-necessity discussion? I do not see the connection.

scorpio810:
> > See: #799
> 
> Sorry, but did you mean by chance another PR/issue or is the "scroll bug" analyzed and rectified in that PR somehow related to the packaging/QT6 version/KF-necessity discussion? I do not see the connection.

Sorry @arummler  It 's >:  https://github.com/qelectrotech/qelectrotech-source-mirror/issues/798

arummler:
> > Correction: similar to the Mac build I did not use the KDE libs but the in-built replacement. Afaik it should be 1:1 so not sure it is necessary to go to the trouble to bundle the KDE libs just for 2-3 widgets?
> 
> https://github.com/search?q=repo%3Aqelectrotech%2Fqelectrotech-source-mirror+KAutoSaveFile&type=pullrequests
> 
> #598
> 
> > Proposed scope
> > Add rotation to the crash-recovery path specifically (not autosave, which intentionally writes the real file on an opt-in schedule):
> 
> > Keep a small, fixed number of rotating snapshots (e.g. 3) instead of one KAutoSaveFile.
> > On each writeBackup() tick, rotate existing generations before writing the new one.
> > Extend checkBackupFiles()'s stale-file scan/prompt to list all generations for a project, letting the user pick which to restore (most recent by default).
> > No new settings UI initially — a fixed small default covers the reported case.
> > Why this shape
> > Autosave intentionally commits to the real file on a user-chosen schedule; adding generations there is a different feature (versioned saves) with different UX questions. The crash-recovery file is already synthetic and out of the user's sight, so extending it from 1 to 3 generations is a contained change to writeBackup()/checkBackupFiles() without touching autosave semantics.

I have to admit that I did not test myself the behaviour WITH_KF/WITHOUT_KF- Looking through #533 I assumed that the files in sources/ui/nokde are a 1:1 drop-in replacement for KColorButton, KColorCombo and KAutoSave.  If that's not the case (ie without KF there is a functionality loss) or of you strongly prefer to package with the KF libraries if feasible at all, the workflows will need to be re-worked a bit.

scorpio810:
Good question, and thanks for digging into #533/#598.

On the KColorButton/KColorCombo/KAutoSaveFile drop-in question: KColorButton and KColorCombo are genuinely 1:1, plain Qt wrappers, no concerns there.

KAutoSaveFile is close but not fully 1:1. The API surface matches and the core mechanism is sound (QLockFile-based crash detection, sidecar metadata for recoverability), but I found two gaps compared to the mature KDE implementation:

1. A TOCTOU race between `staleFiles()` (which tries and immediately releases the lock to test recoverability) and the actual `open()` call that re-locks for real recovery. A double-launched instance could theoretically "win" the same recovery twice.
2. The sidecar `.path` metadata write isn't atomic (no write-temp-then-rename). A crash during that specific write would silently lose that one recovery entry.

Neither is a functional loss a user would normally notice, and I wouldn't block the CI/CD work on this — but it's not rigorously 1:1 either, more a younger implementation with less mileage on edge cases than years of KDE production use. So no need to go through the trouble of bundling the KDE libs just for this — the built-in replacement is solid enough for our use case.

On #598: the proposed split (autosave = user-scheduled real file vs. crash-recovery = synthetic, rotated) makes sense to me, and 3 fixed generations with no new settings UI is a reasonable first scope. One thing to verify before merging the implementation: with the current nokde fallback, 3 generations will multiply the number of `.path` sidecars and lock files per project. The hash+random-suffix naming should handle that without collisions, but I'd want a test pass with 3 generations plus a recovery-after-crash scenario before calling it done.

scorpio810:
On scope: I'd like to merge this work, but I want to keep windows-build.yml and windows-msi.yml untouched, and I'm not looking to replace the macOS arm64 CI/CD I already have running.

To be clear about what's already covered on my side: Debian Qt6+KF6 packaging (stable, unstable, and PPA) is built and signed by me, the Flatpak repo (Qt6+KF6) is also already working and signed, macOS (Qt6+KF6, signed/notarized) is already working on my end too, and there's also a Snap CI/CD already in place.

That said, I'm fine with your CI/CD covering macOS Intel x86_64 Qt6, since I'm not planning to rebuild that one for Qt6 myself and may arm64 bundle not signed for users who can’t wait to try it out.

One thing though: I'd rather not have code signing/notarization run inside the CI/CD pipeline itself. Storing the Developer ID certificate and notarization credentials as CI secrets adds real risk, especially with pull_request-triggered workflows — a fork PR (malicious or just a misconfigured workflow) could end up with access to those secrets, and if the certificate ever leaked we'd have to revoke it, which invalidates every build signed with it, including ones already distributed. That's the same reasoning behind not sending the Windows Qt6 MSI yet to signing via CI, and why Dieter added the fork guard there. I'd prefer to keep signing/notarization for the DMGs (arm64 ) out of your build workflow and handle that step on my side, the same way we do for Windows, Debian, and Flatpak.

So overall I'd see this PR's scope as complementary to what I already have running (Linux/FreeBSD on-pull checks, AppImage, macOS x86_64 arm64 unsigned build, etc.) rather than overlapping it.

scorpio810:
Thinking about this more: I don't think it's useless to compile on every PR, but I do think it's a problem if that's the *only* trigger. On-pull compiles are a quality gate (catch a broken PR before merge) — that's valuable on its own, even if nothing gets published from it. But if macOS/Linux builds only ever run on PR and never get published as an actual nightly, then master can move forward for weeks without ever producing something users can download — all that CI time only serves validation, never users.

So I'd suggest splitting it into two separate triggers rather than one:
- on-pull: compile-only, no publish, just a gate before merge (what you have now)
- chained the same way windows-msi.yml is chained off "Windows Build" today: workflow_run once Windows Build completes successfully, plus workflow_dispatch as a manual fallback. Windows Build itself only runs on a monthly cron (1st of the month) or manual dispatch, not on every merge — so this keeps the same cadence: macOS/Linux nightly publishing happens on that same schedule/manual trigger, tied to the same validated commit as Windows, instead of running independently on its own timing.

That keeps everything on the download page in sync (one validated commit, all platforms), avoids unnecessary workflow runs with secret access outside that cadence, and doesn't require a new schedule of its own. I'd keep this nightly publish unsigned though, same reasoning as before — signing/notarization stays on my side, separate from this pipeline.

arummler:
> Good question, and thanks for digging into #533/#598.
> 
> On the KColorButton/KColorCombo/KAutoSaveFile drop-in question: KColorButton and KColorCombo are genuinely 1:1, plain Qt wrappers, no concerns there.
> 
> KAutoSaveFile is close but not fully 1:1. The API surface matches and the core mechanism is sound (QLockFile-based crash detection, sidecar metadata for recoverability), but I found two gaps compared to the mature KDE implementation:
> 
> 1. A TOCTOU race between `staleFiles()` (which tries and immediately releases the lock to test recoverability) and the actual `open()` call that re-locks for real recovery. A double-launched instance could theoretically "win" the same recovery twice.
> 2. The sidecar `.path` metadata write isn't atomic (no write-temp-then-rename). A crash during that specific write would silently lose that one recovery entry.
> 
> Neither is a functional loss a user would normally notice, and I wouldn't block the CI/CD work on this — but it's not rigorously 1:1 either, more a younger implementation with less mileage on edge cases than years of KDE production use. So no need to go through the trouble of bundling the KDE libs just for this — the built-in replacement is solid enough for our use case.
> 
> On #598: the proposed split (autosave = user-scheduled real file vs. crash-recovery = synthetic, rotated) makes sense to me, and 3 fixed generations with no new settings UI is a reasonable first scope. One thing to verify before merging the implementation: with the current nokde fallback, 3 generations will multiply the number of `.path` sidecars and lock files per project. The hash+random-suffix naming should handle that without collisions, but I'd want a test pass with 3 generations plus a recovery-after-crash scenario before calling it done.

To summarize:  the vendored in KAutoSaveFile has limitations compared to the one the KF lib provides. But it is good enough to use for packaging? 

For me the logical follow-up would be 
a) to document that limitation (outside of MR) in the README
b) to improve the vendored-in KAutoSave (also outside this MR)

arummler:
> On scope: I'd like to merge this work, but I want to keep windows-build.yml and windows-msi.yml untouched, and I'm not looking to replace the macOS arm64 CI/CD I already have running.

You have already a macOS CI/CD? From what I see in the repo you have in the moment only a macOS packaging script which is run manually on your local computer. It is located in misc and not in build-aux or the root like the others. That is also something that I re-ordered in this MR to clean up the structure. At least for me it was kind of all over the place and did not really make sense.

Regarding the windows-build my major changes are:
- trigger / nightly deployment as it maskes more sense to have it as part of common workflow. The main issue is the gh-page deployment that has to happen at the same time to avoid overwrite.
- renaming of the workflow to structure accordingly
- version number: as you might have seen I tried to reduce drastcally the number of locations where you need to edit the version number for a new deployment. Ideally that should be only one location.

> 
> To be clear about what's already covered on my side: Debian Qt6+KF6 packaging (stable, unstable, and PPA) is built and signed by me, the Flatpak repo (Qt6+KF6) is also already working and signed, macOS (Qt6+KF6, signed/notarized) is already working on my end too, and there's also a Snap CI/CD already in place.

Just to make sure that the terminology matches: I do not see any workflows for those. You are referrring to the packaging scripts in the root of the repo which you run manually and locally for Debian, Flatpack and Snap. The same - manual local action - for Mac but the script is under /misc.

That's not really CI/CD as that implies automatization.

> 
> That said, I'm fine with your CI/CD covering macOS Intel x86_64 Qt6, since I'm not planning to rebuild that one for Qt6 myself and may arm64 bundle not signed for users who can’t wait to try it out.

> 
> One thing though: I'd rather not have code signing/notarization run inside the CI/CD pipeline itself. Storing the Developer ID certificate and notarization credentials as CI secrets adds real risk, especially with pull_request-triggered workflows — a fork PR (malicious or just a misconfigured workflow) could end up with access to those secrets, and if the certificate ever leaked we'd have to revoke it, which invalidates every build signed with it, including ones already distributed. That's the same reasoning behind not sending the Windows Qt6 MSI yet to signing via CI, and why Dieter added the fork guard there. I'd prefer to keep signing/notarization for the DMGs (arm64 ) out of your build workflow and handle that step on my side, the same way we do for Windows, Debian, and Flatpak.

Hmm, so basically tagged deployment stays fully manual and local (on your computers); nightlies only for Windows and Mac and those unsigned?  Supposedly one can harden Github secrets by using the project settings as all the DevOps driven projects must tend to do, but okay I can understand that you are nervous about yopur personal credentials.

> 
> So overall I'd see this PR's scope as complementary to what I already have running (Linux/FreeBSD on-pull checks, AppImage, macOS x86_64 arm64 unsigned build, etc.) rather than overlapping it.

arummler:
> Thinking about this more: I don't think it's useless to compile on every PR, but I do think it's a problem if that's the _only_ trigger. On-pull compiles are a quality gate (catch a broken PR before merge) — that's valuable on its own, even if nothing gets published from it. But if macOS/Linux builds only ever run on PR and never get published as an actual nightly, then master can move forward for weeks without ever producing something users can download — all that CI time only serves validation, never users.
> 
> So I'd suggest splitting it into two separate triggers rather than one:
> 
> * on-pull: compile-only, no publish, just a gate before merge (what you have now)

This is ready to use in this MR with separate workflow to Windows MinGW, Windows MSVC, MacOS, Linux, BSD. 

> * chained the same way windows-msi.yml is chained off "Windows Build" today: workflow_run once Windows Build completes successfully, plus workflow_dispatch as a manual fallback. Windows Build itself only runs on a monthly cron (1st of the month) or manual dispatch, not on every merge — so this keeps the same cadence: macOS/Linux nightly publishing happens on that same schedule/manual trigger, tied to the same validated commit as Windows, instead of running independently on its own timing.
> 
> That keeps everything on the download page in sync (one validated commit, all platforms), avoids unnecessary workflow runs with secret access outside that cadence, and doesn't require a new schedule of its own. I'd keep this nightly publish unsigned though, same reasoning as before — signing/notarization stays on my side, separate from this pipeline.

A lot of things are possible designed and realized in different ways. Let's separate nightlies (or monthlies as you prefer) from tagging and define the requirements:
- No signage at all if applicable as you prefer not to have any keys anywhere online.
- Packages: Win EXE, Win MSI, Mac DMG uploaded to the nightlies page
- You mentioned Linux:
   - some options:
   - generating deb package PPA / snap / flatpack in the github action and uploading to nightly gh-page
   - triggering/uploading (depending on packaging format) an update to PPA / snapstore / flatpack on the "nightly/development" channel
   - In any case one has to modify the windows workflow trigger and upload as I did here in order to get everything on one page. Or one has to some tricks like collecting the already uploaded Windows artefacts and re-upload but imho it is more complciated as one has to avoid race conditions. The same applies to the doxygen which I integrated into the workflow as in the moment it is over-written in the current master, ie not acccesible.
   
For version tagging:
- as unsigned the final versions cannot be produced as part of a github hosted CI/CD chain. 
- You will keep compiling them fully locally on your computers from scratch and uploading? Or downloading github artefacts and signing them?
- what about the version numbers? any interest in the unification presented in this MR? 
- There are additional options like triggering from the workflow (still requires some credentials); openbuildservice, etc.

If it helps I can separate out the CI part which was quite uncontested. So that it can get already merged.

scorpio810:
https://qelectrotech.org/forum/viewtopic.php?pid=23280#p23280

scorpio810:
On the "not really CI/CD" point: fair on the terminology for the repo-hosting gap specifically (no automated publish of a Debian/Flatpak repo page yet — that's real and worth fixing). But it's not accurate as a description of the whole pipeline: PPA builds trigger automatically once I push the Debian source packages, and Snap runs on a cron plus builds itself on snapcraft.io's edge channel — I only do a manual promotion edge→stable, which is a deliberate quality gate, not a technical limitation. So "fully manual" undersells what's already automated; the actual gap is narrower than that.

scorpio810:
On the /misc layout and the general file structure: I get the frustration, and you're right that it needs proper reorganizing — it's grown organically over 17 years and it shows. I'm on board with cleaning it up. Just flagging that it's going to be real work on my end, since a bunch of my local scripts (packaging, signing, etc.) are hardcoded against the current paths. So I'll need to go through and update all of those references once we settle on the new structure — not a blocker, just don't expect it to happen in the same breath as the merge.

scorpio810:
https://github.com/qelectrotech/qelectrotech-source-mirror/pull/510#issuecomment-4744073034

arummler:
> [#510 (comment)](https://github.com/qelectrotech/qelectrotech-source-mirror/pull/510#issuecomment-4744073034)

I was aware of the PR and as I commented I had cherry-picked it in order to integrate it into this MR. If you are trying to hint at the GIT LFS limits I saw them as well (being basically the only limit imposed on public repos). Hence, I did not change the reduced cadence of git lfs uploads. But it is free to compile doxygen and upload as gh-pages...or do I miss something? These are different end points depending on the trigger condition.

arummler:
PS At that time I also tried to reduce the file size as >900MB is immense. Actually one can either/and change to svg (although not ideal for QTCreator) or reduce the depth represented in the graphs.

scorpio810:
@arummler maybe return to old Doxyfile?
https://github.com/qelectrotech/qelectrotech-source-mirror/commits/master/Doxyfile

scorpio810:
https://qelectrotech.org/forum/viewtopic.php?pid=23288#p23288

scorpio810:
@arummler please move to Qt6 only, thanks you.
Qt5 is EOL.

arummler:
> @arummler please move to Qt6 only, thanks you.
> Qt5 is EOL.

Do you still see Qt5 being used in the CI draft somewhere? I removed it about two weeks ago when you asked for it. The workflows are Qt6+cmake only.

What one could consider at this point (but not part of this MR) is to update the documentation and strip out the compatibility handling for Qt<6 from the code.

scorpio810:
> > @arummler please move to Qt6 only, thanks you.
> > Qt5 is EOL.
> 
> Do you still see Qt5 being used in the CI draft somewhere? I removed it about two weeks ago when you asked for it. The workflows are Qt6+cmake only.
> 
> What one could consider at this point (but not part of this MR) is to update the documentation and strip out the compatibility handling for Qt<6 from the code.

https://qelectrotech.org/forum/viewtopic.php?pid=23293#p23293

ispyisail:
@arummler — I read this properly, which I should have done before writing #832 and #833. You asked me to look at the proposed workflow before rewriting the action, and you were right; sorry for the duplicated effort.

## The design solves the problem I ran into

I checked rather than took it on trust: there is **exactly one `deploy-pages` caller in the entire branch**, in `package.yml`. `auto-doxygen.yml` no longer deploys at all — it emits a `doxygen-html` artifact that `package.yml` downloads and combines with everything else before a single deploy.

That is precisely the collision @scorpio810 raised on #832, and the structure is better than what I proposed there: every producer emits artifacts, exactly one consumer publishes. The `verify-version` gate followed by the fan-out to macOS / AppImage / Windows exe→msi / snap / flatpak / docs as reusable workflows reads cleanly too.

## The conflicts are an argument for landing it

It is currently `mergeable: false`, and only two files conflict — both the same thing:

| File | master | this PR |
|---|---|---|
| `CMakeLists.txt` | `VERSION 0.200.1` | `VERSION ${QET_VERSION_STRING}` |
| `Doxyfile` | `PROJECT_NUMBER = " 0.200.1-dev"` | placeholder, CI injects from `QET_VERSION` |

@scorpio810 had to hand-edit the version in two separate files this week to get to 0.200.1. So the conflict is caused by exactly the maintenance burden this PR removes — `QET_VERSION`/`QET_RELEASE` as the single source, with `qetversion.cpp` generated by `configure_file`.

The third hunk is master's new macOS bundle-identifier block sitting next to your restructured `add_executable`. All three look mechanically resolvable.

## A suggestion, offered as a question

87 commits across 35 files is a lot to review at once, it has been open a month, and @scorpio810 has said he is short of bandwidth. The conflicts will also keep recurring while it waits, since every version bump touches those same two files.

Would it be worth landing the **version single-source** part on its own first? `QET_VERSION`, `QET_RELEASE`, `qetversion.cpp.in` and the two `configure_file` hookups. It is small, it is the thing currently generating the conflicts, it makes the next version bump one file instead of three, and it is useful whether or not the rest of the CI rework lands in this shape.

I realise you already split out #808 and it did not land, so this may be a road you have been down. Just a thought about sequencing rather than content.

## One small thing that looks already settled

The last exchange has @scorpio810 asking to move to Qt6 only, and you replying that Qt5 was removed a fortnight earlier. For what it is worth I checked the branch: every build job pins `-DQT_VERSION_MAJOR=6`, and there is no Qt5 path left in the workflows. That request looks already satisfied.

Happy to test any slice of this the way I have been testing other PRs — clean merge, Qt6 build, corpus run — if that is useful to you.

arummler:
I still intend to get this in, just needs a bit of work. @scorpio810 You mentioned 2 commits ago the signing token for your existing windows workflow - so you are open to have some secrets living on github? Just to know what I should leave in here and what to remove / postpone for later.

scorpio810:
@arummler Good question. I'm fine with SignPath's token living in GitHub secrets — it's designed for that, the actual certificate stays on their infrastructure and the token just triggers a signing request, so the exposure if it ever leaked would be limited. That said, the SignPath Foundation certificate is still pending validation on their end, so that part isn't active yet anyway — I'll let you know once it's approved.

Apple Developer credentials are a different story though — a full certificate/API key gives much broader access (can sign anything under our identity), and with an open-source repo taking external PRs there's a real risk of a secret like that getting exfiltrated through a malicious workflow. I'd rather not put raw Apple credentials in GitHub secrets for that reason.

So: for now, hold off on wiring up SignPath secrets until the certificate is approved, and let's also hold off on Apple signing in CI — happy to look at safer alternatives (e.g. a self-hosted runner, or signing as a separate manual/gated step) if/when we get there.

scorpio810:
https://msicc.net/automating-apple-builds-a-practical-guide-to-github-secrets/
https://msicc.net/exporting-apple-distribution-certificates-for-cicd-the-right-way/
https://msicc.net/ci-ready-macos-signing-combining-certs-for-github-actions/

@jp2images what do think about?

> Apple Developer credentials are a different story though — a full certificate/API key gives much broader access (can sign anything under our identity), and with an open-source repo taking external PRs there's a real risk of a secret like that getting exfiltrated through a malicious workflow. I'd rather not put raw Apple credentials in GitHub secrets for that reason.

jp2images:
@scorpio810 Short answer: I wouldn't. The risk isn't quite where you've placed it though, and there's an option those articles don't cover.

**The articles don't address your concern.** The first is part one: it lists secret names and stops before the workflow, no security discussion. The third is a keychain-import workaround (merging two certs into one `.p12`). Both also cover App Store distribution — Apple Distribution + 3rd Party Mac Developer Installer. We ship a DMG outside the App Store, so we need Developer ID Application + notarization. Different certs, different flow.

**On fork PRs:** `pull_request` runs from a fork don't receive secrets — GitHub withholds them. The `github.repository ==` guard in `windows-msi.yml` is harmless, but it isn't what protects the token. The triggers that hand secrets to untrusted code are `pull_request_target`, `workflow_run` and `issue_comment`. `windows-msi.yml` is `workflow_run` and holds the SignPath token, so that one deserves a look on its own merits.

The argument I'd make instead: anyone who can merge a workflow change to master can print a secret. That circle is wider than forks, and it's why I wouldn't put a `.p12` there.

**What a leak costs — you have this part right.** A SignPath token leaks → revoke the token, signed builds unaffected. A Developer ID `.p12` leaks → revoke the cert, and builds already in users' hands can start failing Gatekeeper. A Developer ID cert can't be scoped either; it signs anything under our team identity.

**Apple's own answer is cloud signing (Xcode 13+).** The private key is generated and held on Apple's servers and never leaves them. CI sends hashes of the app content, Apple signs, the signature comes back. There is no `.p12` anywhere — nothing to steal. Authentication is an App Store Connect API key (`.p8` + key ID + issuer ID), which is revocable in one click without touching the certificate, can be scoped (Developer role + "Access to Cloud Managed Distribution Certificate" instead of Admin), and logs every signing request server-side. A leaked key lets someone request signatures until we revoke; it does not hand them our identity.

It runs on any CI, not just Xcode Cloud:

```
xcodebuild -exportArchive -archivePath ... -exportOptionsPlist ... \
  -authenticationKeyIssuerID <issuer> -authenticationKeyID <key> \
  -authenticationKeyPath key.p8 -allowProvisioningUpdates
```

Caveat: it goes through `xcodebuild -exportArchive`, so it needs an Xcode archive and we build with CMake. `cmake -G Xcode` could bridge that, but it's work and I haven't tried it here. WWDC21 session 10204 has the mechanism.

**Xcode Cloud is the packaged version of the same idea.** Signing runs on Apple's side, so no Apple credentials touch GitHub at all — it removes the question rather than mitigating it, and the Mac build machines come with the $99/year membership (25 compute hours/month). Per-PR builds would eat that allotment fast and start costing extra; a monthly build fits inside it. It also needs an Xcode project, and I haven't confirmed it can emit a notarized Developer ID DMG rather than only TestFlight/App Store builds.

**My suggestion:** ship @arummler's macOS job unsigned so the PR can move, and leave the `.p12`-in-secrets approach from those articles alone. If CI signing comes back up later, cloud signing is the one worth the effort — and it still wants an Environment with required reviewers, restricted to tag and dispatch triggers.

scorpio810:
@jp2images Thanks for the very clear response and the technical details — the reframing around the triggers (pull_request_target/workflow_run/issue_comment) and the cloud signing option are really helpful. I'll go with your suggestion: we'll ship @arummler's unsigned job to unblock the Qt6/CMake migration, and revisit cloud signing later if needed.

In the meantime, I'll keep providing packages I compile myself on my Mac Mini M2 from time to time, as I already do.

scorpio810:
@jp2images @arummler One thing that changes quite a bit for everything we've just discussed: we're considering removing the KF6 dependency from the builds (see issue #907, in favor of the Qt-only replacements in sources/ui/nokde) and embedding sqlite3 directly in QET rather than relying on the system version.

If that goes through, it should resolve a good chunk of what's made the macOS toolchain painful so far — it was mainly KF6 that caused problems with Homebrew (no reliable tap, modules that don't compile, scripts designed to run as root) and sqlite3 that caused the missing-plugin issue without recompiling Qt. Without those two, pretty much the only external dependency left is Qt6 itself — a brew install qt6 stays well within a reasonable CI budget.

Given that, I'm now leaning toward Xcode Cloud rather than a self-managed signing setup: with KF6/sqlite3 out of the picture, the 25 free compute hours/month should comfortably cover the build, and cloud signing means the Developer ID private key never touches GitHub secrets at all — which was really my main worry with the .p12 approach in those articles. @jp2images, does the cmake -G Xcode bridge you mentioned seem realistic to try, or is there a better way to get from our CMake build to an archivable Xcode project?

Otherwise, I’ll carry on putting together bundles every now and then with my mini M2.

jp2images:
@scorpio810 I checked the part I couldn't confirm last time: Xcode Cloud does support Developer ID. There's a "Notarize — macOS" post-action, and `CI_DEVELOPER_ID_SIGNED_APP_PATH` hands the signed archive to `ci_post_xcodebuild.sh`.

The catch is that the private key never lands in the runner's keychain — `security find-identity` returns nothing in a post-build script ([forum thread 762197](https://developer.apple.com/forums/thread/762197), unanswered since August 2024). Two steps in `misc/MacQetDeploy_arm64_cmake.sh` depend on it:

- Line 540 signs the DMG wrapper.
- Line 149 runs `macdeployqt`, which copies the Qt frameworks in *after* the build. Lines 379-402 then re-sign every framework and dylib.

The second one is the blocker. Xcode Cloud signs at export, before `ci_post_xcodebuild` runs, so that re-sign can't happen. Fixing it means the Xcode target embeds the Qt frameworks itself, in a copy-files phase, and `macdeployqt` is dropped. That rewrite is most of the work.

On your question: `cmake -G Xcode` works. But the generated `.xcodeproj` is a build artifact, so it isn't committed, and Xcode Cloud picks project and scheme from the checkout when you define the workflow. The pattern that works is a `ci_post_clone.sh` that generates it, with a shared scheme name matching the workflow — XcodeGen and Tuist users do this routinely. So yes, realistic.

One option we haven't discussed: a self-hosted runner on your M2. The certificate stays in your keychain, and nothing about the CMake build changes. GitHub advises against self-hosted runners on public repos, so it needs an Environment with required reviewers and tag/dispatch triggers only — the same gate I'd want on cloud signing. Less elegant, much less work.

Either way, I'd still merge @arummler's macOS job unsigned first.

Background if useful: [WWDC23 — Simplify distribution in Xcode and Xcode Cloud](https://developer.apple.com/videos/play/wwdc2023/10224/), [Xcode Cloud custom build scripts](https://www.createwithswift.com/custom-build-scripts-and-environment-variables-for-xcode-cloud/), [Distribute outside the Mac App Store](https://help.apple.com/xcode/mac/current/en.lproj/dev033e997ca.html).

scorpio810:
@jp2images Thank you, I’ve found a report mentioning the same issue on Xcode cloud  [847152](https://developer.apple.com/forums/thread/847152)

Thank you very much for your research and expertise.

PR #724, opened 2026-08-11
Fix wire-name export doubling every conductor's count

ispyisail:
Fixes the bug reported on the forum: https://qelectrotech.org/forum/viewtopic.php?id=3125

## Problem

Using "Export the names of list of wires" (`ConductorNumExport`), a single named conductor was exported twice. Reporter's exact words: *"I have a name conductor named '16AWG' and when I use this command it exports two of the same so: '16AWG' '16AWG'"*.

## Root cause

`ConductorNumExport::fillHash()` incremented the name's tally once per **terminal** instead of once per **conductor** — a separate `if` block for `terminal1` and another for `terminal2`, each independently bumping the same `m_hash` entry. Since an ordinary conductor has two terminals and neither is a folio-report terminal in the common case, both blocks fired and every real conductor was counted twice. `wiresNum()` then faithfully repeats each name `m_hash.value(key)` times, so the doubled count became doubled output lines. No bus, multiple diagrams, or master/slave elements needed — a single ordinary two-terminal named conductor reproduces it every time.

## Fix

Increment once per conductor, skipping it only when *both* ends are folio-report terminals (neither represents a real connection) instead of checking each terminal independently and incrementing twice.

## Testing

Verified with a minimal two-conductor project via the `--export-wires` CLI verb:
- **Before fix:** 2 named conductors → 4 lines exported (exact doubling from the report)
- **After fix:** 2 named conductors → 2 lines exported (correct)

Also did a clean incremental build with no errors/new warnings.


scorpio810:
> Before fix: 2 named conductors → 4 lines exported (exact doubling from the report)
> After fix: 2 named conductors → 2 lines exported (correct)

No!  That’s by design: https://qelectrotech.org/forum/viewtopic.php?pid=10633#p10633

scorpio810:
When you connect a wire, you need a wire marker at each end of that wire....

ispyisail:
opps

So true

scorpio810:
;-)

ispyisail:
Reopening — closed to clear @scorpio810's review backlog rather than on merit. Merged with current `master`, and this one can be demonstrated rather than argued.

`--export-wires` on `examples/industrial.qet`:

| | rows |
|---|---|
| before | **940** |
| after | **595** |
| conductors in the project | **671** |

The old figure exceeding the number of conductors is the bug in one line: the count could not have been counting conductors. It was incrementing once per non-report *terminal*, so an ordinary conductor with two real ends was tallied twice.

The new condition counts a conductor once unless *both* of its ends are folio reports, which leaves the other two cases exactly as they were — one report end still counts once, two report ends still count zero — and brings the total below the conductor count where it belongs.

**Verified:** Qt 6.10.2 / KF 6.10.0, build clean (0 errors), `ctest` 4/4, merges clean with `master`, and the export comparison above.

PR #727, opened 2026-08-12
Fix print window clipping diagram when titleblock on right edge is hidden

zi-mozhuang:
When the titleblock is docked on the right edge, diagramRect() always
subtracted titleBlockRect().height() from the printable height. For a
right-edge titleblock this height equals the full diagram height, so the
print area collapsed to zero and the schematic vanished, leaving only the
border and column/row headers.

Query the docked edge via BorderTitleBlock::titleBlockEdge() and subtract
the dimension that actually carries the titleblock extension: height for a
bottom-edge titleblock, width for a right-edge one.

### DIFF
--- a/sources/bordertitleblock.h
+++ b/sources/bordertitleblock.h
@@ -73,7 +73,7 @@ class BorderTitleBlock : public QObject
 			return(rows_count_ * rows_height_); }
 		/// @return la rows header width, in pixels
 		qreal rowsHeaderWidth() const { return(rows_header_width_); }
-		// @return the edge where title block is docked
+		/// @return the edge where title block is docked
 		Qt::Edge titleBlockEdge() const { return(m_edge); }
 	
 		// border - title block = diagram
--- a/sources/bordertitleblock.h
+++ b/sources/bordertitleblock.h
@@ -94,6 +94,9 @@ class BorderTitleBlock : public QObject
 
 		QRectF titleBlockRect () const;
 
+                // @return the edge where title block is docked
+                Qt::Edge titleBlockEdge() const { return(m_edge); }
+
 		DiagramContext titleblockInformation() const;
 	private:
 		QRectF titleBlockRectForQPainter () const;
--- a/sources/bordertitleblock.h
+++ b/sources/bordertitleblock.h
@@ -73,6 +73,8 @@ class BorderTitleBlock : public QObject
 			return(rows_count_ * rows_height_); }
 		/// @return la rows header width, in pixels
 		qreal rowsHeaderWidth() const { return(rows_header_width_); }
+		// @return the edge where title block is docked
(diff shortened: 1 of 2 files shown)


ispyisail:
Reviewed by applying the patch to current `master`, building it, and testing the print preview. It does what it says, and I could not find anything wrong with it.

## The diagnosis is accurate

`BorderTitleBlock::titleBlockRect()` returns a rect whose meaning flips with the docked edge:

```cpp
if (m_edge == Qt::BottomEdge)
    return QRectF(diagram_rect_.bottomLeft(),
                  QSize(diagram_rect_.width(), renderer->height()));   // height = thickness
else
    return QRectF(diagram_rect_.topRight(),
                  QSize(renderer->height(), diagram_rect_.height()));  // WIDTH = thickness
```

So for a right-edge titleblock `.height()` is the **full diagram height**, and the old code subtracted the entire printable area — exactly as described.

The arithmetic is exact rather than approximate. `borderAndTitleBlockRect()` is `diagram_rect_ | titleBlockRect()`, so the union is `diagram + thickness` on the relevant axis; subtracting that thickness returns precisely `diagram_rect_`, for both branches.

## Tested

Patch applies cleanly to current master and builds with no new warnings. Tested through the PDF export path (deliberately not the physical-printer path), with a project whose titleblock is docked right, and a second one docked bottom:

| Scenario | Stock master | With this PR |
|---|---|---|
| Right edge, titleblock **on** | content visible | content visible |
| Right edge, titleblock **off** | **content vanishes** — border + column/row headers only | **content retained** |
| Bottom edge, titleblock **on** | normal | normal |
| Bottom edge, titleblock **off** | normal | **unchanged** |

The bug reproduces on stock exactly as reported, and the fix resolves it with no bottom-edge regression.

## Supporting checks

- **The `if bottom … else …` shape is safe.** `TitleBlockProperties::display_at` is only ever `Qt::BottomEdge` or `Qt::RightEdge` — the XML and settings round-trip is literally `"bottom"`/`"right"`. It also matches the convention already used at every other `m_edge` site in `bordertitleblock.cpp`.
- **The new getter is justified.** There was no accessor for the edge; the alternative, `exportTitleBlock().display_at`, constructs a whole `TitleBlockProperties` just to read one field.
- **No sibling call sites were missed.** The only other use of `titleBlockRect()` outside the class is a `contains()` hit-test in `diagramview.cpp`, which is unaffected. The SVG/PNG export path hides the titleblock via `displayTitleBlock()` instead of subtracting a rect, so it never had this bug.

## One trivial nit

```cpp
// @return the edge where title block is docked
```

The surrounding declarations use `///` for Doxygen, so this comment won't be picked up. One character — not worth blocking on.

Nice catch on a bug that would be easy to dismiss as a rendering quirk.

PR #728, opened 2026-08-12
Fix bugtracker #248: files passed to a running instance are never opened

ispyisail:
Fixes <https://qelectrotech.org/bugtracker/view.php?id=248>.

## The bug

Opening a project while QElectroTech is already running does nothing at all. The second process hands its file names to the first one and exits; the first one silently discards them.

The report describes this as happening with spaces in the path, but it turns out not to be limited to that — **no file ever arrives, whatever it is called**. Spaces are just how the reporter happened to notice.

There are two defects on that path, and they have to be fixed together.

## 1. `QET::splitWithSpaces()` never returned anything

It split on `"[^\\]?(?:\\\\)* "`, which as a regular expression is `[^\]?(?:\\)* `. The character class is unterminated — `[^\]` escapes the closing bracket instead of closing the class — so `QRegularExpression` rejects the whole pattern and `QString::split()` returns an **empty list for every input**. Qt has been saying so on stderr all along, once per attempt:

```
QString::split: invalid QRegularExpression object
```

With an empty argument list, `QETArguments` finds no project files and `openFiles()` has nothing to do.

Replaced with an explicit scan that mirrors `joinWithSpaces()` / `escapeSpaces()`: a backslash escapes the next character, an unescaped space separates arguments. No regular expression, so there is nothing to silently get wrong.

## 2. Opening the files from the socket handler crashes

With (1) fixed the files arrive — and QElectroTech segfaults:

```
#1  QIODevice::channelReadyRead(int)
#2  libQt5Network                     <- QLocalSocket read handler
#13 main
```

`QETApp::receiveMessage()` is invoked directly from the `QLocalSocket` `readyRead` handler that carried the message, and opening a project runs nested event loops of its own (the "please wait" dialog, then the backup prompt). Unwinding back into Qt's socket code long afterwards is what crashes.

The open is now deferred to the event loop, so the socket handler completes first.

Fixing only (1) would have traded a silent failure for a crash, which is why both are in one commit.

## Testing

With a project already open, sending a second file to the running instance:

| Case | Before | After |
|---|---|---|
| Path with spaces (`/tmp/qet space dir/my test.qet`) | nothing happens | opens |
| Path without spaces | nothing happens | opens |
| Two files at once, one with spaces | nothing happens | both open |
| Crash on the backup prompt afterwards | n/a (unreachable) | no crash |

No `invalid QRegularExpression` warnings remain. Built against Qt 5.15, Release, no new warnings.

## Note

This also means the single-instance file hand-off has been dead for some time, so anything relying on it (desktop file association, `xdg-open`, passing a file while QET is open) has been silently doing nothing.

### DIFF
--- a/sources/qet.cpp
+++ b/sources/qet.cpp
@@ -544,23 +544,51 @@ QString QET::joinWithSpaces(const QStringList &string_list) {
 	@return La liste des sous-chaines, sans echappement.
 */
 QStringList QET::splitWithSpaces(const QString &string) {
-	// les chaines sont separees par des espaces non echappes
-	// = avec un nombre nul ou pair de backslashes devant
-#if TODO_LIST
-#pragma message("@TODO remove code for QT 5.14 or later")
-#endif
-	QStringList escaped_strings = string.split(QRegularExpression("[^\\]?(?:\\\\)* "),
-#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0)	// ### Qt 6: remove
-						   QString
-#else
-						   Qt
-#endif
-						   ::SkipEmptyParts);
-
+		// Substrings are separated by unescaped spaces; spaces belonging to a
+		// substring are escaped by joinWithSpaces()/escapeSpaces(), which also
+		// escapes the backslash itself. Scanning explicitly rather than
+		// splitting on a regular expression: the previous pattern,
+		// "[^\\]?(?:\\\\)* ", contained an unterminated character class
+		// ("[^\\]" escapes the closing bracket), so QRegularExpression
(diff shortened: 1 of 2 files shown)


ispyisail:

ispyisail:
Closing — both fixes here are already on master, merged separately.

Opening a project while QElectroTech is already running now works: the file arrives and opens, with or without spaces in the path. #861 fixed the argument parsing, with a test this branch never had, and #868 moved the opening out of the socket handler, which is the crash this branch's change would have caused.

Verified on master 0258d1a7 — a second instance handed a path with spaces to a running one, which opened it and stayed up.

Still outstanding on bugtracker #248: the window does not come to the front.

PR #729, opened 2026-08-12
Don't drop files handed to a starting instance while the backup prompts are up

ispyisail:
> **Depends on #728.** This branch is built on top of it, so the diff shown here includes that commit. Review #728 first; this PR adds one commit on top (`sources/qetapp.cpp`, +10/-1).

## The bug

`QETApp`'s constructor ends with `checkBackupFiles()`, which opens modal dialogs — the "restore these files?" prompt, and the crash-report offer. Those dialogs run their own event loop, so **the constructor does not return until the user answers them**.

`main()` still has work to do at that point. In particular, a few lines later:

```cpp
QObject::connect(&app, &SingleApplication::receivedMessage,
                 &qetapp, &QETApp::receiveMessage);
```

While the prompts are up, that connection does not exist yet. A second instance launched during that window — double-clicking a project, or `xdg-open`, while the first copy is still asking about restore files — hands its file names over, SingleApplication accepts and delivers them, and **nothing is listening**. The message is discarded, the second process has already exited, and the file is silently lost.

## The fix

Defer `checkBackupFiles()` to the event loop so the constructor returns promptly. `main()` finishes wiring up, and the prompts appear immediately afterwards exactly as before. `checkBackupFiles()` is called from this one place only.

## Testing

With a stale restore file present, sending a project to the running instance *while the restore prompt is displayed*:

| | Before | After |
|---|---|---|
| File sent during the prompt | dropped — never appeared, even after answering the prompt | opens |
| Restore prompt still shown and working | yes | yes |
| Backup prompt still shown and working | yes | yes |

## Why it is separate from #728

This is only *observable* once #728 is in — before that, no file passed to a running instance was opened under any circumstances, so the window described here was invisible. They are independent defects on the same path, kept as separate commits.

### DIFF
--- a/sources/qet.cpp
+++ b/sources/qet.cpp
@@ -544,23 +544,51 @@ QString QET::joinWithSpaces(const QStringList &string_list) {
 	@return La liste des sous-chaines, sans echappement.
 */
 QStringList QET::splitWithSpaces(const QString &string) {
-	// les chaines sont separees par des espaces non echappes
-	// = avec un nombre nul ou pair de backslashes devant
-#if TODO_LIST
-#pragma message("@TODO remove code for QT 5.14 or later")
-#endif
-	QStringList escaped_strings = string.split(QRegularExpression("[^\\]?(?:\\\\)* "),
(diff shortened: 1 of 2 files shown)


ispyisail:

PR #731, opened 2026-08-12
CLI export: hide terminal markers and names in rendered output (add --show-terminals)

philippeagray:
## Problem

Headless exports (`--export-pdf`, `--export-png`, `--export-svg`) render
every element terminal with the editor's red terminal stroke and blue
docking dot (and the terminal name, when one is set to display). A
finished drawing exported from the CLI therefore ships with coloured
editor UI on every terminal — easy to spot on an otherwise monochrome
schematic.

## Root cause

`Terminal::paint()` draws the markers whenever the diagram's
`drawTerminals()` / `drawTerminalNames()` flags are set, and both
default to `true` (`sources/diagram.cpp`, constructor). The GUI export
dialog clears them through `Diagram::applyProperties()`
(`draw_terminals` / `draw_terminal_names` in `ExportProperties`), but
`cli_export.cpp` never does — `renderDiagram()` only toggles the grid
and guides.

## Fix

`renderDiagram()`, `exportPdf()` and `exportImages()` take a
`showTerminals` parameter (default `false`) and thread it through to
`setDrawTerminals()`/`setDrawTerminalNames()` around the render,
exactly like the existing grid/guides toggle, restoring the previous
values afterwards. `run()` recognises a new standalone `--show-terminals`
switch, stripped out before the existing positional-argument parsing so
it can appear anywhere on the command line.

This is a default-off change to the CLI export path only — application
defaults are untouched (the interactive editor and its own export
dialog behave exactly as before), and non-image export modes (BOM,
nets, links, cables, resave, etc.) are unaffected since they never call
`renderDiagram()`. I chose a real flag over unconditionally hiding the
markers so `--show-terminals` stays available for visually debugging an
unconnected pin from the CLI, which is arguably the whole reason the
markers exist in the first place.

Updated the `CLIExport::run()` doc comment in `cli_export.h` to list
the new flag alongside the existing usage examples.

## Testing

Verified by inspection against `Terminal::paint()`'s guard and by
reproducing the issue on 0.100.1-dev r9269 Windows builds: a generated
project exported via `--export-pdf` contains exactly one red stroke
(`1 0 0` stroke colour) and one blue dot (`0 0 1`) per terminal in the
PDF content stream; the same export with `--show-terminals` restores
them; the GUI export of the same project with default export options
contains none either way (this patch doesn't touch that path). I don't
have a local Qt build environment, so I'd appreciate CI/a maintainer
build confirming compilation — the patch only uses the same public
`Diagram` setters/getters the export dialog path already uses, plus a
`QStringList::removeAll` on the incoming args.

Context: we drive the CLI export from an automated drawing-generation
pipeline (electrical lab drawing set), where the coloured markers were
the one thing separating headless output from the GUI export.

### DIFF
--- a/sources/cli_export.cpp
+++ b/sources/cli_export.cpp
@@ -117,11 +117,22 @@ void renderDiagram(Diagram *diagram, QPainter &painter, const QRectF &target)
 	// and restore it afterwards.
 	const bool was_drawing_grid = diagram->displayGrid();
 	const bool was_drawing_guides = diagram->displayGuides();
+	// Terminal markers (red stroke + blue docking dot) and terminal names
+	// are editor UI: Terminal::paint() draws them whenever the diagram's
+	// drawTerminals()/drawTerminalNames() flags are set (default true).
+	// The GUI export path clears them via Diagram::applyProperties();
+	// do the same here so headless exports match.
+	const bool was_drawing_terminals = diagram->drawTerminals();
+	const bool was_drawing_terminal_names = diagram->drawTerminalNames();
 	diagram->setDisplayGrid(false);
 	diagram->setDisplayGuides(false);
+	diagram->setDrawTerminals(false);
+	diagram->setDrawTerminalNames(false);
 	diagram->render(&painter, target, source, Qt::KeepAspectRatio);
 	diagram->setDisplayGrid(was_drawing_grid);
 	diagram->setDisplayGuides(was_drawing_guides);
+	diagram->setDrawTerminals(was_drawing_terminals);
+	diagram->setDrawTerminalNames(was_drawing_terminal_names);
 }
 
 int exportPdf(QETProject &project, const QString &output)
--- a/sources/cli_export.cpp
+++ b/sources/cli_export.cpp
@@ -109,7 +109,16 @@ QString diagramStem(Diagram *diagram, int index)
 }
 
 /// Render @p diagram into @p painter, fitting @p target to the page rect.
-void renderDiagram(Diagram *diagram, QPainter &painter, const QRectF &target)
+/// @p showTerminals: paint terminal markers (red stroke + blue docking
+/// dot) and terminal names, as the interactive editor does. Off by
+/// default — Terminal::paint() draws them whenever the diagram's
+/// drawTerminals()/drawTerminalNames() flags are set (default true, so
+/// headless export used to ship them as editor UI); the GUI export
+/// dialog already defaults to clearing them via
+/// Diagram::applyProperties(). Pass true (--show-terminals) to keep
+/// them, e.g. to visually debug an unconnected pin.
+void renderDiagram(Diagram *diagram, QPainter &painter, const QRectF &target,
+					bool showTerminals = false)
 {
 	const QRect source = diagramRect(diagram);
 	// Export without the editor grid: drawBackground() only paints it when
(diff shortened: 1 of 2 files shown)


ispyisail:
> I don't have a local Qt build environment, so I'd appreciate CI/a maintainer build confirming compilation

**It compiles cleanly** — Qt 5.15, Release, GCC, no errors and no new warnings. I also applied it to current `master` (clean apply), built it, and exercised all three image formats. It behaves exactly as described.

## The analysis is correct

Everything in the root-cause section checks out:

- `Terminal::paint()` guards on the flags at `terminal.cpp:202` (`drawTerminals()`) and `:281` (`drawTerminalNames()`).
- Both default to `true` — `diagram.cpp:69-70`.
- The GUI dialog clears them through `Diagram::applyProperties()` at `diagram.cpp:2322-2332`, and `cli_export.cpp` never did.

One point strengthens the default-off choice more than the description claims: `ExportProperties` itself defaults to `draw_terminals(false)` / `draw_terminal_names(false)` (`exportproperties.cpp:36-37`). So this doesn't just pick a sensible default — it makes the CLI agree with the GUI export default. Divergence was the bug.

## Verified behaviour

Test project with terminals, exported headless (`QT_QPA_PLATFORM=offscreen`):

| Export | Without flag | With `--show-terminals` |
|---|---|---|
| PDF — `1 0 0 SCN` (red stroke) | 0 | 2 |
| PDF — `0 0 1 SCN` (blue dot) | 0 | 2 |
| SVG — red/blue stroke attrs | 0 | 6 |
| PNG — image hash | — | differs |

The PDF content stream with the flag contains precisely what you described: `1 0 0 SCN` stroking a short segment (`-13 0 m / -9 0 l / S`) for the terminal, and `0 0 1 SCN` stroking a zero-length one (`-13 0 m / -13 0.001 l / S`) for the docking dot. Neither appears without the flag.

Also confirmed:

- **Flag position genuinely doesn't matter** — before the export flag, between the positional arguments, and at the end all produce identical output (2 red + 2 blue). The `removeAll()`-before-parsing approach works.
- **Non-image modes are unaffected** — `--export-bom` produces byte-identical CSV with and without the flag.
- **`--show-terminals` reaches `run()` intact.** CLI export is dispatched from `main.cpp:133` before SingleApplication and before `QETArguments`, so there is no earlier option validation to trip over, and `isExportRequest()` only scans for export flags so the new switch doesn't disturb detection.

## Notes, none blocking

1. **The default arguments are redundant.** `renderDiagram()`, `exportPdf()` and `exportImages()` all live in the anonymous namespace with every call site passing `showTerminals` explicitly, so `= false` never binds. Harmless, and it does compile — there are no separate forward declarations, so there is no risk of the "default argument given twice" error. But dropping the defaults would make a future call site that forgets the parameter a compile error rather than a silent behaviour change. Purely your call.

2. **`removeAll()` operates on the whole argument list**, so an output path or a `--set-titleblock` value literally named `--show-terminals` would be swallowed. Contrived enough not to matter; noting it only for completeness.

3. No separate `--help` text lists these flags, so the header doc comment is the documentation — and it is updated correctly, including the "no effect on non-image modes" caveat.

Well-diagnosed and cleanly scoped: the save/restore idiom matches the existing grid/guides handling exactly, and the blast radius is limited to the CLI image-export path.

philippeagray:
Related: filed a separate issue for a conductor-rendering defect found while testing the CLI headless-export path in this PR — #734 (conductors render a short self-retracing spur/tick at docked terminals; reproduces on both a generated file and QET's own shipped `affuteuse_modern.qet` example, and survives QET's own `--resave`).

ispyisail (approved):
Verified every claim in the PR body directly against source: Terminal::paint()'s guard on drawTerminals()/drawTerminalNames(), Diagram's draw_terminals_/draw_terminal_names_ defaults (true), applyProperties()'s existing GUI-side handling of the same flags, and confirmed renderDiagram() has exactly the 3 call sites described (2 in exportImages, 1 in exportPdf) — non-image export modes genuinely never touch it. The --show-terminals extraction runs before the existing positional-arg parsing loop, so it's genuinely order-independent without touching that logic, and there's no header declaration of renderDiagram/exportPdf/exportImages anywhere else, so no default-argument redefinition risk. Default-off keeps the interactive editor's own export dialog untouched. Solid, well-tested contribution.

PR #732, opened 2026-08-12
Fix bugtracker #238: summary lists folios in the wrong order

ispyisail:
Fixes <https://qelectrotech.org/bugtracker/view.php?id=238>.

## The bug

A project summary is a table of contents, so its rows have to follow the page order. They don't: the folios come out sorted alphabetically by whichever column happens to be first in the table — normally the title.

## Cause

`SummaryQueryWidget::queryStr()` built the `ORDER BY` clause inside the same loop that built the column list, so the two were always identical:

```sql
SELECT title, author FROM project_summary_view ORDER BY title, author
```

The folio's position in the project (the view's `pos` column) only took part in the ordering at all if the user happened to add the **Position** column to the *displayed* columns — and even then only as the first sort key if they moved it to the top.

There is no sort control anywhere in this widget. The only list it offers reorders *displayed* columns, so this was not something a user could correct.

## The fix

Order by `pos`. A user who wants something else can still write their own statement with the **Requête SQL** option, which this doesn't touch.

`setQuery()`, which rebuilds the widget state from a stored query, only parses between `SELECT` and `FROM` and ignores the `ORDER BY` entirely — so saved summary configurations keep working.

## Testing

Two-folio project whose titles sort the opposite way to the pages: **"Zebra" on page 1, "Alpha" on page 2**.

The dialog's SQL preview shows the generated statement directly:

| | generated query |
|---|---|
| before | `SELECT title FROM project_summary_view ORDER BY title` |
| after | `SELECT title, author FROM project_summary_view ORDER BY pos` |

And against that data, the old clause returns the pages in the wrong order:

```
page order in the project:  Zebra(1), Alpha(2), Middle(3)

ORDER BY title  ->  Alpha, Middle, Zebra      <- what a summary showed
ORDER BY pos    ->  Zebra, Alpha, Middle      <- the actual page order
```

### DIFF
--- a/sources/dataBase/ui/summaryquerywidget.cpp
+++ b/sources/dataBase/ui/summaryquerywidget.cpp
@@ -60,28 +60,18 @@ QString SummaryQueryWidget::queryStr() const
 	}
 
 		//Made a string list with the columns (keys) choosen by the user
-	QStringList keys = selectedKeys();
+	const QString column = selectedKeys().join(QStringLiteral(", "));
 
-	QString select ="SELECT ";
-	QString order_by = " ORDER BY ";
-
-	QString column;
-	bool first = true;
-	for (auto key: keys) {
-		if (first) {
-			first = false;
-		} else {
-			column += ", ";
-			order_by +=", ";
-		}
-		column += key;
-		order_by += key;
-	}
-
-	QString from = " FROM project_summary_view";
-
-	QString q(select + column + from + order_by);
-	return q;
+		//Order by the folio's position in the project, never by the displayed
+		//columns. This is a summary -- a table of contents -- so its rows have
+		//to follow the page order. The ORDER BY used to be built from the same
+		//loop as the column list, so the folios came out sorted by whichever
+		//column happened to be first in the list (the title, normally), and
+		//"pos" only took part at all if the user had added the Position column
+		//to the display. There is no sort control in this widget, so that
+		//coupling was not something the user could correct.
+	return QStringLiteral("SELECT ") + column
+			+ QStringLiteral(" FROM project_summary_view ORDER BY pos");
 }
 
 /**


scorpio810:
This issue was posted on 2021-06-22 10:36 are you sure this issue as not be fixed since?

ispyisail:
Fair question — I checked before answering, and yes, it still reproduces on master `5f8367975`.

`SummaryQueryWidget::queryStr()` still builds the `ORDER BY` clause inside the same loop as the column list, so the two are always identical:

```cpp
for (auto key: keys) {
    if (first) { first = false; }
    else { column += ", "; order_by += ", "; }
    column   += key;
    order_by += key;      // same loop, same keys
}
```

With the default columns that produces:

```sql
SELECT title, author FROM project_summary_view ORDER BY title, author
```

so the summary is sorted alphabetically rather than by page order.

`project_summary_view` does expose the folio position (`d.pos AS pos`), but the view carries no `ORDER BY` of its own:

```sql
CREATE VIEW project_summary_view AS SELECT ... d.pos AS pos
  FROM diagram_info di, diagram d WHERE di.diagram_uuid = d.uuid
```

so `pos` only affects the ordering if a user happens to add the **Position** column to the *displayed* columns and moves it to the top — and the widget has no sort control to do that deliberately. The only reordering list it offers changes which columns are shown.

Worth saying your scepticism was reasonable: I went through the open bugtracker entries recently and a good number of the older ones have quietly been fixed without being closed. This one hasn't.

Happy to add a before/after screenshot if that's useful.

scorpio810:
https://qelectrotech.org/forum/viewtopic.php?id=2619

ispyisail:
Thanks — that thread is a better argument for this patch than the bugtracker entry is, because **the fix you recommended in it is the one this PR implements**:

> Recommended sorting by "folio position" instead of the number itself

That is exactly `ORDER BY pos`. McKeon agreed in the same thread. So this is a second, independent report of the bug, from users, with the maintainer already having named the right fix — it just never got made.

It also tells me the symptom is broader than bugtracker #238 describes. #238 is "wrong order"; the forum is specifically **"10 sorts before 2"** once a project passes ten folios. Both are the same root cause, and `ORDER BY pos` fixes both. The reason is a schema detail worth putting on the record:

```cpp
// sources/dataBase/projectdatabase.cpp:271
"pos INTEGER)"                       // diagram.pos

// :302  -- every diagram_info key except "date"
diagram_info_table += string += string=="date" ? " DATE" : " VARCHAR(100)";
```

`folio` is text, `pos` is an integer. Sorting on `folio` is therefore lexicographic — that *is* the "10 before 2" bug — while `pos` sorts numerically with no cast needed. Rebuilding the real schema and view in SQLite:

```
ORDER BY folio  ->  1, 10, 11, 12, 2, 3
ORDER BY pos    ->  1, 2, 3, 10, 11, 12
```

## Why `pos` rather than the thread's workarounds

Nardo26's `cast(folio as int)` and LievenC's `folio + 0 ASC` both fix the ten-folio case, and both stay available through the **Requête SQL** option, which this PR does not touch. But they break on the custom folio naming julien.mathieu raises later in that same thread (`0-PG`, `0-SOM`): both labels cast to `0`, so they tie, and the tie is not stable.

```
rows inserted one way:   cast -> [0-PG, 0-SOM, 1]     ORDER BY pos -> [0-PG, 0-SOM, 1]
rows inserted the other: cast -> [0-SOM, 0-PG, 1]     ORDER BY pos -> [0-PG, 0-SOM, 1]
```

The cover page and the contents page swap. `pos` is right in both cases because it is the folio's actual position in the project, not something parsed back out of its name.

## Compatibility, re-checked

- Saved summary configurations keep working: `setQuery()` removes the first 7 characters and truncates at `FROM` (`summaryquerywidget.cpp:100-102`), so it only ever parses the column list and never reads the `ORDER BY`.
- Hand-written queries are untouched: `queryStr()` returns the user's text unchanged when the SQL checkbox is ticked (`:57-59`), so anyone already using the forum workarounds keeps them.

## One thing I should correct in my own description above

The before/after table in the PR body compares `SELECT title` against `SELECT title, author`, which makes it look as though the patch changed the selected columns too. It does not — only the `ORDER BY` changes. Same column list, both sides:

| | generated query |
|---|---|
| before | `SELECT title, author FROM project_summary_view ORDER BY title, author` |
| after | `SELECT title, author FROM project_summary_view ORDER BY pos` |

## One pre-existing wart, not addressed here

If a user removes every displayed column, `selectedKeys()` returns empty and the statement becomes `SELECT  FROM project_summary_view …`, which is a syntax error. I checked: it fails identically with and without this patch, so it is not a regression and I have left it alone rather than widen the diff. Happy to fix it separately if you want it.

ispyisail:

PR #733, opened 2026-08-12
Fix bugtracker #243: allow copying out of a read-only element

ispyisail:
Fixes <https://qelectrotech.org/bugtracker/view.php?id=243>.

## The problem

Opening an element you cannot write — anything from the QET collection, for instance — disables **Copy** along with everything else, so there is no way to reuse a primitive from it. The reporter's workaround was to save the whole element into their own collection first, just to take one shape out of it.

## Cause

Two actions stood in the way in `QETElementEditor::updateAction()`, and neither of them modifies anything:

- **Select All** and **Invert Selection** were in the list disabled outright when read only, so nothing could be selected in the first place;
- **Copy** was enabled only when `!m_read_only && selectedItems().count()`, so even a mouse selection left it greyed out.

## The fix

Both now work on a read-only element. Copy is safe there: `ElementScene::copy()` serialises the current selection to the clipboard and touches neither the element nor the file.

**Cut, Paste, Paste-in-area, Delete, Rotate, Fine-Rotate, Flip, Mirror, the depth actions and the add-primitive tools all stay disabled exactly as before.** This only stops the editor refusing to read out what it is already displaying on screen.

## Testing

With a `chmod 444` element (title bar shows `[lecture seule]`):

| | before | after |
|---|---|---|
| Tout sélectionner | greyed | **enabled** |
| Inverser la sélection | greyed | **enabled** |
| Copier (with a selection) | greyed | **enabled** |
| Couper / Coller / Coller dans la zone | greyed | greyed |
| Supprimer / Rotation / Mirror | greyed | greyed |
| Enregistrer | greyed | greyed |

After Select All → Copy, the clipboard contains the expected element description:

```xml
<definition type="element" link_type="simple" ... >
    <names><name lang="en">ReadOnly copy test</name></names>
    <description>
        <rect .../>
        <line .../>
        <circle .../>
    </description>
</definition>
```

— ready to paste into an element of your own. The source file is untouched afterwards: same permissions (`-r--r--r--`), same checksum.


scorpio810:
An common collection as readonly, user need to save it this element to custom or company collections before edit it!

ispyisail:
Understood — if saving into a custom or company collection first is the intended workflow for a common-collection element, then this is working as designed and the premise of the PR is wrong. Closing it.

For the record, and in case anyone revisits this: while checking your comment I found the change was only partial anyway. `QETElementEditor::setReadOnly()` also does

```cpp
m_view->setInteractive(!ro);
```

so on a read-only element a click on the canvas selects nothing, and this PR does not touch that. I confirmed it headlessly on both refs — a synthesised click selects 1 item when the view is interactive and 0 items in read-only, identically before and after the change. The Copy it enables is therefore only reachable through the Parts dock list, not by clicking the primitive you want.

So the description's claim that "even a mouse selection left it greyed out" was inaccurate — there is no mouse selection available in that state at all. Apologies for the noise.

Happy to revisit later if the read-only behaviour is ever reconsidered, but on your explanation there is nothing to fix here.

ispyisail:

The claim that matters is that copying does not modify the element, and it holds: `ElementScene::copy()` calls `toXml(false)` and writes the result to the clipboard. It touches nothing else, so allowing it on a read-only element cannot alter the element.

The split the change makes is the right one — actions that only move the *selection* around (`Select all`, `Invert selection`, `Deselect all`, `Copy`) become available, while everything that would modify the element (`Cut`, `Delete`, `Paste`, `Paste from file`/`from element`, rotate, flip, mirror, depth changes) stays disabled by `m_read_only` exactly as before. Without this there is no way to select a primitive in a read-only element in order to copy it out, which is what bugtracker #243 asks for.

**Not verified:** the editor's action states interactively on a read-only element.

PR #736, opened 2026-08-13
Fix folio report link picker showing candidates as blank rows

### LINKED ISSUE #735: Folio report link picker: valid candidates render as blank rows (list looks empty)
### Summary

In the folio-report link editor (`LinkSingleElementWidget`, the **Report de folio** tab of a report arrow's properties), a perfectly valid candidate arrow is drawn as a **completely blank row**. The list looks empty, so it appears there is nothing to link to — but the row is there and becomes visible only once you click the empty strip and it highlights.

This makes cross-folio arrow linking look broken to the user.

### Cause

`LinkSingleElementWidget::buildTree()` finds the candidate correctly. I confirmed this with temporary instrumentation in `availableElements()`:

```
DBG256 self_type= 4 m_filter= 2 diagrams= 2
DBG256   cand type= 2 free= true folio= 1
DBG256 buildTree got 1 candidates
```

The problem is purely the column order. In the `AllReport` branch, `setUpHeaderLabels()` puts five **conductor** properties first:

```
N° de fil | Fonction | Tension / Protocole | Couleur du conducteur | Section du conducteur | N° de folio | Position | Titre de folio
```

and `buildTree()` fills all five with empty strings when the candidate has no conductor attached:

https://github.com/qelectrotech/qelectrotech-source-mirror/blob/master/sources/ui/linksingleelementwidget.cpp

```cpp
if (!elmt->conductors().isEmpty())
{
    ConductorProperties cp = elmt->conductors().first()->properties();
    str_list << cp.text;
    ...
}
else
    str_list << "" << "" << "" << "" << "";
```

The three columns that would actually identify the candidate — folio number, position, folio title — are appended *after* those five, and at the default column widths they sit off-screen past the tree's horizontal scrollbar. Net result: a row whose every visible cell is empty.

A report arrow with no conductor yet attached is a completely normal state, so this is easy to hit.

### Steps to reproduce

1. New project, add a second folio.
2. Place a **going** arrow (`next_report`) on folio 2 and a **coming** arrow (`previous_report`) on folio 1. Do not attach conductors.
3. Double-click the coming arrow → **Report de folio** tab.
4. The list appears empty.
5. Click the blank strip just under the header — a row highlights. It was there all along.
6. Right-click it → *Lier l'élément* → Apply. The link works fine and the arrow's `%{label}` resolves correctly.

Tested on `master` (merge-base `7307a59c1`), Qt5, Linux.

### Suggested fix

Any of these would resolve it; the first seems closest to intent:

- Lead the `AllReport` columns with the identifying ones (folio number / position / folio title), then the conductor properties — matching the `Slave` branch, which already leads with `actualLabel()` and so never shows a blank row.
- Or, when the candidate has no conductor, put the folio/position text into the first column instead of an empty string.
- Or call `resizeColumnToContents()` / set a sensible default header state so the identifying columns are visible without horizontal scrolling.

Only the `AllReport` branch is affected — the `Slave` branch puts `actualLabel()` first and is fine.

ispyisail:
Fixes #735.

### Problem

In a report arrow's **Report de folio** tab, a valid candidate arrow was drawn as a completely blank row. The list looked empty, so cross-folio linking appeared impossible — the row was actually there and became visible only once clicked.

`buildTree()` found the candidate correctly (confirmed with temporary instrumentation: `buildTree got 1 candidates` while the UI showed nothing). The cause was purely column order: the `AllReport` branch listed five **conductor** properties first, and `buildTree()` fills all five with empty strings when the candidate has no conductor attached — a perfectly normal state for a freshly placed report arrow. The three columns that identify the candidate (folio, position, folio title) came last and sat off-screen past the horizontal scrollbar.

### Change

Identifying columns first, conductor properties after — matching the `Slave` branch, which already leads with `actualLabel()` and so never shows a blank row.

Two smaller points in the same commit:

- when the candidate has no diagram, three empty strings are appended so the columns stay aligned with the headers instead of shifting left (previously the row was short by three cells);
- the saved header state key becomes `report-state-2`, so a state saved under the old column order isn't restored onto the new one and re-hiding the now-leading columns.

No new translatable strings — the same `tr()` literals, reordered.

### Testing

Built and exercised under Xvfb using the two custom arrow elements attached to bugtracker #256 (`01going_arrow_with_protocol.elmt` / `02coming_arrow.elmt`), with no conductors attached — the case that produced the blank row:

- **Before:** header shown, list visually empty; clicking the blank strip highlighted an invisible row.
- **After:** the row reads `Label de folio: 2/2 | Position: D6`.
- Linking still works end to end — right-click → *Lier l'élément* → Apply, after which the coming arrow's composite text resolves to `2-D6 / ` as expected.

Only the `AllReport` branch is touched; the `Slave` picker is unchanged.

### DIFF
--- a/sources/ui/linksingleelementwidget.cpp
+++ b/sources/ui/linksingleelementwidget.cpp
@@ -85,7 +85,7 @@ LinkSingleElementWidget::LinkSingleElementWidget(Element *elmt,
 		
 		const auto elmt_type{this->m_element->elementData().m_type};
 		if (elmt_type & ElementData::AllReport)
-			settings.setValue(QStringLiteral("link-element-widget/report-state"), qba);
+			settings.setValue(QStringLiteral("link-element-widget/report-state-2"), qba);
 		else if (elmt_type == ElementData::Slave)
 			settings.setValue(QStringLiteral("link-element-widget/slave-state"), qba);
 	});
@@ -304,6 +304,30 @@ void LinkSingleElementWidget::buildTree()
 			QStringList search_list;
 			QStringList str_list;
 			
+				//The folio of the candidate is what identifies it, so it comes first.
+				//The conductor properties below are empty whenever the candidate has
+				//no conductor attached yet, which would otherwise leave the row blank.
+			if (Diagram *diag = elmt->diagram())
+			{
+				if (settings.value(QStringLiteral("genericpanel/folio"), false).toBool())
+				{
+					autonum::sequentialNumbers seq;
+					QString F =autonum::AssignVariables::formulaToLabel(diag->border_and_titleblock.folio(), seq, diag, elmt);
+					str_list << F;
+				}
+				else
+				{
+					str_list << QString::number(diag->folioIndex() + 1);
+				}
+				str_list << diag->convertPosition(elmt->scenePos()).toString();
+				str_list << diag->title();
+			}
+			else
+			{
+				qDebug() << "In method void LinkSingleElementWidget::updateUi(), provided element must be in a diagram";
+				str_list << "" << "" << "";
+			}
+
 			if (!elmt->conductors().isEmpty())
 			{
 				ConductorProperties cp = elmt->conductors().first()->properties();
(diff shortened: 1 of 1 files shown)


ispyisail:

PR #737, opened 2026-08-13
Fix headless CLI hanging forever on version-incompatible projects

ispyisail:
### Problem

`QETProject::readProjectXml()` raises two blocking `QMessageBox` prompts while parsing a project: one when it was saved by a newer QElectroTech than the running build, and one when it was saved by 0.6 or older. Both ask the user whether to open anyway.

In a headless CLI run there is nobody to answer, so the modal never returns and the process hangs until it is killed. Because the prompt fires during *load*, before any export work starts, **every** CLI mode is affected.

The shipped example `examples/schema_indus.qet` (`version="0.3"`) reproduces it. All eight of these hang indefinitely on current master:

```
--info  --export-bom  --export-nets  --export-links
--export-cables  --export-wires  --resave  --export-pdf
```

I found this by running every example project through every CLI mode with a 120s timeout — that one project accounted for 8 timeouts, the only hangs in 185 runs.

### Change

Add `QETProject::setInteractive()`, mirroring the existing `setBackupEnabled()` pattern, and disable interactivity in `main.cpp`'s CLI branch right next to the `setBackupEnabled(false)` call already there for the same class of reason.

With interactivity off, both cases report the version mismatch on stderr via `qWarning()` and continue as if the user had chosen Open, so the export still happens and the reason is visible in the log:

```
Warning: '.../examples/schema_indus.qet' was saved with QElectroTech 0.3 and is only
partially compatible with this build (0.100.1); opening it anyway.
```

The GUI path is untouched — both prompts still appear when a project is opened interactively.

### Testing

On `examples/schema_indus.qet`, all eight modes went from hitting the 120s timeout to completing in under a second, with valid output (`--info` reports 48 elements / 69 conductors, `--export-pdf` writes a 21 KB PDF).

Regression check: opened the same project in the GUI and confirmed the "Avertissement" dialog still appears as before.

### DIFF
--- a/sources/main.cpp
+++ b/sources/main.cpp
@@ -249,6 +249,10 @@ QGuiApplication::setHighDpiScaleFactorRoundingPolicy(QetSettings::hdpiScaleFacto
 			// runs on a background thread referencing the project and races the
 			// process exit (intermittent segfault in QET::writeToFile).
 			QETProject::setBackupEnabled(false);
+			// No blocking prompts either: readProjectXml() asks the user what
+			// to do about version-incompatible projects, and with nobody to
+			// answer, the modal hangs the process instead of exporting.
+			QETProject::setInteractive(false);
 			return CLIExport::run(export_app.arguments());
 		}
 	}
--- a/sources/qetproject.cpp
+++ b/sources/qetproject.cpp
@@ -46,10 +46,16 @@
 static int BACKUP_INTERVAL = 1200000; //interval in ms of backup = 20min
 
 bool QETProject::m_backup_enabled = true;
+bool QETProject::m_interactive = true;
 
 void QETProject::setBackupEnabled(bool enabled)
 {
 	m_backup_enabled = enabled;
+}
+
+void QETProject::setInteractive(bool interactive)
+{
+	m_interactive = interactive;
 }
 
 /**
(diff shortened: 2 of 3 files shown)


ispyisail:
Superseded by #661, which fixes the same hang generally rather than at this one site.

Both start from the same case — `examples/schema_indus.qet` (version 0.3) blocking `QETProject::readProjectXml()` on a modal nobody can answer. This PR adds `QETProject::setInteractive()` and guards that specific prompt; #661 puts the guard in `QET::QetMessageBox` itself, which is where all 52 call sites already funnel, so the whole class is closed rather than the one instance.

That matters because the version warning is only the box this example file happens to trigger. Any modal reachable during a load does the same thing, and the next one added would need another fix like this one.

I'd suggest closing this in favour of #661 — happy to do so, just say the word. If you'd rather take the narrow fix first, this one still stands on its own.

ispyisail:
Closing in favour of #661, which fixes this hang generally rather than at the single `readProjectXml()` site.

Same starting case, wider fix: #661 guards inside `QET::QetMessageBox`, where all the call sites already funnel, so any modal reachable during a load is covered rather than just the version warning this example file happens to trigger.

Nothing here is lost — the diagnosis and the example project carry over, and I've attached the verification from this PR to #661.

PR #739, opened 2026-08-13
Export component info as invisible PDF text annotations

Kellermorph:
Add optional PDF export of element properties (BMK, manufacturer, designation,
description, and all non-formula keys) as invisible text annotations.

### UI
- New checkbox "Inscribe component information" in ProjectPrintWindow

### PDF export pipeline (projectprintwindow.cpp)
- When enabled, iterates all elements (skipping slaves and report elements),
  collects their properties using translated key names (via
  QETInformation::translatedInfoKey()), and creates placeholder link
  annotations via drawHyperlink() during the paint cycle.

### Post-processing (pdf_links.cpp)
New `convertComponentInfoAnnotations()` performs a forward-scan byte-level
conversion of the generated PDF:

- Converts placeholder link annotations (/Subtype /Link) to text annotations
  (/Subtype /Text)
- Removes the /A << /S /URI >> action dictionary (marker URL no longer needed)
- Writes /Contents as UTF-16BE hex string with BOM prefix for full Unicode
  support (Umlauten, special characters)
- Sets an invisible appearance via /AP referencing a shared empty Form XObject
  (/Type /XObject /Subtype /Form /BBox [0 0 0 0] /Length 0), making the
  annotation icon invisible while preserving the popup on click in the
  annotation panel
- Rebuilds the xref table to account for the new objects

`convertUriToGoTo()` now skips component-info annotations that have already
been processed.

### Cleanup
- Removed unused inject_annotations.py (rejected PyMuPDF approach)
- Removed unused struct fields (ComponentInfo.pageNumber, .rect) and dead
  code (pdfRect computation, unused sUri variable)
- Updated comments to reflect current behavior

## How it works

1. During painting, each element's properties are collected and a link
   annotation with marker URL http://componentinfo.local/ is placed at the
   element's position.
2. After the PDF is written, convertComponentInfoAnnotations() finds these
   markers, replaces them with invisible text annotations containing the
   translated property text, and rebuilds the xref table.
3. convertUriToGoTo() then processes remaining cross-reference links,
   skipping the already-converted component-info annotations.

The result: invisible annotations accessible via the PDF viewer's annotation
panel (Okular, Adobe, Brave, etc.) with full Unicode support.

### DIFF
--- a/sources/pdf_links.cpp
+++ b/sources/pdf_links.cpp
@@ -32,6 +32,7 @@
 #include <QGraphicsTextItem>
 #include <QList>
 #include <QRegularExpression>
+#include <QTextCodec>
 #include <QUrl>
 #include <QVector>
 
@@ -238,6 +239,15 @@ void convertUriToGoTo(const QString &pdfPath)
 
 			QByteArray uriVal = data.mid(uriStart, closeParen - uriStart);
 
+			// Skip component-info annotations — handled by
+			// convertComponentInfoAnnotations() in a separate pass.
+			if (uriVal.startsWith("componentinfo://")
+				|| uriVal.startsWith("http://componentinfo.local/")) {
+				out.append(data.mid(found, closeParen + 1 - found));
+				pos = closeParen + 1;
+				continue;
+			}
+
 			// Extract page number: look for #page=N or bare page=N
 			int pageNum = -1;
 			int hashPos = uriVal.lastIndexOf("#page=");
@@ -377,6 +387,202 @@ void convertUriToGoTo(const QString &pdfPath)
 	if (!out.open(QIODevice::WriteOnly | QIODevice::Truncate)) return;
 	out.write(result);
 	out.close();
+}
+
+void convertComponentInfoAnnotations(const QString &pdfPath,
+									const QList<ComponentInfo> &annotations)
+{
+	if (annotations.isEmpty()) return;
+
+	QFile f(pdfPath);
+	if (!f.open(QIODevice::ReadOnly)) return;
+	QByteArray data = f.readAll();
+	f.close();
+
+	const QByteArray marker("http://componentinfo.local/");
+	if (!data.contains(marker)) return;
+
+	int xrefStart = data.lastIndexOf("\nxref\n");
+	if (xrefStart == -1) xrefStart = data.lastIndexOf("\nxref ");
+	if (xrefStart == -1) return;
+	++xrefStart;
+	QByteArray body = data.left(xrefStart);
+
+	// Pre-scan body for highest object number
+	int maxObjNum = 0;
+	{
+		const QByteArray objMarker(" 0 obj");
+		int p = 0;
+		while ((p = body.indexOf(objMarker, p)) != -1) {
+			int numStart = p - 1;
+			while (numStart > 0 && body[numStart - 1] != '\n' && body[numStart - 1] != '\r')
+				--numStart;
+			QByteArray numStr = body.mid(numStart, p - numStart).trimmed();
+			bool ok = false;
+			int num = numStr.toInt(&ok);
+			if (ok && num > maxObjNum)
+				maxObjNum = num;
+			++p;
+		}
+	}
+	// Empty Form XObject used as invisible appearance for all annotations.
+	// All annotations reference this single object.
+	int emptyXObjNum = maxObjNum + 1;
+
+	QByteArray out;
+	out.reserve(data.size());
+
+	int pos = 0;
+	int annotIdx = 0;
+
+	while (pos < body.size()) {
+		int markerPos = body.indexOf(marker, pos);
+		if (markerPos == -1) {
+			out.append(body.mid(pos));
+			break;
+		}
+
+		int uriOpen = body.lastIndexOf("/URI (", markerPos);
+		int closeParen = (uriOpen != -1) ? body.indexOf(')', markerPos) : -1;
+		if (uriOpen == -1 || closeParen == -1 || uriOpen < pos) {
+			out.append(body.mid(pos, markerPos + marker.size() - pos));
+			pos = markerPos + marker.size();
+			continue;
+		}
+
+		// Find /S /URI before /URI (
+		int sUriPos = body.lastIndexOf("/S /URI", uriOpen);
+
+		// Find the /A << that opens the action dict containing /S /URI.
+		// Qt always writes: /A <<\n/S /URI\n/URI (...)\n>>\n>>
+		// We need /Contents at the annotation level, NOT inside /A <<.
+		int aDictOpen = body.lastIndexOf("/A <<", (sUriPos != -1) ? sUriPos : uriOpen);
+		int copyEnd = (aDictOpen != -1 && aDictOpen >= pos) ? aDictOpen
+					: (sUriPos != -1 && sUriPos >= pos) ? sUriPos
+					: uriOpen;
+		out.append(body.mid(pos, copyEnd - pos));
+
+		// Get text from annotations list (matched by order)
+		if (annotIdx >= annotations.size()) {
+			// Keep original action dict intact
+			out.append(body.mid(copyEnd, closeParen + 1 - copyEnd));
+			pos = closeParen + 1;
+			continue;
+		}
+
+		QByteArray contents = annotations[annotIdx].contents.toUtf8();
+		++annotIdx;
+
+		// Replace /Subtype /Link with /Subtype /Text
+		int subTypePos = out.lastIndexOf("/Subtype /Link");
+		if (subTypePos != -1) {
+			out.replace(subTypePos, 14, "/Subtype /Text");
+		}
+
+		// Encode as UTF-16BE hex with BOM for proper Unicode support (Umlauten etc.).
+		// PDF spec: hex strings starting with FE FF are interpreted as UTF-16BE.
+		QTextCodec *codec = QTextCodec::codecForName("UTF-16BE");
+		QByteArray utf16be;
+		utf16be.append('\xfe');
+		utf16be.append('\xff');
+		if (codec) {
+			utf16be += codec->fromUnicode(QString::fromUtf8(contents));
+		} else {
+			// Fallback: manual UTF-16BE encoding
+			QString str = QString::fromUtf8(contents);
+			for (int i = 0; i < str.size(); ++i) {
+				ushort cp = str.at(i).unicode();
+				utf16be.append(static_cast<char>((cp >> 8) & 0xFF));
+				utf16be.append(static_cast<char>(cp & 0xFF));
+			}
+		}
+
+		out += "/Contents <";
+		out += utf16be.toHex();
+		out += ">\n/AP << /N " + QByteArray::number(emptyXObjNum) + " 0 R >>\n";
+
+		// Skip the entire /A << ... >> action dict.
+		// After closeParen ')' we have: \n>>  (closes /A dict)  \n>>  (closes annot dict)
+		int actionDictClose = body.indexOf(">>", closeParen + 1);
+		if (actionDictClose != -1) {
+			pos = actionDictClose + 2;  // skip past >> that closes /A dict
+		} else {
+			pos = closeParen + 1;
+		}
+	}
+
+	if (annotIdx == 0) return;
+
+	// Append empty Form XObject (shared by all annotations for invisible appearance)
+	QByteArray emptyXObj;
+	emptyXObj += QByteArray::number(emptyXObjNum) + " 0 obj\n";
+	emptyXObj += "<< /Type /XObject /Subtype /Form /BBox [0 0 0 0] /Length 0 >>\n";
+	emptyXObj += "stream\nendstream\n";
+	emptyXObj += "endobj\n";
+	out += emptyXObj;
+
+	// Rebuild xref table
+	QMap<int, int> offsets;
+	{
+		const QByteArray objMarker(" 0 obj");
+		int p = 0;
+		while ((p = out.indexOf(objMarker, p)) != -1) {
+			int numStart = p - 1;
+			while (numStart > 0 && out[numStart - 1] != '\n' && out[numStart - 1] != '\r')
+				--numStart;
+			QByteArray numStr = out.mid(numStart, p - numStart).trimmed();
+			bool ok = false;
+			int objNum = numStr.toInt(&ok);
+			if (ok && objNum > 0)
+				offsets[objNum] = numStart;
+			++p;
+		}
+	}
+
+	if (offsets.isEmpty()) return;
+
+	int maxObj = offsets.lastKey();
+
+	QByteArray xref;
+	xref += "xref\n";
+	xref += "0 " + QByteArray::number(maxObj + 1) + "\n";
+	xref += "0000000000 65535 f \n";
+	for (int i = 1; i <= maxObj; ++i) {
+		if (offsets.contains(i)) {
+			xref += QByteArray::number(offsets[i]).rightJustified(10, '0')
+				+ " 00000 n \n";
+		} else {
+			xref += "0000000000 65535 f \n";
+		}
+	}
+
+	QByteArray trailer;
+	{
+		int tPos = data.indexOf("trailer", xrefStart);
+		if (tPos != -1) {
+			int tEnd = data.indexOf("%%EOF", tPos);
+			if (tEnd != -1) tEnd += 5;
+			if (tEnd != -1)
+				trailer = data.mid(tPos, tEnd - tPos);
+		}
+	}
+	if (trailer.isEmpty())
+		trailer = "trailer\n<<>>\n%%EOF";
+
+	QByteArray result;
+	result.reserve(out.size() + xref.size() + trailer.size() + 64);
+	result += out;
+	int newXrefOffset = out.size();
+	result += xref;
+	result += trailer;
+	result += "\nstartxref\n";
+	result += QByteArray::number(newXrefOffset);
+	result += "\n%%EOF\n";
+
+	QFile outF(pdfPath);
+	if (!outF.open(QIODevice::WriteOnly | QIODevice::Truncate)) return;
+	outF.write(result);
+	outF.close();
 }
 
 } // namespace PdfLinks
--- a/sources/pdf_links.cpp
+++ b/sources/pdf_links.cpp
@@ -502,7 +502,11 @@ void convertComponentInfoAnnotations(const QString &pdfPath,
 		int subTypePos = (subTypeInBody != -1 && subTypeInBody < aDictOpen)
 			? out.size() - (aDictOpen - subTypeInBody) : -1;
 		if (subTypePos != -1)
-			out.replace(subTypePos, 14, "/Subtype /Text");
+			// Acrobat draws its own sticky-note icon for /Subtype /Text
+			// whatever /AP says (verified on Reader 5.0). /Square with no
+			// /C and no /IC draws nothing anywhere, and is still a markup
+			// annotation, so the /Contents popup is unaffected.
+			out.replace(subTypePos, 14, "/Subtype /Square");
 
 		// Encode as UTF-16BE hex with BOM for proper Unicode support (Umlauten etc.).
 		// PDF spec: hex strings starting with FE FF are interpreted as UTF-16BE.
--- a/sources/pdf_links.cpp
+++ b/sources/pdf_links.cpp
@@ -32,7 +32,6 @@
 #include <QGraphicsTextItem>
 #include <QList>
 #include <QRegularExpression>
-#include <QTextCodec>
 #include <QUrl>
 #include <QVector>
 
@@ -399,8 +398,8 @@ void convertComponentInfoAnnotations(const QString &pdfPath,
 	QByteArray data = f.readAll();
 	f.close();
 
-	const QByteArray marker("http://componentinfo.local/");
-	if (!data.contains(marker)) return;
+	const QByteArray markerPrefix("http://componentinfo.local/");
+	if (!data.contains(markerPrefix)) return;
 
 	int xrefStart = data.lastIndexOf("\nxref\n");
 	if (xrefStart == -1) xrefStart = data.lastIndexOf("\nxref ");
@@ -433,20 +432,34 @@ void convertComponentInfoAnnotations(const QString &pdfPath,
 	out.reserve(data.size());
 
 	int pos = 0;
-	int annotIdx = 0;
+	bool anyConverted = false;
 
 	while (pos < body.size()) {
-		int markerPos = body.indexOf(marker, pos);
+		// Find next indexed marker: http://componentinfo.local/<N>
+		int markerPos = body.indexOf(markerPrefix, pos);
 		if (markerPos == -1) {
 			out.append(body.mid(pos));
 			break;
 		}
 
+		// Extract index from marker URL
+		int idxStart = markerPos + markerPrefix.size();
+		int idxEnd = idxStart;
+		while (idxEnd < body.size() && body[idxEnd] >= '0' && body[idxEnd] <= '9')
+			++idxEnd;
+		if (idxEnd == idxStart) {
+			// No index — skip malformed marker
+			out.append(body.mid(pos, markerPos + markerPrefix.size() - pos));
+			pos = markerPos + markerPrefix.size();
+			continue;
+		}
+		int annotIndex = body.mid(idxStart, idxEnd - idxStart).toInt();
+
 		int uriOpen = body.lastIndexOf("/URI (", markerPos);
 		int closeParen = (uriOpen != -1) ? body.indexOf(')', markerPos) : -1;
 		if (uriOpen == -1 || closeParen == -1 || uriOpen < pos) {
-			out.append(body.mid(pos, markerPos + marker.size() - pos));
-			pos = markerPos + marker.size();
+			out.append(body.mid(pos, idxEnd - pos));
+			pos = idxEnd;
 			continue;
 		}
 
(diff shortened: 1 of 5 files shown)


ispyisail:
Nice approach — keeping the data out of the visual output while making it machine-readable is the right call, and reusing the existing post-process pass fits how `pdf_links.cpp` already works. A few things from reading through it, one of which is a build breaker.

## Blocker: `#include <QTextCodec>` breaks the Qt6 build

- `CMakeLists.txt` supports `-DQT_VERSION_MAJOR=6` explicitly, and even special-cases `find_package(Qt6 COMPONENTS GuiPrivate)` — which this feature depends on for `QPdfEngine::drawHyperlink`.
- `Qt5Compat` / `Core5Compat` is not among the requested components, and `QTextCodec` moved into that module in Qt6.
- `QTextCodec` is used nowhere else under `sources/`.

The good news is the fix costs nothing: the function already contains a correct manual UTF-16BE fallback immediately below the codec branch. Dropping the codec path and always using the loop produces identical bytes, with no Qt6 breakage and one less dependency.

## Correctness of the rewriting pass

**The trailer's `/Size` is now understated.** The trailer is copied verbatim (`data.mid(tPos, tEnd - tPos)`), but this pass *appends a new object* (the shared Form XObject). `convertUriToGoTo()` copies the trailer the same way, so that pattern is inherited rather than introduced here — but it only ever rewrites objects in place, so its `/Size` stays correct. This is the first pass that adds one. Same story for the duplicate `startxref`: the copied trailer already contains the original one, and a second is appended after. Readers that take the last `startxref` and ignore `/Size` are fine, which is presumably why it tests clean, but stricter validators won't be.

**Contents are matched to annotations by scan position.** `annotations[annotIdx++]` is paired with the n-th `http://componentinfo.local/` found in the byte stream. If a single marker is ever missed, duplicated, or ordered differently from `m_componentInfoList`, every popup after that point silently shows a *different component's* data. That is a bad failure mode for information someone reads off a drawing, and it fails quietly. Putting the index straight in the marker (`http://componentinfo.local/<n>`) and looking it up would remove the whole class of problem for a handful of lines.

**`out.lastIndexOf("/Subtype /Link")` is unbounded.** Cross-reference links are also `/Subtype /Link`. The search walks back through everything accumulated so far, so if the nearest preceding match belongs to a cross-ref annotation, that one gets rewritten to `/Text` — breaking a working cross-reference *and* leaving the component-info annotation as a live external URL. It is also O(n) per annotation over a growing buffer, so O(n²) on a big project. Bounding the search to the enclosing `obj … endobj` addresses both.

**Malformed dict when `/A <<` is not found.** `copyEnd` then falls back to `sUriPos`/`uriOpen`, so `/Contents` and `/AP` are written *inside* the action dictionary while `/S /URI` survives and only the first `>>` is skipped. Unlikely against Qt's writer, but it emits a broken annotation silently; leaving that annotation untouched would be the safer branch.

## Smaller things

- The diff re-indents unrelated lines: the whole `convertUriToGoTo` doc block in `pdf_links.h` moves to column 0, and three existing slot declarations in `projectprintwindow.h` lose their indentation. Worth restoring to keep the diff to what actually changed.
- `dynamic_cast<Element*>` where the codebase generally uses `qgraphicsitem_cast`.
- `contents.toUtf8()` then `QString::fromUtf8(contents)` round-trips something that is already a `QString`.
- The description mentions removing `inject_annotations.py`, but that file is neither in this diff nor on master.

## Overall

The feature is worth having. The Qt6 include is the only thing I would call blocking, and it is a one-line deletion. The `/Size` bump and the indexed marker are both small and would make the pass much harder to get subtly wrong; I would also want the `/Subtype /Link` search bounded before this ships, since as written it can damage cross-reference links that currently work.

Kellermorph:
I've passed on your comments, and the AI says everything has been incorporated.

Kellermorph:
annotIndex was declared inside printDiagram() and reset to 0 for
each diagram page, while m_componentInfoList accumulated entries
across all pages. This caused page 2+ annotations to reference the
wrong ComponentInfo entry (e.g. page 2's first component got the
info from page 1's first component via URL index /0).

Moved annotIndex to requestPaint() and pass it by reference to
printDiagram() so it increments globally across all pages.

scorpio810:
Hi Kellermorph, thanks for pushing through ispyisail's review comments.

Looking at the current diff, the fixes look right on paper: the /Subtype /Link search is now bounded to the current annotation object, /Size is properly bumped, and the xref table is fully rebuilt from a fresh object scan rather than patched incrementally. That addresses the main blocking concern about cross-reference annotations potentially getting clobbered.

One thing I'd like confirmed before merging: convertComponentInfoAnnotations() runs before convertUriToGoTo(), and both do their own file read/parse/rewrite pass. Could you test a project that has both "Inscribe component information" enabled and page references / cross-reference links (#page=N) in the same export? I want to make sure the two passes chain cleanly on a file that's been rewritten in between, since we don't have automated tests covering this byte-level PDF surgery.

Context: I just confirmed in the forum that the existing XREF/GoTo link export works flawlessly even on very old PDF readers (Acrobat Reader 5.0 on Windows 98 SE) — https://qelectrotech.org/forum/viewtopic.php?pid=23217#p23217. That's a compatibility bar I'd like to keep, so I'd rather we catch any regression here than after merge.

If that combined test comes back clean, I'm fine moving forward with this.

ispyisail:
@scorpio810 I ran the combined test you asked for, and it does not come back clean — so I'm glad you asked for it before merge.

Short version: the two passes chain fine. The problem is entirely inside `convertComponentInfoAnnotations()`, and it only shows up when cross-reference links are present, which is exactly why it survived testing so far.

## Method

I couldn't drive the print dialog headlessly, so I tested the byte-level surgery directly:

- both functions lifted **verbatim** from this PR's head (`1780fde`), compiled standalone against Qt Core;
- input PDFs generated by real `QPdfEngine::drawHyperlink()` calls, so the annotation dictionaries are genuine Qt output, not hand-written;
- annotations emitted in QET's real order — `injectCrossRefLinks()` runs first, then the component-info loop, so each page is `[cross-refs][component infos]`;
- passes run in QET's order: `convertComponentInfoAnnotations()` then `convertUriToGoTo()`.

## Result

Page 1 with two cross-refs + two info annotations, page 2 with one of each:

```
obj  8  /Subtype /Text   cross-ref (GoTo)     *** WRONG ***
obj  9  /Subtype /Link   cross-ref (GoTo)     OK
obj 10  /Subtype /Link   info (Contents)      *** WRONG ***
obj 11  /Subtype /Text   info (Contents)      OK
obj 17  /Subtype /Text   cross-ref (GoTo)     *** WRONG ***   <- page 2
obj 18  /Subtype /Link   info (Contents)      *** WRONG ***   <- page 2
```

**On every page that has both features, the first cross-reference link is turned into `/Subtype /Text`** — it keeps its `GoTo` action but a `/Text` annotation is a sticky note, not a clickable link, so the jump is gone. **And the first component-info annotation is left as `/Subtype /Link`** with its `/A << /S /URI >>` already stripped, so it is an inert link rather than a popup.

Control run, same file, `convertUriToGoTo()` alone with no component info:

```
obj  8  /Subtype /Link   GoTo     OK
obj  9  /Subtype /Link   GoTo     OK
obj 18  /Subtype /Link   GoTo     OK
```

So the existing XREF/GoTo export is not at fault — it is untouched and still correct. That compatibility bar you verified on Acrobat 5.0 holds; it is this pass that removes the links before the reader ever sees them.

## Cause

The `/Subtype /Link` search is bounded, but to the wrong thing — to `pos`, the end of the *previously processed marker*, rather than to the annotation object being rewritten:

```cpp
int searchStart = out.size() - (aDictOpen - pos);
int subTypePos  = out.indexOf("/Subtype /Link", searchStart);
```

The chunk from `pos` to `aDictOpen` is everything since the last component-info annotation — which, given the emission order, contains that page's cross-reference annotations, and those are `/Subtype /Link` too. So `indexOf` finds the first cross-ref's subtype and rewrites that one instead.

Only the *first* info annotation on a page is affected, because after it `pos` has advanced past the intervening cross-refs — which is why partial testing looks fine.

## Fix

Anchoring the search to the enclosing object is enough:

```cpp
int objStart = body.lastIndexOf(" 0 obj", aDictOpen);
int subTypeInBody = (objStart != -1)
    ? body.indexOf("/Subtype /Link", objStart) : -1;
int subTypePos = (subTypeInBody != -1 && subTypeInBody < aDictOpen)
    ? out.size() - (aDictOpen - subTypeInBody) : -1;
if (subTypePos != -1)
    out.replace(subTypePos, 14, "/Subtype /Text");
```

With that patch applied to the same extracted code, the same two files come back fully correct:

```
obj  8  /Subtype /Link   cross-ref (GoTo)     OK
obj  9  /Subtype /Link   cross-ref (GoTo)     OK
obj 10  /Subtype /Text   info (Contents)      OK
obj 11  /Subtype /Text   info (Contents)      OK
obj 17  /Subtype /Link   cross-ref (GoTo)     OK
obj 18  /Subtype /Text   info (Contents)      OK
```

I also ran a variant with the two kinds fully interleaved; head fails 4 of 6 there, the patch passes 6 of 6.

## The rest of your list checks out

Verified on the rewritten files rather than by reading:

- `/Size` correctly bumped to `emptyXObjNum + 1`;
- xref fully rebuilt — every `n` entry offset lands exactly on its `N 0 obj`, **0 bad offsets**;
- exactly one `startxref` in the output, duplicate correctly dropped;
- poppler opens all outputs without complaint;
- `/Contents` paired to the right component via the indexed marker (`-K1`, `-K2`, `-Q3` each landed on the right object);
- UTF-16BE with BOM round-trips, umlauts included (`Müller & Söhne`, `Schütz`).

The indexed-marker change did its job — no mispairing at all, including across the page boundary that `1780fde` fixed.

## Caveat

These are synthetic PDFs with genuine Qt-emitted annotations, not a full export from the GUI. The emission order I relied on is read from `projectprintwindow.cpp` — `injectCrossRefLinks()` before the component-info loop. If a real export ever orders them differently the specific counts change, but the defect does not: any cross-ref annotation preceding a component-info marker since the last processed one will be hit.

Worth a real end-to-end export on a project with both features to confirm the counts, but I'd treat the `/Subtype` bound as needing fixing regardless.

Kellermorph:
Thanks for testing.

Fix /Subtype /Link replacement hitting cross-ref annotations

The /Subtype /Link search in convertComponentInfoAnnotations() was
anchored to pos (end of previous marker), which could reach into
cross-ref annotations sitting between markers on the same page.
This caused the first cross-ref per page to be wrongly converted
to /Subtype /Text (killing GoTo jumps) and the first component-info
annotation to be left as an inert /Subtype /Link.

Anchor the search to the enclosing PDF object (" 0 obj") instead,
so only the /Subtype /Link of the current annotation is replaced.

scorpio810:
Hi ispyisail,

Thanks for the thorough review on #739 — really useful catches. Kellermorph pushed two follow-up commits (f0348a7, 1780fde) that look like they address the structural points: the marker is now indexed (http://componentinfo.local/<N>) instead of relying on scan order, the /Subtype /Link replacement is bounded to the current annotation, the trailer /Size gets bumped, the duplicate startxref is stripped, and a missing /A << now leaves the annotation untouched instead of writing a malformed dict.

Could you take another pass when you have a moment, in particular to confirm the QTextCodec/Qt6 include got dropped? I didn't want to merge without your sign-off given how deep you went into the byte-level logic.

scorpio810:
Side note for context: I just tested PDF exports from QET (Qt6) with page references / XREF GoTo links on a genuinely ancient setup — Windows 98 SE, Acrobat Reader 5.0, period-correct PC — during a retrofit job today. Opened fine, links worked perfectly.

That's the existing convertUriToGoTo() mechanism though, not this PR's convertComponentInfoAnnotations() (/Subtype /Text with an empty /AP XObject and UTF-16BE /Contents). Both are old-enough PDF constructs (1.3/1.4) that I'd expect similar compatibility, but worth a real test on the same rig once this lands — I'll try to check it against my use case (PLC program in one window, drawings in another).

scorpio810:
Before this merges, it'd be good to get real confirmation that a PDF exported with both the page-reference/XREF links AND the new "Inscribe component information" checkbox enabled still opens correctly and isn't corrupted on an old PDF reader — the two post-processing passes (convertComponentInfoAnnotations then convertUriToGoTo) both rewrite the file at the byte level, so it's worth checking on something stricter than a modern viewer.

Does anyone here have a Windows 98 SE / Acrobat Reader 5.0 VM (or similarly old setup) and some time to test this before merge? Happy to share a sample export to try.

ispyisail:
@scorpio810 @Kellermorph — re-ran everything against `b8df298`. Taking your three points in order.

## 1. The `/Subtype` fix works

I rebuilt the harness with both functions lifted **verbatim from `b8df298`** (not my patched copy) and re-ran the same two files:

```
real QET ordering  [cross-refs][infos]      fully interleaved
obj  8  /Link  cross-ref (GoTo)   OK        obj  8  /Link  cross-ref (GoTo)   OK
obj  9  /Link  cross-ref (GoTo)   OK        obj  9  /Text  info (Contents)    OK
obj 10  /Text  info (Contents)    OK        obj 10  /Link  cross-ref (GoTo)   OK
obj 11  /Text  info (Contents)    OK        obj 11  /Text  info (Contents)    OK
obj 17  /Link  cross-ref (GoTo)   OK        obj 17  /Text  info (Contents)    OK
obj 18  /Text  info (Contents)    OK        obj 18  /Link  cross-ref (GoTo)   OK
```

6 of 6 in both orderings, where head previously failed 4 of 6 and 2 of 6. Structure still clean: 0 bad xref offsets, `/Size` correct, single `startxref`, poppler opens it, contents paired to the right component.

## 2. `QTextCodec` / Qt6 — confirmed, by compiling rather than grepping

Gone from all five changed files. But since a grep only proves absence of that one symbol, I went further: this box has Qt **6.10.2**, so I compiled the two rewritten functions against it directly.

- `pdf_links.cpp`'s `convertUriToGoTo()` + `convertComponentInfoAnnotations()` — **compile clean and link clean against Qt6Core**.
- Running the Qt6-linked binary over the same input produces output **byte-identical** to the Qt5 build (same SHA-256), so this is not just "it builds", it is "it behaves the same".
- Scanned the whole diff for the other usual Qt6 casualties — `QRegExp`, `QLinkedList`, `QStringRef`, `SkipEmptyParts`, `qVariantValue`, `QDesktopWidget`, `endl`, `QMatrix`, `toSet()`. None present.

So the blocking Qt6 concern from my first review is resolved.

**Unrelated finding while doing this, worth its own issue:** a full Qt6 configure of the tree fails today for reasons that have nothing to do with this PR:

```
tests/qttest/CMakeLists.txt:73 (target_link_libraries):
  Target "qt_unittests" links to: Qt::GuiPrivate  but the target was not found.
```

and with `-DPACKAGE_TESTS=OFF` it then dies on `INTERFACE_QT_MAJOR_VERSION of "Qt5::Core" does not agree with QT_MAJOR_VERSION` for `qelectrotech`. `Qt6GuiPrivateConfig.cmake` *is* present here, so this is the test subproject not doing its own `find_package`, plus a Qt5/Qt6 target mix somewhere in the dependency fetch. I did not chase it further — flagging it because it means nobody is getting an end-to-end Qt6 build of master right now, this PR or not.

## 3. Old-reader compatibility — what I could and could not establish

I do not have a Win98 SE / Acrobat 5.0 rig, so I cannot answer your actual question. What I can narrow down is *what to look for* on it.

The file declares `%PDF-1.4`, and every construct this pass introduces is within Acrobat 5's vocabulary: `/Subtype /Text` (1.0), `/AP` appearance streams (1.2), UTF-16BE hex strings with BOM (1.2), Form XObjects (1.2). Nothing here needs a modern reader.

The part I would actually watch is **the invisibility, not the integrity.** There is no `/F` flag on these annotations — invisibility rests entirely on the reader honouring the empty appearance stream:

```
29 0 obj
<< /Type /XObject /Subtype /Form /BBox [0 0 0 0] /Length 0 >>
stream
endstream
endobj
```

I tested that with poppler, with a control to make sure the test meant anything:

- converted file rendered against the pre-conversion baseline: **0 differing pixels** — genuinely invisible;
- same file with the `/AP` entries blanked out (same byte length, offsets preserved): **806 differing pixels** — poppler falls back to drawing the default sticky-note icon.

So the mechanism works, and it works *because* the reader honours `/AP`. A reader that ignores `/AP` for `/Text` annotations draws its own note icon instead — which is exactly what that 806-pixel control looks like. On the Acrobat 5 rig the failure mode to check for is therefore **not a corrupt file — it is a visible note icon sitting on top of every component on the drawing.**

Two small things if you want to harden it:

- `/F 2` (Hidden) is the spec-level way to guarantee it never renders. The trade-off is that some readers also drop Hidden annotations from the comment panel, which would defeat the feature — so it needs testing on the same rig rather than being applied blind.
- The Form XObject has no `/Resources`. Optional since PDF 1.2 and Acrobat is lenient, but it was *required* in 1.1 and earlier, so it is a cheap thing to add if very old readers are a target.

Happy to generate a sample export with both features enabled for whoever has the VM — the two failure modes to look for are a note icon appearing on the drawing, and a cross-reference jump that no longer responds.

## Where that leaves it

From my side the blocking items are done: the Qt6 include is gone and verified by compilation, and the `/Subtype` bound is fixed and verified against the head. I have no further blocking comments. The remaining question is yours — whether the invisible-annotation construct behaves on the old readers you need to support, and that genuinely needs the hardware.

ispyisail:
"Does anyone here have a Windows 98 SE / Acrobat Reader 5.0 VM" I'm building now. I'm going to want to call a favour on this one.

ispyisail:
@scorpio810 — you asked whether anyone had an old setup to test this on. I built one, and it answers your question, but not the way I expected.

**Adobe Acrobat Reader 5.0 running under Wine.** Adobe's own `ar500enu.exe` from their public FTP (via the Wayback Machine, pinned by SHA-256), extracted, licence dialog pre-accepted in the registry, driven headless under Xvfb. This is not an approximation: Reader 5.0 under Wine runs Adobe's own 2001 PDF parser and renderer. Only the Win32 API layer is emulated, and that layer has nothing to do with PDF interpretation. It is not Windows 98, but it *is* Acrobat 5.

## The good news, confirmed on the real reader

`b8df298` opens cleanly, and the cross-reference links work. I clicked the first one on page 1 and Acrobat navigated to page 2 — status bar reading `2 of 2`. So the `/Subtype` fix genuinely restores the jump on the reader you care about, and nothing about the file upsets a 2001 parser.

## The bad news

**Acrobat Reader 5.0 ignores the empty `/AP` and draws its own yellow sticky-note icon anyway.**

Measured against an annotation-free copy of the *same file*, page 1:

| build | Acrobat Reader 5.0 | poppler splash / cairo |
|---|---|---|
| `b8df298` (current head) | **1884 subpixels — icon drawn** | 0 — clean |
| `b8df298` + `/F 2` | **0 — clean** | 0 — clean |
| pre-fix head | 1951 subpixels — icon drawn | 878–960 — icon drawn |

Both poppler backends honour the empty appearance stream and draw nothing, which is why this never showed up in testing. Acrobat 5 does not. On the page-2 screenshot the icon sits right next to the text, and in a real export there would be **one per component**, on top of the drawing.

That is the thing I would want you to weigh before merging, because it lands precisely on your use case: PLC program in one window, schematic in the other, on the Win98 machine. A dense folio would come up covered in sticky-note icons.

## A fix, and its catch

Adding `/F 2` (Hidden) to the `/Text` annotations makes Acrobat 5 draw nothing at all — 0 differing subpixels against the baseline, verified, not theorised.

The catch is that the PDF spec defines Hidden as "do not render **or allow the user to interact with**" the annotation. So on readers that follow that strictly, `/F 2` may also put the component information out of reach — which is the whole feature. I have not tested that half. It may be that the honest position is: on Acrobat 5 you can have a clean drawing or readable annotations, but not both, while modern readers give you both. That is a design call, and it is yours.

## Tooling, if it is useful

Two scripts, both usable by anyone here:

- `pdf-oldreader-check.py` — a fast gate over any QET export: constructs an old reader cannot parse at all (`/Type /XRef`, `/Type /ObjStm`, `/Encrypt` — PDF 1.5 files do not degrade, they simply do not open), xref/`/Size`/trailer integrity, every `GoTo` still living on a `/Subtype /Link` (it catches the pre-fix regression), and a differential render to spot stray ink. It calibrates each renderer first by re-rendering with `/AP` stripped, because an engine that draws nothing even then cannot detect a stray icon — Ghostscript is exactly that case, and without the calibration it would have passed a file poppler visibly breaks.

- `acro5-render.sh` — sets up the Reader 5.0 prefix and renders or compares PDFs in it.

Worth being blunt about the limit: **the checker passes `b8df298` cleanly.** Every renderer it can reach honours the `/AP`, so it reports no ink. That is a false negative for Acrobat 5, and no number of modern engines would have caught it. Use it as the quick gate; use the Reader 5.0 rig as the authority. I can push both wherever you would like them, or just hand over the sample exports.

## Where I land

The two things I raised are fixed and verified on the real reader. The icon behaviour is new information rather than a defect in Kellermorph's work — the empty-`/AP` approach is entirely reasonable and works on every modern viewer. Whether it is acceptable that the annotations are visible on Acrobat 5, or whether `/F 2` and the loss of interaction there is the better trade, is a product decision for you rather than a review comment from me.

ispyisail:
Follow-up: there is a fix, and it is one word.

## It is not the appearance stream — it is the subtype

I assumed Acrobat 5 was rejecting the degenerate `/BBox [0 0 0 0]` and falling back to its default icon. That is wrong. I rebuilt the file with `/BBox [0 0 1 1]` and `[0 0 2 2]` and Acrobat drew the icon anyway, identically:

```
/BBox [0 0 0 0]   1884 subpixels — icon
/BBox [0 0 1 1]   1884 subpixels — icon
/BBox [0 0 2 2]   1884 subpixels — icon
```

Acrobat special-cases `/Subtype /Text`. A `/Text` annotation is *the* sticky note, and Acrobat draws its own icon for it whatever the appearance stream says. No `/AP` can suppress it.

So change the subtype. Both alternatives I tried are clean, measured against an annotation-free copy of the same file:

| annotation | Acrobat Reader 5.0 | poppler splash / cairo |
|---|---|---|
| `/Text` + empty `/AP` (current head) | **1884 — icon drawn** | 0 |
| `/Text` + `/F 2` (Hidden) | 0 — clean | 0 |
| **`/Square` + empty `/AP`** | **0 — clean** | 0 |
| `/Stamp` + empty `/AP` | 0 — clean | 0 |

## Why `/Square` rather than `/F 2`

`/F 2` works, but the spec defines Hidden as "do not render **or allow the user to interact with**" — it buys invisibility by disabling the annotation, which is the feature.

`/Square` needs no flag. A square annotation with no `/C` (border colour) and no `/IC` (interior colour) has nothing to draw: my checker reports every renderer as unable to draw anything for it *even with `/AP` removed entirely*. It is invisible by construction rather than by the reader's goodwill about an empty appearance stream. It is still a markup annotation, so it keeps its `/Contents` and stays a first-class comment.

I verified the payload survives the change untouched — same objects, same pairing, same UTF-16BE BOM:

```
head   obj 10 /Text    BMK: -K1     fix   obj 10 /Square  BMK: -K1
       obj 11 /Text    BMK: -K2           obj 11 /Square  BMK: -K2
       obj 18 /Text    BMK: -Q3           obj 18 /Square  BMK: -Q3
```

## The patch

In `convertComponentInfoAnnotations()`:

```cpp
-           out.replace(subTypePos, 14, "/Subtype /Text");
+           // Acrobat draws its own sticky-note icon for /Subtype /Text
+           // whatever /AP says (verified on Reader 5.0). /Square with no
+           // /C and no /IC draws nothing anywhere, and is still a markup
+           // annotation, so the /Contents popup is unaffected.
+           out.replace(subTypePos, 14, "/Subtype /Square");
```

The length change from 14 to 16 bytes is safe: the xref is rebuilt from a fresh object scan afterwards, and `subTypePos` is derived from the just-appended chunk each iteration rather than from any absolute offset.

Keeping the empty `/AP` alongside it is worth doing — it costs one shared object and stops any reader that *would* draw a default border for `/Square` from doing so.

## The one thing I could not test

Whether Okular and Adobe still list these in their comment panels once they are `/Square` rather than `/Text`. Both are markup annotations under the spec, and panels enumerate markup annotations by `/Contents`, so they should — but "should" is exactly the word that got us here, and I have neither reader on this machine.

That is a much smaller ask than my last one: open an export with the patch in Okular, and check the component information still appears in the reviews/comments panel. If it does, this is strictly better than the current head on every reader tested. If it does not, `/Stamp` is the fallback — also clean on Acrobat 5, though it depends on `/AP` being honoured rather than having nothing to draw.

@scorpio810 note this also removes the reason to care about `/F 2` and its interaction trade-off — that option was only interesting while the icon looked unavoidable.

Kellermorph:
@ispyisail Done — switched to /Subtype /Square per your recommendation.
Tested on my end: annotations still appear in the comments panel (Okular).
Thanks for the thorough testing and the Win98/Reader 5.0 setup, that caught something none of us would have found otherwise.

scorpio810:
@Kellermorph , @ispyisail ask me when you think it's ready to merge, thanks for your work.

ispyisail:
@scorpio810 — I think it is ready. Re-verified everything against `4572fca31`, not just the diff.

## Verified on this head

| check | result |
|---|---|
| cross-reference links, QET emission order | 3 of 3 intact — `/Subtype /Link` + `GoTo` |
| cross-reference links, fully interleaved | 3 of 3 intact |
| component info annotations | 3 of 3 converted to `/Subtype /Square` + `/Contents` |
| xref integrity | 29 objects, every in-use offset lands on its object, 0 bad |
| `/Size`, `startxref`, `/Root`, `%%EOF` | correct, single `startxref` |
| payload | `-K1`, `-K2`, `-Q3`, UTF-16BE BOM intact |
| Qt5 build | clean |
| Qt6 (6.10.2) compile | clean |

And on Acrobat Reader 5.0 itself, measured against an annotation-free copy of the same file:

```
4572fca31 vs baseline:  0 differing subpixels   — no icon drawn
cross-reference click:  navigates, status bar reads "2 of 2"
```

So on the reader that matters to you: the file opens, the drawing is clean, and the folio jumps work. That is the behaviour you had before this PR, plus the annotations.

Both of my blocking findings are closed, and closed on the real reader rather than on paper:

- the `/Subtype` bound (`b8df298`) — the first cross-reference link on every page is no longer clobbered;
- the sticky-note icon (`4572fca31`) — Acrobat special-cases `/Text` and draws its own icon whatever `/AP` says; `/Square` has no `/C` and no `/IC`, so there is nothing to draw.

@Kellermorph's Okular check closes the last gap I could not test myself — that the annotations still list in the comments panel as `/Square`. That was the one thing standing between "invisible" and "invisible and useless", so thank you for checking it.

## Two things to weigh, neither blocking

**Style nit.** The new call sits under a braceless `if` with the comment block between:

```cpp
if (subTypePos != -1)
	// four lines of comment
	out.replace(subTypePos, 14, "/Subtype /Square");
```

Correct as written — comments are not statements — but the next person to add a line there will silently put it outside the conditional. Braces would cost nothing.

**The empty `/AP` is now belt-and-braces.** `/Square` draws nothing on its own, so the shared Form XObject is no longer load-bearing. I would keep it: it costs one object and it protects against any reader that would otherwise draw a default border. Just worth knowing it is no longer the mechanism.

## What I did not test

All my testing used PDFs generated by real `QPdfEngine::drawHyperlink()` calls with the PR's own functions lifted verbatim — not an export driven through the print dialog. Kellermorph's Okular run covers that side. And Reader 5.0 ran under Wine rather than on Win98 hardware; the PDF engine is Adobe's own either way, but if you want a belt-and-braces check on the real rig before merging, a single export with both features enabled would do it — look for a note icon on the drawing and a cross-reference that does not respond.

With those caveats stated: **no further blocking comments from me, and I would be comfortable seeing this merge.**

PR #740, opened 2026-08-13
Export slave cross-reference labels to DXF

ispyisail:
Addresses part of what @scorpio810 asked for on #718 — cross-references in the DXF export, from the forum report at https://qelectrotech.org/forum/viewtopic.php?id=2481 ("les références croisées n'apparaissent pas dans le dxf").

### Cause

`ExportDialog::generateDxf()` walks the scene and collects items by cast. A slave element's cross-reference label — the `(6-G15)` pointing back to its master — is a plain **`QGraphicsTextItem`** created as a child of a `DynamicElementTextItem` (`dynamicelementtextitem.cpp:1375`). It is neither an `IndependentTextItem` nor a `DynamicElementTextItem`, so it matched no branch in the walker and was silently dropped.

Nothing was wrong with the label; it was simply never collected.

### Change

Collect it through the existing public `DynamicElementTextItem::slaveXrefItem()` accessor, and draw it with the same placement, rotation-accumulation and multi-line handling as the other text items — using `defaultTextColor()`, since a bare `QGraphicsTextItem` has no `DiagramTextItem::color()`.

### Measurements

`examples/industrial.qet`, all 50 folios, compared against the PDF export (which renders the whole scene, so it shows what *should* be there):

| | before | after | PDF |
|---|---|---|---|
| slave xrefs `(n-Xn)` | 0 | **41** | 41 |
| folio/position strings | 317 | 358 | 403 |

The slave cross-references now match the PDF exactly.

### What this does not cover

Two clarifications on the original request:

**Folio reports already export.** I checked before changing anything — the 317 folio/position strings present in the unmodified DXF are the folio-report texts. They are `DynamicElementTextItem`s and were already collected. So that half of the request appears to be already satisfied; if something specific is still missing there, a sample project would help.

**The master-side cross-reference table is still missing** — the remaining 403 − 358 = 45 strings. That is drawn by `CrossRefItem`, a `QGraphicsObject` that paints itself with custom `QPainter` code in three modes (`drawAsCross`, `drawAsContacts`, `drawAsPlcTable`) including contact symbols and rules. Giving it a faithful DXF representation means reproducing that drawing as DXF primitives, which is a substantially larger job than this and worth doing separately — possibly better solved generically with a DXF-writing paint device than per-item. Happy to take it on if you want it.

### DIFF
--- a/sources/exportdialog.cpp
+++ b/sources/exportdialog.cpp
@@ -472,6 +472,10 @@ void ExportDialog::generateDxf(
 	QList<Conductor *> list_conductors;
 	QList<DiagramTextItem *> list_texts;
 	QList<DiagramImageItem *> list_images;
+		//Slave cross-reference labels. They hang off a DynamicElementTextItem
+		//as plain QGraphicsTextItem children, so neither cast below picks them
+		//up and they were missing from the DXF entirely.
+	QList<QGraphicsTextItem *> list_xref_texts;
 	QList<QLineF *> list_lines;
 	QList<QRectF *> list_rectangles;
 	//QList<QRectF *> list_ellipses;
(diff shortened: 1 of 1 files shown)


plc-user:
This is what I get, when exporting to dxf:

grafik

The display errors are probably due to the chosen font, but that wasn't the issue here at all!

The cross-references are carried over into the DXF file!

scorpio810:
Folio Reports are represented by Contacts when exporting the diagram to DXF


Screenshot
Screenshot-1

scorpio810:
Hum, maybe old report folio
Screenshot-2

Screenshot-3

PR #741, opened 2026-08-13
Extend terminal numbering dialog with letter numbering and strip selection

Kellermorph:
This PR extends the automatic terminal numbering dialog with two new features:

"Also number letters" option - When alphanumeric numbering is selected, a new checkbox becomes available. When enabled, skipped numbers are appended to letter-based suffixes (e.g., X1:N4, X1:PE5 instead of X1:N, X1:PE).
Terminal strip selection - All unique terminal strip prefixes (the part before :) are now collected and displayed as checkboxes. Users can exclude specific strips from numbering by unchecking them.
Changes
UI (terminalnumberingdialog.ui):

Added "Numéroter également les lettres" checkbox (indented under the alphanumeric radio button)
Added "Borniers" group box on the right side with scrollable checkboxes for each terminal strip
Restructured layout from vertical to horizontal (left: axis priority + numbering type, right: terminal strips)
Dialog size adjusted to 500×450
Logic (terminalnumberingdialog.h/cpp):

Constructor now accepts QETProject* to populate terminal strip checkboxes on dialog open
Added alsoNumberLetters() and excludedStrips() methods
Numbering logic extended: when "also number letters" is enabled, counter is appended to alphabetical suffixes
Excluded terminal strips are skipped during numbering
Integration (qetdiagrameditor.cpp):

Updated slot_terminalNumbering() to pass project to dialog constructor
Behavior
Before:

X1:1, X1:2, X1:3, X1:N, X1:PE → X1:1, X1:2, X1:3, X1:N, X1:PE (N and PE keep original, numbers 4 and 5 are lost)
After (with "also number letters" enabled):

X1:1, X1:2, X1:3, X1:N, X1:PE → X1:1, X1:2, X1:3, X1:N4, X1:PE5

### DIFF
--- a/sources/qetdiagrameditor.cpp
+++ b/sources/qetdiagrameditor.cpp
@@ -2799,11 +2799,11 @@ void QETDiagramEditor::generateTerminalBlock()
  * Opens the dialog for automatic terminal numbering and applies the generated undo command.
  */
 void QETDiagramEditor::slot_terminalNumbering() {
-	TerminalNumberingDialog dialog(this);
-	if (dialog.exec() == QDialog::Accepted) {
-		QETProject *project = currentProject();
-		if (!project) return;
+	QETProject *project = currentProject();
+	if (!project) return;
 
+	TerminalNumberingDialog dialog(this, project);
+	if (dialog.exec() == QDialog::Accepted) {
 		// Fetch the generated undo command from the dialog logic
 		QUndoCommand *macro = dialog.getUndoCommand(project);
 
--- a/sources/ui/terminalnumberingdialog.cpp
+++ b/sources/ui/terminalnumberingdialog.cpp
@@ -5,18 +5,57 @@
 #include "../qetgraphicsitem/element.h"
 #include "../undocommand/changeelementinformationcommand.h"
 #include <QUndoCommand>
+#include <QCheckBox>
+#include <QVBoxLayout>
+#include <QSet>
 #include <algorithm>
 
 /**
  * @brief TerminalNumberingDialog::TerminalNumberingDialog
  * Constructor
  * @param parent
+ * @param project Pointer to the current project (used to populate terminal strips)
  */
-TerminalNumberingDialog::TerminalNumberingDialog(QWidget *parent) :
-QDialog(parent),
-ui(new Ui::TerminalNumberingDialog)
+TerminalNumberingDialog::TerminalNumberingDialog(QWidget *parent, QETProject *project) :
+    QDialog(parent),
+    ui(new Ui::TerminalNumberingDialog),
+    m_project(project)
 {
     ui->setupUi(this);
+
+    // Connect radio button to enable/disable the "also number letters" checkbox
+    connect(ui->rb_type_alpha, &QRadioButton::toggled, ui->cb_also_alpha, &QCheckBox::setEnabled);
+
+    // Collect all unique terminal strip prefixes from the project
+    if (m_project) {
+        QSet<QString> prefixes;
+        foreach (Diagram *diagram, m_project->diagrams()) {
+            foreach (QGraphicsItem *qgi, diagram->items()) {
+                if (Element *elmt = qgraphicsitem_cast<Element *>(qgi)) {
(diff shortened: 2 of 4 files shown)


ispyisail:
Useful pair of additions — being able to exclude individual strips is something I can see people wanting straight away. One defect in the "also number letters" rule though, plus a couple of smaller inconsistencies.

## The new numbering is not idempotent, and it damages alphanumeric suffixes

I simulated the label logic in `getUndoCommand()` exactly. Re-running the numbering — a normal thing to do after adding a few terminals — appends the counter again on every pass:

```
run 1: X1:N4      run 2: X1:N44      run 3: X1:N444
```

The reason is that on the second pass the suffix is now `N4`, which `toInt()` still rejects, so it takes the alphabetic branch a second time. The worked example in the PR description degrades the same way: `X1:N4, X1:PE5` becomes `X1:N44, X1:PE55`.

The same branch also mangles suffixes that are already meaningfully alphanumeric. Phase terminals, single run:

```
X1:L1 -> X1:L14      X1:L2 -> X1:L25      X1:L3 -> X1:L36
```

Both fall out of the same condition — treating "not a pure number" as "is a letter". Restricting the append to suffixes that are *entirely* alphabetic fixes both at once:

```cpp
} else {
    // Only append to purely alphabetic suffixes (N, PE), so re-running is a
    // no-op and already-numbered suffixes (L1, L2, L3) keep their meaning.
    const bool all_letters = !ti.suffix.isEmpty()
            && std::all_of(ti.suffix.cbegin(), ti.suffix.cend(),
                           [](QChar c){ return c.isLetter(); });
    if (alsoAlpha && all_letters)
        newLabel = ti.prefix + ":" + ti.suffix + QString::number(newNum);
    else
        newLabel = ti.prefix + ":" + ti.suffix;
}
```

which gives:

```
run 1: X1:N4    run 2: X1:N4    run 3: X1:N4      (stable)
X1:L1 -> X1:L1      X1:PE -> X1:PE5               (phases preserved)
```

## Smaller inconsistencies between the two passes

**Colon-less labels get no checkbox.** The constructor only records a prefix when `colonIndex != -1`, but `getUndoCommand()` treats a label with no colon as `prefix = label` (with an empty suffix). So a terminal labelled just `N` never appears in the strip list, can never be excluded, and is silently rewritten to `N:1`. Worth splitting the label the same way in both places.

**`auto_num_locked` is not honoured when building the list.** `getUndoCommand()` skips locked terminals, the constructor does not — so a strip whose terminals are all locked still shows a checkbox that cannot affect anything.

**`m_project`** is kept as a member but only ever used inside the constructor; a local would do.

Minor: dropping the `parent = nullptr` default on the constructor is fine with only one caller, but `(QWidget *parent = nullptr, QETProject *project = nullptr)` would keep it a drop-in.

The enable/disable wiring is correct — `cb_also_alpha` is `enabled=false` in the `.ui` with `rb_type_num` pre-checked, so connecting `toggled → setEnabled` is sufficient and there's no stale-state problem on open. Both passes use `lastIndexOf(':')`, so the prefix used for exclusion matches the prefix used for counting. Excluded strips are skipped before entering `termList`, so they don't perturb the counters. The undo macro is built and discarded correctly when empty.

Caveat on my end: this is a code review plus a faithful simulation of the string handling — I have not built the branch and driven the dialog, so I have not confirmed the integration end to end.

Kellermorph:
Fixed issues:
Idempotent numbering - "Also number letters" now only appends to purely alphabetic suffixes (N, PE). Re-running the numbering no longer duplicates numbers (X1:N4 stays X1:N4, not X1:N44).
Phase terminals preserved - Suffixes like L1, L2, L3 are no longer modified since they contain digits. Only pure letters (N, PE, etc.) get the counter appended.
Colon-less labels - Terminals without a colon (e.g. just "N") now appear in the strip checkbox list.
auto_num_locked honored - Locked terminals are excluded from the strip checkbox list in the constructor.
Member variable removed - m_project is no longer stored as a member; only used locally in constructor.
Default parameters - Constructor signature restored to (QWidget *parent = nullptr, QETProject *project = nullptr).

PR #748, opened 2026-08-14
Fix slave contact groups being trimmed to one when reopening element properties

Kellermorph:
https://qelectrotech.org/bugtracker/view.php?id=339

Summary: When editing a Master element's properties, enabling the "Define slave elements" checkbox without first checking the "Define max slaves" checkbox caused the slave contact groups list to be trimmed to a single entry upon reopening the dialog.

The root cause was in on_m_slave_groups_checkbox_toggled: when toggling the slave groups checkbox on, it forced the max_slaves checkbox on but did not update the spinbox value. The spinbox retained its default value of 1 (from the UI minimum), so populateSlaveGroupsTable() would trim all loaded groups down to one.

The fix sets the spinbox value to the current number of existing groups when the slave groups checkbox forces the max_slaves checkbox on.


IBSYSLevi:
Thanks @Kellermorph , is this then obsolete: #688  ?

Kellermorph:
> Thanks @Kellermorph , is this then obsolete: #688 ?

i dont know exactly but i dont think so

Issue #201, opened 2021-11-26
snap: Update stable release to core20 & introduce branch for stable releases


ppd:
Now that https://github.com/snapcore/snapcraft/pull/3596 has landed, we should do the following things:

- ~[ ] Branch off 0.8 to be able to update the snap packaging for the stable release~
- [x] Port the stable release to core20. Master has the recipe already.
- [x] Change the user visible version in the store to 0.8 without the git suffix
- [x] Introduce a workflow for rebuilding the stable track if security updates require it

ppd:
@scorpio810 Maybe the cleanest way to manage the snaps would be to have a separate snap repo "qelectrotech-snap". There we would copy the packaging for the stable release. The build files for edge stay in the main repo.

ppd:
snapcore/snapcraft#3596 has been released to stable in https://github.com/snapcore/snapcraft/releases/tag/6.0.1

I'll port the stable release soon.

scorpio810:
Great, thanks.

ppd:
I'm putting the recipes & workflows for stable snaps here: https://github.com/qelectrotech/qelectrotech-snap

ppd:
@scorpio810 Can you maybe try 0.8.0 from the candidate channel? `snap refresh --candidate qelectrotech`. It's 0.8.0 built with the current recipe from master, including core20, newer Qt, updated dependencies etc.
For me, it looks good. I'll promote it to stable if I get one or two positive feedbacks.

ppd:
A branch for publishing stable release updates (first to candidate, then we release them to stable by hand): https://github.com/qelectrotech/qelectrotech-source-mirror/tree/snap/latest/candidate

The workflow run for a 0.8.0 rebuild on core20: https://github.com/qelectrotech/qelectrotech-source-mirror/actions/runs/1722623821

ppd:
Released revision 1452 to stable.

PR #751, opened 2026-08-15
Compiled translation file proposal

arummler:
Removing the qm translation files as by default they are rebuild by CMake anyhow. Changing the output directory which pointed to the source directory. That avoids polluting the source directory during out-of-source builds and it is standard that these files should be located inside the build directory during the build step.

Alternative: do not build the qm files by default during build but rely on the checked in files. That's also easy to do.


scorpio810:
No, I use for some of old packagings with qmake and I us thise .qm files yet.

arummler:
And they get picked up hard coded from the source directory? Can we re-visit when qmake has been dropped completely? It is easier not force paths manually in cmake but to go with the presets.

scorpio810:
> No, I use for some of old packagings with qmake and I us thise .qm files yet.

Ok, now for me with Qt6 only packaging.

ispyisail:
I think the blocker on this may have lifted yesterday without anyone noticing.

@scorpio810 objected in August that the checked-in `.qm` files were still needed for old qmake packaging, and @arummler asked whether it could be revisited once qmake was dropped. Yesterday you wrote:

> Ok, now for me with Qt6 only packaging.

If that means the old qmake packaging is retired, this PR is unblocked.

## Concrete evidence that the problem is real

I hit it today, working on something unrelated:

```
error: Your local changes to the following files would be overwritten by checkout:
	lang/qet_tr.qm
Please commit your changes or stash them before you switch branches.
```

I had to `git checkout -- lang/` before I could change branch. Building QET modifies tracked files, so an ordinary build blocks checkouts and leaves noise in every `git status`. That is exactly what the one-line change here fixes: `OUTPUT_LOCATION "${QET_DIR}/lang"` writes generated files back into the source tree, `OUTPUT_LOCATION "lang"` keeps them in the build directory where generated files belong.

## The conflict makes the case for the PR

It is currently `mergeable: false`, 195 commits behind, and **every conflicting file is a `.qm`**:

```
lang/qet_de.qm  lang/qet_en.qm  lang/qet_fr.qm  lang/qet_hu.qm
lang/qet_nl_BE.qm  lang/qet_pt_BR.qm  lang/qet_zh.qm
```

Seven binary build artefacts conflicting in git is the argument for not tracking them, stated by git itself. The resolution is trivial — they are deleted by this PR — but it will keep re-conflicting for as long as it waits, because master keeps regenerating them.

## The question that settles it

@scorpio810 — does "now for me with Qt6 only packaging" mean the checked-in `.qm` files are no longer consumed by anything downstream?

- **If yes**, this can land after a rebase. It is one line plus 33 deletions of files the build recreates anyway.
- **If something still reads them from the source tree**, @arummler's own alternative is the safer variant and he offered it in the description: keep the files checked in, just stop rebuilding them by default. That fixes the dirty-worktree problem without removing anything.

Either way this is a small cleanup with a clear answer, and it is currently costing more in repeated conflicts than it would to decide.

scorpio810:
Thanks @arummler @ispyisail , merged. For the record, here is what came up after the merge; the fixes are in the commits referenced above.

1. Packaging still copied .qm from the source tree. The Windows workflow did cp -r lang plus a find build -name "*.qm" … || true, so a missing translation would have shipped silently. Now it copies from build/lang/ and fails if any .qm is missing. Under MSYS2, $GITHUB_WORKSPACE is a Windows path (D:\a\...), so the check uses find instead of shell globs, whose backslashes broke compgen -G. The macOS script (misc/MacQetDeploy_arm64_cmake.sh) got the same fix.

2. Korean was lost. qet_ko.qm was tracked, but qet_ko.ts had never been listed in TS_FILES, so nothing regenerated it once the tracked file was gone. It was added to TS_FILES in cmake/qet_compilation_vars.cmake. The checks now compare the generated .qm against TS_FILES and fail with the missing language names. A .ts present in lang/ but not listed there only raises a warning.

qet_nl_NL.ts is deliberately left out of TS_FILES: it is fully untranslated, and since langFromSetting() keeps the full locale, an empty qet_nl_NL.qm would take precedence over qet_nl.qm for users with an nl_NL locale and show English instead.

3. macOS: Homebrew @rpath dependencies. Unrelated to this PR, but it surfaced once the script checked macdeployqt's output: recent Homebrew bottles (brotli, webp, sharpyuv) reference their dependencies as @rpath/libX.dylib, which macdeployqt cannot resolve (-libpath doesn't help), and libbrotlicommon kept an absolute /opt/homebrew/... install id. The script now rewrites any remaining /opt/homebrew reference to @rpath, copies missing libraries into Contents/Frameworks, and aborts if anything still points outside the bundle.

Windows Build #201 on master is green with 33 .qm, and the macOS DMG built and notarized fine.

arummler:
I think some off the self implemented packaging scripts and manual cmake parts could be made obsolete by using more idiomatic CMake/Qt approaches. E.g. the the translations can be simply embedded as a ressource into the binary as described here:
https://doc.qt.io/qt-6/cmake-get-started.html

Something like
find_package(Qt6 REQUIRED COMPONENTS Core LinguistTools)
set(I18N_TRANSLATED_LANGUAGES de fr es ...
qt_standard_project_setup()
qt_add_executable(my_app main.cpp)
target_link_libraries(my_app PRIVATE Qt6::Widgets ...
qt_add_translations(my_app)

They should be then available under ":/i18n".

scorpio810:
Agreed in principle. Today's work shows why: after #751 I had to add ~40 lines of shell in both `windows-build.yml` (1f647fa, a45a984) and `MacQetDeploy_arm64_cmake.sh` (c4d2f0c) just to copy and verify `.qm` files, and `qet_ko` still slipped through (7d4cbc3). Embedding would remove those blocks and the `lang/` install step on every platform.

**Measured cost:** the 33 generated `.qm` files total 7.0 MB before rcc compression, so embedding is cheap.

**Bug found on the way:** six of them (`fi`, `no`, `rs`, `sk`, `sl`, `sr`) come from 0%-translated `.ts` files. An empty `.qm` loads "successfully", so `setLanguage()` never fell back to `qet_en` and those users got the French source strings. 99e9ce5acabafbae827485686895e50a7a7ef03d fix that treats an empty translator as not loaded, and logs the QET and Qt `.qm` actually loaded in the startup diagnostics. This also makes globbing `lang/qet_*.ts` safe: e.g. `qet_nl_NL.ts` (0%) would now fall back to `qet_nl` instead of shadowing it.

A few QET specifics for the CMake side:

- **Naming / source language:** files are `lang/qet_XX.ts`, the project is `qelectrotech` and the source language is **French**. Automatic naming via `I18N_TRANSLATED_LANGUAGES` would create `qelectrotech_XX.ts` in the source root, and the default plurals-only file for `en` would clash with our full `qet_en.ts`. So: explicit `TS_FILES` (or a `file(GLOB ... CONFIGURE_DEPENDS)`) plus `NO_GENERATE_PLURALS_TS_FILE`.
- **lupdate:** please pass `SOURCE_TARGETS qelectrotech` explicitly. Since 6.7 the first argument isn't used for extraction, and auto-collection would also pick up the test targets.
- **Runtime:** `QETApp::setLanguage()` needs `:/i18n` as a source. Keep `--lang-dir` first so translators can test a fresh `.qm` without rebuilding (the MSI launcher also passes it).
- **Qt's own catalogs** (`qtbase_XX`) are separate. This morning I made Windows (4b675dd) and macOS (e982778, 2620d34, bc06f7d) bundle them as `lang/qt_XX.qm`. `MERGE_QT_TRANSLATIONS` could replace that, but it needs Qt 6.9 (Debian trixie ships 6.8.2) and qtbase only has regional `pt_PT`/`zh_CN`. So that part stays as is for now, or goes behind a version check.
- **Qt floor:** this means requiring Qt ≥ 6.7 and dropping the Qt5 / 6.0–6.6 branches in CMake.

scorpio810:
PFiouuu....I never would have thought that this small change would take me so long....

PR #752, opened 2026-08-15
Take the modal dialog out of RotateTextsCommand's constructor

ispyisail:
`RotateTextsCommand` called `QDialog::exec()` from **inside its constructor**, so the command could not be built without a human answering a dialog.

That made it untestable headlessly and undrivable from any script or test harness — and it is why my own #312 fix (PR #707) shipped with its save/reload round-trip unverified: the symptom could not be reproduced without a GUI, and driving the GUI for it did not work.

### What changed

The command now takes the angle as a parameter and asks nothing. Two statics carry the interactive half:

```cpp
static bool hasSelectedTexts(Diagram *diagram);   // anything to rotate?
static bool askRotation(qreal &rotation);         // dialog; false if cancelled
```

The single call site in `QETDiagramEditor` asks first, then builds the command.

`askRotation()` deliberately stays in this class rather than moving to the caller: it keeps the `QObject` `tr()` context, so existing translations of *"Orienter les textes sélectionnés"* are not invalidated.

### User-visible behaviour is unchanged

Same dialog, same title, same widget, and still no dialog when the selection is empty. Verified by running the built binary under Xvfb: select all → <kbd>Ctrl</kbd>+<kbd>Space</kbd> → the orientation dialog opens as before.

### Also fixes a latent null dereference

When nothing is selected, the constructor calls `setObsolete(true)` without ever creating `m_anim_group`. `QUndoStack::push()` calls `redo()` **before** it discards an obsolete command, which dereferences a null pointer. Unreachable today only because the action is disabled on an empty selection. `undo()`/`redo()` now guard.

### Verification — bugtracker #312 reproduced and fixed under test, for the first time

With the command finally constructible headlessly, I drove it through a scratch scripted-editing hook (not part of this PR) against `examples/741.qet` — 67 conductors, single folio, `rotation` attributes stripped first — and counted what got written on save:

| build | conductors with `rotation` | with `userx` |
|---|---|---|
| PR #707 applied (`forceRotateByUser`) | **67** | 0 |
| PR #707 reverted (`forceMovedByUser`) | **0** | **67** |

The reverted build writes no rotation at all — lost on reload, exactly bugtracker #312. The fixed build writes all 67.

It also surfaced a **second symptom of that bug** that had gone unnoticed: because the buggy code called `forceMovedByUser(true)`, it did not merely drop the rotation, it spuriously marked every text as manually positioned, writing `userx`/`usery`. Projects saved by an affected build carry position pins the user never set, and #707 does not remove them. Flagging it here rather than acting on it — I have not assessed whether that warrants a migration.

### Note for scripted callers

The rotation *value* is applied through `QPropertyAnimation`, which needs a spinning event loop. Headless, the flag is set and the attribute written, but the value stays at its pre-animation figure. Fine for the save gate; a scripted caller wanting the angle actually applied would need the animation completed or bypassed. Not addressed here.

### Testing

- Clean build, no new warnings.
- Interactive path verified under Xvfb as described above.
- Headless A/B as tabulated above.

### DIFF
--- a/sources/qetdiagrameditor.cpp
+++ b/sources/qetdiagrameditor.cpp
@@ -1631,7 +1631,18 @@ void QETDiagramEditor::selectionGroupTriggered(QAction *action)
 			diagram->undoStack().push(c);
 	}
 	else if (value == "rotate_selected_text")
-		diagram->undoStack().push(new RotateTextsCommand(diagram));
+	{
+			//Ask for the angle first, then build the command: the command
+			//itself no longer opens a dialog. Guarding on the selection keeps
+			//the previous behaviour of showing no dialog when there is
+			//nothing to rotate.
+		if (RotateTextsCommand::hasSelectedTexts(diagram))
+		{
+			qreal rotation = 0;
+			if (RotateTextsCommand::askRotation(rotation))
+				diagram->undoStack().push(new RotateTextsCommand(diagram, rotation));
+		}
+	}
 	else if (value == "find_selected_element" && currentElement())
 		findElementInPanel(currentElement()->location());
 	else if (value == "edit_selected_element")
--- a/sources/undocommand/rotatetextscommand.cpp
+++ b/sources/undocommand/rotatetextscommand.cpp
@@ -26,14 +26,33 @@
 #include "../qtextorientationspinboxwidget.h"
 
 /**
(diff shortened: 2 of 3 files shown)


ispyisail:

PR #753, opened 2026-08-15
Add a non-interactive mode so headless runs cannot block on a modal

ispyisail:
A modal dialog opened during a headless run has nobody to answer it, so the process blocks forever.

This is a recurring **class** of bug rather than a one-off. Two of my own open PRs each fix a single instance of it:

- #661 — command-line tools hanging on a modal message box
- #737 — headless CLI hanging on version-incompatible projects

and the next prompt someone adds to a load path reintroduces it. **This PR is intended to supersede both** — I am happy to close them if you prefer this approach.

### What it does

Adds `QET::setInteractive()` / `QET::isInteractive()`, mirroring the existing `QETProject::setBackupEnabled()` pattern, set `false` from `main.cpp`'s command-line branch. The four `QET::QetMessageBox` helpers consult it: when non-interactive they report the question on stderr and return the caller's `defaultButton` instead of opening a dialog.

That covers the **59 existing `QET::QetMessageBox` call sites at once**, including both version-compatibility prompts in `QETProject::readProjectXml()` that #737 addresses individually. It also routes the one raw `QMessageBox::warning` on the element-load path through the wrapper.

### The default-answer rule

A caller that passes no default gets `QMessageBox::NoButton`, deliberately.

Nearly every call site is written as *"if the user chose the abort option, abort"* — so `NoButton` means the operation proceeds unchanged, which is the safe direction and needs no per-site guesswork about what to auto-answer. The two `readProjectXml` prompts are exactly this shape (`if (ret == QMessageBox::Cancel) discard`), so they now proceed as Open, matching #737's intent without a bespoke fix.

Call sites wanting a specific unattended answer pass an explicit default, as several already do.

### Verification

`examples/schema_indus.qet` is version 0.3 and previously hung **every** CLI verb forever:

| verb | before | after |
|---|---|---|
| `--info` | HUNG (60 s timeout) | ok |
| `--resave` | HUNG | ok |
| `--export-bom` / `-nets` / `-links` | HUNG | ok |
| `--export-pdf` / `-svg` / `-png` | HUNG | ok |

with the suppressed prompt reported on stderr:

```
warning (headless, no user to ask): Avertissement -- Le projet que vous tentez
d'ouvrir est partiellement compatible avec votre version 0.100.1 ...
```

Controls: a normal project (`examples/741.qet`) round-trips with its conductor count unchanged, and the GUI still shows its dialogs — interactive mode is the default and only the CLI branch turns it off. Verified under Xvfb.

### Not covered, on purpose

This handles **confirmation** dialogs, which have a sensible unattended answer. It does not help **parameter-gathering** dialogs — ones whose whole purpose is to obtain a value with no default. Suppressing those just yields "cancelled". Those need the parameter lifted into the API instead; #752 does that for `RotateTextsCommand`. There are very few of them.

### DIFF
--- a/sources/main.cpp
+++ b/sources/main.cpp
@@ -127,6 +127,11 @@ QGuiApplication::setHighDpiScaleFactorRoundingPolicy(QetSettings::hdpiScaleFacto
 			// runs on a background thread referencing the project and races the
 			// process exit (intermittent segfault in QET::writeToFile).
 			QETProject::setBackupEnabled(false);
+			// Nobody is watching a headless run, so a modal prompt has no way
+			// to be answered and blocks the process forever. Suppress them:
+			// QET::QetMessageBox reports the question on stderr and returns
+			// the caller's default instead of opening a dialog.
+			QET::setInteractive(false);
 			return CLIExport::run(export_app.arguments());
 		}
 	}
--- a/sources/qet.cpp
+++ b/sources/qet.cpp
@@ -713,6 +713,30 @@ bool QET::eachStrIsEqual(const QStringList &qsl) {
 	@param c QetCollection value to convert
 	@return The QetCollection enum value converted to a QString
 */
+namespace {
+		/// @see QET::setInteractive
+	bool interactive_mode = true;
+}
+
+/**
+	@brief QET::setInteractive
+	@param interactive : false to suppress every blocking user prompt.
+	Set once, early, from the headless command-line path.
+*/
+void QET::setInteractive(bool interactive)
+{
+	interactive_mode = interactive;
+}
+
+/**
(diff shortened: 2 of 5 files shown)


ispyisail:
Closing as a duplicate of #661, which I opened on 4 August and which already does exactly this: the same `main.cpp` hook beside `setBackupEnabled(false)`, the same non-interactive guard on all four `QET::QetMessageBox` helpers, the same stderr reporting.

#661 also handles the unattended answer better than this PR did — it prefers an affirmative button (Ok/Open/Yes/…) so a batch run completes, where this one returned `NoButton` and relied on call sites treating that as "carry on".

My apologies for the noise. #661 is the one to review.

Issue #758, opened 2026-08-17
Ctrl+G (Atteindre un élément) bypasses ShortcutManager: not listed, not rebindable, not conflict-checked


ispyisail:
`Édition > Atteindre un élément` is bound to `Ctrl+G`, but the binding is
hardcoded rather than registered with `ShortcutManager`, so it does not appear
in *Préférences → Raccourcis*.

`sources/qetdiagrameditor.cpp:797`:

```cpp
m_jump_to_element = new QAction(tr("Atteindre un élément"), this);
m_jump_to_element->setShortcut(Qt::CTRL | Qt::Key_G);
```

Every other shortcut in the application goes through
`ShortcutManager::instance().registerAction(...)`, which is what makes it
listed, rebindable and included in conflict detection.

### Consequences

1. **The user cannot see or change it.** `Ctrl+G` is not in the Shortcuts page,
   so it cannot be discovered there or rebound.
2. **Conflict detection cannot see it either.** If a user assigns `Ctrl+G` to
   some other action in the Shortcuts page, the page reports no conflict —
   `checkConflicts()` only compares registered entries — and the two bindings
   then collide silently at runtime.

`Ctrl+G` is currently free among the 95 registered shortcuts, so nothing clashes
today. The failure only appears once a user chooses it, which makes it easy to
miss and confusing when it happens.

### Scope — this is the only instance

The whole source tree contains exactly two `setShortcut` calls:

```
sources/qetdiagrameditor.cpp:798   m_jump_to_element->setShortcut(Qt::CTRL | Qt::Key_G);
sources/qetdiagrameditor.cpp:2053  act->setShortcut(QKeySequence());     // clears, not sets
```

The second clears a shortcut, so `m_jump_to_element` is the single action in
QElectroTech that gets a key outside the registry.

### How it was found

By dumping every `QAction` from the running application (menu path, shortcut,
owning window) and cross-checking against a static scan of the source. Of 98
actions carrying a shortcut at runtime: 95 matched a `registerAction` call, 2
were Qt built-ins (`Shift+F1`, *"What's This?"*), and this was the only one that
was neither.

### Suggested fix

Replace the `setShortcut` call with a registration, matching the surrounding
style:

```cpp
m_jump_to_element = new QAction(tr("Atteindre un élément"), this);
ShortcutManager::instance().registerAction(
    m_jump_to_element, "diagrameditor.jump_to_element",
[...]

PR #759, opened 2026-08-17
Scope shortcut conflict detection to the category (fixes #757)

### LINKED ISSUE #757: Shortcuts config page reports 60 of its own 95 default bindings as conflicts
The Shortcuts configuration page marks most of its own default bindings as
conflicts. On master `7307a59c1`, **60 of the 95 registered shortcuts** are
highlighted red with a *"Ce raccourci est aussi utilisé par : …"* tooltip —
for bindings that are correct.

### Cause

`ShortcutManager` ids are per-window, and QElectroTech deliberately registers
the same key once per editor. For example `Ctrl+Shift+S`:

```
elementeditor.save_as_file      "Enregistrer dans un fichier"
diagrameditor.save_file_as      "Enregistrer sous"
titleblockeditor.save_as_file   "Enregistrer vers un fichier"
```

`Undo`, `Redo`, `New`, `Open` and `Save` are each registered three times the
same way. That is correct — they act on different windows.

But `ShortcutsConfigPage::checkConflicts()` builds a single global map of
sequence → rows and flags any sequence used by more than one row:

```cpp
QHash<QString, QList<int>> sequence_to_rows;
for (int row = 0; row < m_rows.size(); ++row) {
    const QString sequence_text = m_rows.at(row).edit->keySequence().toString();
    if (!sequence_text.isEmpty())
        sequence_to_rows[sequence_text] << row;
}
...
const bool conflicted = !sequence_text.isEmpty() && conflicting_rows.size() > 1;
```

There is no notion of scope, so every legitimate cross-window reuse is reported
as a collision.

Counting the current registrations: **23 sequences are shared by more than one
action, covering 60 rows.**

### Reproduction

Open *Préférences → Raccourcis* on a stock build; most rows are red.

Isolated repro, compiling only `shortcutmanager.cpp` and
`shortcutsconfigpage.cpp` against Qt5 and registering QElectroTech's real ids:

```cpp
m.registerAction(&a1,"diagrameditor.save_file_as",   "Éditeur de schémas",   QKeySequence("Ctrl+Shift+S"));
m.registerAction(&a2,"elementeditor.save_as_file",   "Éditeur d'élément",    QKeySequence("Ctrl+Shift+S"));
m.registerAction(&a3,"titleblockeditor.save_as_file","Éditeur de cartouche", QKeySequence("Ctrl+Shift+S"));
m.registerAction(&b1,"diagrameditor.undo",           "Éditeur de schémas",   QKeySequence::Undo);
m.registerAction(&b2,"elementeditor.undo",           "Éditeur d'élément",    QKeySequence::Undo);
m.registerAction(&b3,"titleblockeditor.undo",        "Éditeur de cartouche", QKeySequence::Undo);
m.registerAction(&c1,"mainwindow.fullscreen",        "Général",              QKeySequence("Ctrl+Shift+F"));
```

Output (run under `-platform offscreen`):

```
     ok     Général | Plein écran
  CONFLICT  Éditeur d'élément | Annuler (element)
  CONFLICT  Éditeur d'élément | Enregistrer dans un fichier
  CONFLICT  Éditeur de cartouche | Annuler (cartouche)
  CONFLICT  Éditeur de cartouche | Enregistrer vers un fichier
  CONFLICT  Éditeur de schémas | Annuler (schemas)
  CONFLICT  Éditeur de schémas | Enregistrer sous
  -> 6 of 7 rows flagged
```

Only the one genuinely unique binding escapes.

### Why it matters

A conflict indicator that fires on almost every row carries no information, and
it trains users to ignore it — including when a real conflict appears. It also
makes the page look broken on first open.

### Suggested fix

Give each registry entry a scope — the window or class that owns it, which the
category string already approximates — and have `checkConflicts()` compare only
within a scope. Two actions in different editors sharing `Ctrl+S` is not a
conflict; two in the same editor is.

If the category is considered a good enough proxy, this is a small change:
key the hash on `(category, sequence)` rather than `sequence` alone. That would
resolve all 60 current false positives.

Happy to open a PR if you would like it done that way, or a different way if
you prefer a scope independent of the display category.

ispyisail:
Fixes #757.

`ShortcutsConfigPage::checkConflicts()` compared key sequences across the whole
registry. But QElectroTech deliberately registers one key per editor window —
`Undo`, `Redo`, `New`, `Open`, `Save` and `Ctrl+Shift+S` each exist three times,
once for the diagram, element and titleblock editors. Those are not collisions:
they act on different windows.

The effect was that **60 of the 95 shipped default bindings displayed as
conflicts**, so the red highlight carried no information and the page looked
broken on first open.

### The change

Conflicts are keyed on `(category, sequence)` instead of `sequence` alone:

```cpp
QString conflictKey(const QString &category, const QString &sequence)
{
	return category + QLatin1Char('\x1f') + sequence;
}
```

`Row` gains a `category` field, populated from the `ShortcutInfo` the page
already receives. **No change to `ShortcutManager` or its API**, and no new
concept — the category is the registry's existing per-window grouping.

+22 / −3 across two files.

### Verification

Compiling `shortcutmanager.cpp` and `shortcutsconfigpage.cpp` standalone against
Qt5 and registering QElectroTech's **real** ids, plus one deliberately planted
same-category collision (`Doublon A` / `Doublon B`, both `Ctrl+Alt+D` in
`Éditeur de schémas`):

**Before — 8 of 9 rows flagged**

```
  flagged: Annuler (element)              <- false: different window
  flagged: Enregistrer dans un fichier    <- false
  flagged: Annuler (cartouche)            <- false
  flagged: Enregistrer vers un fichier    <- false
  flagged: Annuler (schemas)              <- false
  flagged: Enregistrer sous               <- false
  flagged: Doublon A                      <- genuine
  flagged: Doublon B                      <- genuine
  FAIL cross-window duplicates still flagged
  PASS genuine same-category collision still flagged
```

**After — 2 of 9 rows flagged**

```
  flagged: Doublon A
  flagged: Doublon B
  PASS cross-window duplicates not flagged
  PASS genuine same-category collision still flagged
```

The second assertion is the one that matters: a change that merely stopped
flagging things would also clear the false positives and would be worse than
the bug. The planted collision is still detected, so the indicator is scoped
rather than weakened.

Full application builds and links cleanly (496/496, Qt5, Debug).

### One design note

Using the display category as the scope key is a deliberate **minimal** choice:
it works because the category *is* the per-window grouping today, and it needs
no registry change. The tradeoff is that conflict logic now depends on a
display string.

If you would rather have a scope that is independent of the category — an
explicit window/owner recorded at registration — that is a larger change to
`ShortcutManager`, and I am happy to do it that way instead. I went with the
smaller version rather than assume.

Related: #758 (`Ctrl+G` bypasses `ShortcutManager` entirely, so it is invisible
to this page and to conflict detection). Both are ways the Shortcuts page
currently misrepresents the real state of the application's shortcuts.


scorpio810:
Thanks for the update! One process note: once a PR is open for review, please avoid rebasing/force-pushing (git push --force) on it — it breaks links to commits already referenced in the conversation and complicates incremental review. If you need to address feedback, an additional commit (plain git push, no --force) works fine — history gets squashed on merge anyway.

Also, this push pulled in a lot of unrelated changes (translation files, autoNum, diagnostics report, etc.) that aren't part of this fix — looks like the branch picked up commits from master that have nothing to do with the shortcut conflict scoping. Could you rebase cleanly against current master so the diff only shows your actual change? I won't review/merge until the PR is back to just the intended fix.



> ispyisail force-pushed the fix-shortcut-conflict-scope branch from 053903a to dab4c6c

https://github.com/qelectrotech/qelectrotech-source-mirror/compare/053903ad67179749b6f8f67e5c76a0920ec41a3e..dab4c6c42491922eddb839240d397f04ca972f84

ispyisail:
Both fair, and I've adjusted:

**No more force-pushing open PRs.** You're right that it breaks the commit links reviewers already have and makes incremental review harder — from now on I'll push additional commits instead and let the base drift until merge time (or a maintainer squash). Sorry for the churn here.

**On the "unrelated changes" concern — I think that's the force-push compare link, not the PR's actual state.** The auto-generated note above ("force-pushed from 053903a to dab4c6c") links a raw two-dot compare between those two exact SHAs. `053903a` was built on the branch's original base from when I opened this (`7307a59c1`, Aug 6); `dab4c6c` is rebased onto current master (`759b197dc`). There are 240 commits between those two points — including exactly the translation and autoNum churn you're seeing (German/Hungarian translation updates, a signal/slot migration pass, etc.) — so a raw compare between the pre- and post-rebase SHAs will always show that whole gap. It's not that this PR picked up those changes; it's that the link diffs two arbitrary commits from very different points in history rather than diffing the PR against its base.

The PR's actual diff (Files changed tab, and `gh pr diff 759`) is still exactly the one intended change: `sources/ui/configpage/shortcutsconfigpage.cpp`, +20/-2, one commit. I don't think there's anything left to clean up — happy to be shown otherwise if GitHub is rendering something different for you, but from the API and the diff itself this PR only ever touched that one file.

scorpio810:
Thanks for the explanation — and you're right, I was going off the raw 053903a..dab4c6c compare link, not the PR's actual diff. Checked the Files changed tab and it's exactly what you describe: one file, +20/−2, nothing else. Sorry for the false alarm on that part.

The force-push point stands as agreed — appreciate you adjusting. This one's ready for a proper look now.

ispyisail:

ispyisail:

The reasoning holds up: QElectroTech deliberately registers the same sequences once per editor window — Undo, Redo, New, Open, Save and Ctrl+Shift+S each exist three times, for the diagram, element and title-block editors. Those are not collisions, because the actions belong to different windows and can never be live at the same time. Comparing sequences across the whole registry flagged 60 of the 95 shipped defaults, which makes the conflict indicator worthless precisely when it should be trusted.

Keying on `category + sequence` uses the registry's own existing per-window grouping as the scope, so it needs no new concept. The `\x1f` separator is a sensible choice — a unit separator cannot appear in either a category name or a `QKeySequence::toString()` result, so the two fields cannot run together and produce a false match.

**Not verified:** the indicator's appearance in the preferences page interactively.

PR #761, opened 2026-08-18
Change element editor coordinates display

ChuckNr11:
Commit: Change live cursor-coordinate readout on running ESEvent
During an ES event in the element editor (e.g., drawing a line), the coordinates for the coordinate display are not formatted using 'snapToGrid'. As a result, the coordinates of the help cross do not match the displayed coordinates. The signal for transmitting the coordinates has been moved to the `updateHelpCross` function within the event interface. Therfore, position formatted with 'snapToGrid' is used for the display across all events.

Commit:  move  element editor coord display to center of statusar
The coordinate display has been moved to the center of the status bar because it is more easily visible there than in the bottom-right corner.

### DIFF
--- a/sources/editor/elementscene.cpp
+++ b/sources/editor/elementscene.cpp
@@ -121,7 +121,7 @@ void ElementScene::mouseMoveEvent(QGraphicsSceneMouseEvent *e)
 {
 	if (m_event_interface) {
 		if (m_event_interface -> mouseMoveEvent(e)) {
-			emit mouseMoved(e -> scenePos());
+			emit mouseMoved(snapToGrid(e->scenePos()));
 			if (m_event_interface->isFinish()) {
 				delete m_event_interface;
 				m_event_interface = nullptr;
--- a/sources/editor/ui/qetelementeditor.cpp
+++ b/sources/editor/ui/qetelementeditor.cpp
@@ -55,6 +55,7 @@
 #include <QActionGroup>
 #include <QFileDialog>
 #include <QSvgGenerator>
+#include <QHBoxLayout>
 
 /**
  * @brief QETElementEditor::QETElementEditor
@@ -1200,7 +1201,18 @@ void QETElementEditor::initGui()
 		//Live cursor position readout, in the same scene coordinates as the parts' X/Y properties
 	m_position_label = new QLabel(this);
 	m_position_label->setMinimumWidth(120);
-	statusBar()->addPermanentWidget(m_position_label);
+	
+	// Layout
+	QHBoxLayout *coordDisplayLayout = new QHBoxLayout();
+	coordDisplayLayout->setContentsMargins(0, 0, 0, 0);
+	coordDisplayLayout->addWidget(m_position_label);
+	coordDisplayLayout->addStretch();
+	// Widget
+	QWidget *coordDisplay = new QWidget;
+	coordDisplay->setLayout(coordDisplayLayout);
+	
+	statusBar()->addPermanentWidget(coordDisplay);
+	
 	connect(m_elmt_scene, &ElementScene::mouseMoved, this, [this](const QPointF &pos) {
 		m_position_label->setText(tr("X: %1  Y: %2")
 			.arg(pos.x(), 0, 'f', 1)


ispyisail:
Nice catch on the coordinate mismatch — the bug is real. In `ElementScene::mouseMoveEvent()` the ES-event branch emits the raw position while the normal path emits a snapped one:

```cpp
if (m_event_interface) {
    if (m_event_interface->mouseMoveEvent(e)) {
        emit mouseMoved(e->scenePos());        // raw
        ...
QPointF event_pos = e->scenePos();
if (!(e->modifiers() & Qt::ControlModifier))
    event_pos = snapToGrid(event_pos);
emit mouseMoved(event_pos);                    // snapped
```

so the readout and the help cross disagree while drawing. Coverage of the new emit looks right too — `updateHelpCross()` is called at the top of `mouseMoveEvent()` in every `ESEvent*` class, so the readout still updates on every move.

A few things I would raise before this lands.

**1. Emitting another object's signal**

```cpp
// eseventinterface.cpp
emit m_scene->mouseMoved(point);
```

This compiles, but a signal is meant to be emitted only by the class that declares it. I suspect you were pushed into it because `ESEventInterface` is not a `QObject` and so cannot declare a signal of its own.

The same result fits in one line inside the scene, with no cross-object emit and no change to `eseventinterface.cpp`:

```cpp
if (m_event_interface->mouseMoveEvent(e)) {
    emit mouseMoved(snapToGrid(e->scenePos()));   // was e->scenePos()
```

`updateHelpCross()` snaps with `m_scene->snapToGrid(p)`, so the value is identical and the cross and readout still agree — but the change is one file instead of four.

**2. Does the status bar change actually centre it?**

`QStatusBar::addPermanentWidget()` adds to the *right-hand* side. Wrapping the label in a `QWidget` whose layout is `addWidget(label); addStretch();` left-aligns the label inside that widget, but the widget still reports a label-sized `sizeHint` and still sits on the right. Did the centring show up as you expected when you ran it? If not, a normal `addWidget()` plus stretch behaves differently from the permanent-widget area.

**3. Minor**

- `#include "qboxlayout.h"` in the header uses the lowercase private path; `<QHBoxLayout>` is the public spelling, and a forward declaration would be enough since only pointers are stored.
- `m_coord_display_layout` need not be a member — the widget owns it once `setLayout()` is called.
- The snap fix and the status-bar relocation are independent. Splitting them would let the fix go in on its own merit while the placement is discussed, and the fix looks uncontroversial.

Only the first point is a correctness concern; the rest are style and scope. Thanks for digging into this — the mismatch has been there a while.

ChuckNr11:
Hi ispyisail,
Thanks for your feedback.
That approach if   " emit mouseMoved(snapToGrid(e->scenePos()));"  was my initial thought too, but I ended up preferring the solution using ESEventInterface.
My programming skills are somewhat limited, and I wasn't aware of that particular aspect—but I’m always happy to learn. I’ve adjusted the code based on your recommendation.

Centering the coordinate display works fine; I’ve actually been using it that way for a while in my personal version of Qelectrotech.
I’ve updated the code here as well, following your recommendations.

ispyisail:
Thanks for turning that around so quickly — and no need to apologise for the first approach, spotting the mismatch at all was the hard part.

I built the branch at `84a4b22` and measured the status-bar behaviour in isolation, so the notes below are from running it rather than from reading it.

## 1. The snap fix — resolved, and it is correct

`emit mouseMoved(snapToGrid(e->scenePos()))` is right, and I checked the thing I had only assumed last time: `updateHelpCross()` snaps with `m_scene->snapToGrid(p)` **unconditionally**, with no `Ctrl` test, and all five `ESEventAdd*` classes snap unconditionally too:

```
eseventaddline     line.setP2(m_scene->snapToGrid(event->scenePos()));
eseventaddrect     QRectF rect(..., m_scene->snapToGrid(event->scenePos()));
eseventaddellipse  QPointF pos = m_scene->snapToGrid(event->scenePos());
eseventaddpolygon  m_polygon->setLastPoint(m_scene->snapToGrid(...));
eseventaddarc      m_mouse_pos = m_scene->snapToGrid(event->scenePos());
```

So geometry, help cross and readout now agree in every ES event. Deliberately *not* mirroring the `Ctrl` bypass from the normal path is the correct call here — `Ctrl` does not bypass the grid while drawing in the first place, so honouring it in the readout would have reintroduced the very mismatch you set out to fix.

This commit looks finished to me.

## 2. The status bar — it does move, but it is not centred, and it costs 18px of height

Your "it works" and my "it will not centre" are both partly right, which is why I measured it. I rebuilt your exact arrangement (`QLabel` → `QHBoxLayout` + `addStretch` → wrapper `QWidget` → `addPermanentWidget`) in a standalone Qt 5.15 program and printed the geometry:

| window width | label spans | label centre | window centre | off by |
|---|---|---|---|---|
| 700 | 348..482 | 415 | 350 | +65 |
| 1000 | 498..632 | 565 | 500 | +65 |
| 1600 | 798..932 | 865 | 800 | +65 |

The pattern is exact: the label's **left edge** lands on `width/2` every time. The wrapper claims the right half of the bar and `addStretch()` pins the label to its left edge. So the text *starts* at the centre and runs rightwards — which is a real improvement over master (label centre at 909 of 1000, hard against the right edge), and I can see why it reads as centred in daily use. But it is not centring, and the error grows with the text: with `X: -12345.6  Y: -98765.4` the offset goes from +58 to +89px, so the readout is not sitting still relative to the window as you draw.

**The part I would ask you to change**, which is a genuine regression rather than a matter of taste:

```
PR #761  statusbar height = 47   (wrapper layout margins = 9,9,9,9)
master   statusbar height = 29
```

A `QHBoxLayout` gets 9px default margins on all four sides, so wrapping the label makes the status bar **62% taller** for the whole element editor. One line fixes it:

```cpp
coordDisplayLayout->setContentsMargins(0, 0, 0, 0);
```

If you do want true centring, this measured as off-by-11px rather than off-by-65, with no wrapper, no extra height, and no new member:

```cpp
m_position_label->setAlignment(Qt::AlignCenter);
statusBar()->addWidget(m_position_label, 1);   // normal area, stretch 1
```

Either way I would not block on the placement — it is a taste call for the maintainers. The extra 18px I would fix before merge.

## 3. Small leftovers

- `m_coord_display` is only ever touched inside `initGui()`, so it does not need to be a member either — the same reason `m_coord_display_layout` did not. `addPermanentWidget()` reparents it and the status bar owns it from then on.
- `QHBoxLayout` still has no `#include`. It compiles today only because `elementscene.h:26` does a blanket `#include <QtWidgets>`; that is QET's wart, not yours, but an explicit `#include <QHBoxLayout>` in the `.cpp` would make it stand on its own. (Dropping the lowercase `qboxlayout.h` from the header was the right move.)
- The two commits are cleanly separable (`2803dcf` snap fix, `84a4b22` placement). If the placement discussion runs long, the snap fix is uncontroversial and could go in on its own.

Nice work on the fix itself — that one has been wrong for a while.

ChuckNr11:
> The pattern is exact: the label's left edge lands on width/2 every time. The wrapper claims the right half of the bar and addStretch() pins the label to its left  edge

I was aware that it is not in the center.

> PR #761  statusbar height = 47   (wrapper layout margins = 9,9,9,9)
master   statusbar height = 29

You're right, of course; I hadn't even noticed that the status bar had become so much larger. I've adopted your suggested correction.

> m_coord_display_layout did not.

I could have figured that out on my own. I've corrected it.

ispyisail:
Built `5c093e840` and re-measured. Both points from last time are resolved, and **I owe you a correction on the centring — you were right and I was wrong about which behaviour is better.**

## Status bar height — fixed

| | statusbar height |
|---|---|
| master | 29 |
| PR before (`9,9,9,9` margins) | 47 |
| **PR now** | **29** |

Back to parity. `setContentsMargins(0,0,0,0)` did it.

## The centring — I was pushing you the wrong way

I said the label wasn't centred, and it isn't: its left edge sits at `width/2` and the text grows rightward. What I did not check last time was what happens as the readout *changes width*, which it does on every mouse move.

Measured, same window, three different coordinate strings:

| text | your arrangement | truly centred |
|---|---|---|
| `X: 0.0  Y: 0.0` | **489**..609 | **429**..549 |
| `X: 123.4  Y: 567.8` | **489**..623 | **422**..556 |
| `X: -1234.5  Y: -6789.0` | **489**..653 | **407**..571 |

A genuinely centred label **moves horizontally every time the numbers change width** — 429 → 422 → 407. On a readout that updates continuously while you drag, that is constant shimmer.

Yours anchors the left edge at 489 and grows to the right, so the text never jitters. That is the better behaviour for a live readout, and a better justification for the layout than "it is centred" — which, as you already noted, it is not. I would leave it exactly as it is.

## One small thing left

`QHBoxLayout` is used in `qetelementeditor.cpp`, but neither that file nor its header includes it — the earlier `#include "qboxlayout.h"` went away with the header change. It builds today (I get a clean 472/472) because something else pulls it in transitively, but that is the kind of thing that breaks later when an unrelated header is tidied.

Adding this to the `.cpp` includes would make it robust:

```cpp
#include <QHBoxLayout>
```

Entirely optional and not a blocker.

## Overall

The snap fix is correct and complete — geometry, help cross and readout now agree in every ES event. The status-bar change no longer costs any height, the widget and layout are local instead of members, and the header is untouched, so the diff is down to two files and +13/-2.

Looks good to me. Thanks for working through three rounds of feedback on this — the readout mismatch had been there a long time and nobody else spotted it.

ChuckNr11:
> Adding this to the .cpp includes would make it robust:
#include QHBoxLayout

done
Thanks a lot

PR #766, opened 2026-08-21
PLC terminal names: fix transfer to slaves and display in properties widgets

Kellermorph:
Bug Fix - PLC terminal labels not transferred from Master to Slave

When linking a PLC Slave to a PLC Master from the Master side (m_element->linkType() == Element::Master), the terminal labels (Anschlüsse) defined in PlcIO.terminals were never propagated to the slave. The code unconditionally used m_slave_contact_groups for terminal labels, which is empty for PLC masters. Additionally, ELMT_PLC_TC and ELMT_PLC_T1..T4 were missing from the Master-case PLC variable propagation.

Fixed in sources/undocommand/linkelementcommand.cpp by:

Branching on m_master_type == PLC to use plcMasterData().ios[group_idx].terminals for PLC masters
Adding the missing ELMT_PLC_TC and ELMT_PLC_T1..T4 values to the slave's DiagramContext
Feature - Display PLC terminal names in properties widgets

PLC terminal names defined in the element editor were not visible in the schematic editor's GUI. Added a read-only "Bornes" column to both:

MasterPropertiesWidget - shows comma-separated terminal names when editing a PLC Master's properties
PlcLinkWidget - shows actual terminal names instead of just the count when connecting a Slave to a Master

### DIFF
--- a/sources/properties/elementdata.h
+++ b/sources/properties/elementdata.h
@@ -99,6 +99,19 @@ class ElementData : public PropertiesInterface
 			int terminalCount = 1;    ///< Number of terminals for this IO (1-4)
 			QStringList terminals;    ///< Terminal values T1, T2, ... (size = terminalCount)
 
+			/**
+			 * @brief Return terminal labels, generating defaults (T1, T2...) if empty
+			 */
+			QStringList effectiveTerminals() const {
+				if (!terminals.isEmpty())
+					return terminals;
+				int count = qMax(terminalCount, 1);
+				QStringList defaults;
+				for (int i = 0; i < count && i < 4; ++i)
+					defaults << QStringLiteral("T%1").arg(i + 1);
+				return defaults;
+			}
+
 			bool operator==(const PlcIO &other) const {
 				return type == other.type
 					&& address == other.address
--- a/sources/qetgraphicsitem/element.cpp
+++ b/sources/qetgraphicsitem/element.cpp
@@ -38,6 +38,15 @@
 #include "dynamicelementtextitem.h"
 #include "elementtextitemgroup.h"
 #include "iostream"
+
+#include <QCollator>
+
+static const QString plcTerminalKeys[] = {
+	QETInformation::ELMT_PLC_T1,
+	QETInformation::ELMT_PLC_T2,
+	QETInformation::ELMT_PLC_T3,
+	QETInformation::ELMT_PLC_T4
+};
 #include "../qetxml.h"
 #include "../qetversion.h"
 #include "qgraphicsitemutility.h"
@@ -1511,13 +1520,7 @@ void Element::setElementData(ElementData data)
 					{
 						QString val = (t < io.terminals.size())
 							? io.terminals.at(t) : QString();
-						ctx.addValue(
-							QStringList({
-								QETInformation::ELMT_PLC_T1,
-								QETInformation::ELMT_PLC_T2,
-								QETInformation::ELMT_PLC_T3,
-								QETInformation::ELMT_PLC_T4
-							}).at(t), val);
+						ctx.addValue(plcTerminalKeys[t], val);
 					}
 					slave->setElementInformations(ctx);
 
--- a/sources/qetgraphicsitem/terminal.cpp
+++ b/sources/qetgraphicsitem/terminal.cpp
@@ -823,17 +823,15 @@ QUuid Terminal::uuid() const
 QString Terminal::name() const
 {
 	if (d->m_use_master_label && parent_element_) {
-		// Find the master element in the slave's linked elements
 		for (Element *elmt : parent_element_->linkedElements()) {
 			if (elmt->linkType() == Element::Master) {
 				int group_idx = elmt->groupIndexForElement(parent_element_);
 				if (group_idx >= 0) {
-					// For PLC masters, use io.terminals as labels
 					if (elmt->elementData().m_master_type == ElementData::PLC) {
 						const auto &plc_data = elmt->elementData().plcMasterData();
 						if (group_idx < plc_data.ios.size()) {
 							int label_idx = d->m_master_label_index;
-							const QStringList &labels = plc_data.ios.at(group_idx).terminals;
+							const QStringList labels = plc_data.ios.at(group_idx).effectiveTerminals();
 							if (label_idx >= 0 && label_idx < labels.size()) {
 								return labels.at(label_idx);
 							}
--- a/sources/ui/masterpropertieswidget.cpp
+++ b/sources/ui/masterpropertieswidget.cpp
@@ -979,14 +979,7 @@ void MasterPropertiesWidget::plcUpdateDisplaySettings()
 	ElementData::PlcMasterData plc_data = ed.plcMasterData();
 	plc_data.ios.clear();
 
-	// Build address -> original IO lookup to correctly reattach terminal data
-	// after row reorder (move up/down) or row removal
-	QHash<QString, int> addr_to_orig_idx;
-	for (int i = 0; i < ed.plcMasterData().ios.size(); ++i) {
-		addr_to_orig_idx[ed.plcMasterData().ios.at(i).address] = i;
-	}
-
-	// Read IOs from table, preserving terminal data from original IOs
+	// Read IOs from table, preserving terminal data from original IOs by row index
 	for (int row = 0; row < m_plc_table->rowCount(); ++row) {
 		ElementData::PlcIO io;
 
@@ -1010,9 +1003,12 @@ void MasterPropertiesWidget::plcUpdateDisplaySettings()
 		if (crossref_item)
 			io.crossRef = crossref_item->text();
 
-		// Preserve terminal data by looking up original IO via address
-		if (addr_to_orig_idx.contains(io.address)) {
-			const auto orig_io = ed.plcMasterData().ios.at(addr_to_orig_idx.value(io.address));
+		// Preserve terminal data by matching row index directly.
+		// This avoids address-based lookup which fails when all addresses
+		// are empty (common in PLC masters) — a hash collision would cause
+		// only the last IO's terminals to be used for all rows.
+		if (row < ed.plcMasterData().ios.size()) {
+			const auto &orig_io = ed.plcMasterData().ios.at(row);
 			io.terminalCount = orig_io.terminalCount;
 			io.terminals = orig_io.terminals;
 		}
--- a/sources/ui/masterpropertieswidget.cpp
+++ b/sources/ui/masterpropertieswidget.cpp
@@ -525,10 +525,10 @@ void MasterPropertiesWidget::updateUi()
 
 			// Table
 			m_plc_table = new QTableWidget(m_plc_widget);
-			m_plc_table->setColumnCount(5);
+			m_plc_table->setColumnCount(6);
 			m_plc_table->setHorizontalHeaderLabels({
 				tr("Type"), tr("Adresse"), tr("Fonction"),
-				tr("Commentaire"), tr("Réf. croisée")
+				tr("Commentaire"), tr("Réf. croisée"), tr("Bornes")
 			});
 			m_plc_table->horizontalHeader()->setSectionResizeMode(QHeaderView::Stretch);
 			m_plc_table->setSelectionBehavior(QAbstractItemView::SelectItems);
@@ -592,6 +592,11 @@ void MasterPropertiesWidget::updateUi()
 			auto *crossref_item = new QTableWidgetItem(io.crossRef);
 			crossref_item->setFlags(crossref_item->flags() & ~Qt::ItemIsEditable);
 			m_plc_table->setItem(row, 4, crossref_item);
+
+			// Anschlüsse (read-only)
+			auto *terminals_item = new QTableWidgetItem(io.terminals.join(QStringLiteral(", ")));
+			terminals_item->setFlags(terminals_item->flags() & ~Qt::ItemIsEditable);
+			m_plc_table->setItem(row, 5, terminals_item);
 		}
 
 		m_plc_table->blockSignals(false);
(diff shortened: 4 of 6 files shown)


ispyisail:
Read through this one carefully and built it. The core diagnosis is right, and it fixes something real: `m_slave_contact_groups` genuinely is empty for a PLC master, so the terminal labels never had anywhere to come from. Branching on `m_master_type == PLC` to read `plcMasterData().ios[idx].terminals` is the correct shape for the fix.

Four things below. One I think should be in this PR; the rest are your call.

## 1. The same bug is left behind in the branch you forked from

This is the one I'd fix here, because after this PR the file contains four of these sorts and exactly one of them is the odd one out.

On current `master` there are already three:

| line | branch | key |
|---|---|---|
| 417 | PLC (slave side) | `baseName()` |
| 440 | non-PLC (slave side) | `baseName()` |
| 511 | inside `makeLink()` | **`name()`** |

Your patch adds a fourth at 514 using `baseName()` — correct — but the `else` you split off keeps `name()`:

```cpp
std::sort(slave_terms.begin(), slave_terms.end(),
    [](Terminal *a, Terminal *b) { return a->name() < b->name(); });
```

So it ends up three-to-one, and the outlier is the one the patch moved but didn't convert.

It matters because `Terminal::name()` is not an accessor. When `m_use_master_label` is set it walks `parent_element_->linkedElements()`, finds the master, and returns `labels.at(m_master_label_index)` (`terminal.cpp:823-856`). The sort key therefore depends on whether the slave is *already* linked: re-link a slave that carries labels from a previous master and it sorts by label text rather than by `T1, T2, T3`, and the indices permute.

`baseName()` exists precisely to avoid that, and says so on master:

```cpp
/**
    @brief Terminal::baseName
    Return the original terminal name (T1, T2...) without master label override.
    Used for sorting when linking.
*/
```

The comment sitting directly above the `name()` call — "Sort terminals by name (T1, T2, T3...) to match label order" — is describing `baseName()` semantics, not `name()` semantics. One-word change, and all four sites agree.

## 2. The sort is lexicographic, and terminal numbering goes past 9

Both branches compare terminal names as strings. `terminaldata.h:139` documents the range explicitly:

```cpp
/// Index into the master's contact group labels (T1=0, T2=1, ..., T20=19)
```

With 20 terminals a string sort gives:

```
T1, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T2, T20, T3, T4, ... T9
```

which assigns the wrong label index to **19 of the 20 terminals**. The PLC path is protected today only because `terminalCount` is capped at 4, so it is latent there; the contact-group path is reachable now. A comparator on the numeric suffix (or `QCollator` with `setNumericMode(true)`) fixes both.

Pre-existing rather than introduced by you — but the PR is already rewriting these two loops, so it is a natural place to deal with it.

## 3. The `plc_tc` / `plc_t1..t4` values you propagate cannot currently be displayed

Not introduced by you — `master` already writes these same keys at `linkelementcommand.cpp:470` and `element.cpp:1508` — but this PR adds a third write site, so it seems worth surfacing.

`ELMT_PLC_TC` and `ELMT_PLC_T1..T4` are declared in `qetinformation.h:133-137` but are **not** in `QETInformation::elementInfoKeys()`. That list stops at `ELMT_PLC_CROSSREF` (`qetinformation.cpp:193-197`). And membership of that list is what makes a key usable:

```cpp
QString QETInformation::elementInfoToVar(const QString &info)
{
    if (QETInformation::elementInfoKeys().contains(info))
        return infoToVar(info);
    else
        return (QString ("%{void}"));      // <- plc_t1..t4 land here
}
```

The same list also drives the element information widget (`elementinfowidget.cpp:204,266`), the dynamic-text source model (`dynamicelementtextmodel.cpp:1976`), composite text editing (`compositetexteditdialog.cpp:68`) and the Search & Replace variable list (`diagram.cpp:2574,2613`).

The values *do* survive a save — `DiagramContext::toXml` iterates `keys()` with no whitelist — so they sit in the `.qet` file. They just have no surface that can read them back out. Registering the five keys (plus their `translatedInfoKey` names) would make all three write sites actually do something.

I'd suggest that as a **separate** PR rather than folding it in here: `elementInfoKeys()` feeds the project database columns as well, so it is a wider change than it looks and deserves its own diff.

## 4. Smaller things

**Removing the terminal-count spinbox is safe, and it is worth saying why**, because it looks like a feature removal and is not. On master the table is created with `setColumnCount(5)`, but `plcAddRow()` writes the spinbox to column 5 and the cross-ref item to column 6. Qt silently drops both — I checked rather than assumed:

```
columnCount after writing to col 5 and 6 : 5
item(0,6)       = (null)
cellWidget(0,5) = (null)
```

So that spinbox was never reachable by a user. You are deleting dead code, not a control.

**`tr("Anschlüsse")` → `tr("Bornes")` is a bigger fix than the description suggests.** French is the `tr()` source language throughout this codebase, so a German source string becomes the msgid that every other translation is keyed against — it would have shown up untranslated in every language, not just looked odd in the French UI.

**Micro:** the per-iteration `QStringList({ELMT_PLC_T1, ..., ELMT_PLC_T4}).at(t)` builds and destroys a four-element list on every pass; a `static const` array reads better and costs nothing. And `linkelementcommand.cpp` uses `std::sort` without including `<algorithm>` — it compiles through transitive includes today, but it is one header reshuffle away from not doing so. Also pre-existing.

**Left alone deliberately:** `plcAddRow()` still creates column 4 (Réf. croisée) as a plain editable item while `updateUi()` makes the same column read-only, so a freshly added row behaves differently from a reloaded one. Adjacent to your change but not caused by it.

## Build

Builds clean against current `master` (Qt 5.15, Release), no new warnings in any of the three changed files.

Nice catch on the underlying bug — the PLC master/slave label path being empty is not obvious from reading the code, and the `Bornes` column makes the data visible where it actually needs to be.

scorpio810:
Ask me when thinking is ready to merge, thanks @Kellermorph @ispyisail .

scorpio810:
master type in makeLink()) is a real and valuable fix.

Two things worth checking before merge though:

Translation string change: in plclinkwidget.cpp, the header label changed from tr("Anschlüsse") to tr("Bornes"). The German string "Anschlüsse" was actually a bug — it should never have been left in the source as an untranslated/hardcoded German string in the first place. So switching it to tr("Bornes") is the right call, but since "Anschlüsse" was presumably already picked up by the translation files as a translatable string, this change should be flagged so the .ts files get updated and the old string doesn't linger as a stale/orphaned entry.
Sorting inconsistency: for PLC masters, terminals are now sorted by baseName(), while the existing (non-PLC) branch still sorts by name(). Was this difference intentional? If not, it could cause a mismatch between how labels are ordered/assigned depending on master type, and might introduce a subtle bug in terminal-to-label mapping for PLC elements specifically.
Everything else looks consistent with the existing elementdata.h structures (terminals/terminalCount), so this is close — just want these two points clarified before merging to avoid introducing a regression.

ispyisail:
Both good questions. I went and checked them rather than reasoning about it, and the answers are more comfortable than they look — one of them is essentially free, and the other is the reverse of the way round it reads.

## 1. The string change costs almost nothing, because the destination already exists

I was expecting to have to say "yes, this orphans a string". It doesn't, because **`Bornes` is already a source string in the project**, used for the same concept at `contactgroupselectiondialog.cpp:57`:

```cpp
<< tr("Bornes");
```

so it is already present in **31 of the 34** `.ts` files — and already translated where it matters:

| | `<source>Anschlüsse</source>` (going away) | `<source>Bornes</source>` (already there) |
|---|---|---|
| de | `Anschlüsse` | **`Anschlüsse`** |
| en | `Connections` | **`Terminals`** |
| fr | `Connexions` | source string |
| other 28 | `type="unfinished"` | `type="unfinished"` |

So after the rename a German user still sees *Anschlüsse* — but reached through the translation table instead of through a hardcoded German source string, which is the actual fix. English arguably improves, since this column is terminals rather than connections. The 28 unfinished entries are unfinished either way, so nothing is lost there.

The only casualty is the French `Connexions` attached to the German source — and that entry is itself the anomaly, since French is the `tr()` source language here, so it should never have been translating *from* German in the first place.

Worth adding: this also makes the two dialogs agree. `contactgroupselectiondialog.cpp` already calls its equivalent column *Bornes*; `plclinkwidget.cpp` was the odd one out.

The `.ts` files will still want an `lupdate` pass to drop the now-unreferenced entry, as you say — but it is a stale-entry cleanup, not lost translator work.

## 2. The sorting difference is real, and the risk is the other way round

> it could ... introduce a subtle bug in terminal-to-label mapping for PLC elements specifically

It is the opposite: the PLC branch is the correct one, and the branch that keeps `name()` is the one carrying the bug. `master` already agrees with the PLC branch in two other places:

| file/line | branch | key |
|---|---|---|
| `linkelementcommand.cpp:417` | PLC | `baseName()` |
| `linkelementcommand.cpp:440` | non-PLC | `baseName()` |
| `linkelementcommand.cpp:511` | inside `makeLink()` | **`name()`** |

So three of the four sort sites after this PR use `baseName()`, and the one that does not is the pre-existing straggler at 511 that the patch happened to move.

The reason `baseName()` is right is that `Terminal::name()` is not an accessor. When `m_use_master_label` is set it walks `parent_element_->linkedElements()`, finds the master and returns `labels.at(m_master_label_index)` (`terminal.cpp:823-856`). The sort key therefore depends on whether the slave is *already* linked — re-link a slave that still carries labels from a previous master and it sorts by label text rather than by `T1, T2, T3`, and the mapping permutes. `baseName()` returns `d->m_name` unconditionally, so it is stable.

`baseName()` exists for exactly this, and says so:

```cpp
/**
    @brief Terminal::baseName
    Return the original terminal name (T1, T2...) without master label override.
    Used for sorting when linking.
*/
```

The comment directly above the `name()` call — *"Sort terminals by name (T1, T2, T3...) to match label order"* — is describing `baseName()` behaviour, not `name()` behaviour.

**So: not intentional, but harmless as merged, and the safe change is to move line 511's `else` branch onto `baseName()` too** rather than to move the PLC branch back. That makes all four agree and removes a latent re-link bug that predates this PR.

## One thing neither of us has mentioned

While checking the above I noticed the sort is lexicographic in all four places, and `terminaldata.h:139` documents the range as `T1=0, T2=1, ..., T20=19`. A string sort of `T1..T20` gives

```
T1, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T2, T20, T3, ... T9
```

which assigns the wrong label index to 19 of the 20 terminals. PLC is shielded today only because `terminalCount` is capped at 4; the contact-group path is not. Also pre-existing, and probably its own small PR — a comparator on the numeric suffix, or `QCollator` with `setNumericMode(true)`.

For what it is worth I built the PR as-is against current `master` (Qt 5.15, Release): clean, no new warnings in any of the three changed files.

ispyisail:
@scorpio810 you asked to be told when this is ready — from my side it is, once two one-liners land. Neither changes what the PR does.

**1. `linkelementcommand.cpp`** — the sorting question you raised. The fix is to bring the non-PLC branch up to the PLC one, not the reverse:

```diff
-						[](Terminal *a, Terminal *b) {
-							return a->name() < b->name();
-						});
+						[](Terminal *a, Terminal *b) {
+							return a->baseName() < b->baseName();
+						});
```

All four sort sites then agree, which is what `baseName()`'s doc comment says it exists for.

**2. `masterpropertieswidget.cpp`** — two guards left over from the old 7-column layout, now that the table is 6:

```diff
-	if (!m_plc_table || row < 0 || col < 0 || col > 6)
+	if (!m_plc_table || row < 0 || col < 0 || col > 5)
```

```diff
 			for (int c = 0; c < cells.size(); ++c) {
-				if (c > 6) break;
+				if (c > 5) break;
```

Without them a horizontal paste of six or more columns writes into the read-only *Bornes* cell. Cosmetic only — `plcUpdateDisplaySettings()` reads back columns 1-4 and keeps `io.terminals` from the model, so the data is not affected and the next `updateUi()` repaints it correctly.

With those in, good to merge from my side: builds clean against current `master`, no new warnings in the three changed files.

Thanks @Kellermorph — the master-side label transfer was the hard part to spot.

Kellermorph:
Problem:
When linking a PLC slave to a PLC master, the terminal names (e.g.
AI0, AI1, ...) defined in the master's plcMasterData XML were not
transferred to the slave element. Instead, the slave's default
terminal names (T1, T2, ...) were kept.

Root cause:
The address-based lookup in MasterPropertiesWidget::savePlcTable()
collided when all PLC IO entries had empty addresses (common for
PLC masters like DEOS modules). A QHash<QString,int> mapped every
IO's empty address "" to the same key, keeping only the last index.
On every save/load cycle, all IO rows received the terminal data
of the last IO (e.g. AI7), corrupting the display and label
transfer for all other IOs.

Additionally, the terminal sort in linkelementcommand.cpp used
simple string comparison (baseName() < baseName()), which sorted
"T10" before "T2". For non-PLC masters it also used name() instead
of baseName(), ignoring the master label.

Fixes:
- masterpropertieswidget.cpp: Replace address-based IO lookup with
  direct row-index matching to preserve terminal data correctly
  even when all addresses are empty
- linkelementcommand.cpp: Use effectiveTerminals() for PLC label
  transfer (with T1,T2... fallback when terminals list is empty);
  add QCollator with numericMode for natural terminal sorting;
  move PLC variable population into the PLC branch; fix broken
  indentation from previous iteration
- terminal.cpp: Use effectiveTerminals() for PLC master name
  resolution
- elementdata.h: Add PlcIO::effectiveTerminals() method that
  returns the stored terminals or generates T1,T2... defaults
- element.cpp: Extract plcTerminalKeys[] static array to avoid
  repeated QStringList construction

scorpio810:
@Kellermorph  can you make an little video how to use?

Kellermorph:
The procedure is the same as described here:

https://qelectrotech.org/forum/viewtopic.php?id=3173

except that, so far, some updates have caused the display of the port name to stop working again.

plc-user:
@Kellermorph:
In one of your previous PRs for the PLC Master, there is a compilation warning:
```../../sources/ui/masterpropertieswidget.cpp: In member function ‘void MasterPropertiesWidget::plcUpdateDisplaySettings()’:
../../sources/ui/masterpropertieswidget.cpp:1011:37: warning: possibly dangling reference to a temporary [-Wdangling-reference]
 1011 |                         const auto &orig_io = ed.plcMasterData().ios.at(row);
      |                                     ^~~~~~~
../../sources/ui/masterpropertieswidget.cpp:1011:72: note: the temporary was destroyed at the end of the full expression ‘ElementData::plcMasterData() const().ElementData::PlcMasterData::ios.QList<ElementData::PlcIO>::at(((qsizetype)row))’
 1011 |                         const auto &orig_io = ed.plcMasterData().ios.at(row);
      |                                               ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~

```
Are you going to fix that, too?

Kellermorph:
> @Kellermorph: In one of your previous PRs for the PLC Master, there is a compilation warning:
> 
> ```
> ../../sources/ui/masterpropertieswidget.cpp:1011:37: warning: possibly dangling reference to a temporary [-Wdangling-reference]
>  1011 |                         const auto &orig_io = ed.plcMasterData().ios.at(row);
>       |                                     ^~~~~~~
> ../../sources/ui/masterpropertieswidget.cpp:1011:72: note: the temporary was destroyed at the end of the full expression ‘ElementData::plcMasterData() const().ElementData::PlcMasterData::ios.QList<ElementData::PlcIO>::at(((qsizetype)row))’
>  1011 |                         const auto &orig_io = ed.plcMasterData().ios.at(row);
>       |                                               ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
> ```
> 
> Are you going to fix that, too?

i will take a look

PR #769, opened 2026-08-22
Fix dock widget size/position not being restored on Qt6

Kellermorph:
Summary:

After migrating to Qt6, dock widget sizes and positions (AutoNumbering, elements panel, properties editor) were not restored on application restart. The values were saved correctly to QSettings but restoreState() failed to apply them silently.

Root cause: In Qt6, QMainWindow::restoreState() requires the window to be visible. The code called restoreState() before show() in the constructor, which worked in Qt6 but no longer functions correctly.

Fix: Split readSettings() in QETDiagramEditor into two calls:

readSettings() — restores geometry and view mode (before show())
readSettingsState() — restores dock/toolbar state (after show())
This ensures restoreState() is called after the window is shown, as Qt6 requires.

Files changed:

sources/qetdiagrameditor.cpp — split readSettings, add readSettingsState()
sources/qetdiagrameditor.h — declare readSettingsState()

### DIFF
--- a/sources/editor/ui/qetelementeditor.cpp
+++ b/sources/editor/ui/qetelementeditor.cpp
@@ -967,11 +967,9 @@ void QETElementEditor::readSettingsState()
 
 	QVariant state = settings.value("elementeditor/state");
 	if (state.isValid()) {
-#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
-		restoreState(state.toByteArray());
-#else
-		restoreState(state.toByteArray());
-#endif
+		if (!restoreState(state.toByteArray())) {
+			settings.remove("elementeditor/state");
+		}
 	}
 }
 
--- a/sources/editor/ui/qetelementeditor.cpp
+++ b/sources/editor/ui/qetelementeditor.cpp
@@ -952,7 +952,13 @@ void QETElementEditor::readSettings()
 
 	QVariant state = settings.value("elementeditor/state");
 	if (state.isValid()) {
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+		if (!restoreState(state.toByteArray())) {
+			settings.remove("elementeditor/state");
+		}
+#else
 		restoreState(state.toByteArray());
+#endif
 	}
 
 	auto data = m_elmt_scene->elementData();
--- a/sources/editor/ui/qetelementeditor.cpp
+++ b/sources/editor/ui/qetelementeditor.cpp
@@ -77,8 +77,9 @@ QETElementEditor::QETElementEditor(QWidget *parent) :
 	//ui->m_display_menu->insertMenu(ui->m_zoom_in_action, menu);
 
 	setWindowState(Qt::WindowMaximized);
-	readSettings();
+	readSettings();  // restoreGeometry before show()
 	show();
+	readSettingsState();  // restoreState() must be called after show() in Qt6
 }
 
 /**
@@ -950,20 +951,28 @@ void QETElementEditor::readSettings()
 		restoreGeometry(geometry.toByteArray());
 	}
 
+	auto data = m_elmt_scene->elementData();
+	data.m_drawing_information = settings.value("elementeditor/default-informations", "").toString();
+	m_elmt_scene->setElementData(data);
+}
+
+/**
+ * @brief QETElementEditor::readSettingsState
+ * Restore the window state (docks, toolbars).
+ * Must be called AFTER show() in Qt6 for restoreState() to work correctly.
+ */
+void QETElementEditor::readSettingsState()
+{
+	QSettings settings;
+
 	QVariant state = settings.value("elementeditor/state");
 	if (state.isValid()) {
 #if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
-		if (!restoreState(state.toByteArray())) {
-			settings.remove("elementeditor/state");
-		}
+		restoreState(state.toByteArray());
 #else
 		restoreState(state.toByteArray());
 #endif
 	}
-
-	auto data = m_elmt_scene->elementData();
-	data.m_drawing_information = settings.value("elementeditor/default-informations", "").toString();
-	m_elmt_scene->setElementData(data);
 }
 
 /**
--- a/sources/editor/ui/qetelementeditor.h
+++ b/sources/editor/ui/qetelementeditor.h
@@ -116,6 +116,7 @@ class QETElementEditor : public QMainWindow
 	private:
 		bool canClose();
 		void readSettings();
+		void readSettingsState();
 		void writeSettings() const;
 		void setupActions();
 		void updateAction();
--- a/sources/qetapp.cpp
+++ b/sources/qetapp.cpp
@@ -1975,6 +1975,7 @@ void QETApp::openTitleBlockTemplate(const TitleBlockTemplateLocation &location,
 	qet_template_editor -> setOpenForDuplication(duplicate);
 	qet_template_editor -> edit(location);
 	qet_template_editor -> show();
+	qet_template_editor -> readSettingsState();  // must run after show() in Qt6
 }
 
 /**
@@ -1986,6 +1987,7 @@ void QETApp::openTitleBlockTemplate(const QString &filepath) {
 	QETTitleBlockTemplateEditor *qet_template_editor = new QETTitleBlockTemplateEditor();
 	qet_template_editor -> edit(filepath);
 	qet_template_editor -> show();
+	qet_template_editor -> readSettingsState();  // must run after show() in Qt6
 }
 
 /**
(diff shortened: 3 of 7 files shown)


ispyisail (changes requested):
The diagram-editor half of this fix is correct, but `QETElementEditor` and `QETTitleBlockTemplateEditor` don't actually get fixed by the `#if QT_VERSION >= 6 ... settings.remove(...)` wrapper — their `restoreState()` calls still run before `show()`:

- `QETElementEditor`'s constructor: `readSettings()` at `qetelementeditor.cpp:80`, `show()` at line 81 — same order as before.
- `QETTitleBlockTemplateEditor`: `readSettings()` runs inside the constructor; `show()` is called separately afterward by three different call sites (`QETApp::openTitleBlockTemplate()` x2, `newTemplate()`) — same ordering problem.

Given the root cause here (Qt6 requires the window visible for `restoreState()`), these two will keep failing to restore exactly as before, except now the `settings.remove(...)` guard will delete the saved `elementeditor/state` / `titleblocktemplateeditor/state` value on every single Qt6 launch, since the call deterministically fails every time. That's worse than the pre-patch behavior — the setting doesn't just stay inert, it gets wiped, so even a later correct fix finds nothing left to restore for existing users.

I built and verified a fix applying the same split-call pattern this PR already uses for `QETDiagramEditor`, extended to the other two windows (`QETTitleBlockTemplateEditor` needed it threaded through all 3 of its `show()` call sites, since it isn't shown from its own constructor). Compiles clean against current master, no new warnings. Diff:

```diff
diff --git a/sources/editor/ui/qetelementeditor.cpp b/sources/editor/ui/qetelementeditor.cpp
index 0a05d64b4..ea5847158 100644
--- a/sources/editor/ui/qetelementeditor.cpp
+++ b/sources/editor/ui/qetelementeditor.cpp
@@ -77,8 +77,9 @@ QETElementEditor::QETElementEditor(QWidget *parent) :
 	//ui->m_display_menu->insertMenu(ui->m_zoom_in_action, menu);
 
 	setWindowState(Qt::WindowMaximized);
-	readSettings();
+	readSettings();  // restoreGeometry before show()
 	show();
+	readSettingsState();  // restoreState() must be called after show() in Qt6
 }
 
 /**
@@ -950,6 +951,20 @@ void QETElementEditor::readSettings()
 		restoreGeometry(geometry.toByteArray());
 	}
 
+	auto data = m_elmt_scene->elementData();
+	data.m_drawing_information = settings.value("elementeditor/default-informations", "").toString();
+	m_elmt_scene->setElementData(data);
+}
+
+/**
+ * @brief QETElementEditor::readSettingsState
+ * Restore the window state (docks, toolbars).
+ * Must be called AFTER show() in Qt6 for restoreState() to work correctly.
+ */
+void QETElementEditor::readSettingsState()
+{
+	QSettings settings;
+
 	QVariant state = settings.value("elementeditor/state");
 	if (state.isValid()) {
 #if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
@@ -960,10 +975,6 @@ void QETElementEditor::readSettings()
 		restoreState(state.toByteArray());
 #endif
 	}
-
-	auto data = m_elmt_scene->elementData();
-	data.m_drawing_information = settings.value("elementeditor/default-informations", "").toString();
-	m_elmt_scene->setElementData(data);
 }
 
 /**
diff --git a/sources/editor/ui/qetelementeditor.h b/sources/editor/ui/qetelementeditor.h
index 579d9bab4..b164017f8 100644
--- a/sources/editor/ui/qetelementeditor.h
+++ b/sources/editor/ui/qetelementeditor.h
@@ -116,6 +116,7 @@ class QETElementEditor : public QMainWindow
 	private:
 		bool canClose();
 		void readSettings();
+		void readSettingsState();
 		void writeSettings() const;
 		void setupActions();
 		void updateAction();
diff --git a/sources/qetapp.cpp b/sources/qetapp.cpp
index d8afcc34b..0463be2ee 100644
--- a/sources/qetapp.cpp
+++ b/sources/qetapp.cpp
@@ -1975,6 +1975,7 @@ void QETApp::openTitleBlockTemplate(const TitleBlockTemplateLocation &location,
 	qet_template_editor -> setOpenForDuplication(duplicate);
 	qet_template_editor -> edit(location);
 	qet_template_editor -> show();
+	qet_template_editor -> readSettingsState();  // must run after show() in Qt6
 }
 
 /**
@@ -1986,6 +1987,7 @@ void QETApp::openTitleBlockTemplate(const QString &filepath) {
 	QETTitleBlockTemplateEditor *qet_template_editor = new QETTitleBlockTemplateEditor();
 	qet_template_editor -> edit(filepath);
 	qet_template_editor -> show();
+	qet_template_editor -> readSettingsState();  // must run after show() in Qt6
 }
 
 /**
diff --git a/sources/titleblock/qettemplateeditor.cpp b/sources/titleblock/qettemplateeditor.cpp
index 6277a25d5..2249b9be8 100644
--- a/sources/titleblock/qettemplateeditor.cpp
+++ b/sources/titleblock/qettemplateeditor.cpp
@@ -358,6 +358,7 @@ void QETTitleBlockTemplateEditor::newTemplate()
 	QETTitleBlockTemplateEditor *qet_template_editor = new QETTitleBlockTemplateEditor();
 	qet_template_editor -> edit(TitleBlockTemplateLocation());
 	qet_template_editor -> show();
+	qet_template_editor -> readSettingsState();  // must run after show() in Qt6
 }
 
 /**
@@ -645,8 +646,19 @@ void QETTitleBlockTemplateEditor::readSettings()
 	// window size and position
 	QVariant geometry = settings.value("titleblocktemplateeditor/geometry");
 	if (geometry.isValid()) restoreGeometry(geometry.toByteArray());
+}
+
+/**
+	@brief QETTitleBlockTemplateEditor::readSettingsState
+	Restore the window state (docks, toolbars).
+	Must be called AFTER show() in Qt6 for restoreState() to work correctly
+	-- callers are responsible for calling this after show(), since this
+	window isn't shown by its own constructor.
+*/
+void QETTitleBlockTemplateEditor::readSettingsState()
+{
+	QSettings settings;
 
-	// window state (toolbars, docks...)
 	QVariant state = settings.value("titleblocktemplateeditor/state");
 	if (state.isValid()) {
 #if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
diff --git a/sources/titleblock/qettemplateeditor.h b/sources/titleblock/qettemplateeditor.h
index 67ac3c231..cfb6709b7 100644
--- a/sources/titleblock/qettemplateeditor.h
+++ b/sources/titleblock/qettemplateeditor.h
@@ -108,6 +108,7 @@ class QETTitleBlockTemplateEditor : public QETMainWindow {
 
 	public slots:
 	void readSettings();
+	void readSettingsState();
 	void writeSettings();
 	void selectedCellsChanged(const QList<TitleBlockCell *>&);
 	void duplicateCurrentLocation();
```

I don't have a Qt6 environment handy to reproduce the original symptom end-to-end, so this is verified by compilation + structural parity with the diagram-editor fix, not a live before/after on Qt6 — worth a maintainer or your own Qt6 build confirming the restore actually kicks in for these two windows before merging.

Kellermorph:
Thanks for the thorough review and the detailed diff.

I've applied all your suggestions:

QETElementEditor: Split readSettings()/readSettingsState(), added readSettingsState() call after show() in the constructor
QETTitleBlockTemplateEditor: Same split, added readSettingsState() after show() in newTemplate() and both QETApp::openTitleBlockTemplate() overloads
Removed settings.remove() guards: As you pointed out, these would destructively delete the saved state on every Qt6 launch since restoreState() deterministically fails when called before show()
Builds clean against current master. Please re-review.

ispyisail (approved):
Verified this directly — fetched the updated branch and rebuilt (473/473, clean, no new warnings). All three `readSettingsState()` calls now genuinely run after `show()`: `QETElementEditor`'s constructor, and all three of `QETTitleBlockTemplateEditor`'s external call sites (`newTemplate()` and both `QETApp::openTitleBlockTemplate()` overloads). This resolves the bug for all three windows now, not just the diagram editor.

Two small, non-blocking nits, not worth another round:
1. Removing the `settings.remove()`-on-failure guard from `QETElementEditor` and `QETTitleBlockTemplateEditor` left the `#if QT_VERSION >= QT_VERSION_CHECK(6,0,0) / #else / #endif` wrapper in place around it — both branches now just call `restoreState(state.toByteArray());` with nothing else, so the conditional is dead code. Harmless, just noise; could collapse to one plain call whenever convenient.
2. `QETDiagramEditor` still keeps its remove-on-failure guard (correctly — it's safe now that the call runs after `show()`), so there's a minor inconsistency where one of the three windows self-heals on a corrupted saved state and the other two don't.

Neither affects correctness. Thanks for the quick turnaround.

Kellermorph:
Good catches on both nits. I've cleaned them up:

Dead #if QT_VERSION removed: Both branches in QETElementEditor and QETTitleBlockTemplateEditor were calling restoreState(state.toByteArray()) unconditionally — collapsed to a single call.
Consistent settings.remove() guard: Added to all three editors so they all self-heal on a corrupted saved state, not just the diagram editor.
Builds clean, branch is updated.

PR #685, opened 2026-08-07
Revert "Auto-break conductor"

scorpio810:
Reverts qelectrotech/qelectrotech-source-mirror#639

PR #770, opened 2026-08-22
Fix paste breaking existing conductors

Kellermorph:
Copying a component and pasting it on the same page broke existing conductors and reconnected them through the pasted element. This was unwanted behavior — the pasted element should be completely independent.

Root cause: autoBreakConductors() was called in PasteDiagramCommand::redo() (added in commit 39b1e836b). When pasting at the same position, the pasted element's terminals overlap with the original's conductor endpoints, causing autoBreakConductors() to detect those conductors as "needing a break" and reconnecting them through the new element.

Fix: Removed the autoBreakConductors() call from PasteDiagramCommand. This function is only needed when placing new elements from the collection (DiagramEventAddElement, ElementsMover), not when pasting copied elements. This restores the pre-39b1e836b behavior where pasted elements are fully independent.

### DIFF
--- a/sources/diagramcommands.cpp
+++ b/sources/diagramcommands.cpp
@@ -17,9 +17,7 @@
 */
 #include "diagramcommands.h"
 
-#include "autobreakconductor.h"
 #include "diagram.h"
-#include "qetproject.h"
 #include "qetgraphicsitem/conductortextitem.h"
 #include "qetgraphicsitem/element.h"
 #include "qetgraphicsitem/elementtextitemgroup.h"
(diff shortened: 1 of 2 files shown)


ispyisail (approved):
Reviewed against the auto-break-conductor feature's full history (#639 → #685 reverted → e8f80697f reapplied → 39b1e836b extended into PasteDiagramCommand → #700). This PR removes exactly the PasteDiagramCommand scope-creep from 39b1e836b, restoring auto-break to its originally-scoped element-placement case. Pure deletion, no new surface — looks correct.

PR #771, opened 2026-08-22
Add configurable distance between label and slave XRef

Kellermorph:
Add a SpinBox to the XRef properties widget to allow users to configure
the distance (in pixels) between the element label and the slave cross
reference text. The offset applies to all 5 position modes (Top, Bottom,
Left, Right, Text alignment).

Previously the distance was hardcoded to 0 with no way to adjust it.
The new setting ranges from -50px to +100px (default 0), is stored in
both project XML and QSettings, and is available per element type
(coil, protection, commutator, PLC).

Changed files:
- xrefproperties.h/cpp: add m_slave_offset property with serialization
- xrefpropertieswidget.ui/cpp: add SpinBox for the new property
- dynamicelementtextitem.h/cpp: apply slave_offset in setXref_item()
- elementtextitemgroup.h/cpp: apply slave_offset in adjustSlaveXrefPos()

### DIFF
--- a/sources/editor/ui/dynamictextfieldeditor.cpp
+++ b/sources/editor/ui/dynamictextfieldeditor.cpp
@@ -245,7 +245,8 @@ void DynamicTextFieldEditor::fillInfoComboBox()
 	else {
 		strl = QETInformation::elementInfoKeys();
 
-		bool is_plc_slave = (type == ElementData::Slave
+		bool is_slave = (type == ElementData::Slave);
+		bool is_plc_slave = (is_slave
 							 && ed.m_slave_type == ElementData::PLCSlave);
 
 		if (is_plc_slave) {
@@ -273,6 +274,10 @@ void DynamicTextFieldEditor::fillInfoComboBox()
 			strl.removeAll(QETInformation::ELMT_PLC_T2);
 			strl.removeAll(QETInformation::ELMT_PLC_T3);
 			strl.removeAll(QETInformation::ELMT_PLC_T4);
+		}
+
+		if (is_slave) {
+			strl.prepend(QETInformation::ELMT_XREF);
 		}
 	}
 
--- a/sources/properties/xrefproperties.cpp
+++ b/sources/properties/xrefproperties.cpp
@@ -36,6 +36,7 @@ XRefProperties::XRefProperties()
 	m_master_label = "%f-%l%c";
 	m_slave_label = "(%f-%l%c)";
 	m_offset = 0;
+	m_slave_offset = 0;
 	m_xref_pos = Qt::AlignBottom;
 }
 
@@ -56,6 +57,8 @@ void XRefProperties::toSettings(QSettings &settings,
 	settings.setValue(prefix % "snapto", snap);
 	int offset = m_offset;
 	settings.setValue(prefix % "offset", offset);
+	int slave_offset = m_slave_offset;
+	settings.setValue(prefix % "slave_offset", slave_offset);
 	QString master_label = m_master_label;
 	settings.setValue(prefix % "master_label", master_label);
 	QString slave_label = m_slave_label;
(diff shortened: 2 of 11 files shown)


ispyisail (approved):
Checked all 5 position-mode branches in DynamicElementTextItem::setXref_item() — the offset sign is applied consistently in the "away from element" direction in every case, and both of that class's call sites were updated to pass xrp.slaveOffset(). Serialization (toSettings/fromSettings/toXml/fromXml/operator==) is complete and matches the sibling m_offset field's pattern exactly.

One gap worth a follow-up: ElementTextItemGroup::adjustSlaveXrefPos() has a second call site inside updateAlignment() that isn't updated to pass the offset (that function has no XRefProperties in scope). updateAlignment() fires from addToGroup()/removeFromGroup()/setAlignment()/setVerticalAdjustment()/AddElementTextCommand — all realistic user actions — so for elements using ElementTextItemGroup, the configured slave offset will reset to 0 on those triggers until the next updateXref() call restores it. Cosmetic and self-correcting, not a crash or data-loss risk, but visible enough to be worth a one-line fix threading the offset through that call too.

ChuckNr11:
The references are often quite far removed from the label, so the idea of ​​an adjustable distance isn't bad. 
However, wouldn't it be much more advantageous to have a free text element to which you can assign the reference (similar to your texts for terminals)? Such a text could then be freely formatted and positioned.

Kellermorph:
> Checked all 5 position-mode branches in DynamicElementTextItem::setXref_item() — the offset sign is applied consistently in the "away from element" direction in every case, and both of that class's call sites were updated to pass xrp.slaveOffset(). Serialization (toSettings/fromSettings/toXml/fromXml/operator==) is complete and matches the sibling m_offset field's pattern exactly.
> 
> One gap worth a follow-up: ElementTextItemGroup::adjustSlaveXrefPos() has a second call site inside updateAlignment() that isn't updated to pass the offset (that function has no XRefProperties in scope). updateAlignment() fires from addToGroup()/removeFromGroup()/setAlignment()/setVerticalAdjustment()/AddElementTextCommand — all realistic user actions — so for elements using ElementTextItemGroup, the configured slave offset will reset to 0 on those triggers until the next updateXref() call restores it. Cosmetic and self-correcting, not a crash or data-loss risk, but visible enough to be worth a one-line fix threading the offset through that call too.

Thanks for the thorough review and for catching this!

You're right — updateAlignment() (called from addToGroup/removeFromGroup/setAlignment/setVerticalAdjustment/AddElementTextCommand) has a second call to adjustSlaveXrefPos() that was missing the offset. I've fixed it by threading the offset through from the project's XRefProperties, matching the same pattern used in updateXref().

The fix is in elementtextitemgroup.cpp:265-275 — it now queries the master element's type, retrieves the XRefProperties, and passes slave_offset to adjustSlaveXrefPos().

Kellermorph:
> The references are often quite far removed from the label, so the idea of ​​an adjustable distance isn't bad. However, wouldn't it be much more advantageous to have a free text element to which you can assign the reference (similar to your texts for terminals)? Such a text could then be freely formatted and positioned.

Yes, the next step could be to add another item to the dropdown menu and include that variable in the dynamic text. Good idea.

Kellermorph:
Add a new 'Champ de texte' (Text field) option to the XRef slave position
dropdown in project properties. When selected, the cross-reference text
is stored in element informations under the 'xref' key instead of being
displayed as a separate graphics item next to the label.

This allows users to place the cross-reference text anywhere on the
element by adding a dynamic text field with info 'Réf. croisée'
(Cross-reference) in the element editor.

Changes:
- Add 'Champ de texte' to m_xrefpos_cb using Qt::AlignHCenter as
  internal value (backward compatible, unused flag)
- Add ELMT_XREF ('xref') key to QETInformation with translation
- Show 'Réf. croisée' in dynamic text info combo for all Slave types
- Handle AlignHCenter mode in DynamicElementTextItem::updateXref():
  calculate xref text and store in elementInformations instead of
  creating m_slave_Xref_item; set up connections for future updates
- Same handling in ElementTextItemGroup::updateXref()
- Clean up old m_slave_Xref_item when switching to text field mode

Kellermorph:
Hi @scorpio810 

Sorry to bother you, but is there a reason why the pull request isn't being merged? I'd like to use that feature in my next plan.

PR #772, opened 2026-08-22
Add PDF page import as image

Kellermorph:
This PR adds a new toolbar action "Ajouter un PDF" next to the existing "Ajouter une image" button, allowing users to import individual PDF pages directly onto the diagram as rasterized images.

Changes
New files:

sources/diagramevent/diagrameventaddpdf.h/.cpp — Diagram event handling PDF import (mirrors the existing image import flow)
sources/ui/pdfpagesdialog.h/.cpp — Page selection dialog with live preview
Modified files:

cmake/qet_compilation_vars.cmake — Added Qt::Pdf module dependency and new source files
sources/qeticons.h/.cpp — Added adding_pdf icon (document-import.png)
sources/qetdiagrameditor.cpp — Added "Ajouter un PDF" QAction and trigger handler
How it works
Click "Ajouter un PDF" in the toolbar
Select a PDF file via file dialog
If the PDF has multiple pages: a dialog shows the page count, a spinbox for page selection, and a live preview of the selected page
The selected page is rendered at 150 DPI with a white background
The image follows the cursor; left-click to place, right-click to rotate, Ctrl+scroll to scale
Uses Qt6's QPdfDocument for PDF rendering — no external dependencies required.


plc-user:
Two comments:

I think 150 dpi (which is less than for a fax!) is too low in some cases. In my opinion, a limited selection of dpi options (150, 300, 600) would make sense, considering that the pages should also be clear to read when printed.

Regarding QPdfDocument, search results indicate that Qt5 support is questionable, and you yourself mention that your function was created for/with Qt6: 
Have we completely discontinued Qt5 support for QET in the meantime?
Of course, it makes sense to write new features for the more modern system, but as long as we haven’t completely dropped Qt5, ...

Kellermorph:
thanks for the feedback!

DPI options: Good point. I'll add a DPI combo box (150 / 300 / 600) to the page selection dialog. 150 DPI remains the default as a reasonable compromise for screen use, but users who need print-quality output can pick 300 or 600.

Qt5 support: The project still supports Qt5 (CMake defaults to QT_VERSION_MAJOR 5, there are 22+ #if QT_VERSION guards across the codebase). QPdfDocument is Qt6-only, so the PDF import feature will be conditionally compiled with #if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0). On Qt5 builds the action simply won't appear – everything else continues to work as before.

The page preview in the dialog will also adapt dynamically to the selected DPI setting.

pkess:
I read the description only and I would rallye like to get this. 

Some idea:
You descrived that you can select the Page for import after selecting the Page. In my opinion sometimes you want to get all pdf Pages and use them as background for a Set of new Pages. This would allow to import a falls schematic of an existing installation and add changes by use of qet. You can then add more and more qet symbols to replace the old schematic on demand.


Thank you for the pr

scorpio810:
Humm, you can't save  big PDF  in XML QET 's files .....to base64 encoder/decoder...

pkess:
The question here is what is considered as *big* pdf file. When the import makes it a size of multiple of 100mb i would expect the system to get slower in performance. But the alternative for me is to make screenshots and use them for the same purpose. This would come with the same disadvantages as if the Page are imported in one action. 

Summarized: I think the user should decide if he wants to try to import a set of pdf pages. We could add a warning if more than e.g. 10 pages are imported in one session but there must not be a hard limit.

scorpio810:
I "m think that didn't work with PDF export, no?

scorpio810:
BTW , you can use my workarounds...
https://youtu.be/RSRjqiLXuiw
https://youtu.be/1QzODcMEjr0

scorpio810:
New patch 
[pdf-import-qtpdf-optional.patch](https://github.com/user-attachments/files/31346375/pdf-import-qtpdf-optional.patch)

Edit: merged!

plc-user:
I changed the initial directory to “Documents” and adjusted a variable name.

Die Deutsche Übersetzung überlasse ich dem Ersteller des PR. :wink:

Kellermorph:
> I changed the initial directory to “Documents” and adjusted a variable name.
> 
> Die Deutsche Übersetzung überlasse ich dem Ersteller des PR. 😉

mache ich, sobald alles gemerged wurde. also alles wie gehabt :D

scorpio810:
DPI combo box (150 / 300 / 600) , maybe add 1200 DPI for some users will want to print In A0  printing tracers?

PR #774, opened 2026-08-22
Fix PLC table dangling reference warning and row deletion

Kellermorph:
Two fixes for the PLC master element table:

Fix dangling reference warning (masterpropertieswidget.cpp): Cache the result of plcMasterData() in a local variable to avoid creating temporaries in the loop. The previous code took a const reference to an element of a temporary that was destroyed at the end of the expression, triggering -Wdangling-reference.
Fix row deletion in the element editor (elementpropertieseditorwidget.cpp): Change setSelectionBehavior from SelectItems to SelectRows. The - button called selectedRows() which always returned empty when only individual cells were selectable, making row deletion silently do nothing.

### DIFF
--- a/sources/editor/ui/elementpropertieseditorwidget.cpp
+++ b/sources/editor/ui/elementpropertieseditorwidget.cpp
@@ -690,7 +690,7 @@ void ElementPropertiesEditorWidget::createPlcConfigWidgets()
 	m_plc_table->horizontalHeader()->resizeSection(2, 150);
 	m_plc_table->horizontalHeader()->resizeSection(3, 150);
 	m_plc_table->horizontalHeader()->resizeSection(4, 100);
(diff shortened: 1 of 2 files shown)


ispyisail (approved):
Confirmed both root causes directly. plcRemoveRow() calls selectionModel()->selectedRows(), which per Qt's model/view semantics only returns non-empty when the selection behavior includes whole-row selection — SelectItems never produces that, so the fix is correct. And the original `ed.plcMasterData().ios.at(row)` reference genuinely dangles: C++ temporary lifetime extension doesn't cover a reference obtained through a member-function call (.at()) on a temporary, so the PlcMasterData temporary (and its owned ios vector) is destroyed at the end of the statement while orig_io still points into it. That's a real use-after-free -Wdangling-reference is right to flag, not just noise to silence. Good fixes, both minimal and correct.

PR #777, opened 2026-08-24
Fix folio-report link picker rendering valid candidates as blank rows

### LINKED ISSUE #735: Folio report link picker: valid candidates render as blank rows (list looks empty)

ispyisail:
Fixes #735.

## Problem

In the Report de folio tab of a report arrow's properties
(`LinkSingleElementWidget`, `AllReport` branch), a valid candidate arrow
renders as a completely blank row when it has no conductor attached yet
— the normal state right after placing a new folio-reference arrow.

`buildTree()` was correctly finding the candidate (confirmed in the
original report via temporary instrumentation), but `setUpHeaderLabels()`
put five conductor-property columns first (N° de fil, Fonction, Tension/
Protocole, Couleur, Section), and every one of those is legitimately
empty before a conductor is linked. The three columns that would
actually identify the row — folio number/label, Position, Titre de
folio — were appended last, past the default column widths, so the row
looked entirely empty until clicked.

## Fix

Reorder both `setUpHeaderLabels()` and `buildTree()`'s `AllReport`
branch to lead with the three always-populated columns, matching the
existing `Slave` branch (which leads with `actualLabel()` and doesn't
have this problem). Conductor properties — legitimately blank pre-link
— now trail instead of leading. Column count and the settings-driven
folio-label-vs-number choice are unchanged; only the order moved, and
header/data order were kept in lockstep in both settings branches.

## Testing

Builds clean, no new warnings. Verified by static read that the header
label list and the `str_list` population in `buildTree()` produce the
same 8-column order in both `genericpanel/folio` settings states.

Note for anyone who already has a saved column-width state for this
tree (`link-element-widget/report-state` in QSettings): since the
column count is unchanged, the saved widths still apply positionally,
just to a different column than before the reorder — a one-time
cosmetic adjustment, not a regression.

### DIFF
--- a/sources/ui/linksingleelementwidget.cpp
+++ b/sources/ui/linksingleelementwidget.cpp
@@ -298,12 +298,37 @@ void LinkSingleElementWidget::buildTree()
 	}
 	
 	else if (m_element->elementData().m_type & ElementData::AllReport)
-	{	
+	{
 		for(const auto &elmt : elmt_vector)
 		{
 			QStringList search_list;
 			QStringList str_list;
-			
+
+				//Identifying columns first: a candidate with no conductor
+				//attached yet (the normal state for a freshly placed report
+				//arrow) would otherwise render as an all-blank row, since
+				//every conductor-property column below can legitimately be
+				//empty. Folio number/position/title never are.
+			if (Diagram *diag = elmt->diagram())
+			{
+				if (settings.value(QStringLiteral("genericpanel/folio"), false).toBool())
+				{
+					autonum::sequentialNumbers seq;
+					QString F =autonum::AssignVariables::formulaToLabel(diag->border_and_titleblock.folio(), seq, diag, elmt);
+					str_list << F;
+				}
+				else
+				{
+					str_list << QString::number(diag->folioIndex() + 1);
+				}
+				str_list << diag->convertPosition(elmt->scenePos()).toString();
+				str_list << diag->title();
+			}
+			else
+			{
+				qDebug() << "In method void LinkSingleElementWidget::updateUi(), provided element must be in a diagram";
+			}
+
 			if (!elmt->conductors().isEmpty())
 			{
 				ConductorProperties cp = elmt->conductors().first()->properties();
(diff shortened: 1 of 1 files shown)


ispyisail:

scorpio810:
I 'm think that @Kellermorph was worked on it and already merged, no?

Kellermorph:
Are you asking whether the cross-reference to a slide reference is preserved when copying? If so, yes, I've already fixed that. When I build the current version, the cross-reference is empty after copying.

ispyisail:
Right that it is already fixed — though it was @enesgursoy6110 rather than @Kellermorph: #822, merged 9 September, which closed #735.

Same idea done better. #822 moves the visual sections and leaves the logical column IDs alone, so saved header layouts stay valid; mine reordered them logically, which is why it had to carry a caveat about saved column widths landing on different columns. Nothing here worth keeping.

Closing.

PR #779, opened 2026-08-24
Make Diagram::toXml() serialize elements and conductors deterministically

### LINKED ISSUE #754: Saving an unmodified project is not reproducible: element order and legacy terminal ids change on every save
Saving a project without editing it produces a **different file every time**. The
same input, resaved repeatedly with the same binary, yields a different byte
stream on each run — element blocks are emitted in a different sequence and
legacy `<terminal id=...>` values shuffle with them.

This makes `.qet` files impractical to keep under version control: every
open-and-save produces a large spurious diff, and two users saving an unmodified
project get different files.

### Reproduction

QElectroTech `0.100.1-dev` (master `7307a59c1`), Qt5, Linux, each run in an
isolated `HOME`/`XDG_CONFIG_HOME`/`XDG_DATA_HOME`:

```bash
for i in $(seq 6); do
  qelectrotech -platform offscreen --resave examples/ShellyParts.qet /tmp/o$i.qet
done
md5sum /tmp/o*.qet | awk '{print $1}' | sort -u | wc -l
```

**Expected:** `1` — resaving an unmodified project is idempotent.
**Actual:** `6` — six runs, six distinct outputs.

`ShellyParts.qet` is a good minimal case because it contains **zero
conductors**, so this is not conductor-related.

**Control, same binary and conditions:** `examples/pinball_williams_em.qet` is
stable (3 runs, 1 md5), so it is not environmental — it depends on the project's
item layout.

A representative diff between two runs of the same input:

```
< <element type="...shelly_rgbw2.elmt" x="480" y="330" .../>
> <element type="...shelly1pm.elmt"    x="330" y="160" .../>
<     <terminal x="-13.5673" y="40.9602" id="12" orientation="2"/>
```

### Cause

`Diagram::toXml()` serialises by walking `items()`:

```cpp
// sources/diagram.cpp:968
for(QGraphicsItem *qgi : items())
```

`QGraphicsScene::items()` returns items in stacking order, and items sharing a
Z value are ordered by the scene's internal index rather than by any stable,
content-derived key. That order is not guaranteed to be reproducible across
processes, so the element sequence in the output varies run to run.

The legacy terminal ids inherit the same instability. `sources/diagram.cpp:1039`
builds

```cpp
QHash<Terminal *, int> table_adr_id;
```

from scratch on every save, keyed by **pointer** and filled in that same
`items()` order. Pointer-keyed `QHash` iteration depends on heap addresses, so
the assigned integers differ between processes: three resaves of one file wrote
`terminal1="30"`, `"11"` and `"25"` for the same conductor.

### Suggested direction

Serialise from a deterministically ordered sequence rather than raw `items()` —
e.g. sort the collected items by a stable content key (uuid where present,
otherwise `(x, y, type)`) before writing. The same ordering would then make
`table_adr_id` assignment reproducible.

I have not opened a PR for this: the right ordering key is a design decision for
the maintainers, and changing serialisation order touches file compatibility, so
it seemed better to report the defect with a reproduction than to impose a fix.
Happy to prepare a patch if you can indicate the preferred key.

### Why it is worth fixing

- `.qet` files cannot be meaningfully diffed or version-controlled today.
- It blocks any automated round-trip/regression testing of the save path, since
  a byte comparison of "unchanged" output always fails.

ispyisail:
Fixes #754.

## Problem

Saving an unmodified project produced a different byte stream on every
run. `QGraphicsScene::items()` returns items in stacking order, and
ties between same-Z items follow the scene's internal index rather
than any content-derived order — not reproducible across process runs.
The legacy terminal-id table inherits the same instability, since ids
are assigned sequentially in element order.

## Fix, part 1: deterministic order

Sort `list_elements` and `list_conductors` into a deterministic order
before serializing, using a key built from **position**, not
`uuid()`. This was a deliberate choice, not the first thing I tried:
for an item with no persisted `uuid` attribute, `fromXml()` invents a
fresh random one on every load, so sorting by uuid would still be
non-deterministic across process runs for exactly the legacy files
this issue is about. Position (`pos().x()`/`pos().y()`) is what's
actually written to and read from the file — never regenerated.

## Fix, part 2: found while verifying part 1

Even with order fixed, resaves still weren't byte-identical.
`Conductor::toXml()` unconditionally writes `m_uuid` back out —
including the synthetic value `fromXml()` just invented when the file
had no `uuid` attribute. Checked every example project with
conductors: **none of them have a persisted conductor uuid at all**,
so this alone meant no project with conductors could ever resave
identically, independent of ordering.

`Conductor` gets a `m_persist_uuid` flag (false only when the uuid
it's holding was synthesized rather than loaded from the file), and
`toXml()` stops writing a value that was never meant to be permanent.
`newUuid()` (used when duplicating a conductor to avoid a primary-key
collision in the wiring database) sets the flag back to true, since
that uuid *is* meant to stick.

## Deliberately NOT doing the same for Element

I checked whether the identical fix applies to `Element::uuid()` and
concluded it doesn't, safely: element uuids are cross-referenced by
*other* elements' `<links_uuids>` blocks for master/slave/report
linking, matched on load via `elmt->uuid() == stored_uuid`
(`element.cpp`, `tmp_uuids_link`). Making an element's own uuid
non-persistent would silently break that match for any linked element
that doesn't already have one — a real regression, not a theoretical
one. Left as a smaller, separate residual (see Testing below) rather
than risk it in this PR.

## Testing

8 example projects (everything with conductors, plus the two
zero-conductor control cases already used in `FINDINGS.md` F002), 5
resaves each in isolated `HOME`/`XDG_CONFIG_HOME`/`XDG_DATA_HOME`
environments:

- Element and conductor **order**: 0 churning sections across the
  whole corpus. Before the fix, the majority of diagrams in
  `industrial.qet`, `m_000.qet` and `tremie_vibrante.qet` churned on
  effectively every run.
- Conductor uuid **values**: 0 churn (before: every conductor in
  every project, since none have a persisted uuid).
- **6 of 8 projects are now byte-for-byte identical (md5) across all
  5 runs.** The remaining 2 (`industrial.qet`, `m_000.qet`) differ
  only in the handful of element uuids covered by the Element
  residual above — confirmed by checking those specific uuids, not
  assumed from the md5 mismatch alone.
- Element/conductor counts before and after resave match exactly on
  every project — the sort doesn't drop or duplicate anything.

Full project builds clean, no new warnings.

### DIFF
--- a/sources/diagram.cpp
+++ b/sources/diagram.cpp
@@ -40,8 +40,58 @@
 #include "undocommand/addelementtextcommand.h"
 #include "qetinformation.h"
 #include "qetproject.h"
+#include <algorithm>
 #include <cassert>
 #include <math.h>
+
+namespace {
+	/// Format a position as a string that sorts the same way the numbers do
+	/// (fixed precision, so "-" and decimals compare correctly as text).
+	QString positionKey(const QPointF &pos)
+	{
+		return QStringLiteral("%1|%2")
+				.arg(pos.x(), 0, 'f', 4)
+				.arg(pos.y(), 0, 'f', 4);
+	}
+
+	/// Sort key for Diagram::toXml()'s <elements> block: the element's own
+	/// diagram-local position, exactly what it's already saved as (x/y),
+	/// never invented or regenerated. uuid() is deliberately NOT used here:
+	/// for an element with no persisted uuid attribute, fromXml() invents a
+	/// fresh random one on every load, so sorting by uuid would still be
+	/// non-deterministic across process runs for any legacy file.
+	QString elementSortKey(Element *elmt)
+	{
+		return positionKey(elmt->pos());
+	}
+
+	/// Sort key for a terminal: its parent element's position, then the
+	/// terminal's own position local to that element (from the .elmt
+	/// definition, fixed regardless of where the element is placed).
+	QString terminalSortKey(Terminal *terminal)
+	{
+		if (!terminal)
+			return QString();
+		Element *parent = terminal->parentElement();
+		return (parent ? positionKey(parent->pos()) : QStringLiteral("?"))
+				+ QLatin1Char(':') + positionKey(terminal->pos());
+	}
+
+	/// Sort key for Diagram::toXml()'s <conductors> block. Built from both
+	/// endpoints' terminalSortKey(), not Conductor::uuid(): in every example
+	/// project checked, conductors have no persisted uuid attribute at all,
+	/// so uuid() is a freshly-minted random value on every load -- exactly
+	/// as unusable for cross-run determinism as the element case above, just
+	/// with no persisted fallback to reach for instead. Canonicalised
+	/// (smaller key first) since a conductor's two ends are unordered for
+	/// this purpose.
+	QString conductorSortKey(Conductor *cond)
+	{
+		QString a = terminalSortKey(cond->terminal1);
+		QString b = terminalSortKey(cond->terminal2);
+		return (a <= b) ? (a + QLatin1Char('>') + b) : (b + QLatin1Char('>') + a);
+	}
+}
 
 int Diagram::xGrid  = 10;
 int Diagram::yGrid  = 10;
@@ -1013,6 +1063,20 @@ QDomDocument Diagram::toXml(bool whole_content, bool is_copy_command) {
 			}
 		}
 	}
+
+		// items() returns items in stacking order, which is not guaranteed
+		// reproducible across processes (ties between same-Z items follow
+		// the scene's internal index, not any content-derived order) -- so
+		// without this, saving an unmodified project produces a different
+		// byte stream on every run. Elements and conductors are the two
+		// blocks observed to actually churn across the example corpus;
+		// sort them into a deterministic, content-derived order before
+		// serializing. This also fixes the legacy terminal-id churn below,
+		// since those ids are assigned sequentially in element order.
+	std::stable_sort(list_elements.begin(), list_elements.end(),
+			  [](Element *a, Element *b) { return elementSortKey(a) < elementSortKey(b); });
+	std::stable_sort(list_conductors.begin(), list_conductors.end(),
+			  [](Conductor *a, Conductor *b) { return conductorSortKey(a) < conductorSortKey(b); });
 
 	// correspondence table between the addresses of the terminals and their ids
 	// table de correspondance entre les adresses des bornes et leurs ids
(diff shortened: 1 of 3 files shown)


scorpio810:
Thanks for this — wanted to flag some prior history here since it's directly relevant and you may not have come across it.

This PR tackles tag order (elements/conductors within <diagram>), which complements two earlier pieces of work aimed at the same underlying goal — making QET files diff-friendly under version control:

QT_HASH_SEED=0 (PR #415, plc-user, merged into 0.100 via main.cpp): fixed attribute order randomization caused by Qt's hash-seed randomization (a security mitigation against algorithmic-complexity attacks that had the side effect of shuffling XML attribute order on every save). Confirmed still present in the 0.100 stable tag.
Further back, De-Backer's "QElectroTech Element Updater" (2020-2021, started at plc-user's request) explicitly sorted attributes alphabetically and normalized .elmt structure for the same reason — this predates #415 and was aimed specifically at .elmt files rather than .qet projects.
Joshua's comment on the #415 thread is worth quoting for context on why uuid exists at all: "By the past the attribute was sorted in qet [...] but Qt changed the way the dom xml writer works [...] This is why we have an 'element definition' uuid. Uuid is much much better for qet (not for git diff)."

So your fix here is the missing piece: #415 handles attribute order within a tag, plc-user's work handled .elmt normalization, and this PR handles tag order within .qet projects. Together they cover the three main sources of spurious diff noise for anyone tracking QET files in git — which several of us (myself included) actually do for real documentation projects, so this is genuinely useful outside of just closing #754.

ispyisail:
Your welcome

ispyisail:
Pushed a fix for a sort-order bug in `positionKey()` (b2f4ef5d2). Plain fixed-precision formatting (`%.4f`) does not sort in numeric order once the integer part has a different digit count -- e.g. `"15.0000"` sorted before `"5.0000"` as text, even though 15 > 5 -- which silently broke the determinism goal of this branch for any diagram with coordinates spanning more than one digit width. Confirmed with a quick numeric check before and after. Now shifts into a non-negative range and zero-pads to a fixed width so the string sorts the same way the number does, including negatives. Builds cleanly.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

scorpio810:
Thanks for the fix, this looks solid. The width margin in coordinateKey() is more than enough for any realistic diagram coordinate, and qRound64 preserves the original .4f precision so there's no drift for already-stored values.

Could you open this as a PR (rather than a direct commit) so we get CI coverage on it? It'd also be good to add a small unit test locking in the sort order — a couple of positions spanning different digit widths (including negatives) compared before/after positionKey(), so this class of bug can't silently reappear.

Edit: merged.

ispyisail:
Opened #815 for the unit test you asked for, since this PR is already merged: adds tst_diagramsortkeys.cpp covering digit-width and negative-boundary cases, with positionKey()/coordinateKey() pulled into a small header so the test exercises the exact same code toXml() uses. Verified it builds and passes.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

PR #784, opened 2026-08-24
fix problem with PDF-links in files with brackets in name

plc-user:
When I see it correctly, the source of links marked as "/URI (...)" fills the whole line in a PDF.
So when replacing that by a "/GoTo ..." - link we need to search not only for a closing bracket, but for a closing bracket followed by a NewLine "\n".
In this case we can handle filenames with round brackets as well.
This works for me on my Debian GNU/Linux stable and unstable.
Could someone try this with win to confirm the fix?

### DIFF
--- a/sources/pdf_links.cpp
+++ b/sources/pdf_links.cpp
@@ -228,7 +228,7 @@ void convertUriToGoTo(const QString &pdfPath)
 
 			// Find closing ')' of the URI value
 			int uriStart = found + sUri.size();
-			int closeParen = data.indexOf(')', uriStart);
+			int closeParen = data.indexOf(")\n", uriStart);
 			if (closeParen == -1) {
 				// Malformed — copy rest verbatim
 				out.append(data.mid(found));


ispyisail (approved):
Verified this with a direct A/B rather than just reading it — built both the current master and this patch, then exported the same project to a PDF path with a paren in it (`Projects (2026)/schema.pdf`, a very plausible real-world case: date-suffixed or versioned folder names).

**Before (current master):** all 4 cross-reference links fail to convert — `/S /URI` markers: 4, `/S /GoTo` markers: 0.

**After (this patch):** all 4 convert correctly — `/S /URI` markers: 0, `/S /GoTo` markers: 4.

Root cause matches what you describe exactly. `injectCrossRefLinks()`'s own comment a few lines above documents "Qt always writes exactly this: `/S /URI\n/URI (file:///...#page=N)\n`" — so the `\n`-terminated pattern was already the documented, intended contract; `convertUriToGoTo()` just wasn't matching on it. The URI value comes from `QUrl::fromLocalFile(outputFileName)`, i.e. the PDF's own output path — so a literal `(`/`)` anywhere in that path (Qt escapes it as `\(`/`\)` per PDF string-literal syntax, e.g. `Projects%20\(2026\)`) was always going to trip the old naive `indexOf(')')`, since that search doesn't know about PDF's escaping and matches the first `)` byte regardless — which is the *escaped* one, not the real terminator. Confirmed that's exactly what was happening by inspecting the raw PDF bytes before and after.

Also ran the full `pdflink-check` regression suite (this project's own gate for this exact mechanism) against the patched binary on the normal example corpus — clean, 0 findings, so this doesn't regress the already-working (no-parens) case.

On your own open question — Windows line endings: Qt's PDF content-stream writer builds the stream as raw bytes via `QIODevice`, not text-mode file I/O, so it shouldn't be subject to the OS's LF→CRLF translation, and I'd expect `\n` to be what's actually emitted on Windows too. I don't have a Windows environment to confirm directly, so treat that as reasoning rather than verification — but the mechanism this depends on (Qt not doing text-mode translation on an explicitly binary content stream) is a safe bet, not a coincidence specific to Linux.

Small, correctly scoped, matches the code's own documented format. Approving.

PR #785, opened 2026-08-24
Make auto_num_locked/potential_isolating/exclude_from_bom case-insensitive

ispyisail:
## Problem

`auto_num_locked`, `potential_isolating` and `exclude_from_bom` are stored as
plain text and compared against the literal `"true"` everywhere they are
read — a SQL predicate in `element_nomenclature_view`, and case-sensitive
`==` checks in `terminal.cpp`, `terminalnumberingdialog.cpp` and
`elementinfowidget.cpp`. The only writer (the checkbox in
`ElementInfoWidget`) always writes lowercase `"true"`/`"false"`, so this is
invisible from the GUI. But the value can also come from a hand-authored
`.elmt`/`.qet` file, where `"True"`, `"TRUE"` or `"tRuE"` read as entirely
plausible spellings of yes and silently do nothing — no error, no warning,
the checkbox just never takes effect.

## Fix

Added `QETInformation::isInfoFlagTrue()` as the one place this comparison is
made, used at all four C++ read sites, so a future fifth site can't
reintroduce the case-sensitivity. The SQL side (`element_nomenclature_view`)
gets `COLLATE NOCASE`, since a C++ helper can't reach into a view definition.

## Verification

Built and tested against `examples/tremie_vibrante.qet` (98 components) via
`--export-bom`:

| stored value | BOM rows | before this fix |
|---|---|---|
| *(absent)* | 98 | 98 |
| `false` | 98 | 98 |
| `true` | **93** | 93 |
| `True` | **93** | 98 *(no effect)* |
| `TRUE` | **93** | 98 *(no effect)* |
| `tRuE` | **93** | 98 *(no effect)* |
| `FALSE` | 98 | 98 |
| `False` | 98 | 98 |

Marking 5 elements `exclude_from_bom="True"` had no effect before this
change (98 rows, unchanged) and now correctly excludes them (93 rows). Same
for `TRUE`/`tRuE`; the `false`-family variants correctly remain no-ops.

Full corpus regression: `--export-bom` output is byte-identical to master on
21 of 22 comparable example projects (`schema_indus.qet` excluded — a known,
unrelated startup-modal hang). The one difference, on `photovoltaique.qet`,
is three rows tied on `ORDER BY label` landing in a different order — same
row count, empty diff once sorted, and a pre-existing non-determinism (the
query has never defined a tiebreaker) unrelated to this change.

### DIFF
--- a/sources/dataBase/projectdatabase.cpp
+++ b/sources/dataBase/projectdatabase.cpp
@@ -583,7 +583,10 @@ void projectDataBase::createElementNomenclatureView()
 						 "di.folio AS folio,"
 						 "e.pos AS position "
 						 " FROM element_info ei, diagram_info di, element e, diagram d"
-						 " WHERE ei.element_uuid = e.uuid AND e.diagram_uuid = d.uuid AND di.diagram_uuid = d.uuid AND (ei.exclude_from_bom IS NOT 'true')");
+						 // COLLATE NOCASE: the writer always stores lowercase "true"/"false",
+						 // but a hand-authored file can spell it "True"/"TRUE" -- those used
+						 // to compare unequal to 'true' and stay in the BOM silently.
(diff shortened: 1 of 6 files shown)


ispyisail:

PR #787, opened 2026-08-24
Cascade-delete a diagram's element/terminal/conductor rows on removal

ispyisail:
## Problem

`removeDiagram()` only ever deleted the diagram's own row. No foreign key in
this schema is declared `ON DELETE CASCADE` (and SQLite foreign-key
enforcement is never turned on for this connection anyway), so removing a
diagram left every `element`, `element_info`, `terminal` and `conductor` row
that belonged to it behind in the database — silently, since nothing reads
them until the next full `updateDB()` rebuild papers over it.

## Why this had never crashed anything

`Diagram::~Diagram()` explicitly walks and deletes its top-level items
through `removeItem()` (which does call `dataBase()->removeElement()`
correctly), but deliberately skips conductors — because a conductor's
destructor touches both of its terminals
(`terminal1->removeConductor(this)`), and those terminals may belong to an
element already destroyed earlier in the same sweep. Conductors are instead
destroyed as a side effect of `Terminal::~Terminal()`'s `qDeleteAll()` on its
own conductor list, which is a plain C++ delete that never goes through
`Diagram::removeItem()` and therefore never calls
`dataBase()->removeConductor()` at all. So the object graph is torn down
safely, but the database is never told about the conductors or their
terminals.

## Fix

Added the missing bulk deletes to `projectDataBase::removeDiagram()` itself,
run while the diagram (and its live scene) still exist. Confirmed
`QETProject::detachDiagram()` emits `diagramRemoved()` (which this class's
constructor connects to this slot) synchronously, before the `Diagram`
object is scheduled for destruction via `deleteLater()`, so nothing here
races the C++ teardown described above. Order matters: `element_info` and
`terminal` have no `diagram_uuid` column of their own, so both are scoped
through a subquery on `element` and must run before `element` itself is
deleted.

## Verification

Called `projectDataBase::removeDiagram()` directly against
`examples/industrial.qet` (50 diagrams) and compared table counts before and
after, with no intervening `updateDB()` call to mask a gap:

```
element=354->335  element_info=354->335  terminal=1087->1033  conductor=671->626  diagram=50->49
```

Every delta matches a direct SQL count for that diagram's own rows exactly
(19 elements, 54 terminals — later cross-checked with a diagram picked for
having the most conductors: 34 elements dropped to 335, terminal count
dropped by exactly 54, matching a `SELECT COUNT(*) FROM terminal WHERE
element_uuid IN (SELECT uuid FROM element WHERE diagram_uuid=...)` run
independently). Both "orphan rows still referencing the removed diagram"
checks read 0 afterward — so the cascade is complete and, just as
importantly, scoped: nothing belonging to the other 49 diagrams moved.

## Separate finding, not fixed here

While building the verification probe above, found that
`QETProject::removeDiagram(Diagram*)` (the synchronous, non-undoable
variant, not the usual GUI `ProjectView::removeDiagram()` path) segfaults if
the enclosing `QETProject` is destroyed before an event-loop iteration lets
its pending `deleteLater()` run. Reproduces identically on unmodified
master, so it predates and is unrelated to this change — a headless caller
is the only realistic way to hit it, which is how this surfaced. Happy to
open a separate issue/PR for it if useful.

### DIFF
--- a/sources/dataBase/projectdatabase.cpp
+++ b/sources/dataBase/projectdatabase.cpp
@@ -225,7 +225,40 @@ void projectDataBase::addDiagram(Diagram *diagram)
 
 void projectDataBase::removeDiagram(Diagram *diagram)
 {
-	m_remove_diagram_query.bindValue(":uuid", diagram->uuid().toString());
+	const QString uuid_str = diagram->uuid().toString();
+
+		//Order matters: element_info and terminal are scoped through a
+		//subquery on element, so they must run before element itself is
+		//deleted below. Without this, a removed diagram left its rows
+		//behind in every one of these tables until the next full
+		//updateDB() rebuild -- invisible day to day, since nothing reads
+		//them meanwhile, but a real inconsistency between the live scene
+		//and the database in between.
+	m_cascade_remove_element_info_query.bindValue(":uuid", uuid_str);
+	if (!m_cascade_remove_element_info_query.exec()) {
+		qDebug() << "projectDataBase::removeDiagram element_info cascade error : "
+				 << m_cascade_remove_element_info_query.lastError();
+	}
+
+	m_cascade_remove_terminal_query.bindValue(":uuid", uuid_str);
+	if (!m_cascade_remove_terminal_query.exec()) {
+		qDebug() << "projectDataBase::removeDiagram terminal cascade error : "
+				 << m_cascade_remove_terminal_query.lastError();
+	}
+
+	m_cascade_remove_conductor_query.bindValue(":uuid", uuid_str);
+	if (!m_cascade_remove_conductor_query.exec()) {
+		qDebug() << "projectDataBase::removeDiagram conductor cascade error : "
+				 << m_cascade_remove_conductor_query.lastError();
+	}
+
+	m_cascade_remove_element_query.bindValue(":uuid", uuid_str);
+	if (!m_cascade_remove_element_query.exec()) {
+		qDebug() << "projectDataBase::removeDiagram element cascade error : "
+				 << m_cascade_remove_element_query.lastError();
+	}
+
+	m_remove_diagram_query.bindValue(":uuid", uuid_str);
 	if (!m_remove_diagram_query.exec()) {
 		qDebug() << "projectDataBase::removeDiagram delete error : " << m_remove_diagram_query.lastError();
 	} else {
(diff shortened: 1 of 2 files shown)


scorpio810:
Reviewed the diff — the cascade logic itself is correct and the ordering (element_info/terminal via subquery on element, before element, before diagram) matches the FK dependency chain described in the commit message. Verification against a 50-diagram project with exact row-count deltas is solid.

One thing worth addressing before merge:

The four cascade deletes + the final diagram delete aren't wrapped in a transaction. Each exec() call is checked individually and logs via qDebug() on failure, but execution continues to the next query regardless — including the final DELETE FROM diagram. If, say, the terminal cascade fails (SQLite busy/locked, unexpected constraint, etc.) while element_info and conductor succeed, the database is left in a partially-cascaded state: some orphaned rows removed, others not, and the diagram row itself still gets deleted on top of that inconsistency. Given the whole point of this PR is closing a silent-inconsistency gap, an all-or-nothing transaction around the five statements would make the fix complete rather than just less-likely-to-fail:


```
m_data_base.transaction();
// ... four cascade execs + final remove_diagram exec ...
if (/* any exec failed */) {
    m_data_base.rollback();
} else {
    m_data_base.commit();
}
```
Not a blocker — the current behavior (leaving orphan rows on any failure) is still strictly better than pre-patch master (leaving orphan rows on every call) — but worth closing while this code path is already open.

ispyisail:
Reviewed this diff. One correctness concern in the new cascade-delete logic:

**`sources/dataBase/projectdatabase.cpp:236-262`** — the new cascade deletes (`element_info`/`terminal`/`conductor`/`element` before deleting a `diagram` row) only `qDebug()`-log errors from each `exec()` and continue regardless. If any cascade query fails (e.g. a locked DB), the parent `diagram` row is still deleted, leaving orphaned child rows behind — the same kind of inconsistency this patch is meant to fix, just triggered by a different failure path (a query error instead of missing cascade logic). Worth checking `exec()`'s return value and aborting/rolling back the delete on failure rather than continuing.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

ispyisail:
Pushed a fix for the swallowed-error issue noted above (7c5f8c6a4): `removeDiagram()` now wraps the cascade delete in a transaction and rolls back + bails on the first failed `exec()`, instead of logging and continuing. A mid-cascade failure can no longer leave the `diagram` row deleted while its `element`/`terminal`/`element_info`/`conductor` rows survive. Builds cleanly.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

PR #788, opened 2026-08-24
Make ProjectAutoNumConfigPage follow ProjectConfigPage's init() contract

ispyisail:
## Problem

`ProjectConfigPage::init()` is documented as *"Typically, you should call
this function in your subclass constructor"* — it runs `initWidgets()`,
`initLayout()`, and (if a project is set) `readValuesFromProject()` and
`adjustReadOnly()`, in that order. `ProjectMainConfigPage`'s constructor
follows this. `ProjectAutoNumConfigPage`'s did not: it called
`initWidgets()`, its own `buildConnections()`, and `readValuesFromProject()`
directly — skipping both `initLayout()` and `adjustReadOnly()` entirely, and
calling `readValuesFromProject()` with no null-project guard.

In practice this is harmless today — this subclass's `initLayout()` and
`adjustReadOnly()` overrides are both empty, and every construction site
happens to pass a real project — but it's exactly the kind of latent
inconsistency called out in the refactor notes for this class ("remove
inconsistent virtual method usage... allow subclasses independent
implementation"). The day someone fills in `adjustReadOnly()` for this page
(e.g. to disable auto-numbering editing on a read-only project, which is
what the empty override's own doc comment says it's for), the constructor
path would silently never call it.

## Fix

Narrow, behaviour-preserving: the constructor now calls `init()` like its
sibling does. `buildConnections()` moves to the end of `initWidgets()`, the
same relative position it held in the constructor, so behaviour for the
paths already exercised is unchanged.

This is deliberately the safe half of the refactor note. Removing the
`init()`/`initWidgets()`/`initLayout()`/`readValuesFromProject()` scaffolding
itself, so subclasses are free to sequence things however they want, is a
real redesign of the `ConfigPage` contract and needs a decision on what
should replace it — not attempted here.

## Verification

GUI capture rather than reading: opened Project Properties on
`examples/industrial.qet`, selected "Numérotation auto", and confirmed:

- the Management tab renders with its saved policy (Conductor/Element
  "Both", "Apply to Entire Project"),
- the Conducteurs tab's combo box comes up pre-populated with the project's
  saved context ("de la nouvelle numérotation"), which on selection
  correctly fills the Type/Valeur/Formule fields.

That confirms both `readValuesFromProject()` and the `buildConnections()`
signal wiring still work end-to-end through the new call sequence.


ispyisail:

ispyisail:
Reopening. Closed in the batch that cleared @scorpio810's review backlog rather than on merit; merged with current `master` (no conflict) and re-reviewed before bringing it back.

The thing worth checking here is the one that would make this dangerous: `buildConnections()` moves into the end of `initWidgets()`, which would double every signal connection if `initWidgets()` were ever called more than once. It is not — the only two call sites are `ProjectConfigPage::init()` and the constructor this PR replaces, so after the change it runs exactly once.

Behaviour is otherwise identical, because both calls `init()` adds are verified empty:

| old | new |
|---|---|
| `initWidgets()` | `initWidgets()`, ending with `buildConnections()` |
| `buildConnections()` | `initLayout()` — `{}` inline in the header |
| `readValuesFromProject()` | `readValuesFromProject()`, now inside `if (m_project)` |
| | `adjustReadOnly()` — empty body |

The null guard is defensive rather than a fix for anything live: `readValuesFromProject()` dereferences `m_project` four times unguarded, but the sole construction site (`projectpropertiesdialog.cpp:42`) always receives a real project from a `ProjectView`. Nothing crashes today.

To be clear about what this is worth: it is code hygiene, not a user-visible fix. The value is that `adjustReadOnly()` — whose own doc comment says it exists to disable editing on a read-only project — would never have been called on this page, so whoever eventually implements it would have found a silent no-op.

**Verified:** Qt 6.10.2 / KF 6.10.0, build clean (0 errors), `ctest` 4/4, merges clean with `master`. The original GUI check described above still stands: Project Properties → "Numérotation auto" renders with the project's saved policy, and the Conducteurs combo comes up pre-populated and fills Type/Valeur/Formule on selection — which exercises both `readValuesFromProject()` and the `buildConnections()` wiring through the new sequence.

PR #790, opened 2026-08-24
Complete the Brazilian Portuguese translation

cezlom:
Brings Brazilian Portuguese from **87.2% to 100%** of the 2850 messages in `lang/qet_pt_BR.ts`.

Rebased on current `master`, so this includes the strings added by the two recent *Update translations files* commits.

### What changed

The 364 messages still marked `unfinished` are now complete:

| | |
|---|---|
| Empty, translated here | 228 |
| Linguist suggestions reviewed and kept | 122 |
| Linguist suggestions corrected | 14 |

These contexts were previously untranslated in full: `ContactGroupSelectionDialog`, `PlcLinkWidget`, `TerminalNumberingDialog`, `ShortcutsConfigPage`, `BackupDialog`, `DiagnosticsReportDialog`, `GuidesPropertiesWidget`, `PdfPagesDialog`, `EdzArchive`.

### Terminology

Follows what the file already established: *borne* → "terminal", *bornier* → "régua de terminais", *folio* → "página", *cartouche* → "bloco de legenda", *schéma* → "esquema", *maître/esclave* → "mestre/escravo", *pivoter* → "girar". PLC terms use the Brazilian abbreviation **CLP**, and `NO`/`NC` contacts use **NA**/**NF**.

### Two fixes worth pointing out

Both were pre-existing suggestions that were wrong, not new translations:

- **`Annuler` read "Desfazer"** (undo) where it is a dialog button next to OK — now "Cancelar".
- **The degree symbol used `U+00BA` MASCULINE ORDINAL INDICATOR** (º) instead of `U+00B0` DEGREE SIGN (°).

### Scope

Only `<translation>` elements are touched — `<source>`, `<location>` and `<comment>` are byte-for-byte unchanged, and no already-approved translation was overwritten. Placeholders (`%1`, `%n`), HTML tags and plural forms were checked against their sources across all 364 messages.

`lang/qet_pt_BR.qm` is regenerated with `lrelease` (2850 finished, 0 unfinished). Only the pt_BR pair is included — the other languages' `.qm` files are stale relative to their `.ts` sources in `master`, but that is out of scope here.


ispyisail (approved):
I don't speak Brazilian Portuguese, so I can't judge translation quality or wording the way I'd review code — flagging that limitation up front rather than pretending otherwise. What I *can* verify mechanically, I did, by diffing the actual before/after `.ts` XML programmatically rather than trusting the PR description:

- **2850 messages before and after, 0 added/removed/reordered.**
- **`unfinished` count: 364 → 0**, exactly matching the PR. Exactly 364 messages have a changed `<translation>` — not one more, not one less — so nothing outside the stated scope was touched.
- **`<source>` and `<comment>` are byte-for-byte identical for every message**, confirmed by diffing all 2850, not sampling.
- **Placeholders (`%1`, `%n`), HTML tag names, and `numerusform` structure** all checked across all 364 changed messages by extracting and comparing sets against the source — 0 mismatches on any of the three.
- **The "Annuler" → "Cancelar" fix** is real: confirmed in `ContactGroupSelectionDialog`, previously translated "Desfazer" (Undo).
- **The `.qm` regeneration claim** checks out: ran `lrelease` on the new `.ts` myself — "2850 finished, 0 unfinished", matching what's stated.

One thing worth a precise callout rather than a blanket pass: **the degree-symbol fix is real but only 1 of 5.** Checked every message where the source is a bare `°` (U+00B0 DEGREE SIGN): `TerminalEditor` is the one this PR fixes (º → °). Four more with the identical mistranslation (source `°`, translated as the wrong `º` MASCULINE ORDINAL INDICATOR) are still in the file this PR ships: `GeneralConfigurationPage`, `IndiTextPropertiesWidget`, `ReplaceConductorDialog`, `TextEditor`. Not a blocker — these are pre-existing, outside the 364-message scope this PR is actually about — but worth knowing since the PR text reads as if the degree-symbol class of mistake was addressed, and four instances of it remain. (For contrast: the several "Nº" abbreviations elsewhere in the diff, e.g. "Nb. contacts" → "Nº contatos", are the *correct* use of that character — Portuguese's ordinal indicator for "número" — not the same bug; only the bare-`°`-source cases are wrong.)

Core deliverable — completing the 364 unfinished messages without touching anything else — is solid and precisely scoped. Approving.

cezlom:
Thanks for the careful review — especially for verifying the diff mechanically instead of taking the description at face value.

You were right about the degree sign: the fix was 1 of 5. I pushed a follow-up commit (8ca62c4a1) that corrects the four remaining messages whose `<source>` is a bare `°` (U+00B0 DEGREE SIGN) but were translated with `º` (U+00BA MASCULINE ORDINAL INDICATOR):

- `GeneralConfigurationPage`
- `IndiTextPropertiesWidget`
- `ReplaceConductorDialog`
- `TextEditor`

(`TerminalEditor` was already correct — that was the one fixed in the first commit.) All four are pre-existing, `git blame` dates them to 2024-03-28, so they were outside the 364-message scope, but since the PR is still open it seemed better to close the whole class of mistake here than to leave four instances behind.

And to confirm your other point: the `Nº` occurrences elsewhere (`Nº contatos`, `Nº terminais`, `Nº Pág:`) are the correct Portuguese ordinal indicator for "número" and were deliberately left alone.

The same commit also fixes punctuation in the two long `SelectAutonumW` help texts, also pre-existing: a stray space in `"N ° página"` / `"n ° da página"`, and two unbalanced quotation marks (`O tipo Página"` and `“Número da página`).

Re-ran the same structural checks on the new commit:

- 2850 messages before and after, none added, removed or reordered; 0 unfinished
- `<source>` and `<comment>` byte-for-byte identical across all 2850
- exactly 6 messages have a changed `<translation>` (7 diff lines — the two conductor-help fixes are in the same message)
- 0 messages left where the source is `°` and the translation uses `º`
- `.qm` regenerated with `lrelease`: `2850 finished, 0 unfinished`

Also verified visually in a local build with `--lang-dir` pointed at this branch: the rotation spin box suffixes in Settings → Texts now render `0°` instead of `0º`.

Issue #782, opened 2026-08-24
Non-finite (NaN/Inf) coordinates are never validated on load, and propagate into saved files


ispyisail:
## Summary

Non-finite (`NaN`/`Inf`) coordinate values are never validated when loading a project. They're accepted silently, written straight back out on save, and in at least one case make two consecutive resaves of the *same* file disagree with each other.

Related: #781 — the same missing validation, but landing on an item whose insertion triggers a collision test, hangs QET completely rather than just corrupting a value. That's the severe case; this issue is the general one.

## Reproduction — passthrough

```bash
python3 -c "
text = open('examples/grafcet.qet', encoding='utf-8').read()
old = '<dynamic_elmt_text rotation=\"0\" y=\"5\" x=\"-30\"'
new = '<dynamic_elmt_text rotation=\"0\" y=\"nan\" x=\"-30\"'
open('grafcet-nan-text.qet', 'w', encoding='utf-8').write(text.replace(old, new, 1))
"
qelectrotech --resave grafcet-nan-text.qet out.qet
grep -o 'y="nan"' out.qet
# -> y="nan"   (present in the output; never rejected or clamped)
```

## Reproduction — non-idempotent resave

Found via a mutation sweep (different seed file, `-inf` on a different element type) rather than the exact repro above:

> `dynamic_texts value differs` between two consecutive resaves of the same once-mutated file — a value that makes two resaves of an *unmodified* file disagree is a distinct, sharper problem than passthrough alone. Root cause: `NaN != NaN` under IEEE 754, so any code path comparing "did this field change" against a NaN-holding value concludes "yes" every time it's checked.

I haven't yet reduced that specific case to a copy-pasteable repro the way the passthrough case above is reduced — flagging that this exists rather than asserting a guaranteed recipe.

## Root cause / suggested direction

Same as #781: reject or clamp non-finite values for coordinate attributes during XML parsing, before they reach anything that stores or compares them. One fix closes both issues.

## What I haven't verified

Whether every non-finite coordinate eventually reaches a hang-capable code path given enough scene complexity, or whether some are genuinely inert — only a handful of cases observed, not enough to characterize where the line is.

Found via a mutation-based sweep of the example corpus, not a user report.

ispyisail:
Fixed in 68e507a5 — non-finite element and terminal coordinates are now rejected on load. The commit names this issue but did not auto-close it.

PR #792, opened 2026-08-25
Fix #781, #782: reject non-finite element/terminal coordinates on load

### LINKED ISSUE #781: A NaN coordinate on any element hangs QET forever while loading the project
## Summary

A `NaN` coordinate on an element hangs QET forever — CPU pegged at 100%, no error, no timeout — while **loading** the project. This isn't limited to export: `--resave`, every `--export-*` mode, and (per the call stack below) plain interactive open all go through the exact same code path.

## Reproduction

Any small project with elements *and conductors* works. Using the shipped `examples/grafcet.qet`:

```bash
python3 -c "
text = open('examples/grafcet.qet', encoding='utf-8').read()
old = '<element freezeLabel=\"false\" z=\"10\" y=\"60\" x=\"220\" prefix=\"\"'
new = '<element freezeLabel=\"false\" z=\"10\" y=\"60\" x=\"nan\" prefix=\"\"'
assert text.count(old) == 1
open('grafcet-nan.qet', 'w', encoding='utf-8').write(text.replace(old, new, 1))
"
qelectrotech --resave grafcet-nan.qet out.qet
```

Result: hangs indefinitely. `user` CPU time tracks wall time 1:1 (confirmed to 20s+ before being killed) — this is a busy-spin, not a blocked wait. Reproduced 3/3 independent runs.

The resulting file is well-formed XML (`x="nan"` is a syntactically valid, if semantically bogus, attribute value) — this is not a parser-rejects-malformed-input case, it's silently accepted and only breaks downstream.

**A project with elements but no conductors does not reproduce this** (tried `ShellyParts.qet` first — no hang) — see root cause below for why.

## Root cause

Stack trace, captured via `gdb -batch`, `SIGINT` sent mid-spin on the standalone repro:

```
QPainterPath::QPainterPath(QPainterPath const&)                      [Qt5Gui]
QPathClipper::QPathClipper(QPainterPath const&, QPainterPath const&) [Qt5Gui]
QPainterPath::intersects(QPainterPath const&) const                  [Qt5Gui]
QGraphicsItem::collidesWithPath(...)                                 [Qt5Widgets]
QGraphicsItem::collidesWithItem(...)                                 [Qt5Widgets]
Conductor::calculateTextItemPosition()
Conductor::itemChange(QGraphicsItem::GraphicsItemChange, QVariant const&)
QGraphicsScene::addItem(QGraphicsItem*)                              [Qt5Widgets]
Diagram::addItem(QGraphicsItem*)
Diagram::fromXml(QDomElement&, QPointF, bool, DiagramContent*)
QETProject::readDiagramsXml(QDomDocument&)
QETProject::openFile(QFile*)
CLIExport::run(QStringList const&)
main()
```

This is why a conductor-free file doesn't trigger it: `Conductor::itemChange()` runs a collision test (`calculateTextItemPosition()`) as a side effect of the conductor being inserted into the scene during load. If a nearby element carries a NaN coordinate, the `QPainterPath` built from it contains NaN-valued points; feeding that into Qt's `QPathClipper` (a sweep-line polygon-clipping algorithm) spins forever instead of terminating — NaN breaks the ordering comparisons the algorithm's termination depends on.

**Practical impact:** any project file with a NaN coordinate — from a prior bug, external tooling, or a manual edit — becomes permanently unopenable in both the GUI and every CLI mode, with no error message distinguishing "hung" from "just slow."

## Suggested direction

Reject or clamp non-finite (`NaN`/`±Inf`) values for coordinate attributes during XML parsing (`Element::fromXml`, `Terminal::fromXml`, and equivalents), before any `QPainterPath`/geometry gets built from them. This is the same gap behind #782 (non-finite coordinates passing through unvalidated) — one fix closes both.

## What I haven't verified

- Which specific neighboring item the NaN-bearing element collides against during insertion — not isolated to a single pair.
- Whether `Inf`/`-Inf` reaches this same code path — the two `Inf` cases I found (see the linked issue) happened to land on elements that didn't trigger this collision test, which may be about *where* they landed rather than evidence `Inf` is safe.
- Not independently reproduced via the interactive GUI — inferred from the identical call chain through `QETProject::openFile()`, which both paths share.

Found via a mutation-based sweep of the example corpus (`simulator/` in the harness this account also maintains), not a user report.

ispyisail:
## Problem

`QString::toDouble()` reports a **successful** conversion for `"nan"`,
`"inf"`, and `"-inf"` — confirmed directly:

```cpp
bool ok;
QString("nan").toDouble(&ok);  // ok == true, value is NaN
```

`Element::valideXml()` and `Terminal::valideXml()` already check the
conversion's `ok` flag as a validity gate before an element/terminal is
allowed to load, but since `ok` is `true` for non-finite values, that
check never catches them.

**#781** — a NaN-positioned element reaching the scene can hang
`QGraphicsScene::addItem()` forever. An existing conductor's
`itemChange()` runs a collision test
(`calculateTextItemPosition()` → `QGraphicsItem::collidesWithPath()`)
whose underlying `QPathClipper` spins without terminating when fed a
NaN-valued `QPainterPath` — NaN breaks the ordering comparisons the
clipping algorithm's termination depends on. Every CLI verb and
interactive open goes through the same `QETProject::openFile()` →
`Diagram::fromXml()` path, so any project containing a NaN coordinate —
from a prior bug, external tooling, or manual editing — becomes
completely unopenable, with no error message.

**#782** — short of triggering a hang, a non-finite coordinate that
doesn't happen to hit a collision test simply passes straight through
to the resaved file, with nothing to stop it.

## Fix

`Element::valideXml()` and `Terminal::valideXml()` now also check
`qIsFinite()` on the parsed `x`/`y`, rejecting the whole item the same
way a missing attribute already does.

`DynamicElementTextItem::fromXml()` has no such reject-the-item gate
(`void` return, no caller check), so its `x`/`y` are clamped to `0`
instead — the same fallback the attribute lookup already uses when
`x`/`y` is missing entirely.

## Testing

Verified against both issues' exact repro steps:

- **#781**: `Habitat-Unifilaire.qet` with `x="nan"` on one element —
  hung (killed by a 25s timeout) on an unfixed build; resaves cleanly
  (exit 0) with this fix.
- **#782**: `grafcet.qet` with `y="nan"` on a `dynamic_elmt_text` — the
  value passed straight through to the resaved file on an unfixed
  build; clamped to `0` with this fix. The specific field is now
  stable (`y="0"` on two consecutive resaves) where it read `"nan"`
  both times before.

  (`grafcet.qet` has an unrelated, already-known, not-yet-merged fix
  for element/terminal-order non-determinism, so a whole-file diff
  across two resaves still differs for reasons unconnected to this
  change — checked the specific once-NaN field in isolation instead of
  the whole file.)

- Also checked `-inf` on `affuteuse_250h.qet`: same rejection, same
  result.

Full database/document consistency sweep of the unmutated example
corpus (23 projects, load/export/resave cross-checks), 0 regressions.

Closes #781, #782.


ispyisail:

ispyisail:

The premise is correct and easy to overlook: `QString::toDouble()` treats `"nan"`, `"inf"` and `"-inf"` as *successful* conversions, so the existing `ok` checks do not catch them. A corrupted or hand-edited file can therefore hand a non-finite coordinate straight to the scene, where it propagates through bounding rects and item positions.

The two treatments are appropriately different. `Element::valideXml()` is a gate that can reject the whole element, so rejecting is right there. `DynamicElementTextItem::fromXml()` has no such gate — there is nothing to refuse at — so falling back to 0 matches exactly what already happens when the attribute is missing entirely, which is the least surprising behaviour available.

**Verified:** Qt 6.10.2 / KF 6.10.0, build clean (0 errors), `ctest` 4/4, merges clean with `master`. `--info` output is unchanged across all 24 shipped example projects, as expected — none of them contain non-finite coordinates, so the new guards are inert on well-formed files.

**Not verified:** loading a deliberately corrupted project containing `nan`/`inf` coordinates, to watch the guards fire.

PR #799, opened 2026-08-30
Fix issue #798: element editor crash on scroll-wheel zoom

ispyisail:
## Problem

Issue #798: in the element editor, holding the mouse scroll wheel to zoom in/out closes the application completely (reported on Windows, 0.200.1-r9435).

## Cause

`ElementView::wheelEvent()` calls `zoomIn()` / `zoomOut()` on every wheel notch, and those call `scale()` with no bound on the resulting view transform. `zoomIn()` multiplies the scale by 4/3 each notch; after enough notches the transform scale (`m11`) overflows the `double` range and the `QTransform` becomes non-invertible. `mapToScene()` then returns NaN/inf geometry, which is handed to the raster paint engine on the next `drawBackground()` pass and aborts the editor. `zoomOut()` is symmetric — `adjustSceneRect()` grows the scene rect without limit until it overflows.

The attached diagnostic report is consistent with this: an `EventLoopWatchdog` stall and a paint/glyph error immediately before the log ends, with no C++ exception backtrace.

`DiagramView::zoom()` (the main diagram editor) had the same unbounded `scale()` and is fixed here too.

## Fix

- `ElementView`: new `scaleClamped()` helper that applies the scale factor only while the resulting `m11` stays within `[m_min_zoom, m_max_zoom]` (0.1 .. 200); all four zoom entry points (`zoomIn`, `zoomOut`, `zoomInSlowly`, `zoomOutSlowly`) go through it.
- `DiagramView::zoom()`: same clamp applied inline before the existing scale logic (range 0.01 .. 200).

Zoom behaviour inside the allowed range is unchanged; `zoomReset()` / `zoomFit()` are untouched. Small and self-contained (4 files).

### DIFF
--- a/ChangeLog_full.md
+++ b/ChangeLog_full.md
@@ -20,6 +20,7 @@ All notable changes to QElectroTech are documented here.
 
 ### 🐛 Bug Fixes
 
+- Fix #798: clamp element-editor and diagram-view zoom to prevent view-transform overflow crash on scroll-wheel zoom ([3ca5d4a](../../commit/3ca5d4ab2))
 - Fix(windows-msi): inject rev into MSI Version Build field ([e19f523](../../commit/e19f5232277efb37435cb65a83563d73333d62ec))
 - Fix #391: use wide-char path for pugixml on Windows to handle Unicode paths ([31edf30](../../commit/31edf30c619213368e9b592b51be6ca8190db831))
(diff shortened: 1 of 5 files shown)


scorpio810:
Nice fix, merged. One thing I want to double-check: `m_min_zoom` is 0.1 in `ElementView` but 0.01 in `DiagramView::zoom()` - was the lower bound difference intentional (schematics can be much larger than a single element, so more zoom-out headroom makes sense), or just not harmonized between the two? Want to make sure before someone reports inconsistent zoom-out behavior between the two editors.

PR #805, opened 2026-09-02
Select newly added diagram in project tree

Kellermorph:
Summary: When adding a new page to a project, the diagram view correctly switches to the new page, but the project tree (ElementsPanel) kept the previous page selected. This happened because GenericPanel::diagramAdded() rebuilt the tree but never selected the newly added diagram item — the diagram parameter was marked as Q_UNUSED.

The fix iterates backwards through the project's child items after the tree rebuild and selects the last QET::Diagram type item, which is the newly added diagram.

### DIFF
--- a/sources/genericpanel.cpp
+++ b/sources/genericpanel.cpp
@@ -730,11 +730,24 @@ void GenericPanel::projectInformationsChanged(QETProject *project) {
 /**
 	@brief GenericPanel::diagramAdded
 	@param project
-	@param diagram (unused)
+	@param diagram the newly added diagram to select in the tree
 */
 void GenericPanel::diagramAdded(QETProject *project, Diagram *diagram) {
-	Q_UNUSED(diagram)
 	addProject(project, nullptr, GenericPanel::AddChildDiagrams);
+	if (diagram) {
+		QTreeWidgetItem *projectItem = itemForProject(project);
+		if (projectItem) {
+			for (int i = projectItem->childCount() - 1; i >= 0; --i) {
+				QTreeWidgetItem *child = projectItem->child(i);
+				if (child && child->type() == QET::Diagram) {
+					clearSelection();
+					setCurrentItem(child);
+					child->setSelected(true);
+					break;
+				}
+			}
+		}
+	}
 	emit(panelContentChanged());
 }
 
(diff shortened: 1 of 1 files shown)


scorpio810:
Thanks for the fix! I tried to reproduce this on my Debian build (Qt6, current master) and the newly added diagram is already selected in the project tree as expected — no fix needed on my side. Could you tell me exactly how you're adding the page (toolbar button, context menu, keyboard shortcut, or something else), and which commit of master you branched from? Want to make sure we're testing the same code path before deciding whether this is still needed.

scorpio810:
QElectroTech V 0.200.1-dev
Compilation : GCC 16.2.0
Built with Qt 6.10.2 - x86_64 - Date : Sep 2 2026 : 04:18:35  

Build with KF6

scorpio810:
Confirmed — reproduces with Ctrl+T. Whatever action I tried first (toolbar button) must already call setCurrentItem() explicitly somewhere else, which masked the bug. Ctrl+T goes straight through GenericPanel::diagramAdded() without that extra call, so your fix is needed for that path. Thanks for tracking it down — looks good to merge.

scorpio810:
Looks good to merge on my end — confirmed the bug on the Ctrl+T path. One thing worth double-checking before we merge: the fix selects the last QET::Diagram item after the tree rebuild, which assumes the new diagram always ends up last. Does that hold for every path (duplicating a folio, pasting/importing multiple folios at once, undo/redo of an add)? If there's any case where the new diagram doesn't land at the end, selecting by the diagram parameter itself (currently Q_UNUSED) rather than by position would be more robust. Let me know what you've tested — happy to merge once we're confident on that.

Kellermorph:
Oh, thanks for pointing that out. I hadn't tested it that way. The change has been incorporated into the fix. Thank you very much, Laurent.

plc-user:
@Kellermorph:
Eine Anmerkung: Nutzt Du bitte für jeden Commit mehr als ein Wort für die Beschreibung? 
Es muss kein Roman sein, aber ein paar Wörter mehr sollten es schon sein.
Sonst haben wir in der Commit-Auflistung viele Male "fix", die sich auf den ersten Blick nicht zuordnen lassen...
Danke!

For our international users:

Please note: Could you please use more than one word for the description in each commit?
It doesn’t have to be a whole novel, but it should be a few more words.
Otherwise, we’ll end up with lots of “fix” entries in the commit list that can’t be assigned to specific changes at first sight...
Thanks!

Issue #408, opened 2025-10-31
Support for arcs on main Tools toolbar


rmelotte:
A feature I'm currently missing in QElectroTech is the ability to add arcs from the main Tools toolbar, similarly to the already existing arc tool in the element editor.

It is not uncommon to add arcs for wires on electrical diagram, like it can be seen on the following sample diagram (I borrowed the picture from the following link, just to have an example: https://electrical-engineering-portal.com/download-center/books-and-guides/electrical-engineering/electrical-wiring-home)
:

<img width="1456" height="751" alt="Image" src="https://github.com/user-attachments/assets/a563b235-a8b3-400b-8bd2-7d16c8863188" />

Using arcs makes it easier to distinguish the wires from the rest of the diagram.

A current workaround is to use a polyline for that, but the result is less visually pleasing.

Has this feature ever been considered?
Does anyone have any idea about the complexity of adding such a feature?

ispyisail:
Hi @rmelotte, good request — this has come up before and is worth doing properly.

**Current state:** QET's element editor already has a full arc implementation (`PartArc` with bounding-rect + start/span angle, interactive handles, XML round-trip). The main diagram's shape item (`QetShapeItem`) supports Line, Rectangle, Ellipse, and Polygon — but not Arc. So the infrastructure exists, it just hasn't been bridged across.

**Three ways to approach it:**

**Option A — Extend `QetShapeItem` with an Arc type**
Add `Arc = 16` to the `ShapeType` enum, add `startAngle`/`spanAngle` properties, and mirror what the element editor already does. The toolbar gets a new "Add arc" action alongside the existing four. Complexity: moderate — ~300–400 lines across `qetshapeitem`, the diagram editor, and the properties panel.

**Option B — Fixed-angle shortcut first**
A simpler first step: the draw interaction produces a half-circle (180° span) that the user can resize. Angles would be editable after placement in the properties panel. Lower interaction complexity, full flexibility later.

**Option C — Three-point arc**
User clicks start point, end point, then a point on the curve. Most natural for the "jump over a crossing wire" use case, but requires fitting a circle through three points — more maths, more novel interaction.

@scorpio810 — what's your preference on direction? Option A follows the element-editor precedent exactly and gives users the full control. Option B gets something usable faster. Option C is the most ergonomic for the wire-crossing use case but the most work. Happy to take this on if there's a direction you'd like to go.

scorpio810:
[...]

PR #809, opened 2026-09-05
Add editable resize/rotate/skew handles for shapes and images

arummler:
Shapes and images can now be resized, rotated, and skewed directly on the canvas, not just moved. Both share one small transform struct (rotation, then skew, then scale, anchored on a movable pivot) and one handle widget, so a corner drag, an edge skew, or grabbing the rotate handle behaves the same way and runs through the same matrix math everywhere, instead of every item type reinventing its own.

Shapes also gained a proper pen tool (bezier paths, corner/smooth/ symmetric nodes), arc support, and mirroring. The former addresses issue #408. Images gained non-destructive cropping and colour-keyed transparency, both remember their own settings, so reopening the dialog picks up where you left off instead of starting over.

Properties dialogs for both were extended to match (position, size, angle, skew), with undo/redo wired through for every handle drag. This goes further than PR #749.

Old XML files can read easily as the transformation is only added if needed and the old syntax is still used and understood if it is not needed.

Attached files show one using the capabilities before this MR (Qt5 based) and one opened with the changes, modified using the new capabilities and saved. I had to rename the ending for github.
[test-graphs-v100.qet.xml](https://github.com/user-attachments/files/31865362/test-graphs-v100.qet.xml)
[test-graphs-v210-new-edit-functions-image-triggered.qet.xml](https://github.com/user-attachments/files/31865364/test-graphs-v210-new-edit-functions-image-triggered.qet.xml)

One icon added from the current breeze svg collection. Could be consolidated in #690  

Disclaimer: mostly vib coded with Claude but iterative, using manual tests and some hand coding.


scorpio810:
Thanks for this, tested it locally and the resize/rotate/skew handles feel solid — nice unification of the transform math across shapes and images.

One thing worth fixing before merge: in ImageTransparentColorDialog, the tolerance slider is wired to QSlider::valueChanged rather than sliderReleased, and each tick runs applyColorKey() against the full-resolution m_sourceImage. On a larger image (a scanned schematic background, several Mpx) dragging the slider is going to lag noticeably, since it's a full O(width×height×numColors) pixel loop on every intermediate value, not just on release. Could you either switch to sliderReleased, or build the live preview off a downsampled copy of the source image and only run the full-resolution pass in resultPixmap() (which already does the right thing, computing full-res once on accept)?

Everything else I looked at (undo/redo wiring, XML backward-compat via the two-tier fallback, PromoteShapeCommand's DOM cloning, rebuildHandles() vs repositionHandles() split) looks careful and well-reasoned — appreciated given the size of the change.

Given this is a single squashed commit touching almost every graphics item (+6760/-409 across 29 files), would you be open to splitting it into a few logical PRs (transform infra / pen tool / image crop+transparency)? It'd make future review and bisecting a lot easier, for this and any follow-up fixes.

scorpio810:
Thanks for this, tested it locally and the resize/rotate/skew handles feel solid — nice unification of the transform math across shapes and images.

One thing worth fixing after merge: in ImageTransparentColorDialog, the tolerance slider is wired to QSlider::valueChanged rather than sliderReleased, and each tick runs applyColorKey() against the full-resolution m_sourceImage. On a larger image (a scanned schematic background, several Mpx) dragging the slider is going to lag noticeably, since it's a full O(width×height×numColors) pixel loop on every intermediate value, not just on release. Could you either switch to sliderReleased, or build the live preview off a downsampled copy of the source image and only run the full-resolution pass in resultPixmap() (which already does the right thing, computing full-res once on accept)?

Everything else I looked at (undo/redo wiring, XML backward-compat via the two-tier fallback, PromoteShapeCommand's DOM cloning, rebuildHandles() vs repositionHandles() split) looks careful and well-reasoned — appreciated given the size of the change.

PR #810, opened 2026-09-05
Image speed improvement and additional transparency feature

arummler:
As requested in the previous review the preview picture is now downscaled which improves the speed. I also decided to add a sensitivity channel to each colour.


ispyisail:
Reviewed this diff (diagramimageitem.cpp/.h, imagetransparentcolordialog.cpp/.h). This is a clean refactor of the transparent-color feature from a single shared tolerance to a per-colour tolerance (`PickedColor` struct).

- Legacy XML load/save fallback for old single-tolerance files is handled correctly.
- Index-based add/remove/update logic stays in sync via full `rebuildSwatches()` calls.
- The downsampled-preview vs full-res-result split looks intentional and is documented.

No runtime-correctness issues found (no inverted conditions, missing null checks, dropped guards, swallowed errors, or copy-paste variable mixups). No dead code or duplicated logic introduced.

PR #811, opened 2026-09-06
Enable PDF import in Flatpak by adding QtWebEngine BaseApp

Kellermorph:
The PDF import feature ("Ajouter un PDF") was not available in the
Flatpak build because Qt6Pdf was not found by CMake.

The KDE Platform 6.11 runtime does not include Qt6Pdf — it lives in
the qtwebengine source tree. By adding `io.qt.qtwebengine.BaseApp` as
the base app (which is the standard approach for KDE Flatpak apps
needing QtWebEngine/QtPdf), the Qt6Pdf libraries and CMake files are
available at /app/lib, and `find_package(Qt6 COMPONENTS Pdf)` succeeds.

Changes:
- Add "base": "io.qt.qtwebengine.BaseApp" and "base-version": "6.11"
- Add cleanup-commands for BaseApp (required by BaseApp convention)

### DIFF
--- a/build-aux/flatpak/org.qelectrotech.QElectroTech.json
+++ b/build-aux/flatpak/org.qelectrotech.QElectroTech.json
@@ -3,6 +3,8 @@
   "runtime": "org.kde.Platform",
   "runtime-version": "6.11",
   "sdk": "org.kde.Sdk",
+  "base": "io.qt.qtwebengine.BaseApp",
+  "base-version": "6.11",
   "command": "qelectrotech",
   "rename-desktop-file": "org.qelectrotech.qelectrotech.desktop",
   "rename-appdata-file": "qelectrotech.appdata.xml",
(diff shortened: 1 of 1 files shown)


scorpio810:
Thanks you Thomas, I just compiled and pushed new Flatpak to QET 's remote Fllatpak repo.
Please try it.

Kellermorph:
Hi, thanks for rebuilding! The PDF button is now showing up and working.

Just to explain what happened on my end: I had two installations of the Flatpak on my system — an older user-level install from the qelectrotech-origin remote (which still had the old qt6-pdf module approach without the BaseApp) and a newer system-level install from qet-devel (which had the updated manifest with io.qt.qtwebengine.BaseApp). Flatpak always prioritizes user-level installs over system-level ones, so it kept launching the old version even after the update.

After removing the old user install with flatpak uninstall --user org.qelectrotech.QElectroTech, the correct version from qet-devel takes over and the PDF import works as expected.

scorpio810:
Hi thanks for your feedback.

PR #820, opened 2026-09-08
Reject non-finite coordinates while loading XML

### LINKED ISSUE #781: A NaN coordinate on any element hangs QET forever while loading the project

### LINKED ISSUE #782: Non-finite (NaN/Inf) coordinates are never validated on load, and propagate into saved files

enesgursoy6110:
Non-finite XML coordinates currently pass through Qt's numeric conversion. An element with an `x="nan"` value can then reach `QPainterPath` collision handling while conductors are loaded and keep the application busy indefinitely.

This change adds a shared finite-number reader for graphical XML attributes. Invalid or non-finite values fall back to the same defaults used for missing attributes and produce a diagnostic containing the tag, attribute, and rejected value. The reader is applied to element positions, text items, groups, shapes, images, table positions, pen widths, and conductor positions. Required real-valued definition attributes now reject non-finite values, and invalid conductor segment lengths are ignored so the existing automatic routing fallback can run.

Validation:

- Built the updated `master` checkout in debug mode with Qt 5.15.19 and MinGW.
- Resaved the shipped `grafcet.qet` example after mutating element and dynamic-text coordinates to `nan`, `inf`, and `-inf`.
- All six malformed fixtures completed within the timeout and retained their element and conductor counts.
- Covered defaults, finite values, text rotation/width, polygon coordinates, and transforms with local Qt tests.

Fixes #781.
Fixes #782.

### DIFF
--- a/sources/qet.cpp
+++ b/sources/qet.cpp
@@ -239,7 +239,7 @@ bool QET::attributeIsAReal(
 	// verifie la validite de l'attribut
 	bool ok;
 	qreal tmp = e.attribute(nom_attribut).toDouble(&ok);
-	if (!ok) return(false);
+	if (!ok || !qIsFinite(tmp)) return(false);
 	if (reel != nullptr) *reel = tmp;
 	return(true);
 }
--- a/sources/qetgraphicsitem/ViewItem/qetgraphicstableitem.cpp
+++ b/sources/qetgraphicsitem/ViewItem/qetgraphicstableitem.cpp
@@ -626,12 +626,8 @@ void QetGraphicsTableItem::fromXml(const QDomElement &dom_element)
 	}
 
 	this->setPos(
-				dom_element.attribute(
-					"x",
-					QString::number(10)).toDouble(),
-				dom_element.attribute(
-					"y",
-					QString::number(10)).toDouble());
+				QETXML::finiteAttribute(dom_element, "x", 10),
+				QETXML::finiteAttribute(dom_element, "y", 10));
 	//Size is not set now because will change during the whole process of opening a project from the xml
 	m_pending_size = QSize(
 				dom_element.attribute(
--- a/sources/qetgraphicsitem/conductor.cpp
+++ b/sources/qetgraphicsitem/conductor.cpp
@@ -16,6 +16,7 @@
 	along with QElectroTech.  If not, see <http://www.gnu.org/licenses/>.
 */
 #include "../qetgraphicsitem/conductor.h"
+#include "../qetxml.h"
 #include "../qetproject.h"
 #include "../QPropertyUndoCommand/qpropertyundocommand.h"
 #include "../autoNum/numerotationcontextcommands.h"
@@ -1022,8 +1023,8 @@ bool Conductor::fromXml(QDomElement &dom_element)
 		m_uuid = QUuid::createUuid();
 	}
 
-	setPos(dom_element.attribute("x", nullptr).toDouble(),
-		   dom_element.attribute("y", nullptr).toDouble());
+	setPos(QETXML::finiteAttribute(dom_element, "x"),
+		   QETXML::finiteAttribute(dom_element, "y"));
 
 	bool retval = pathFromXml(dom_element);
 
@@ -1152,7 +1153,7 @@ bool Conductor::pathFromXml(const QDomElement &e) {
 		// cette longueur doit etre un reel
 		bool ok;
 		qreal segment_length = current_segment.attribute("length").toDouble(&ok);
-		if (!ok) continue;
+		if (!ok || !qIsFinite(segment_length)) continue;
 
 		if (current_segment.attribute("orientation") == "horizontal") {
 			segments_x << segment_length;
--- a/sources/qetgraphicsitem/conductortextitem.cpp
+++ b/sources/qetgraphicsitem/conductortextitem.cpp
@@ -16,6 +16,7 @@
 	along with QElectroTech.  If not, see <http://www.gnu.org/licenses/>.
 */
 #include "conductortextitem.h"
+#include "../qetxml.h"
 
 #include "../diagram.h"
 #include "../diagramcommands.h"
@@ -70,12 +71,12 @@ Conductor *ConductorTextItem::parentConductor() const
 */
 void ConductorTextItem::fromXml(const QDomElement &e) {
 	if (e.hasAttribute("userx")) {
-		setPos(e.attribute("userx").toDouble(),
-			   e.attribute("usery").toDouble());
+		setPos(QETXML::finiteAttribute(e, "userx"),
+			   QETXML::finiteAttribute(e, "usery"));
 		moved_by_user_ = true;
 	}
 	if (e.hasAttribute("rotation")) {
-		setRotation(e.attribute("rotation").toDouble());
+		setRotation(QETXML::finiteAttribute(e, "rotation"));
 		rotate_by_user_ = true;
 	}
 }
--- a/sources/qetgraphicsitem/diagramimageitem.cpp
+++ b/sources/qetgraphicsitem/diagramimageitem.cpp
@@ -16,6 +16,7 @@
 	along with QElectroTech.  If not, see <http://www.gnu.org/licenses/>.
 */
 #include "diagramimageitem.h"
+#include "../qetxml.h"
 
 #include "../PropertiesEditor/propertieseditordialog.h"
 #include "../QPropertyUndoCommand/qpropertyundocommand.h"
@@ -1409,8 +1410,8 @@ bool DiagramImageItem::fromXml(const QDomElement &e)
 	// <transform> element (independent scaleX/scaleY, and/or a custom
 	// pivot), the same two-tier fallback QetShapeItem's own fromXml()
 	// already uses for its identical <transform> element.
-	m_transform.rotation = e.attribute("rotation").toDouble();
-	m_transform.scaleX = e.attribute("size").toDouble();
+	m_transform.rotation = QETXML::finiteAttribute(e, "rotation");
+	m_transform.scaleX = QETXML::finiteAttribute(e, "size", 1);
 	m_transform.scaleY = m_transform.scaleX;
 	m_transform.pivot = boundingRect().center();
 	m_pivotIsCustom = false;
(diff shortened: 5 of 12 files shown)


scorpio810:
Thanks for the thorough writeup and testing on this one.

Quick heads-up before this goes further: this ground has actually been covered twice already. ispyisail opened #682 back on Aug 7 — the earliest one — which traces the NaN hang to a shared helper (QET::attributeIsAReal(), used at ~26 call sites) rather than patching valideXml() directly, and also fixes a separate but related crash: illegal XML control bytes segfaulting inside libQt5Xml itself before QET code even runs (something this PR and #792 don't cover). Then #792 (Aug 25) independently re-fixed the same NaN issue with a narrower patch.

So we've now got three open PRs converging on the same root problem. I'd rather consolidate around #682 as the base — it has the broader scope and the more thorough gdb-verified root-cause analysis — than merge overlapping fixes. There's also an unresolved thread on #682 about whether to prioritize this at all given the Qt6 migration, worth reading before continuing here.

I'll hold off reviewing this one until that's sorted out — appreciate the effort though, and the non-finite-value handling for the broader attribute set (text/groups/shapes/etc.) you added here may still be useful once we land on one approach.

ispyisail:
Reviewed this by building it and reproducing the bug rather than reading the diff alone. **It works, and the claims in the description hold up.**

## Verified

Qt 5.15.18, Ubuntu 26.04, gcc 15.2.0. Same binary, same file, only this PR differing:

| | master | with #820 |
|---|---|---|
| `--resave` of `grafcet.qet` with one element `x="nan"` | killed at 25 s, **25 s of user CPU** — a busy spin, as #781 describes | **0.309 s**, exit 0 |
| Diagnostic | none | `Invalid numeric XML attribute "element" "x" "nan" - using 0` |
| Data | — | 56 elements / 35 conductors, identical to the clean original |

The 22 loadable projects in `examples/` also still load and export with no crash or hang.

## What I think is well judged

Hardening `QET::attributeIsAReal` is the part I'd highlight. It covers every *required* real-valued definition attribute in one place — including **terminal x/y**, which matters most here, because terminals are where conductors attach and therefore what feeds the `QPainterPath` collision handling that spins in the first place. `finiteAttribute()` then covers the optional positional attributes. Two mechanisms, each where it belongs, rather than one blanket helper.

The defaults are also chosen individually rather than uniformly zero — `scaleX`/`scaleY` to 1, `text_width` to -1, `z` to the current `zValue()`. That is easy to get wrong and it has not been.

## Three things worth raising, none of them blocking

**1. One behaviour change that is not in the description.** `DiagramImageItem::fromXml()` was `e.attribute("size").toDouble()`, which yields **0** when the attribute is absent, and is now `finiteAttribute(e, "size", 1)`. So a missing `size` changes from scale 0 (invisible) to scale 1. Almost certainly a fix rather than a regression — and I checked all 123 `<image>` elements in `examples/` carry `size`, so the practical impact today is nil — but it is a semantic change riding along unmentioned, and worth a line in the description so a future bisect does not have to rediscover it.

**2. A coverage gap, which I tested rather than assumed.** `factory/elementpicturefactory.cpp` still reads element *drawing* coordinates unguarded — polygon points, text position and rotation, rect `rx`/`ry`, `painter.translate`. I mutated a polygon `x1` to `nan` in a real project and it does **not** hang (0.307 s), so this is a latent robustness gap rather than the reported bug. Reasonable follow-up, not a reason to hold this PR.

**3. A question about the validation, not a challenge.** The branch is based on `a583b3c43`, which includes `6b577ee75` — and since that commit, master does not compile with the default `QT_VERSION_MAJOR=5` (`QLabel::pixmap()` returns a pointer in Qt5; 5 errors in `imagetransparentcolordialog.cpp`, see #824). The description says the updated master checkout was built with Qt 5.15.19 and MinGW, so I could not reconcile the two — I had to apply #824 before this branch would build at all. Did something differ in your setup, or was the build from an earlier checkout? Asking mainly because it bears on how much the six malformed fixtures demonstrate, and because **anyone else verifying this PR will hit the same wall** until #824 lands.

## One small suggestion

`qWarning()` fires per bad attribute, so a badly corrupted file will emit one line per coordinate. Arguably correct for a diagnostic, but a per-load summary might read better if that turns out noisy in practice.

Nice piece of work — the fix is in the right place and it does what it says.

enesgursoy6110:
Thanks for the detailed review and validation. I agree that maintaining a third overlapping fix would make the eventual resolution harder to review and maintain, so I am closing this PR in favor of #682 as the consolidation base.

The wider optional-attribute coverage here, the missing-image-size fallback behavior, and the remaining element drawing-coordinate coverage can be reconsidered as focused follow-ups after the shared root fix is settled. I am deliberately not expanding this overlapping PR further.

PR #821, opened 2026-09-08
Scope shortcut conflicts to overlapping editors

### LINKED ISSUE #757: Shortcuts config page reports 60 of its own 95 default bindings as conflicts

enesgursoy6110:
The shortcut configuration page treats every identical key sequence as a global conflict. QElectroTech registers common defaults such as Save, Undo, and Redo once per editor window, so legitimate bindings in separate editors are reported as conflicts.

This change derives stable scopes from shortcut IDs and compares bindings only where those scopes can overlap. Actions in the same editor still conflict; main-window actions overlap every editor; depth actions overlap the diagram and element editors where they are installed. Translated category labels remain presentation data.

Validation:

- Built the updated `master` checkout in debug mode with Qt 5.15.19 and MinGW.
- Verified that matching Save shortcuts in the diagram, element, and title-block editors are accepted.
- Verified that duplicate shortcuts in one editor, global main-window conflicts, and shared depth-action conflicts remain detected.

Fixes #757.

### DIFF
--- a/sources/ui/configpage/shortcutsconfigpage.cpp
+++ b/sources/ui/configpage/shortcutsconfigpage.cpp
@@ -269,6 +269,25 @@ void ShortcutsConfigPage::applyFilter()
 	m_count_label->setText(tr("%n action(s)", nullptr, visible_actions));
 }
 
+// IDs use stable context prefixes; translated display categories are not scopes.
+// Main-window actions are shared by the editors, and depth actions by the
+// diagram and element editors. Panel-specific shortcuts retain their own scope.
+static bool shortcutScopesOverlap(const QString &first_id, const QString &second_id)
+{
+	const QString first = first_id.section(QLatin1Char('.'), 0, 0);
+	const QString second = second_id.section(QLatin1Char('.'), 0, 0);
+	if (first == second || first == QLatin1String("mainwindow")
+			|| second == QLatin1String("mainwindow"))
+		return true;
+
+	const auto has_depth = [](const QString &scope) {
+		return scope == QLatin1String("diagrameditor")
+				|| scope == QLatin1String("elementeditor");
+	};
+	return (first == QLatin1String("depth") && has_depth(second))
+			|| (second == QLatin1String("depth") && has_depth(first));
+}
+
 /**
 	@brief ShortcutsConfigPage::checkConflicts
 	Highlight every row whose currently-edited sequence is shared, non-empty,
@@ -287,8 +306,12 @@ void ShortcutsConfigPage::checkConflicts()
 	for (int row = 0; row < m_rows.size(); ++row) {
 		Row &current_row = m_rows[row];
 		const QString sequence_text = current_row.edit->keySequence().toString();
-		const QList<int> &conflicting_rows = sequence_to_rows.value(sequence_text);
-		const bool conflicted = !sequence_text.isEmpty() && conflicting_rows.size() > 1;
+		QList<int> conflicting_rows;
+		for (int other : sequence_to_rows.value(sequence_text)) {
+			if (other != row && shortcutScopesOverlap(current_row.id, m_rows.at(other).id))
+				conflicting_rows << other;
+		}
+		const bool conflicted = !conflicting_rows.isEmpty();
 
 		current_row.conflicted = conflicted;
 


scorpio810:
Thanks for looking into this — #757 has been a real annoyance on the shortcuts page.

Just a heads-up: the shortcuts registry and config page this touches (#587) was built by ispyisail as part of a broader keyboard-shortcuts effort (see discussion #574). That work is mostly merged now, but #585 (Tab/Shift+Tab selection cycling, same area) is still open. I'd rather get that settled first before layering more changes onto the shortcut-conflict logic, to avoid stepping on each other. I'll come back to this once #585 is resolved.

ispyisail:
Read this one against the source rather than building it, since it is UI-state logic with no headless path.

**The problem is real and the approach is right.** `elementeditor.save`, `titleblockeditor.save` and `diagrameditor.save_file` all bind `QKeySequence::Save` in three separate top-level windows, so today they are reported as a conflict when nothing is actually ambiguous. Deriving the scope from the ID prefix rather than the translated category is the correct instinct — the categories are presentation data and would have been the wrong key.

I also checked the guard that disappears in the diff: dropping `!sequence_text.isEmpty()` is safe, because the map is only populated for non-empty sequences a few lines above, so `value("")` returns an empty list and `conflicted` stays false. Worth stating explicitly since it looks like a removed check at a glance.

## One gap, and I think it matters

`shortcutScopesOverlap()` knows three scopes. There are **eight** in use:

```
33 diagrameditor   28 elementeditor   19 titleblockeditor   8 panel
 4 depth            3 richtext         2 mainwindow         1 autonum
```

Two of the unhandled ones are **docks inside the diagram editor window**, not separate windows:

- `panel.*` — `ElementsPanelWidget`, `qetdiagrameditor.cpp:182` `addDockWidget(Qt::LeftDockWidgetArea, qdw_pa)`
- `autonum.*` — `AutonumberingDockWidget`, `qetdiagrameditor.cpp:271` `addDockWidget(Qt::RightDockWidgetArea, m_autonumbering_dock)`

And nothing in the tree calls `setShortcutContext`, so every action keeps the default `Qt::WindowShortcut` — meaning a dock's actions share the diagram editor window's shortcut namespace rather than being scoped to the dock widget.

There is a live collision in the shipped defaults today:

| ID | Sequence |
|---|---|
| `panel.delete_diagram` | `Qt::Key_Delete` |
| `diagrameditor.delete_selection` | `Qt::Key_Delete` |

Same window, both `Qt::WindowShortcut`. That is the classic Qt ambiguous-shortcut case, it is reported as a conflict today, and after this change `shortcutScopesOverlap("panel", "diagrameditor")` returns false so it is **silently no longer reported**.

So the PR trades a set of false positives for a smaller set of false negatives. The false positives are worse and more numerous, so this is still a clear improvement — but the fix looks like a two-line addition: treat `panel` and `autonum` as overlapping `diagrameditor`, the same way `depth` already overlaps the two editors it is installed in.

For the rest: `titleblockeditor` is its own top-level window (`titleblock/qettemplateeditor.cpp`) and `richtext` is its own editor (`richtext/richtexteditor.cpp`), so leaving those isolated looks right to me.

## Minor

Inside the `if (conflicted)` branch, the tooltip loop still filters `if (other_row != row)`. `conflicting_rows` now excludes the current row by construction, so that test is redundant — harmless, just dead.

## Suggestion

The scope table is the load-bearing part of this change, and it is currently three string comparisons in a static function that will silently do the wrong thing when a ninth scope appears. It might be worth a comment naming the eight known scopes and which are windows versus docks, so the next person adding a `registerAction` prefix knows there is a table to update.

Good change otherwise — the diagnosis is right and the mechanism is the appropriate one.

ispyisail:
@enesgursoy6110 — an apology is owed here. I merged #759 earlier today, which fixes the same issue (#757) in a worse way, without checking whether a PR for it already existed. Yours had been open since 8 September. That is my mistake, and the conflict this PR now has with `master` is my doing, not yours.

Your approach is the better one, and your comment that *"translated display categories are not scopes"* is a direct and correct criticism of what I merged. #759 keys conflict detection on `category`, which is a `tr()` string — `tr("Panneau des éléments")`, `tr("Général")`. Yours keys on the ID prefix, which is stable and untranslated.

I checked this against the real registry (94 registered actions, plus the four `depth.*` ones registered through `QObject::tr`) rather than by reading, and the difference is concrete.

On the shipped defaults the two behave identically: all 24 shared sequences are legitimate cross-editor duplicates — `diagrameditor.undo` / `titleblockeditor.undo` / `elementeditor.undo` and so on — and neither flags them. So #759 does fix the reported symptom.

They diverge on shortcuts a user assigns in the preferences page:

| scenario | #759 (merged) | #821 |
|---|---|---|
| diagram-editor action given the main-window F1 | **missed** | flagged |
| element-editor action given the main-window F1 | **missed** | flagged |
| diagram action given a `depth.*` shortcut | **missed** | flagged |
| element action given a `depth.*` shortcut | **missed** | flagged |
| two actions in the same editor collide | flagged | flagged |
| diagram vs title-block editor (legitimate) | not flagged | not flagged |

Those four are real conflicts. The main window is live while an editor is open, and the depth actions are installed into both the diagram and element editors — which is exactly what your `shortcutScopesOverlap()` encodes and what keying on the category cannot express.

The reason #759 looked adequate is that the scope prefix currently maps one-to-one onto the translated category for all seven scopes, so same-scope detection happens to come out the same. It only fails where scopes legitimately overlap, which is precisely the case your version handles.

I am reverting #759. Once that lands this should apply cleanly again, and it is the version that should go in. Sorry for the wasted round trip.

ispyisail:
A question about `shortcutScopesOverlap()`, asked because you have clearly thought about the overlap relationships more carefully than I did — I may be missing the reason.

`panel` is treated as its own isolated scope, so a `panel.*` shortcut is never compared against a `diagrameditor.*` one. But the elements panel is a dock inside the diagram editor window:

```cpp
// qetdiagrameditor.cpp:181
qdw_pa->setWidget(pa = new ElementsPanelWidget(qdw_pa));
addDockWidget(Qt::LeftDockWidgetArea, qdw_pa);
```

and `elementspanelwidget.cpp` never calls `setShortcutContext`, so those actions keep the default `Qt::WindowShortcut` — which is scoped to the *window*, and the window is the diagram editor. By that reading `panel` would overlap `diagrameditor` in the same way `depth` does.

What I could not settle is whether those shortcuts are actually live. The panel's actions are parented to the widget but only ever reach a `QAction` through `context_menu->addAction(...)` — I found no `addAction()` on the panel widget itself — and a `QAction` needs to be added to a visible widget for its shortcut to fire. If they are only ever context-menu entries then their key sequences are inert, `panel` is correctly isolated, and there is nothing to change.

Nothing in the shipped defaults depends on the answer: no sequence is currently shared between `panel` and any other scope, so this is latent either way. I did not want to raise it as a defect when it may well be deliberate.

Same question, less strongly, for `richtext` and `autonum`.

PR #823, opened 2026-09-08
Restore Qt5 no-KF qmake builds

enesgursoy6110:
The supported Qt 5 no-KF qmake configuration currently fails for two independent reasons:

- `CONFIG+=no_kf5` defines `BUILD_WITHOUT_KF5`, while the no-KF compatibility code is guarded by `BUILD_WITHOUT_KF`.
- On Qt 5, `QLabel::pixmap()` returns a pointer, but the transparent-color dialog uses the Qt 6 value-returning API.

This change defines both compatibility macros for the no-KF build and uses the dialog's existing `m_displayImage` value for null and dimensions checks. That representation is already the pixmap being displayed and avoids API differences between Qt 5 and Qt 6.

Validation:

- Configured with Qt 5.15.19, MinGW, `CONFIG+=debug CONFIG+=no_kf5 CONFIG-=release`.
- Built the debug application successfully.

### DIFF
--- a/qelectrotech.pro
+++ b/qelectrotech.pro
@@ -256,7 +256,7 @@ QT += xml svg network sql widgets printsupport concurrent gui-private
 # sources/ui/nokde is only added to the include path in that configuration, so
 # a normal KF5 build is unaffected.
 no_kf5 {
-    DEFINES     += BUILD_WITHOUT_KF5
+    DEFINES     += BUILD_WITHOUT_KF BUILD_WITHOUT_KF5
     INCLUDEPATH += sources/ui/nokde
     HEADERS     += $$files(sources/ui/nokde/*.h)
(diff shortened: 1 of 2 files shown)


scorpio810:
Thanks for taking the time to put this together, and for the clean testing writeup.

We've recently completed our migration to CMake/Qt6, and the qmake no_kf5 build path is being phased out as part of that — Qt5 support is being sunset across the project (see the forum announcement on the 0.9/0.100.1 dev builds). Given that, I don't think it makes sense to merge this: qelectrotech.pro's no_kf5{} block has no CMake equivalent, and the pixmap() change only matters for Qt5's pointer-returning API — under Qt6 the original code already works fine as-is, so there's nothing here that needs porting over to the CMake side.

I'll close this for now, but I appreciate the contribution — feel free to open something against CMakeLists.txt if you run into build issues there instead.

PR #824, opened 2026-09-08
Fix Qt5 build: QLabel::pixmap() returns a pointer, not a value

ispyisail:
`ClickableImageLabel::mousePressEvent()` calls `pixmap().isNull()` and `pixmap().width()`. That is the Qt6 signature — in Qt5 `QLabel::pixmap()` returns `const QPixmap *`, so the code does not compile:

```
sources/ui/imagetransparentcolordialog.cpp:68:59: error: request for member 'isNull' in
'((ClickableImageLabel*)this)->ClickableImageLabel::QLabel.QLabel::pixmap()',
which is of pointer type 'const QPixmap*' (maybe you meant to use '->' ?)
```

`CMakeLists.txt` defaults `QT_VERSION_MAJOR` to 5 when it is not specified:

```cmake
if(NOT DEFINED QT_VERSION_MAJOR)
    set(QT_VERSION_MAJOR 5)
```

so **a default configuration of master has not built since 6b577ee75** (5 errors, all in this one function).

## Fix

Read the pixmap once into a local, version-guarded the way the rest of the codebase already handles this split (`qgimanager.cpp`, `exportpropertieswidget.cpp`). That also removes four repeated `pixmap()` calls from the same expression.

## Testing

Qt 5.15.18, Ubuntu 26.04, gcc 15.2.0: master fails with 5 errors before the change and builds clean after it. The built binary starts and reports its version. No behaviour change on Qt6, where `pixmap()` already returned a value.

### DIFF
--- a/sources/ui/imagetransparentcolordialog.cpp
+++ b/sources/ui/imagetransparentcolordialog.cpp
@@ -65,7 +65,14 @@ ClickableImageLabel::ClickableImageLabel(const QImage &sourceImage, QWidget *par
 */
 void ClickableImageLabel::mousePressEvent(QMouseEvent *event)
 {
-	if (event->button() != Qt::LeftButton || pixmap().isNull())
+		// QLabel::pixmap() returns a pointer in Qt5 and a value in Qt6.
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+	const QPixmap label_pixmap = pixmap() ? *pixmap() : QPixmap();
+#else
+	const QPixmap label_pixmap = pixmap();
+#endif
+
+	if (event->button() != Qt::LeftButton || label_pixmap.isNull())
 		return;
 
 	// The label may be larger than its pixmap (layout stretching); the
@@ -73,9 +80,9 @@ void ClickableImageLabel::mousePressEvent(QMouseEvent *event)
 	// so the click has to be re-based against the pixmap's own rect
 	// within the label, not the label's own top-left.
 	const QRect pixmapRect(
-			(width() - pixmap().width()) / 2,
-			(height() - pixmap().height()) / 2,
-			pixmap().width(), pixmap().height());
+			(width() - label_pixmap.width()) / 2,
+			(height() - label_pixmap.height()) / 2,
+			label_pixmap.width(), label_pixmap.height());
 	if (!pixmapRect.contains(event->pos()))
 		return;
 


ispyisail:
Some extra context that turned up right after opening this, in case it helps judge priority. Not a nudge — the PR is minutes old.

**It is actively blocking other work.** I went to verify #661 today by cherry-picking it onto master, and could not produce a binary at all until I applied this fix first. So anyone reproducing, reviewing or bisecting on a default configuration is currently stopped at the compile step, not just inconvenienced.

**Why it went unnoticed for four days.** As far as I can see from the workflows on master, the only build check is Windows, and it is explicitly Qt6-only:

```yaml
# Job 1: Windows build (Qt6/KF6 — sole track since the Qt5 track was removed)
...configured (-DQT_VERSION_MAJOR=6)
```

while `CMakeLists.txt` does this:

```cmake
if(NOT DEFINED QT_VERSION_MAJOR)
    set(QT_VERSION_MAJOR 5)
```

So the **default** build configuration has no CI coverage on any platform, which is why a Qt5-only compile error introduced in 6b577ee75 survived to master and sat there. The error is not subtle once you hit it — 5 hard errors in one function — it simply had nothing building that path.

That is an observation, not a proposal: whether the answer is a Qt5 build check, or flipping the default to 6 now that Qt6 is the tested track, is your call and out of scope for this one-liner. I mention it only because the same class of break can recur silently until the default and the tested configuration agree.

Measured today on master `a583b3c43`, Qt 5.15.18, gcc 15.2.0, Ubuntu 26.04: 5 errors before, clean build after, binary starts and reports its version.

ispyisail:
@scorpio810 @arummler — a question about the default Qt version rather than a proposal, and it may make #824 unnecessary rather than more urgent.

## The mismatch

`CMakeLists.txt` still defaults to Qt5 when nothing is specified:

```cmake
if(NOT DEFINED QT_VERSION_MAJOR)
    set(QT_VERSION_MAJOR 5)
endif()
```

But every build that actually runs is Qt6. The Windows workflow on master says so itself:

> `# Job 1: Windows build (Qt6/KF6 — sole track since the Qt5 track was removed)`

and every job in @arummler's CI overhaul (#722) pins `-DQT_VERSION_MAJOR=6` — Linux, macOS, FreeBSD, MSVC and MinGW alike.

So the configuration that nobody builds is the one a contributor gets by default. That is how the break in #824 reached master and sat for four days: it is a genuine compile error, but only on the path nothing tests.

## Master builds fine on Qt6 today

I checked rather than assumed, on master `a583b3c43` with Qt 6.10.2, gcc 15.2.0, Ubuntu 26.04:

- configure and build clean: **788/788 targets, zero errors**
- the binary links `libQt6Core.so.6` and reports `0.200.1-dev`
- `--export-bom examples/industrial.qet` succeeds and produces output **byte-identical** to the Qt5 build of the same commit

So this is not a porting exercise. Qt6 is already healthy; only the default disagrees with it.

## The question

Which is intended?

1. **Qt6 is the supported default.** Then flipping that one line makes the default configuration the tested one, and #824 becomes much less important — the Qt5 path would no longer be what people get by default. (#824 would still be worth taking for anyone who does build Qt5, but it stops being a trap for newcomers.)
2. **Qt5 is still supported.** Then one Qt5 job somewhere would prevent this recurring, since #722 as it stands would not have caught the #824 break either.

What does not work well is the current middle position: defaulting to 5 while testing only 6.

I have deliberately not sent a patch for this. It is a release-infrastructure decision — distributions still ship Qt5, and packagers may be relying on that default in ways I cannot see from here. It seemed more useful to bring the measurement and let you decide.

Two smaller things noticed while looking, in case they are news:

- **The doxygen docs at download.qelectrotech.org are a version behind** (0.100.0-dev vs 0.200.1-dev). `auto-doxygen.yml` triggers only on tag pushes, so ordinary commits never rebuild it, and its last run (30 April) failed at the **Create Pull Request** step rather than in doxygen itself — which looks like a token problem rather than a documentation problem.
- **Nightly builds stop at 2026-05-08** (0.100.1). Same window as the doxygen failure, so possibly one outage rather than two.

scorpio810:
For Doxygen see https://github.com/qelectrotech/qelectrotech-source-mirror/pull/510#issuecomment-4744073034

>Nightly builds stop at 2026-05-08 (0.100.1). Same window as the doxygen failure, so possibly one outage rather than two.

You ask to https://download.qelectrotech.org/qet/builds/nightly/ 
Yes we moved to Windows CI/CD see on Download page. I 'm stopped to cross-compiling my  own MXE with old Digikam C.Gilles script on my Debian to produce Windows builds. see https://invent.kde.org/graphics/digikam/-/commit/68a36dbe875fca1d3296a721443a14c0fa56a948

1. It  was long to compile all toolchain framework environment ~ 1hours on my Threadripper 24C/48 Threads with make -j48, before with Ryzen 1700X it"s take more 5 hours by arch X86_64 et 32 bits.
2. Two: you need to recompile again when Qt fournish new Framawork version.

ispyisail:
Thank you — that explains it, and **I got the cause wrong above.** Correcting it plainly since I stated it with more confidence than I had earned.

I wrote that the last doxygen run "failed at the **Create Pull Request** step — which looks like a token problem rather than a documentation problem." The step was right; the reason was not. That step pushes LFS objects, so it was almost certainly failing because the LFS bandwidth was already exhausted, exactly as you describe. I inferred a cause from a step name instead of checking, and it pointed at the wrong fix — a token refresh would have done nothing.

## One thing that may still be worth knowing

The `.qch` files were deleted from the tree in 8884e7913 (15 July), but **nothing stops them being regenerated**:

| | State on master |
|---|---|
| `.qch` files in the tree | gone ✓ |
| `Doxyfile` → `GENERATE_QHP` | still `YES` |
| `Doxyfile` → `QCH_FILE` | still `../QElectroTech.qch` |
| `.gitattributes` | still `*.qch filter=lfs …` |
| `auto-doxygen.yml` | still runs `git lfs track "*.qch"` |

So if the action were un-paused today it would rebuild the ~990 MB `.qch`, LFS-track it, and consume the allowance again. The deletion cleaned up what was there; the generator that produces them is untouched.

## Offer

I have the rework you described ready as a small change, if it is useful:

- `Doxyfile`: `GENERATE_QHP = NO`
- `.gitattributes`: drop the `*.qch` LFS rule
- `auto-doxygen.yml`: drop the `git lfs track` step, drop the `Create Pull Request` step (its `add-paths: doc/*.qch` meant the `.qch` was the only thing it ever committed), and drop `qhelpgenerator-qt5` from the apt install

What survives is the half people actually read: doxygen builds `doc/html/` and `upload-pages-artifact` + `deploy-pages` publish it to GitHub Pages. No LFS involvement at all.

The trade-off is explicit: **Qt Assistant help files would no longer be produced.** I read your comment as having already decided that, but I would rather ask than assume — say the word and I will open the PR, or leave it entirely with you if you would prefer to do it yourself.

I am verifying locally that a doxygen run with these settings produces no `.qch` and still produces the HTML, and will confirm before opening anything.

scorpio810:
It 's easy to compile doxyfile in your local git.
`time doxygen Doxyfile`

```
Patching output file 1205/1213
Patching output file 1206/1213
Patching output file 1207/1213
Patching output file 1208/1213
Patching output file 1209/1213
Patching output file 1210/1213
Patching output file 1211/1213
Patching output file 1212/1213
Patching output file 1213/1213
Running qhelpgenerator...
Building up file structure...
Insert custom filters...
Insert help data for filter section (1 of 1)...
Insert files...
Insert contents...
Insert indices...
Documentation successfully generated.
type lookup cache used 14704/65536 hits=56280 misses=15544
symbol lookup cache used 48038/65536 hits=309641 misses=48038
finished...

real    2m33,006s
user    12m44,653s
sys     7m34,617s
```

```
laurent@debian:~/Qet-svn/git/qet/doc$ du -sh html/
985M    html/
```
HTML Doxygen was so big to run on each PR, we need to return in doxyfile history and search light version.
```
laurent@debian:~/Qet-svn/git/qet/doc$ du -h QElectroTech.qch 
737M    QElectroTech.qch
```

scorpio810:
I used to run Doxygen locally and synchronise the files to the remote server using rsync, but that requires me to remember to do it, and sometimes I don’t have the time.

ispyisail:
I measured this rather than guessing, since you asked for a lighter Doxyfile. Two things: one finding that saves you a search, and the numbers.

## There is no lighter version in the history to return to

> we need to return in doxyfile history and search light version

`Doxyfile` has **exactly one commit** in this repository — `a8e2a7acf` (19 June 2026), which **added** it whole at 2838 lines:

```
$ git log --oneline -- Doxyfile
a8e2a7acf 2026-06-19 concat QString with '%'

$ git log --diff-filter=A --stat -- Doxyfile
 Doxyfile | 2838 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```

So there is nothing earlier to go back to. Measuring forward is the only route, which is what I did.

## Numbers

Doxygen 1.15.0, graphviz 14.1.2, Ubuntu 26.04, on current `master`. Each row is a full clean run; only the named settings differ:

| Settings | Size | Time | Files | Images |
|---|---|---|---|---|
| **baseline, as shipped** | **1012 MB** | 134 s | 29 410 | 8 574 |
| `CALL_GRAPH=NO` `CALLER_GRAPH=NO` | **405 MB** | 43 s | 9 244 | 1 852 |
| above **+ `DOT_IMAGE_FORMAT=svg`** | **211 MB** | 22 s | 10 920 | 3 528 |
| `HAVE_DOT=NO` (no graphs at all) | **74 MB** | 8 s | 3 856 | 343 |

My baseline of 1012 MB lines up with your 985 MB — different doxygen and graphviz versions, same ballpark — so the rest of the table should transfer to your machine.

## What that says

**Call and caller graphs are the whole problem.** Turning off just those two takes it from 1012 MB to 405 MB and from 134 s to 43 s — a 60 % size cut and 3× faster, from two lines. They are generated *per function*, which is also why your CPU time was five times your wall time.

**SVG is close to free and halves it again.** `DOT_IMAGE_FORMAT=svg` gets 405 MB down to **211 MB** and 43 s down to 22 s, while keeping every remaining graph — class, collaboration, include, hierarchy, directory. Vector output also scales properly in a browser, so the diagrams are arguably better rather than merely smaller.

**Dropping graphs entirely gives 74 MB**, but I would not: the class and collaboration diagrams are the genuinely useful part of a Doxygen build.

## Suggestion

`CALL_GRAPH=NO`, `CALLER_GRAPH=NO`, `DOT_IMAGE_FORMAT=svg` gives **1012 MB → 211 MB (−79 %) and 134 s → 22 s (6× faster)** while keeping the diagrams people actually look at. Three lines.

If you would rather keep call graphs in some form, `DOT_GRAPH_MAX_NODES` (currently 50) and `MAX_DOT_GRAPH_DEPTH` (currently 0, meaning unlimited) are the softer dials — I can measure a couple of those combinations too if it would help.

One small correction to the earlier discussion: **`SOURCE_BROWSER` is already `NO`** (Doxyfile line 1157), so it is not contributing here. The cost is entirely the dot graphs.

Happy to send this as a PR, or leave it to you — it is your call, and it is a visible change to the documentation rather than a pure cleanup.

arummler:
I am still not sure what the issue with having the doxygen HTML on gh-pages is. I understand that committing it to commit it git LFS is a rate limited I did not even see a reason for that.  

In my comprehensive CI/CD rework #722 @ispyisail I did actually that in: on-pull only gh-pages is updated. I left git LFS for tagging but we still have not decided how the workflow should look like. I don't insist on it but please have a look at the proposed workflow before re-writing the action. And we had already there the discussion that by reducing the depth / switch to svg the size can be reduced by a factor 10.

@scorpio810 You mention the web server: I assume that refers to where your eternal web page is hosted? This is also something that could be easily automatized. But at the same time I think it would investigate the heavy duplication of ressources and sketch a goal how it should look like. A  lot of the pages are outdated and it imposes quite a maintenance burden.
- github wiki vs qelektrotech wiki (both rudimentary)
- git issues vs bug tracker (with different accounts)
- doxygen API documentation hosted on github vs hosted on remote server
- manual hosted on github vs hosted on remote server
    - forum links an old 0.9 english version and an even older 0.8 czech version
    - web page itself links to a new generated version, the old 0.8 chech version and a bunch of pdf files where only the entry page of the manual was auto trasnlated to a bunch of languages in 2020 - not sure what the purpose was

I would propose to sketch at least an outline before applying fixes to that... But best in a new githib discussion or issue rather than attached to a barely related PR.

ispyisail:
I understand the problem is $ cost.  Running too many services costs money

ispyisail:
> A lot of the pages are outdated and it imposes quite a maintenance burden.

Whats your recomendation?

arummler:
> I understand the problem is $ cost. Running too many services costs money

This is something where I am not sure we are talking about the same "services". Microsoft (==github) is giving us a lot of things for free ("free" as in "free beer"). Naturally this can change in the future at any moment and one can have political opinions about the hidden cost for these gifts...

But in the moment basically all github actions are free/unlimited and from the tools which this project uses the only one with limitations likely to be approached is git LFS. That's the one where Laurent ran into the montly transmission bandwidth limitations by pushing to git LFS on each merge.

arummler:
> > A lot of the pages are outdated and it imposes quite a maintenance burden.
> 
> Whats your recomendation?

Considering that Laurent is the only administrator of this project afaik and lacks time I would recommend to consolidate and automatize as much as possible. My impression is that a lot of time has to be spent to collect information from github/forum/bug tracker and re-post on the others, making a release would require a check list and multi day manual process. But I might be mistaken, please take with a grain of salt.

I already mentioned some proposal here and there, particularly in my very long MR. I can try to write it down later once more in an issue or discussion to have it in one place.

scorpio810:
https://github.com/qelectrotech/qelectrotech-source-mirror/pull/833#issuecomment-5612121751

@arummler @ispyisail Previously, on Tuxfamily.org, we had to keep an eye on the disk quota; unfortunately, I was forced to switch to a paid server as Tuxfamily was on the verge of shutting down.

I don’t think we need to run Doxygen on every commit, and as I’ve got hundreds of gigabytes of free space on the server, and building the HTML and QCH takes me less than five minutes – not to mention uploading it to the server via rsync – I’ll be favouring this solution going forward.

scorpio810:
>Considering that Laurent is the only administrator of this project afaik and lacks time.

Yes, after a massive project this August, I’m absolutely knackered and couldn’t take any more, so I took a few weeks off to rest and maybe go to the beach, but you’re all asking me for things a lot at the moment. ;-)

ispyisail:
> > Considering that Laurent is the only administrator of this project afaik and lacks time.
> 
> Yes, after a massive project this August, I’m absolutely knackered and couldn’t take any more, so I took a few weeks off to rest and maybe go to the beach, but you’re all asking me for things a lot at the moment. ;-)

Hugs

PR #825, opened 2026-09-08
Count master contacts in one place, and count contacts not elements

ispyisail:
First step of the contact-counting work discussed in #819. No user-visible feature yet — this puts the counting in one place and fixes a miscount, so the displays that follow cannot disagree with each other.

> **Stacked on #824.** This branch contains that one-line Qt5 build fix as its parent commit, because master does not currently compile with the default `QT_VERSION_MAJOR=5`. Review only the second commit here; the diff shrinks to it once #824 merges.

## The bug

`MasterElement::isFull()` decided whether a coil had room left with:

```cpp
return connected_elements.size() >= max_slaves;
```

That counts linked **elements**. A slave stands for as many contacts as its `number` kind information declares, so a 4-pole contact consumed one contact from the coil's budget instead of four — a coil declaring 4 contacts would accept four 4-pole contacts, i.e. 16. 36 elements in the standard collection declare a `number` between 2 and 4, so this is reachable rather than theoretical.

## The change

`ContactUsage` (`sources/contactusage.h`) holds the two rules that are easy to get wrong:

- **a slave counts once per contact it declares**, not once per element;
- **a changeover is counted once**, as `sw`, never as one NO plus one NC. `CrossRefItem::NOElements()` and `NCElements()` both return changeovers, so any count built by adding those two lists reports one changeover as two contacts.

`MasterElement::contactUsage()` is the thin wrapper that feeds it the linked elements, and `isFull()` now reads it. The per-type displays proposed in #819 — the used count in the element's General tab, and the per-type budget on the cross-reference — need exactly this count, which is why it lives in one place rather than being written out three times.

The header carries no graphics dependency, following the `diagramsortkeys.h` pattern, so the rules are unit tested on their own rather than behind a `QGraphicsItem`.

## Behaviour change

Only for coils that **opt in** to a limit: `max_slaves` defaults to `-1` (unlimited), and no project in `examples/` sets it. For those that do, the cap now counts correctly, which means it is reached sooner than before — by design, since the old number was wrong.

Per #819, whether that cap should warn instead of refusing the link is a separate question still open with @scorpio810, and is deliberately **not** part of this PR.

## Testing

- `tests/qttest/tst_contactusage.cpp` — 9 cases covering both rules plus the degenerate contact counts; runs under `ctest` alongside the existing tests.
- Both rules mutation-checked: reverting to element-counting fails 2 cases, counting a changeover as both NO and NC fails 3.
- All 23 projects in `examples/` still load, re-save (`--resave`) and export (`--export-bom`) with no crash or hang.
- `qet-lint` reports no regressions against its baseline.

Qt 5.15.18, Ubuntu 26.04, gcc 15.2.0.

### DIFF
--- a/sources/contactusage.h
+++ b/sources/contactusage.h
@@ -1,0 +1,85 @@
+/*
+	Copyright 2006-2026 The QElectroTech Team
+	This file is part of QElectroTech.
+
+	QElectroTech is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 2 of the License, or
+	(at your option) any later version.
+
+	QElectroTech is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with QElectroTech.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#ifndef CONTACTUSAGE_H
+#define CONTACTUSAGE_H
+
+#include <algorithm>
+
+/**
+	@brief The ContactUsage struct
+	How many slave contacts a master element currently uses, broken down
+	by contact type.
+
+	Header-only and free of any graphics dependency so that the counting
+	rules can be unit tested on their own. MasterElement::contactUsage()
+	is the thin wrapper that feeds it the linked elements.
+
+	Two rules are easy to get wrong, and both live here so that every
+	caller gets them right:
+	 - a slave stands for as many contacts as its "number" kind
+	   information says, so a 4 pole contact counts as 4, not as 1
+	 - a changeover contact is counted once, as sw. CrossRefItem's
+	   NOElements() and NCElements() both return it, so adding those two
+	   lists together would count it twice.
+*/
+struct ContactUsage
+{
+	/**
+		Contact types a slave can declare. Mirrors
+		ElementData::SlaveState, which is not used directly so that this
+		header stays free of the element data dependencies and can be
+		unit tested on its own. MasterElement::contactUsage() maps
+		between the two.
+	*/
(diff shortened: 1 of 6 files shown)


ispyisail:
**Correction: the premise of the original commit here was wrong, and I have pushed a commit reverting that part.** Flagging it prominently rather than quietly, since the first commit's message states the wrong reasoning.

I claimed `MasterElement::isFull()` was under-counting because it compared linked *elements* against `max_slaves`. It was not. **`max_slaves` is a number of slots, not of contacts**, and the original comparison was correct:

- it sizes the contact-group table in the element editor, one row per slot — `ElementPropertiesEditorWidget::populateSlaveGroupsTable()` grows and trims the group list to exactly the spinbox value;
- `ContactGroupSelectionDialog` disables a group unless `g.contactCount == slaveData.m_contact_count`, so a 4-pole slave needs a group *declaring* 4 and occupies that single group;
- each slave stores exactly one group index (`Element::setGroupIndexForElement`), and used indices are struck out one per linked slave.

So a coil declaring 4 slots accepts 4 slaves whatever their pole count. My change made a single 4-pole slave fill a 4-slot coil on its own and refuse three further links that should have been allowed — a regression, not a fix. Reverted in the commit just pushed.

Worth saying plainly: **@IBSYSLevi read this correctly in #819** and I implied otherwise. He wrote that the property is "simply used to specify how many contacts you want to define on the coil master", which is exactly what it does. I corrected him on a different point — that it does still block linking, which is true — and in doing so wrongly cast doubt on his reading of its meaning.

### What this PR is now

`ContactUsage`, a per-type contact tally with its unit tests, consumed by the displays in #826 and #827. It no longer changes any behaviour: `isFull()` is back to the original comparison.

The tally itself is unaffected by the mistake and still answers the question @scorpio810 asked — *how many contacts must an auxiliary block provide* — which is a genuinely different question from *how many slots are occupied*, and only the former wants the pole multiplier. The header now says so explicitly, so the next reader does not repeat my conflation.

I would understand if you would rather I closed this and folded the tally into #826; say the word.

PR #826, opened 2026-09-08
Show the NO/NC/SW contact breakdown in a master's General tab

ispyisail:
Implements @scorpio810's request from #819: after drawing a schematic, see how many NO, NC and changeover contacts a coil ended up using, so you can pick an auxiliary block that satisfies it — without counting rows on the cross reference by hand.

Shown in the element's **General tab**, as asked, not on the cross reference.

> **Stacked on #825** (which is itself stacked on #824, a one-line Qt5 build fix). Only the last commit belongs to this PR; the diff shrinks to it as the parents merge.

## What it looks like

```
Nombre maximum de contacts esclaves définis : non défini
Nombre de contacts esclaves utilisés : 8
    NO : 6, NC : 1, inverseurs : 1, autres : 0
```

## Three changes to that block

**The breakdown line**, printed only when the master actually has contacts to break down.

**The existing total now counts contacts, not linked elements.** That line's label already read *"Nombre de contacts esclaves utilisés"* while the value was `linkedElements().count()`, so a slave standing for several contacts was under-reported. It now reads `MasterElement::contactUsage()` — the same count `isFull()` uses, so the dialog and the limit can no longer disagree.

**A declared limit of `-1` prints as "non défini"** instead of `-1`. `-1` is the sentinel for "no limit set", and showing it raw reads as a bad value. This is the display half of the confusion @IBSYSLevi raised in #819 about what `max_slaves` actually means.

## One open question, for @scorpio810

You wrote *"only for coil type I'm think"*. **This PR currently shows the breakdown for every master type, not only coils.** My reasoning: the count is equally accurate for a protection relay or any other master, and gating on `kindInformations()["type"] == "coil"` would hide it from those for no real benefit — while adding a magic string comparison. The line only appears when a master has linked contacts, so it stays out of the way otherwise.

Happy to restrict it to coils if you would rather stick to the narrower scope — it is a two-line change, so say the word.

## Testing

- The counting itself is unit tested in #825 (`tst_contactusage`, 9 cases, both counting rules mutation checked).
- `static_cast<MasterElement *>` is safe here: only `MasterElement` passes `Element::Master` to the base constructor, and `ElementFactory` returns `MasterElement` for that link type.
- 22 of the 23 projects in `examples/` load, re-save (`--resave`) and export (`--export-bom`) with no crash or hang. The 23rd is `schema_indus.qet`, which hangs on any headless verb for the unrelated reason in #661.
- Qt 5.15.18, Ubuntu 26.04, gcc 15.2.0.

**Not yet visually confirmed on screen.** I drove the GUI harness far enough to load a project with this build, but not far enough to open a specific master's properties dialog, so the rendered wording is unverified by screenshot. The logic behind it is covered by the tests above. If the wording or layout is not what you want, it is a one-line change.

New strings are French, per the project's `tr()` source language. `.ts` files are untouched, since `lupdate` is a manual target.

### DIFF
--- a/sources/contactusage.h
+++ b/sources/contactusage.h
@@ -1,0 +1,85 @@
+/*
+	Copyright 2006-2026 The QElectroTech Team
+	This file is part of QElectroTech.
+
+	QElectroTech is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 2 of the License, or
+	(at your option) any later version.
+
+	QElectroTech is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with QElectroTech.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#ifndef CONTACTUSAGE_H
+#define CONTACTUSAGE_H
+
+#include <algorithm>
+
+/**
+	@brief The ContactUsage struct
+	How many slave contacts a master element currently uses, broken down
+	by contact type.
+
+	Header-only and free of any graphics dependency so that the counting
+	rules can be unit tested on their own. MasterElement::contactUsage()
+	is the thin wrapper that feeds it the linked elements.
+
+	Two rules are easy to get wrong, and both live here so that every
+	caller gets them right:
+	 - a slave stands for as many contacts as its "number" kind
+	   information says, so a 4 pole contact counts as 4, not as 1
+	 - a changeover contact is counted once, as sw. CrossRefItem's
+	   NOElements() and NCElements() both return it, so adding those two
+	   lists together would count it twice.
+*/
+struct ContactUsage
+{
+	/**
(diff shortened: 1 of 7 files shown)


ispyisail:
**Correction pushed**, following the mistake described in #825: I had been stacking two different units in the same block.

`max_slaves` is a number of **slots**, not of contacts (details in #825). Reporting the line below it in contacts made a coil with one 4-pole slave read:

```
Nombre maximum de contacts esclaves définis : 4
Nombre de contacts esclaves utilisés : 4        <- wrong: 3 slots still free
```

The total now goes back to counting linked elements, which is the original behaviour of that line and matches the unit of the line above it. So this PR no longer changes that line at all.

The per-type breakdown keeps the pole multiplier, because that is the question it answers — how many contacts an auxiliary block must provide — and is now prefixed `Contacts :` so the two units cannot be mistaken for one another:

```
Nombre maximum de contacts esclaves définis : non défini
Nombre de contacts esclaves utilisés : 5
    Contacts : NO : 6, NC : 1, inverseurs : 1, autres : 0
```

Used 5 slots, needing 8 contacts. Those are both true and they are different numbers, which is exactly why they now carry different labels.

Re-verified after the change: 10 unit tests pass, and 22 of the 23 example projects load, re-save and export with no crash or hang (`schema_indus.qet` excluded, #661).

The coil-only scope question from #819 is still open and unaffected by this.

PR #827, opened 2026-09-08
Show used against declared capacity where a master declares contact groups

ispyisail:
Second half of the display work in #819, and the counterpart to @Kellermorph's contact-group model. Where a coil declares what contacts it provides, the General tab reports each type as used against declared rather than as a bare count:

```
Nombre maximum de contacts esclaves définis : 8
Nombre de contacts esclaves utilisés : 4
    NO : 3/4, NC : 1/2, inverseurs : 0/1, autres : 0/0
```

> **Stacked on #826 → #825 → #824.** Only the last commit belongs to this PR; the diff shrinks as the parents merge.

## How it works

`MasterElement::contactCapacity()` sums `contactCount` per type over the element's `SlaveContactGroup` list, reusing the same `ContactUsage` tally that the used count is built on. The mapping from `ElementData::SlaveState` onto that tally's own type is factored into a single helper, so the used count and the declared capacity cannot classify the same contact type differently — with two separate switches they eventually would.

A type used beyond what is declared reads as **`1/0`**. That is deliberate rather than a formatting accident: it says this contact does not fit the part.

## Please read this before merging

**Nothing in the standard collection declares contact groups.** I checked: 0 of the 6,918 `.elmt` files in `10_electric`, 0 of the example projects, and no `.elmt` anywhere in the tree contains `slaveContactGroups`.

Two consequences worth being explicit about:

- **This changes no existing display.** `capacity.total() > 0` is false for every element that exists today, so every current project falls back to the plain count from #826. The risk to existing users is essentially nil.
- **It is equally invisible to them.** Until elements are authored that declare groups, this display cannot appear for anybody.

So this makes the capability *usable* — plausibly part of why it is unused, since until now nothing showed a capacity back to you — but it will not light up on its own. If you would rather it landed alongside a worked example element, or after @Kellermorph's manufacturer/part-number idea has settled the shape of that data, that is a reasonable call and I am happy to hold it.

## Scope

**Display only.** Whether a declared capacity should also feed `MasterElement::isFull()` — and whether that cap should warn rather than refuse — is the open question in #819 and is deliberately untouched here.

## Testing

No existing element exercises this path, so I built a fixture: a coil declaring two NO groups of two contacts, one NC group of two, and one changeover group of one. Loaded through QET with temporary instrumentation, `contactCapacity()` reported:

```
groups=4  NO=4  NC=2  SW=1  other=0  total=7
```

matching the declaration exactly, which confirms the whole path — XML parsed into `m_slave_contact_groups`, type mapping, and per-type summing including two groups of the same type adding together. The instrumentation was removed before committing.

- `tst_contactusage` gains a case for capacity summed across groups: 10 cases, all passing.
- The 22 loadable projects in `examples/` still load, re-save (`--resave`) and export (`--export-bom`) with no crash or hang (`schema_indus.qet` excluded — it hangs on any headless verb for the unrelated reason in #661).
- Qt 5.15.18, Ubuntu 26.04, gcc 15.2.0.

**Not visually confirmed on screen**, same as #826: the wording above is what the code produces, not a screenshot. If the French reads badly I would rather fix it before merge.

### DIFF
--- a/sources/contactusage.h
+++ b/sources/contactusage.h
@@ -1,0 +1,85 @@
+/*
+	Copyright 2006-2026 The QElectroTech Team
+	This file is part of QElectroTech.
+
+	QElectroTech is free software: you can redistribute it and/or modify
+	it under the terms of the GNU General Public License as published by
+	the Free Software Foundation, either version 2 of the License, or
+	(at your option) any later version.
+
+	QElectroTech is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with QElectroTech.  If not, see <http://www.gnu.org/licenses/>.
+*/
+#ifndef CONTACTUSAGE_H
+#define CONTACTUSAGE_H
+
+#include <algorithm>
+
+/**
+	@brief The ContactUsage struct
+	How many slave contacts a master element currently uses, broken down
+	by contact type.
+
+	Header-only and free of any graphics dependency so that the counting
+	rules can be unit tested on their own. MasterElement::contactUsage()
+	is the thin wrapper that feeds it the linked elements.
+
+	Two rules are easy to get wrong, and both live here so that every
+	caller gets them right:
+	 - a slave stands for as many contacts as its "number" kind
+	   information says, so a 4 pole contact counts as 4, not as 1
+	 - a changeover contact is counted once, as sw. CrossRefItem's
+	   NOElements() and NCElements() both return it, so adding those two
+	   lists together would count it twice.
+*/
+struct ContactUsage
+{
+	/**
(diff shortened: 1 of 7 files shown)


ispyisail:
**Correction merged down** from #825 and #826 — see #825 for the underlying mistake (`max_slaves` counts slots, not contacts).

Nothing in this PR's own logic changes. `contactCapacity()` sums `contactCount` across the element's groups, which is a count of **contacts**, and the used side of the ratio is contacts too, so the `3/4` form was consistent all along. Only the label has changed, to keep it distinct from the slot count above it:

```
Nombre maximum de contacts esclaves définis : 8
Nombre de contacts esclaves utilisés : 5
    Contacts : NO : 3/4, NC : 1/2, inverseurs : 0/1, autres : 0/0
```

Still display-only, and still true that no element in the standard collection declares contact groups today, so this changes nothing for existing projects.

Re-verified after the merge: the purpose-built fixture still parses and reports `NO=4 NC=2 SW=1 total=7` as declared, 10 unit tests pass, and 22 example projects load, re-save and export with no crash or hang.

scorpio810:
Screenshot
Great, thanks Shane.

PR #828, opened 2026-09-09
Take the slot count from the contact groups when an element declares them

ispyisail:
An element can declare contact groups and a `max_slaves` that disagree with each other, and nothing reconciles them.

> **Stacked on #824** only, because master does not currently compile with the default `QT_VERSION_MAJOR=5`. This needs nothing from #825–#827 and can be reviewed and merged independently of them.

## The problem

The element editor keeps the two in step — `max_slaves` sizes the contact-group table, one row per slot (`ElementPropertiesEditorWidget::populateSlaveGroupsTable`). Nothing does so on load.

So a hand-written or generated file can carry five groups and `max_slaves=2`. It loads without complaint. `MasterElement::isFull()` then caps linking at two, while `ContactGroupSelectionDialog` still offers all five groups — the user is shown groups that cannot be linked to, with nothing to explain why.

## The fix

When groups are declared, **they are the slots**: a slave occupies exactly one, and the selection dialog offers exactly these. So the limit comes from the group count, which is also the number the user can actually see. `max_slaves` stays as the fallback for elements that declare no groups.

## Scope

**No element in the standard collection declares contact groups** — 0 of the 6,918 `.elmt` files — so this changes nothing for any existing project. It matters once elements are authored that do declare them, which is the direction discussed in #819.

## Testing

No real element exercises either path, so I built two fixtures and confirmed both branches with temporary instrumentation (removed before committing):

| Fixture | Result |
|---|---|
| coil declaring 5 groups, `max_slaves=2` | takes the **group** path — limit 5, matching what the dialog offers |
| coil with `max_slaves=2` and no groups | takes the **fallback** path, behaviour unchanged |

Also: 22 of the 23 projects in `examples/` load and export (`--export-bom`) with no crash or hang. The 23rd is `schema_indus.qet`, which hangs on any headless verb for the unrelated reason in #661.

Qt 5.15.18, Ubuntu 26.04, gcc 15.2.0.


ispyisail:

ispyisail:
Reopening — closed to clear @scorpio810's review backlog rather than on merit. Merged with current `master` and reviewed with particular care, because I got `MasterElement::isFull()` wrong once before: in #825 I changed it to count *contacts*, which was incorrect, and had to revert it after @IBSYSLevi read it correctly and I doubted them.

**This change is not that mistake.** It still counts slots — it only changes where the slot count comes from. Two things make the group count the right limit, and both are checkable:

- **A slave occupies exactly one group.** `m_group_indices` maps `Element*` to a single `int`, and `Element::groupIndexForElement()` returns one index per slave.
- **A group holds at most one slave.** `LinkSingleElementWidget` builds `used_indices` from every already-linked slave, and `ContactGroupSelectionDialog` renders those rows struck out as *"(déjà assigné)"*.

So groups and slaves are one-to-one, and the number of groups *is* the maximum number of slaves. `max_slaves` remains the limit for elements that declare no groups, which is the only case it can still describe correctly.

The bug is real rather than theoretical: nothing reconciles `max_slaves` with the group table when a file is loaded, so an element carrying five groups and `max_slaves=2` would have the dialog offer five slots while linking silently stopped at two.

**One thing worth flagging about the merge.** This branch also carried a Qt5/Qt6 `QLabel::pixmap()` fix for `imagetransparentcolordialog.cpp`, written before `master` gained its own. Master's version is the better one — it uses `pixmap(Qt::ReturnByValue)` rather than dereferencing the deprecated pointer overload — and the merge kept it. The net diff against `master` is now only the 16 lines in `masterelement.cpp`, which is what this PR should be.

**Not verified:** linking more than `max_slaves` slaves to a group-declaring master interactively.

PR #830, opened 2026-09-09
Add smart device metadata to the existing BOM export

enesgursoy6110:
## Summary

- Adds smart device metadata fields through the existing element information/property flow.
- Extends the existing SQL nomenclature view and existing GUI/CLI BOM export instead of introducing a second BOM implementation.
- Uses one shared query for GUI and CLI exports; Simple and Master elements are included, while slave contacts, terminals and excluded entries are omitted.
- Keeps ungrouped rows as the default and preserves the existing grouped export option.
- Writes semicolon-separated UTF-8 CSV with a BOM, RFC-style quote escaping, and atomic `QSaveFile` replacement.

## Metadata and compatibility

The added fields are model, category, voltage rating, current rating and notes. Existing designation, manufacturer and manufacturer reference fields remain the source for description, manufacturer and part number. Data continues to live in the existing element information XML map, so there is no project format version bump. Older projects load with empty/default values.

## Tests

- Qt 6.11.2, CMake, Debug, `BUILD_WITH_KF=OFF`: full application build succeeds.
- CTest: 4/4 tests pass, including the explicitly registered `tst_smart_device` target.
- The smart-device test covers Simple/Master inclusion, slave/terminal exclusion, `exclude_from_bom` boolean spellings, stable ordering, non-ASCII/non-Latin UTF-8 data, CSV quoting/newlines and atomic writing.
- A real `qelectrotech.exe --export-bom` run against a two-device sample produced the expected rows and UTF-8 BOM.

## Scope

This PR intentionally does not add terminal planning, cable management, PLC I/O, panel layout or catalogue integration.


docs/smart-device-bom.md
```diff
@@ -0,0 +1,136 @@
...
+import also preserves leading zeros.
+`QSaveFile` commits atomically; open, write and commit errors are reported.
+Cancel does not write a file. Read-only projects can still be exported.
+
+## Verification
+
+On the configured Windows/MSYS2 Qt5 system, from the repository root:
+
+```powershell
+.\tests\qttest\run-smart-device-tests.ps1
```
arummler: The test is only foreseen to run under a Windows system as it is written in power shell? Why? Is there anything OS specific? The function should work cross-platform.

docs/smart-device-bom.md
```diff
@@ -0,0 +1,136 @@
...
+are exported verbatim so device tags and part numbers are not altered. CSV
+quoting does not disable spreadsheet formulas: only export trusted projects,
+or import all columns as **text** when inspecting untrusted project data. Text
+import also preserves leading zeros.
+`QSaveFile` commits atomically; open, write and commit errors are reported.
+Cancel does not write a file. Read-only projects can still be exported.
+
+## Verification
+
+On the configured Windows/MSYS2 Qt5 system, from the repository root:
```
arummler: Can you test with Qt6? Qt5 is end of life, has been removed from the CI and I would expect it to be removed soonish from the code as well.

docs/smart-device-bom.md
```diff
@@ -0,0 +1,136 @@
...
+```powershell
+.\tests\qttest\run-smart-device-tests.ps1
+```
+
+The runner builds the application with `debug` and `no_kf5`, then links a QtTest
+executable against the real application objects. Use `-SkipApplicationBuild`
+only after building the current application sources. Results and generated
+sample `.qet`, CSV and properties screenshot are under
+`build-debug/smart-device-tests/`. CMake's explicit source list includes the
+new files; qmake discovers them through its existing globs.
```
arummler: In tests/qttest/CMakeLists.txt I do not see globbing used (and that's also better like that). It would need to be listed explicitly. I do not know what the future of the .pro files in this project is but since the swap to Qt6 CMake is the build system default; mostly because Qt intends to move fully to CMake long term.

docs/smart-device-bom.md
```diff
@@ -0,0 +1,136 @@
...
+1. Open an existing project and edit a normal or master element's information
+   tab. Existing fields and the new Model/Category/ratings/Notes rows must show.
+2. Enter manufacturer, part number, quantity and new properties. Apply, undo,
+   redo, save, close and reopen. Verify the same values are present.
+3. Give two elements the same part number on different pages. Export through
+   Project > Export BOM as CSV. Expect separate rows with their own tags/pages.
+4. Enable the existing exclusion checkbox for one element and export again.
+   Its row must be absent; changing a property without saving must affect CSV.
+5. Open CSV with a UTF-8 CSV reader (import text columns when using Excel).
+   Check Turkish characters, leading-zero part numbers, commas and quotes.
```
arummler: I assume it says turkish here because that's your personal target language but in the instructions it should probably read sth. like non-ASCII/ideally even non-latin language to verify that UTF-8 is read properly.

ispyisail:
Reviewed this by building it rather than only reading the diff. The code itself is careful — but there are three structural points I think are worth settling before this goes further, and the first is a question for the maintainers rather than something to fix in the code.

## Verified

Built on **Linux with Qt 6.10.2**, gcc 15.2.0, Ubuntu 26.04 — the configuration CI actually uses, and a different one from the Qt5/MinGW build in the description:

- **794/794 targets, zero errors**, binary links and reports `0.200.1-dev`.

So the code is portable even though the test harness is not (see below). Worth knowing, since you could only exercise one platform.

## 1. QET already has a BOM export

This is the part I would settle first, because it is a design decision rather than a defect.

`exportBom()` in `sources/cli_export.cpp:320` already exists and is reachable as `--export-bom`. The two implementations now differ in almost every respect:

| | existing | this PR |
|---|---|---|
| Data source | `element_nomenclature_view` (project database, SQL) | walks `project.diagrams()` → `diagram->elements()` |
| Separator | `;` | `,` |
| Line ending | `\n` | `\r\n` |
| Encoding | plain UTF-8 | UTF-8 with BOM |
| Entry point | CLI verb | Project menu |
| Columns | label, designation, manufacturer, manufacturer_reference, quantity, location, function, title, folio | No, Device tag, Manufacturer, Part number, Model, Description, Category, Quantity, Page, Notes |

They are substantially the same report, produced by two independent code paths that will drift apart. A user is likely to ask which one is *the* BOM, and a bug fixed in one will not reach the other.

To be clear about what is already right: this PR **does** honour `exclude_from_bom` (line 399), which is exactly the kind of thing a reimplementation usually misses. One subtlety though — the existing view filters in SQL with `ei.exclude_from_bom IS NOT 'true'`, while this uses `QVariant::toBool()`. Those disagree for a stored `"1"`, so even the shared flag can diverge between the two exports.

The question for @scorpio810: should the GUI action drive the **existing** `exportBom()` (extending its column set where the new fields add value), rather than standing up a second implementation? That would keep one BOM with one definition of what a device is.

## 2. The tests cannot run outside Windows

The 11 tests look genuine and well chosen, but they are wired as:

- `run-smart-device-tests.ps1` — PowerShell, invoking `qmake-qt5`, `mingw32-make`, `$env:WINDIR`
- `smart_device.pro` — a qmake project file
- and `tests/qttest/CMakeLists.txt` is **not** touched

So they will not run under `ctest`, will not run in CI, and cannot run on Linux, macOS or FreeBSD at all. The project's convention is CMake plus ctest — `tst_diagramsortkeys` in the same directory is the pattern to copy, and it is about six lines:

```cmake
add_executable(tst_smart_device tst_smart_device.cpp)
add_test(NAME tst_smart_device COMMAND tst_smart_device)
target_include_directories(tst_smart_device PRIVATE ${QET_DIR}/sources)
target_link_libraries(tst_smart_device PRIVATE Qt::Test)
```

As it stands the tests exist but nobody else can run them, including reviewers.

## 3. The `.ts` files should not be in the diff

`lang/qet_en.ts` and `lang/qet_tr.ts` carry 108 added lines, including hand-written Turkish translations. In this project French is the `tr()` source language and the `.ts` files are regenerated by the manual `update_translations` (lupdate) target — hand edits get overwritten on the next run and conflict in the meantime.

The new `tr()` strings in the sources are exactly right; it is only the generated `.ts` changes that should come out.

## Smaller notes

- `docs/smart-device-bom.md` is a genuinely good addition — the column-mapping table makes the intent easy to check.
- Voltage and current ratings being stored but deliberately left out of the CSV is a sensible call, and saying so in the description was helpful.

None of this is a reflection on the quality of the code, which reads well. The duplication question is the one I would resolve first, since the answer changes how much of the rest matters.

arummler:
Regarding the Claude analysis: I seems that the export part should be rather consolidated into the existing export (make it accessible from GUI, options for file format).

ispyisail:
> Regarding the Claude analysis: I seems that the export part should be rather consolidated into the existing export (make it accessible from GUI, options for file format).

Agree

docs/smart-device-bom.md
enesgursoy6110: Addressed in afb5feb55: the PowerShell/qmake runner was removed. The test is now a normal cross-platform CMake/CTest target.

docs/smart-device-bom.md
enesgursoy6110: Addressed in afb5feb55. I configured and built the full application with Qt 6.11.2, CMake, Debug, and BUILD_WITH_KF=OFF; the full build succeeds and all four registered tests pass.

docs/smart-device-bom.md
enesgursoy6110: Addressed in afb5feb55: tst_smart_device is explicitly registered in tests/qttest/CMakeLists.txt. The separate .pro file was removed.

docs/smart-device-bom.md
enesgursoy6110: Addressed in afb5feb55: the documentation and test now describe and cover non-ASCII/non-Latin UTF-8 input, including Japanese text.

enesgursoy6110:
Thanks for the architectural feedback. I consolidated the feature into QET's existing BOM path in `afb5feb55`:

- removed the parallel project-menu exporter and device-walking collector;
- extended `element_nomenclature_view` and made the existing `BOMExportDialog` and `--export-bom` use the same shared query/CSV writer;
- retained ungrouped output as the default and the existing grouped checkbox as an option;
- normalized `exclude_from_bom` values (`true`, `1`, `yes`, `on`);
- removed generated `.ts`/`.qm` changes;
- replaced the Windows-only qmake runner with an explicitly registered CMake/CTest test;
- validated a full Qt 6.11.2 Debug/no-KF build and 4/4 CTest tests.

The PR description now documents the consolidated architecture and verification.

scorpio810:
This branch has conflicts that must be resolved

enesgursoy6110:
Follow-up: I rebased the branch onto current `master` (`99e9ce5ac`) after the translation and wiring-list changes landed. The current head is `77f8e26ff`.

During the rebase I retained upstream's nomenclature element-type filter and combined it with the normalized `exclude_from_bom` handling. Generated translation files remain outside the diff. A fresh Qt 6.11.2 Debug/no-KF full build succeeds and CTest passes 4/4 on the rebased head.

PR #832, opened 2026-09-10
Stop generating the Qt help file, so doxygen no longer needs Git LFS

ispyisail:
The rework @scorpio810 described in #510: stop the doxygen job producing the Qt help file, so it no longer touches Git LFS.

> Unfortunately, I therefore had to pause the GitHub 'Auto-build Doxygen docs' actions [...] I need to rework this file and delete the .qch files (990 MB) containing Qt help files for QET code, to reduce bandwidth usage whilst compiling LFS.

Opening this because it is ready, not to pre-empt you — if you would rather do it yourself, close it and I will not be offended.

## Why the deletion was not enough

The `.qch` files were removed from the tree in 8884e7913, but nothing stopped them being regenerated:

| | State before this PR |
|---|---|
| `.qch` files in the tree | gone |
| `Doxyfile` → `GENERATE_QHP` | `YES` |
| `Doxyfile` → `QCH_FILE` | `../QElectroTech.qch` |
| `.gitattributes` | `*.qch filter=lfs diff=lfs merge=lfs -text` |
| `auto-doxygen.yml` | `git lfs track "*.qch"` |

Un-pausing the action would have rebuilt the ~990 MB help file, pushed it through LFS, and consumed the allowance again.

## What this changes

- **`Doxyfile`** — `GENERATE_QHP = NO`
- **`.gitattributes`** — drop the `*.qch` LFS rule
- **`auto-doxygen.yml`** — drop the Git LFS step; drop the **Create Pull Request** step; drop `qhelpgenerator-qt5` from the apt install

Removing the Create Pull Request step is safe because its `add-paths` was `doc/*.qch` — the help file was the only thing it ever committed. With no `.qch` produced, that step had nothing left to do.

What remains is the half that gets read: doxygen builds `doc/html/`, and `upload-pages-artifact` + `deploy-pages` publish it. **No LFS involvement anywhere in the workflow.**

Net: **+2 / −25**.

## The trade-off, stated plainly

**Qt Assistant help files are no longer produced.** I read your comment as having already decided that, but it is a feature going away, so it should be a deliberate choice rather than a side effect. If you would rather keep the `.qch` and solve the bandwidth another way, this is the wrong PR and I will close it.

## Testing

Ran doxygen 1.15.0 locally against this Doxyfile:

- **zero `.qch` files generated**
- `doc/html/index.html` still produced, 159 MB of HTML

(`graphviz` was not installed locally so the call graphs were skipped; the workflow installs it, and that part is unchanged.)

## Two things this does *not* fix

Worth being explicit so this is not mistaken for a full repair of the documentation pipeline:

1. **The published docs at download.qelectrotech.org stay stale** (0.100.0-dev against 0.200.1-dev on master). This workflow deploys to GitHub Pages, not to that host.
2. **The trigger is still tags only** (`on: push: tags`), so ordinary commits to `master` will not rebuild the documentation even once the action is resumed.

Both are separate decisions from the LFS problem, and I have not touched either.


scorpio810:
https://github.com/qelectrotech/qelectrotech-source-mirror/pull/824#issuecomment-5611188796

scorpio810:
Thanks for this — the LFS fix itself looks solid and the trade-off (no more .qch) is fine by me, per #510.

One thing worth flagging before merging, unrelated to your diff but relevant to what re-enabling this workflow will actually do: the doxygen job's upload-pages-artifact + deploy-pages steps are still there, untouched, and that's actually a problem now — not because of anything in this PR, but because of timing.

Since this action was paused (before windows-msi.yml existed), our nightly Windows build pipeline started deploying its own download page to GitHub Pages, using the exact same actions (upload-pages-artifact@v5 + deploy-pages@v4). Actions-based Pages deployment isn't additive — each run fully replaces whatever's currently live in the github-pages environment. There's no merge, no per-path coexistence.

So once this doxygen action is un-paused, the first time it runs (tag push) it'll overwrite the nightly builds page, and the next Windows nightly build will overwrite the doxygen docs right back. Whichever ran last wins. This never showed up before simply because the doxygen job never ran while the Windows pipeline existed.

This PR isn't the cause and I don't think it needs to block on it, but before we un-pause the action we should sort out one of:

combine both into a single artifact (e.g. doxygen output under a docs/ subpath, generated/merged before one shared deploy-pages step), or
only ever have one job in the repo actually calling deploy-pages, called by the other workflow, or
move doxygen publishing off Actions-based Pages entirely (e.g. push straight to a gh-pages branch subdirectory).

Happy to sort out the actual mechanism — just didn't want this to land and quietly nuke the nightly builds page on the next tag.

scorpio810:
Thought about this a bit more — rather than juggling two independent deploy-pages jobs, I think the cleaner fix is to fold doxygen's output into windows-msi.yml's existing deploy step, so there's exactly one place in the repo that ever calls deploy-pages.

Concretely:

In auto-doxygen.yml: drop the upload-pages-artifact / deploy-pages steps entirely. Just upload doc/html/ as a regular build artifact (or attach it to a tag/release) so it's fetchable elsewhere.
In windows-msi.yml's deploy-pages job: before the Upload GitHub Pages artifact step, add a step that grabs the latest successful auto-doxygen.yml artifact and unpacks it into gh-pages/docs/. If none exists yet (or the last run failed), just skip — same graceful-degradation pattern already used for MSI_URL and the LEGACY_* assets in generate-page.py.
generate-page.py gets a small conditional block (same shape as msi_block/legacy_block) that adds an "API documentation" link only when gh-pages/docs/index.html is present in the fetched artifact.
That way it doesn't matter which workflow runs last (tag push vs. nightly Windows build) — there's only ever one deploy-pages call, and it always republishes the full site (nightly downloads + docs) rather than a partial one that clobbers the other.

Trade-off: the published docs can lag a tag behind whatever's on master between Windows nightly runs, but that's already a limitation you called out above, so I don't think it's a regression.

If that direction sounds reasonable I can put together the actual diff for generate-page.py + both workflow files rather than leaving it as a suggestion on your PR.

scorpio810:
Good catch from both sides — I don't think these numbers actually disagree. @ispyisail's own note explains the gap: his 159 MB run had graphviz missing locally, so CALL_GRAPH/CALLER_GRAPH/etc. were skipped. The CI job does install graphviz, so the real in-workflow output should land much closer to the 985 MB I 'm measured locally with graphviz present, not the 159 MB local test here.

So the LFS/.qch fix in this PR is good to merge as-is — that part's solved regardless. But the underlying "985 MB of HTML, several minutes per run" problem is still there and untouched by this diff, since it's driven by the graph generation (HAVE_DOT + CALL_GRAPH/CALLER_GRAPH/COLLABORATION_GRAPH/etc.) and SOURCE_BROWSER, not by the .qch/LFS path.

Might be worth a follow-up issue to trim those — e.g. disabling call/caller graphs (or capping DOT_GRAPH_MAX_NODES/MAX_DOT_GRAPH_DEPTH) and seeing what that does to du -sh html/ — separate from this PR.

ispyisail:
You are right, and I checked rather than taking it on trust — both workflows do deploy to the same `github-pages` environment:

| Workflow | Artifact path | Actions |
|---|---|---|
| `windows-msi.yml:354` | `gh-pages/` | `upload-pages-artifact@v5` + `deploy-pages@v4` |
| `auto-doxygen.yml` | `doc/html/` | `upload-pages-artifact@v3` + `deploy-pages@v4` |

And your account of why it has never bitten holds up: doxygen was paused before `windows-msi.yml` existed, so the two have never been live in the same period. The collision is latent rather than new.

**This was a gap in my review, not only in my patch.** I read `auto-doxygen.yml` closely enough to delete two steps from it and never asked what else in the repo calls `deploy-pages`. I was looking at the file rather than at the system, and you caught something I should have.

## One thing worth weighing before folding it into `windows-msi.yml`

The single-deploy-point principle is clearly right. The difficulty is that the two jobs run on very different cadences:

- `auto-doxygen.yml` → `on: push: tags` — at release
- `windows-msi.yml` → `workflow_run` after every successful **Windows Build on master** — continuously

So folding doxygen *generation* into that pipeline means building the documentation on every nightly — several minutes and, with graphviz present, close to the 985 MB you measured — to publish something that only meaningfully changes at a tag. And the alternative, having the MSI job publish a doxygen artifact produced elsewhere, needs that artifact stored somewhere in the meantime, which starts to resemble the storage problem we have just removed.

Your **third option** sidesteps both: push doxygen output straight to a `gh-pages` branch subdirectory instead of using Actions-based Pages. Each publisher then owns its own path, neither can replace the other, and the two keep their natural schedules. That is the one I would build if it is my call — but it is your infrastructure and you know the constraints better than I do.

## On the size difference

Agreed, and thank you for measuring it properly. My 159 MB is not the number that matters — as you say, `graphviz` was missing locally so all the graph generation was skipped. Your ~985 MB with `dot` present is the real figure, and it is driven by `HAVE_DOT` plus the call/caller/collaboration graphs and `SOURCE_BROWSER`, none of which this diff touches.

A follow-up trimming those sounds right to me. `CALL_GRAPH`/`CALLER_GRAPH` are usually the expensive pair, and `DOT_GRAPH_MAX_NODES` / `MAX_DOT_GRAPH_DEPTH` give a cheaper dial if you would rather keep some graphs. Happy to measure a few combinations against `du -sh doc/html` and report the numbers, if that would be useful groundwork for whoever takes it.

## This PR

Taking you at your word that the LFS part is good as-is, I will leave this diff alone rather than growing it — the Pages question deserves its own change with its own review, and mixing them would make both harder to reason about. Say the word if you would rather I close this and fold everything into one piece of work instead.

ispyisail:
Closing this — @arummler is right that I should have looked at #722 before rewriting the action, and @scorpio810 has since solved the underlying problem a different way.

**#722 already reworks this exact file, and better.** I checked after your comment: it moves `auto-doxygen.yml` to `workflow_call`, which gives exactly one caller of `deploy-pages` and so resolves the Pages collision @scorpio810 raised here; and it gates the Create Pull Request step on `if: startsWith(github.ref, 'refs/tags/')` rather than removing it. That covers what this PR was for, inside a rework that has thought about the whole pipeline instead of one file.

**And the problem has moved on.** @scorpio810's rsync to the server, and building locally rather than in Actions, means the LFS path this PR was trimming is not on the critical route any more.

I made the same mistake twice in this thread: reading the file on master without asking who else was already changing it. That is how I missed the `deploy-pages` collision, and it is how I ended up duplicating #722. Sorry for the noise — it cost you both review time you did not need to spend.

Nothing here is worth carrying over: **#722 supersedes it.**

For the record, since it may be useful to whoever picks up the size question: `Doxyfile` has exactly one commit in this repository (`a8e2a7acf`, added whole at 2838 lines), so there is no lighter historical version to return to — measuring forward is the only route. Numbers are in #824 if wanted.

PR #838, opened 2026-09-11
Do not install PugiXML and Googletest static libraries. 

arummler:
Do not install 3rd party libs which are only used as static libs. Basically the libs folder in the install order vanishes.

### DIFF
--- a/cmake/fetch_pugixml.cmake
+++ b/cmake/fetch_pugixml.cmake
@@ -26,7 +26,7 @@ if(BUILD_PUGIXML)
     pugixml
     GIT_REPOSITORY https://github.com/zeux/pugixml.git
     GIT_TAG        v1.15)
-
+  set(PUGIXML_INSTALL OFF CACHE INTERNAL "")
   FetchContent_MakeAvailable(pugixml)
 else()
   find_package(pugixml REQUIRED)
--- a/tests/googlemock/CMakeLists.txt
+++ b/tests/googlemock/CMakeLists.txt
@@ -65,7 +65,7 @@ FetchContent_Declare(
   GTest
   GIT_REPOSITORY https://github.com/google/googletest.git
   GIT_TAG        release-1.10.0)
-
+set(INSTALL_GTEST OFF CACHE INTERNAL "")
 FetchContent_MakeAvailable(GTest)
 
 include(../../cmake/fetch_kdeaddons.cmake)
--- a/tests/googletest/CMakeLists.txt
+++ b/tests/googletest/CMakeLists.txt
@@ -65,7 +65,7 @@ FetchContent_Declare(
   GTest
   GIT_REPOSITORY https://github.com/google/googletest.git
   GIT_TAG        v1.17.0)
-
+set(INSTALL_GTEST OFF CACHE INTERNAL "")
 FetchContent_MakeAvailable(GTest)
 
 include(../../cmake/fetch_kdeaddons.cmake)


ispyisail:
The GTest change is correct — `INSTALL_GTEST` is the option googletest exposes (`gtest-src/CMakeLists.txt:20`, guarding the install rules at `googletest/CMakeLists.txt:90`), and setting it `CACHE INTERNAL` *before* `FetchContent_MakeAvailable` is the right idiom in the right place. The problem is real on master: the build tree installs `libgtest.a`, `libgtest_main.a`, `libgmock.a` and `libgmock_main.a`.

Two things from testing it, one of which affects the description.

**PugiXML is still installed.** The title and body mention it and say "the libs folder in the install order vanishes", but the diff only touches GTest. `cmake --install` on a Qt6 build of current master gives:

```
/lib/libpugixml.a
/lib/pkgconfig/pugixml.pc
/lib/cmake/pugixml/pugixml-config.cmake
/lib/cmake/pugixml/pugixml-config-version.cmake
/lib/cmake/pugixml/pugixml-targets.cmake
/lib/cmake/pugixml/pugixml-targets-debug.cmake
/include/pugixml.hpp
/include/pugiconfig.hpp
```

`cmake/fetch_pugixml.cmake` calls `FetchContent_MakeAvailable(pugixml)` with nothing suppressing its install rules, so `lib/` does not end up empty. pugixml has no `INSTALL_PUGIXML` option, so it needs a different approach from the GTest one — `PUGIXML_NO_INSTALL`, or `EXCLUDE_FROM_ALL` on the fetch. Either extending the PR or adjusting the title would avoid someone merging this expecting `lib/` to be clean. `share/ECM/cmake/ECMConfig.cmake` also escapes, from the KDE Frameworks fetch.

**Heads-up, unrelated to this PR:** `cmake --install` currently fails on master before it gets that far:

```
CMake Error at _deps/kcoreaddons-build/cmake_install.cmake:46 (file):
  file INSTALL cannot find ".../ECMPoQm/af/kcoreaddons6_qt.qm": No such file or directory
```

It aborts there, so nothing later in the install runs — which makes this PR awkward to verify end to end until that is fixed.

Minor, pre-existing: `tests/googlemock/CMakeLists.txt` declares `GIT_TAG release-1.10.0` for the same FetchContent name `GTest` that `tests/googletest/CMakeLists.txt` declares as `v1.17.0`. `tests/CMakeLists.txt` adds `googletest` first, so the first declaration wins and the 1.10.0 tag is dead config — `git describe` in the fetched source confirms v1.17.0. Your second hunk is harmless belt-and-braces; I would keep it.

Tested on Qt 6.10.2 / KF 6.10.0, GCC 15.

ispyisail:
**Correction to my previous comment — the install failure I reported was my own error, not a problem on master.** Apologies for the noise.

I had built with `ninja qelectrotech` (a single target) and then run `cmake --install`, which expects what the *default* target produces. The 109 kcoreaddons `.qm` files are built by plain `ninja`, not by the `qelectrotech` target, so they were legitimately absent. With a full build: build exit 0, 109 `.qm` files present, **install exit 0, no errors**. Nothing to fix, and nothing blocking verification of this PR. Please disregard that paragraph.

The pugixml point stands, and re-running on a clean full install shows the situation is broader than I first described. Everything currently landing in `lib/`:

```
libpugixml.a                         pkgconfig/pugixml.pc
libKF6CoreAddons.a                   cmake/pugixml/          (4 files)
libKF6WidgetsAddons.a                cmake/KF6CoreAddons/    (5 files)
objects-Debug/KF6WidgetsAddons_resources_1/...qrc_kcharselect-data_init.cpp.o
x86_64-linux-gnu/libgtest.a          x86_64-linux-gnu/pkgconfig/gtest.pc
x86_64-linux-gnu/libgtest_main.a     x86_64-linux-gnu/pkgconfig/gtest_main.pc
x86_64-linux-gnu/libgmock.a          x86_64-linux-gnu/pkgconfig/gmock.pc
x86_64-linux-gnu/libgmock_main.a     x86_64-linux-gnu/pkgconfig/gmock_main.pc
                                     x86_64-linux-gnu/cmake/GTest/ (4 files)
```

plus `include/pugixml.hpp`, `include/pugiconfig.hpp`, `include/gmock/`, `include/gtest/` and `share/ECM/cmake/`.

So this PR correctly removes the GTest column — that part is real and verified. After it lands, `lib/` would still contain two KF6 static libraries, `libpugixml.a`, fourteen CMake package files, a stray object file and a pkg-config file.

That does not detract from the change; it just means the description promises more than one PR can deliver. If the goal is genuinely an empty `lib/`, the KDE Frameworks fetches need the same treatment as GTest and pugixml — `cmake/fetch_kdeaddons.cmake` already sets `BUILD_TESTING`, `BUILD_QCH` and friends inside `qet_make_kf_available()`, so that is probably the natural place for it.

Tested on Qt 6.10.2 / KF 6.10.0, GCC 15, full `ninja` then `cmake --install --prefix`.

ispyisail:
Here is a concrete suggestion for the pugixml half, built and verified rather than proposed from reading.

**Correcting myself first:** I said earlier that pugixml would need a different approach from GTest. That was wrong — I had not looked. pugixml v1.15 exposes `PUGIXML_INSTALL` (`CMakeLists.txt:45`, default `ON`), so it is exactly the same one-line idiom as your GTest change:

```diff
--- a/cmake/fetch_pugixml.cmake
+++ b/cmake/fetch_pugixml.cmake
@@ -25,6 +25,7 @@ if(BUILD_PUGIXML)
     GIT_REPOSITORY https://github.com/zeux/pugixml.git
     GIT_TAG        v1.15)
 
+  set(PUGIXML_INSTALL OFF CACHE INTERNAL "")
   FetchContent_MakeAvailable(pugixml)
 else()
   find_package(pugixml REQUIRED)
```

It belongs in `cmake/fetch_pugixml.cmake` rather than beside the GTest lines, and only inside the `if(BUILD_PUGIXML)` branch — when `BUILD_PUGIXML=NO` the system pugixml is found rather than built, and nothing is installed anyway.

**Verified**, master + your two hunks + the diff above, full `ninja` then `cmake --install --prefix`: build exit 0, install exit 0, and `lib/` goes from 28 files to 12. `libpugixml.a`, `pkgconfig/pugixml.pc`, the four `cmake/pugixml/` files, both `include/pugi*.hpp`, all four gtest/gmock static libs, their four `.pc` files and the four `cmake/GTest/` files are all gone.

**What remains after that, and why I am not suggesting a one-liner for it:**

```
lib/libKF6CoreAddons.a
lib/libKF6WidgetsAddons.a
lib/cmake/KF6CoreAddons/        (5 files)
lib/cmake/KF6WidgetsAddons/     (4 files)
lib/objects-Debug/KF6WidgetsAddons_resources_1/...qrc_kcharselect-data_init.cpp.o
```

kcoreaddons and kwidgetsaddons have no equivalent switch — they call `install()` unconditionally against the `KDE_INSTALL_*` paths. The obvious lever, `EXCLUDE_FROM_ALL` on `FetchContent_Declare`, needs CMake >= 3.28, and this project declares `cmake_minimum_required(VERSION 3.5...4.2)`, so adopting it would quietly raise the floor. That seems like a decision for a separate change rather than something to fold in here.

So with the diff above, this PR would do what its title says for the two dependencies that make it easy, and the KDE Frameworks leftovers become a known, separate item.

Tested on Qt 6.10.2 / KF 6.10.0, CMake 4.2.3, GCC 15.

arummler:
Thanks, when bisecting I lost the patch for PUGIXML. Just now re-added.

ispyisail:
That does it — the PR now matches its title.

The change I verified was identical apart from leading whitespace, which CMake ignores, so the result carries over: full `ninja` then `cmake --install --prefix` gives build exit 0, install exit 0, and `lib/` drops from 28 files to 12. `libpugixml.a`, `pugixml.pc`, the four `cmake/pugixml/` files, both `include/pugi*.hpp`, the four gtest/gmock static libs, their four `.pc` files and the four `cmake/GTest/` files all stop being installed.

Tiny cosmetic thing if you touch it again: the new line in `fetch_pugixml.cmake` sits at column 0 while `FetchContent_Declare` and `FetchContent_MakeAvailable` around it are indented two spaces inside the `if(BUILD_PUGIXML)` block. No functional difference at all.

The KDE Frameworks leftovers (`libKF6CoreAddons.a`, `libKF6WidgetsAddons.a`, nine CMake files and a stray object file) are the separate item I mentioned — not something this PR should carry.

PR #848, opened 2026-09-12
Guard debug output in titleblocktemplate

arummler:
Produces useless debug message by default.

### DIFF
--- a/sources/titleblocktemplate.cpp
+++ b/sources/titleblocktemplate.cpp
@@ -428,7 +428,9 @@ void TitleBlockTemplate::parseColumns(const QString &cols_string) {
 	rel_col_size_format.setPattern("^([rt])([0-9]+)%$");
 	rel_col_size_format.setPatternOptions(QRegularExpression::CaseInsensitiveOption);
 	bool conv_ok;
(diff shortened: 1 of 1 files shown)


ispyisail:
Good catch, and thank you — I had been filtering this line out of my own test output all day without registering it as something that should not be there.

Checked it rather than eyeballed it, and the fix is exactly right:

- `TITLEBLOCK_TEMPLATE_DEBUG` is the file's own convention, declared (commented out) at line 25 and already guarding the debug output at lines 100, 112, 209, 346 and others.
- Walking every `qDebug()` in `titleblocktemplate.cpp` while tracking `#ifdef` depth, **line 431 is the only unguarded one of the ten**. So this is not a partial fix; it brings the single stray into line with the rest.
- The message fires once per project load on current `master`, on stderr, for anything with a title block.

**Verified:** Qt 6.10.2 / KF 6.10.0, build clean (0 errors), `ctest` 4/4. The stray line goes from 1 per `--info` run to 0, and the `--info` JSON on stdout is unchanged — worth confirming rather than assuming, since something could in principle have been parsing the combined output.

No overlap with #846, which landed a few minutes ago in `qetproject.cpp`.

Merging.

PR #851, opened 2026-09-12
New feature: reuse automatic numbering rules from another project

### LINKED ISSUE #850: Users 's want to store formula rules for others projects
https://qelectrotech.org/forum/viewtopic.php?id=3186

ispyisail:
Closes #850.

If you have spent time setting up wire and element numbering in one project, you currently have no way to use it again in the next one. The only answer is to open both `.qet` files in a text editor and copy the XML across by hand.

This adds a button to the auto-numbering page of the project properties dialog: **"Import from another project…"**. Pick a `.qet`, tick the numberings you want, done.

- Lists every numbering found, grouped by conductor / element / folio
- Names that already exist here are marked and left unticked
- A "replace same-named numberings" option for when overwriting is what you want

The source file is read as plain XML, not opened as a project, so a file saved by an older QElectroTech does not raise the version dialog.

**Tested:** importing `industrial.qet` into `ArduinoLCD.qet` puts all five numberings in place with their formulas intact; importing a project into itself correctly marks all five as already present. Qt 6.10.2, build clean, ctest 4/4.

Not included: the project-template idea also raised on the forum thread. That needs a decision on where templates live and what else they carry.

### DIFF
--- a/sources/autoNum/ui/folioautonumbering.cpp
+++ b/sources/autoNum/ui/folioautonumbering.cpp
@@ -54,6 +54,10 @@ FolioAutonumberingW::~FolioAutonumberingW()
 */
 void FolioAutonumberingW::setContext(QList <QString> autonums)
 {
+		// Replace the list rather than append to it: this is called again
+		// whenever the project's numberings change (import, for instance),
+		// and appending gave a second copy of every name.
+	ui->m_autonums_cb->clear();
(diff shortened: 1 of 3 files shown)


ispyisail:
Follow-up: the project-template idea from the same forum thread is now #852. Kept separate from this PR so the design questions do not sit in a review thread.

PR #853, opened 2026-09-13
Fix macro drag-and-drop placement and preview for Qt6

Kellermorph:
Three bugs fixed in the macro drag-and-drop workflow:

Qt6 drop position (diagramview.cpp): QDropEvent::position() returns viewport coordinates, not scene coordinates. Wrapped with mapToScene() so dropped elements/macros land at the correct position instead of jumping to the upper-left corner.
Macro preview shows grid (diagrameventaddmacro.cpp): The preview pixmap was rendered from a dummy diagram that still drew its background grid. Added setDisplayGrid(false) before rendering so the preview is transparent and can be aligned with guide lines.
Macro labels (BMK) lost on placement (diagrameventaddmacro.cpp): PasteDiagramCommand::redo() erases element labels by default. Temporarily disables this setting during macro placement so designators, formulas and comments are preserved from the macro XML.
Both Qt5 and Qt6 paths are handled via #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0).


sources/diagramview.cpp
```diff
@@ -211,10 +211,10 @@ void DiagramView::handleElementDrop(QDropEvent *event)
 	}
 
 	QPointF drop_pos;
-	#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)	// ### Qt 6: remove
+	#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
 	drop_pos = mapToScene(event->pos());
```
scorpio810: Hi Thomas, please remove flag if QT_VERSION < QT_VERSION_CHECK(6, 0, 0) QET is now support only Qt6 framework.

Thanks,
Laurent

PR #856, opened 2026-09-13
Fix text position shift when duplicating diagram pages

Kellermorph:
When duplicating a page via the project tree context menu, center and right-aligned dynamic texts on PLC slave elements were shifted to the wrong position.

The root cause is that duplicateDiagram() calls setElementInformations() after fromXml() to clear stale PLC data from copied slave elements. This triggers elementInfoChanged() → setPlainText() → finishAlignment(), which recalculates the text position from a wrong starting point.

Element::fromXml() avoids this by wrapping its setElementInformations() call with m_block_alignment = true, preventing finishAlignment() from running during the text update. The fix applies the same pattern to the post-processing PLC cleanup in duplicateDiagram().

Changes:

sources/elementspanelwidget.cpp — duplicateDiagram(): use toXml(true, true) instead of manual correctTextPos/restoreText, wrap PLC cleanup's setElementInformations() with m_block_alignment
sources/qetgraphicsitem/element.h — add public clearPendingLinks() method to prevent copied elements from linking back to source


ispyisail:
The fix is right, but the description points at the wrong cause.

Master's `duplicateDiagram()` has no `setElementInformations()` call, no PLC cleanup and no `linkType()` check — that block is added by this PR, so it cannot be what breaks master.

The actual bug is `new_diagram->restoreText(elmt)`. Every `setPlainText()` in `restoreText()` is wrapped in `m_block_alignment` except the last one, so it runs `finishAlignment()` on elements whose positions came straight from the XML. Removing that call and using `toXml(true, true)` fixes it. Worth correcting the text so the next reader is not hunting code that is not there.

Two things beyond the title, worth stating explicitly since they change what duplicating a page produces:

- `clearPendingLinks()` also drops links to masters on folios that were *not* duplicated, which would have resolved correctly.
- `ELMT_LABEL` and `xref` are stripped from every slave unconditionally.

Both may be what you want (it matches paste), but neither is "text position".

Also: the comments explaining why `newUuid()` is needed — uuid being the database PRIMARY KEY, rows vanishing from nomenclature — are deleted with it. Please keep them.

Tested: Qt5 and Qt6 build clean. Duplicating a 51-text folio gives byte-identical text positions on master and on this branch, so no regression. I could not reproduce the original symptom: it needs a text that is both non-left-aligned and `ElementInfo`/`CompositeText`, and there are zero of those across all 23 example projects. If you have a project that shows it, that would be worth attaching.

Kellermorph:
Thanks for the thorough review. You're right on all counts — the original description was misleading. Here's what changed:

Description corrected: The actual bug was restoreText() being called on every newly loaded element in duplicateDiagram(). Each setPlainText() inside restoreText() is wrapped in m_block_alignment except the last one, so finishAlignment() ran on elements whose positions came straight from the XML — shifting center and right-aligned texts. Using toXml(true, true) and dropping the manual restoreText() call fixes it.

clearPendingLinks() and PLC cleanup: Both are intentional and match the paste behaviour (PasteDiagramCommand). You're right that neither is "text position" — they address the other symptoms of page duplication (stale master links, leftover PLC data). I've called this out explicitly in the description now.

Comments: Restored the newUuid() comments explaining the PRIMARY KEY reasoning.

scorpio810:
Thanks to both of you — the corrected description and the restored `newUuid()` comments make this much easier to follow.

The `restoreText()` / `toXml(true, true)` part is a clear bug fix and I have no objection to it.

The two behaviour changes are a different matter, and I would rather they did not ride along with a fix titled "text position".

`clearPendingLinks()` also drops links to masters on folios that were not duplicated — links that would have resolved correctly. And `ELMT_LABEL` and `xref` are stripped from every slave unconditionally, so a user duplicating a page loses labels that were there before.

The justification given is that it matches `PasteDiagramCommand`, but duplicating a page inside the same project is not the same intent as pasting. When I paste, I know I am moving to a new context. When I duplicate a folio, I expect cross-references to masters on the other folios of the same project to survive. I am not saying the current behaviour is right — stale PLC data on a duplicated slave is a real problem — but this is a product decision about what duplication should produce, and it deserves its own discussion rather than being settled by symmetry with paste.

Could you split this into two PRs? The alignment fix can go in quickly. The link and label handling I would like to think about, and hear from users on.

One more thing: @ispyisail could not reproduce the original symptom — it needs a text that is both non-left-aligned and `ElementInfo`/`CompositeText`, and there are none in the 23 example projects. Could you attach a minimal .qet that shows the shift? Without one we are merging a fix for a bug nobody has seen, and there is nothing to test against if it regresses later.

scorpio810:
There is an existing preference for this: the "do not keep element labels when copying and pasting" checkbox. On my installation it is unchecked, meaning I keep my labels.

Does this PR read that setting? From the description, ELMT_LABEL is stripped from every slave unconditionally, which would override the user's choice — and if the justification for the behaviour is that it matches PasteDiagramCommand, then it should match paste on this point too: paste honours the preference, so duplication should as well.

Same question for xref.

Kellermorph:
Thanks for the feedback — you're right on all counts. I've split this down to the pure alignment fix:

toXml(true, true) handles correctTextPos/restoreText internally for Slave and Report elements only
restoreText() is now called on the target for those same element types to recalculate positions for the resolved text
No data is cleared, no links are touched, no behaviour changes
The clearPendingLinks() and PLC slave data cleanup are intentionally left out. I'll open a separate PR for those once this is merged — happy to discuss the product decisions around label handling and link preservation there. I'll also attach a minimal .qet test file with that PR.

Kellermorph:
an example qet-file
[2026-09-14-18-02_ASP01.xml](https://github.com/user-attachments/files/32205366/2026-09-14-18-02_ASP01.xml)

ispyisail:
Qt6 build clean, ctest 6/6. Corpus still can't reach it — 0 of 7 executable lines, it's GUI-only.

The split addressed both objections; this is the alignment fix alone now.

The new guard is character-for-character `diagram.cpp:1020` and `:1098`, and flipping `is_copy_command` to true has exactly those two effects and nothing else — that flag appears nowhere else in `toXml()`.

The old bug is visible in `restoreText()`: its CompositeText branch consumes state that `correctTextPos()` stashed in `text()`, and master called it on *every* duplicated element, including ones never corrected.

The attached ASP01 carries 88 non-left-aligned ElementInfo/CompositeText texts — the precondition the 23 example projects lack.

**Blocking:** none.
**Non-blocking:** merge master (33 behind).

ispyisail:
Ran it this time instead of only reading it, using the project you attached.

Two binaries built separately (master `7fa720a1`, this PR merged onto master `1ee6142c`), ASP01 opened under Xvfb in each, folio 2 duplicated from the panel context menu, saved, then compared with `qet-folio-diff.py` — its `--self-test` passes first, so the comparison is known to be able to fail.

- master: **1 difference** — one text at `-40.44` vs `-40.4399`
- this PR: **identical**

So the PR does make duplication exactly geometry-preserving, which is a real improvement. But 0.0001 units is not the visible shift the title describes: folio 2 of your own sample doesn't reproduce the symptom. Which folio, and which steps, show it? I got no clean run on folio 1, so that one is still untested.

**Blocking:** none.
**Non-blocking:** merge master (33 behind); a repro recipe for the visible shift, so there is something to test against if it regresses.

Kellermorph:
When I copy page 5 from the example using the current Flatpak version, the elements shift as shown in the images. This doesn't happen with the new version.
orginial Page 5
Copy Flatpak

PR #858, opened 2026-09-13
Add Hide linked elements checkbox to PLC link widget

Kellermorph:
Add a checkbox Masquer les éléments connectés to the top of PlcLinkWidget that allows hiding IO entries already linked to a slave. When checked:

Linked IO entries are hidden instead of shown as struck-through/greyed out
Masters whose entire IO table is fully linked are hidden completely (including the master label)
The checkbox state is persisted via QSettings (plclinkwidget/hideLinked) so it survives across sessions.


ispyisail:
Builds clean on Qt5 and Qt6, ctest 4/4, current with master. One bug.

**The search box undoes the checkbox.** `buildPlcTree()` now sets `setHidden()`, but so does `on_m_search_field_textEdited()`, and nothing reconciles them:

```cpp
child->setHidden(!match);   // match == text.isEmpty() -> true once cleared
parent->setHidden(!any_child_visible);
```

Tick the box, type anything in the search, then clear it: `match` is true for every child, so every linked entry comes back and the fully-linked masters with them. The checkbox still reads as ticked. Only toggling it again restores the hiding.

Two narrower cases: searching a master's label explicitly un-hides all its children (`parent->child(j)->setHidden(false)`), and any linked entry matching the search text reappears.

Worth routing both through one visibility decision — roughly `match && !(hide_linked && is_linked)` — rather than two functions each writing `setHidden` from their own view.

Two things I checked that are fine: `setChecked()` in the constructor fires `toggled` before `setElement()`, but `buildPlcTree()`'s early return covers the null element; and the vacuous `all_children_hidden` case cannot happen because empty IO tables are already skipped.

I could not run any of this. `ElementData::toXml` writes `<plcMasterData>` for a PLC master and there are zero across all 24 example projects, so the PLC link widget is unreachable with shipped data — the bug above is from reading the two functions, not from using the dialog. If you have a project with a PLC master, attaching it would help anyone reviewing this area.

---

Verdict, added after the fact for consistency with later reviews:

**Blocking:** the search/checkbox interaction — fixed by @Kellermorph before merge, routed through `isChildLinked()` and a shared `m_linked_children`, which is the single visibility decision this asked for.

**Non-blocking:** no project in the corpus contains a PLC master, so this widget stays unreachable with shipped data and the findings above were never run.

Kellermorph:
The search field now respects the hide-linked state — both visibility decisions are routed through isChildLinked() using a shared QSet<QTreeWidgetItem*> m_linked_children, so clearing the search no longer restores hidden linked entries.

Thanks for catching the search/checkbox interaction bug.

scorpio810:
@Kellermorph @ispyisail It'sready to merge?

Kellermorph:
in my opinion, yes

PR #861, opened 2026-09-14
Partial fix for bugtracker #248: second-instance file arguments are lost

ispyisail:
> **Partial fix.** @scorpio810 built this branch and the reported symptom persists: with QET running, clicking a project in a file manager still does nothing. This fixes one real defect in that path and does not close bugtracker #248. What is still outstanding is in the comments below.

Bugtracker #248 reports that opening a project while QET is already running does nothing. It is reported for filenames containing spaces, which is how it gets noticed, but plain filenames fail identically — and the cause below is only part of why.

`QET::splitWithSpaces()` splits on `QRegularExpression("[^\\]?(?:\\\\)* ")`. That pattern is invalid — `[^\\]` opens a character class whose `\\]` is an escaped bracket, so the class never closes. Qt reports `isValid() == false`, `QString::split()` warns *"invalid QRegularExpression object"*, and the function returns an **empty list for every input**.

It is the receiving half of the SingleApplication handshake: a second instance sends `"launched-with-args: " + joinWithSpaces(args)`, and the running instance parses it in `QETApp::receiveMessage()` before calling `openFiles()`. An empty split means no arguments arrive.

**Why not just fix the regex.** The separator is a space preceded by an even-length run of backslashes. PCRE2 has no variable-length lookbehind, so that run cannot go in a lookbehind, and matching it by consumption eats the character before the space — which is what the `[^\\]?` was doing. An explicit scan is correct and easier to read.

**Test.** `tst_qetstrings` asserts `splitWithSpaces(joinWithSpaces(x)) == x` for plain names, embedded spaces, embedded backslashes, a trailing backslash and a mixture, plus the specific regression that a plain argument list does not come back empty.

I checked the test actually fails without the fix: 9 of 11 cases fail on the old implementation, 11/11 pass with it. Full suite 5/5 on **Qt 6.10.0** — the version master defaults to — and 5/5 on Qt 5.15.18. The five includes the new `tst_qetstrings`.

**What this changes, measured.** Starting a primary instance and launching a second one with a project path as its argument, then checking whether the primary loads it: master opens it 0 times, this branch 1 time. So the message now arrives. The window still does not come to the front, and on @scorpio810's machine the symptom is unchanged — see the discussion below for how to tell those two apart.


scorpio810:
https://qelectrotech.org/forum/viewtopic.php?pid=23301#p23301

scorpio810:
I can reproduce it here: start QET, then click a .qet project in the file manager — nothing happens. A space in the filename is not needed, a plain name behaves exactly the same.

This looks like a regression introduced when we moved to Qt5 and replaced QtSingleApplication with the vendored SingleApplication.

https://qelectrotech.org/forum/viewtopic.php?pid=9044#p9044

scorpio810:
I built your branch and the problem is still there: with QET already running, clicking a .qet project in the file manager does nothing. Same result with a plain filename and with one containing a space.

So the invalid regex in splitWithSpaces() was real, but it does not seem to be the whole story.

ispyisail:
I looked into the raise and it turns out not to be the answer — but the investigation turned up something more useful for narrowing this down.

**First, what this branch does fix.** Starting a primary instance and launching a second one with a project path as its argument, then checking whether the primary loads it:

| | primary opens the file |
|---|---|
| master | **0 times** |
| this branch | **1 time** |

**The raise idea doesn't hold.** I implemented `raise()` + `activateWindow()` in `QETApp::openProjectFiles()` — matching what `openElementFiles()` already does — and tested it under a real window manager with a competing window in front:

| | active window afterwards |
|---|---|
| with the raise | competing window (unchanged) |
| without it | competing window (unchanged) |

No difference. X11 focus-stealing prevention denies the activation, because the application has no recent user-interaction timestamp. Making it work needs a proper `_NET_ACTIVE_WINDOW` message with a valid timestamp, which is a real piece of work rather than a two-line addition. So I've left it out of this PR rather than add something I can't show works.

**The part that might help you.** `BackupDialog` is unconditional in `openAndAddProject()` — there's no setting that suppresses it. So when the message does arrive, the user gets a modal *"Créer une copie de sauvegarde ?"* dialog, which is hard to miss.

That gives a clean split for what you're seeing on your build:

- **No dialog at all** → the message isn't being delivered, and the remaining bug is upstream of this change. That would point at SingleApplication itself, matching your QtSingleApplication theory.
- **Dialog appears** → the IPC works and what's left is purely that the window doesn't come to the front.

Which of those is it on your machine?

One caveat on my numbers: they're from a headless Linux test (`QT_QPA_PLATFORM=offscreen`) with a second instance launched from a shell, not from a file manager in a real desktop session. That difference could matter and I can't rule it out from here.

scorpio810:
Built your branch. Good news first: it works. The project opens (`Project "industrial.qet" (2557 KiB) opened in 4.312 seconds`) and the BackupDialog does appear, so the message is delivered and openFiles() gets its argument. Your fix is correct.

Bad news: right after the backup XML export, the primary instance segfaults. Backtrace, Qt 6.10.2, Debian sid, Wayland:

```
#0  0x00007ffff561bbf7 in ?? () from libQt6Core.so.6
#1  QIODevice::channelReadyRead(int) () from libQt6Core.so.6
#2  QAbstractSocketPrivate::canReadNotification() () from libQt6Network.so.6
#3  ?? () from libQt6Network.so.6
#4  QApplicationPrivate::notify_helper(QObject*, QEvent*)
#5  QCoreApplication::notifyInternal2(QObject*, QEvent*)
...
#12 QCoreApplication::exec()
#13 main ()
```

Note there is no QET frame above the crash: our slot has already returned and QMetaObject::activate dies while continuing to emit. That points at the sender being destroyed during the slot.

My reading: receiveMessage() runs openFiles() straight from the readyRead path. Opening the project takes 4 seconds and BackupDialog is modal, so a nested event loop runs inside the socket's signal emission. During that time the secondary instance has already exited, the connection closes, and SingleApplication destroys the QLocalSocket. When the nested loop unwinds, activate() touches freed memory.

That would explain why you don't see it headless: no modal dialog, no nested event loop, no window for the socket to die in.

Two things worth looking at:

1. Defer the work out of the readyRead path — QTimer::singleShot(0, ...) or a queued connection in receiveMessage() — so the socket stack is fully unwound before the project is opened. This looks like the real fix and it stays in QET code.
2. Check the vendored SingleApplication for `delete socket` where it should be `socket->deleteLater()`.

This is not a regression from your PR: the path was simply unreachable before, since the argument list always came back empty. Your fix should go in, and this deserves its own issue. I can open one with the full backtrace if you prefer to keep the PR focused.

scorpio810:
Tested on master (2785e255) with 
```
wget https://github.com/qelectrotech/qelectrotech-source-mirror/pull/861.patch
git apply 861.patch
```
, Qt 6.10.2, Debian sid, Wayland session, both instances started from a shell rather than the file manager:

```
terminal 1: gdb --args ./qelectrotech

terminal 2: ./qelectrotech ../examples/industrial.qet
```

The second instance hands over and exits immediately, and at that exact moment the primary logs:

ispyisail:
Thank you — that backtrace is worth more than the fix itself, and yes please, a separate issue would be ideal. I'd rather not widen this PR into a crash it did not cause.

**I could not reproduce it.** Qt 5.15.18, Xvfb/X11, `industrial.qet`. The path is genuinely exercised — the project opens, the backup XML export runs, and the EventLoopWatchdog records main-thread stalls of 663 ms and 463 ms, which is the nested event loop you describe. The primary survives. So it looks specific to Qt 6.10.2 or to Wayland, and I cannot verify a fix here; I could only show it does not break anything.

**Your diagnosis holds up against the source regardless.** I checked each step:

- `SingleApplicationPrivate::slotDataAvailable()` emits `receivedMessage` synchronously from inside the `readyRead` lambda (`singleapplication_p.cpp:452`), so `openFiles()` runs with the socket handler still on the stack.
- `BackupDialog` is unconditional in `openAndAddProject()`, so a modal nested loop is guaranteed, not incidental.
- The stall measurements above confirm the loop actually nests.
- No QET frame above the crash matches a slot that returned while `activate()` carried on emitting.

**One correction on your second suggestion.** The vendored SingleApplication already uses `deleteLater()` — `singleapplication_p.cpp:331`, on the `disconnected` signal. So there is no `delete` to convert. That does not weaken the diagnosis; it narrows it, because it means the deferred delete is not enough on its own once a nested loop is in play. Which leaves your first suggestion as the actual fix.

Deferring out of the `readyRead` path is also just correct independently of the crash: four seconds of project loading and a modal dialog do not belong inside a socket read handler.

I'd rather not write that fix blind, since I cannot demonstrate it works on the platform where the crash happens. If you open the issue with the backtrace I'm happy to prepare the change for you to test, or to review yours.

ispyisail:
Correction to what I said above: I assumed the difference was Qt 5 versus Qt 6. It isn't.

I rebuilt the branch against **Qt 6.10.2** — the same version you are on — and ran the same reproduction. It does not crash. The path is genuinely exercised: the project opens and the EventLoopWatchdog records nine main-thread stalls, so the nested event loop really does nest.

That rules out the Qt version, which was my leading theory. The remaining difference between us is the platform:

| | you | me |
|---|---|---|
| Qt | 6.10.2 | 6.10.2 |
| session | Wayland | Xvfb / X11 (xcb) |
| result | segfault | survives |

**Could you try the same build with `QT_QPA_PLATFORM=xcb`?** One command, and it splits the remaining possibilities cleanly:

- **No crash under xcb** → it is Wayland-specific, and the fix is much easier to reason about. A modal `QDialog::exec()` nests differently there — the xdg-shell mapping and compositor round-trip mean the nested loop processes a different set of events than under xcb, which would be enough to get the socket's deferred delete run inside the nested loop rather than after it.
- **Still crashes under xcb** → something else in your environment is the variable and my reproduction is simply not representative, in which case I would stop guessing and work from your backtrace alone.

Either answer is useful. None of this changes the diagnosis in your earlier comment, which I verified against the source and still think is right — only where the fix has to be careful.

ispyisail:
Reproduced your crash and opened #868 with a fix.

The step I was missing is dismissing the backup dialog. Twice I left it open, the stack never unwound, and nothing failed — which is why I reported it as unreproducible. With the dialog dismissed it segfaults reliably, right after the backup XML export, exactly as you described.

It is also neither Wayland-specific nor a Qt 6 regression: I hit it on **Qt 6.10.2 under X11/xcb**, and the same test survives under a headless Wayland compositor and under Qt 5.15.18 whenever the dialog is left open. The deciding factor is the unwind, not the platform.

Your first suggestion was the right one. #868 defers `openFiles()` out of the `readyRead` path with a zero-timer: segfault twice out of two before, survives twice out of two after, project still opens, ctest green.

Your second suggestion is already satisfied — `singleapplication_p.cpp:331` uses `deleteLater()`. That is what makes the nested loop the whole story rather than a missing `deleteLater`.

#868 needs this PR to be reachable at all, so the order is this one first.

PR #862, opened 2026-09-14
Fix bugtracker #108: the junction dot vanishes on a wide conductor

ispyisail:
`Conductor::paint()` drew every junction as a fixed 3.0-unit ellipse regardless of the conductor carrying it. The width is user-settable from 0.4 to 20.0, so above about 3.0 the dot is narrower than the line and disappears — exactly when a junction most needs to be legible.

The dot now scales with `m_properties.cond_size`, **floored at the historic 3.0 so nothing changes at or below the default width of 1.0**. Only the wide conductors the report is about are affected.

`cond_size` rather than the pen width, because the pen is inflated by 4 while the mouse is over the conductor and the junction should not grow on hover.

**Measured** with a temporary trace over `examples/741.qet`:

| conductor width | dot diameter |
|---|---|
| 1.0 (default) | 3.00 — unchanged |
| 5.0 | 15.00 |

A PNG export of the widened project shows two junctions that were invisible under the line rendering as clear dots.

ctest 4/4 on **Qt 6.10.0** — the version master defaults to — and 4/4 on Qt 5.15.18.


scorpio810:
Please add some captures

ispyisail:
Here they are. `examples/741.qet` with `condsize="5"`, exported to PNG with the same build before and after the change, zoomed 4x on a pair of junctions.

junction dot before and after

Left, the junction dot is the fixed 3.0 units: narrower than the conductor, so it does not read as a junction at all. Right, it scales with the conductor width.

At the default width of 1.0 the diameter stays 3.00 and nothing changes — the floor in the patch makes that explicit, so existing drawings are untouched.

Full folios if you want them in context: [before](https://raw.githubusercontent.com/ispyisail/qelectrotech-source-mirror/pr-assets/assets/bug108-full-orig.png) · [after](https://raw.githubusercontent.com/ispyisail/qelectrotech-source-mirror/pr-assets/assets/bug108-full-fixed.png)

scorpio810:
Thanks for the captures — that shows it clearly.

The approach is sound. The floor at 3.0 means existing drawings are untouched, and using `cond_size` rather than the pen width is the right call: the pen is inflated by 4 on hover and the junction should not grow under the mouse.

Two things before I merge.

The factor of 3 at the top of the range. At `cond_size=5` the dot is 15 units, and at the maximum of 20 it would be 60 — a large blob. Drawing convention usually puts a junction dot at around two to three times the line thickness, so 5 looks fine, but could you post a capture at `cond_size=20` so we can see what the extreme actually looks like? If it is too heavy, capping the diameter or dropping the factor would be easy enough.

Does `Conductor::boundingRect()` (and `shape()`) account for the larger dot? If the bounding rect still assumes 3.0, wide junctions will leave repaint artefacts when the conductor is moved or the view scrolled.

On the description: the `#108` reference renders as a link to PR #108 in this repository. A full URL to the bugtracker entry would be clearer for anyone reading this in two years.

The `pr-assets` branch for the images is a good idea — keeps them out of the main history.

PR #864, opened 2026-09-14
Fix bugtracker #238: summary table ordered by its columns, not by folio

ispyisail:
`SummaryQueryWidget::queryStr()` built its `ORDER BY` from the columns the user chose to display, in the order they chose them:

```cpp
column   += key;
order_by += key;
```

So a summary whose first column is Title came out sorted alphabetically by title, and one starting with Author sorted by author. A table of contents lists the folios of a project; its order is the project's order, not whatever the first column happens to be.

It now orders by `pos`, the folio position `project_summary_view` already exposes from `diagram.pos`. That column is an `INTEGER`, so the sort is numeric and folio 10 does not land between folio 1 and folio 2. One row per folio means `pos` fully determines the order, so no secondary key is needed.

**Demonstrated** against a stand-in view holding four folios:

| query | result |
|---|---|
| `ORDER BY title, pos` (old) | Apple(2) Banana(3) Mango(10) Zebra(1) |
| `ORDER BY pos` (new) | Zebra(1) Apple(2) Banana(3) Mango(10) |

The hand-written query path (`m_edit_sql_query_cb`) returns before this and is untouched, so anyone wanting a different order still has one.

ctest 4/4 on **Qt 6.10.0** — the version master defaults to — and 4/4 on Qt 5.15.18.


scorpio810:
The firts column need to be pos or user add order by pos

`SELECT folio, title, author, date FROM project_summary_view ORDER BY pos, title, author, date`

https://qelectrotech.org/forum/viewtopic.php?id=2619
https://qelectrotech.org/forum/viewtopic.php?id=3033

Others workarounds and howto:
https://qelectrotech.org/forum/viewtopic.php?id=2656
https://qelectrotech.org/forum/viewtopic.php?pid=20541#p20541
https://qelectrotech.org/forum/viewtopic.php?id=2847
https://qelectrotech.org/forum/viewtopic.php?pid=21231#p21231

PR #868, opened 2026-09-14
Fix a use-after-free: forwarded files are opened inside the socket handler

ispyisail:
Fixes the segfault @scorpio810 found while testing #861 — his backtrace and diagnosis are there: https://github.com/qelectrotech/qelectrotech-source-mirror/pull/861#issuecomment-5660456859

`QETApp::receiveMessage()` called `openFiles()` directly. That slot runs inside SingleApplication's socket handling — `SingleApplicationPrivate::slotDataAvailable()` emits `receivedMessage` synchronously from the `readyRead` lambda (`singleapplication_p.cpp:452`). `openFiles()` then loads a project, which on a large one takes seconds, and `openAndAddProject()` puts up a modal `BackupDialog` whose `exec()` runs a nested event loop **while the socket handler is still on the stack**.

During that nested loop the secondary instance exits, the connection closes and the `QLocalSocket` is deleted. When the dialog is dismissed and the stack unwinds, `QMetaObject::activate()` carries on emitting on the freed sender.

A zero-timer returns to the event loop first, so the socket stack is fully unwound before the project is opened. This is @scorpio810's first suggestion; his second — looking for `delete socket` that should be `deleteLater()` — turned out to be already satisfied at `singleapplication_p.cpp:331`, which is why the deferred delete alone is not enough here.

**Reproduction.** The step that matters is dismissing the dialog. I twice failed to reproduce this by leaving it open — the stack never unwinds, so it never fails.

<details><summary>Script</summary>

```bash
QET=./build/qelectrotech
PROJ=examples/industrial.qet          # large: the load must be slow

Xvfb :80 -screen 0 1280x900x24 & sleep 2
export DISPLAY=:80 QT_QPA_PLATFORM=xcb
openbox & sleep 2

$QET & PRIM=$!                        # primary
sleep 16
$QET "$PROJ"                          # secondary forwards its argument, exits
sleep 10

W=$(xdotool search --onlyvisible --name "sauvegarde" | head -1)
xdotool windowactivate --sync "$W"; xdotool key --clearmodifiers Return
sleep 10
kill -0 $PRIM 2>/dev/null && echo alive || echo died
```

</details>

| | crash on unwind | project opens | ctest |
|---|---|---|---|
| before | SIGSEGV, twice | yes | 5/5 |
| after | survives, twice | yes | 5/5 |

Qt 6.10.2, X11/xcb. Worth noting it is **not** Wayland-specific and **not** a Qt 6 regression: I ran the same test under a headless Wayland compositor and under Qt 5.15.18, and the deciding factor in every case was whether the dialog was dismissed.

**This needs #861 to be reachable.** Without it `splitWithSpaces()` returns an empty list, no project opens, and nothing enters this path — which is why the crash was never seen before.


scorpio810:
Reviewed the diff. The two things I wanted to check are both right: `singleShot(0, this, ...)` uses a context object, so the lambda is cancelled rather than fired on a dead object, and `deferred_args` is a copy made before the lambda, properly detached from the local `args_list`. No new dangling reference introduced.

One cosmetic nit: the comment block is indented one level deeper than the surrounding code.

The comment itself is long but I would keep it. Without it, someone will eventually "simplify" the zero-timer away and bring the crash back.

The reproduction script is the part I'd like to see go further. It runs unattended — Xvfb, openbox, xdotool to dismiss the dialog — so it could become a CI regression test. Neither #861 nor this one has automated coverage of the IPC path end to end, and this is exactly the kind of bug that comes back silently. Worth its own PR if you're up for it.

Merging this.

ispyisail:
Built the regression test. It crashes an unfixed master 3 times out of 3 (exit 139), and survives with this PR 3 times out of 3.

Three things I did not expect, each of which first gave me a passing test against a build that crashes reliably:

**Qt 5 never reproduces it.** An unfixed Qt 5 build survives every attempt. Only Qt 6 — I used 6.10.2, same as you. The test now refuses to run on a Qt 5 binary rather than report a pass that proves nothing.

**Opening a project raises "Merci de patienter" about six seconds before the backup prompt, and both stay up.** Answering the first one looks exactly like a successful test run. That was my first false pass.

**It is a race.** Single attempts caught it 2 times in 3 — the second instance's disconnect has to land inside the nested loop. It now runs three attempts and fails on any crash.

I tested this PR cherry-picked onto current master, since the branch itself predates #861; that combination is what the numbers above refer to.

The test lives in my Docker harness rather than upstream — upstream `tests/` is unit tests with no Xvfb, openbox or xdotool, and I did not want to push those CI dependencies into the project for one test. Happy to propose it upstream if you would rather have it there.

PR #869, opened 2026-09-14
Clear stale xref data from PLC slave elementInformations on unlink and position change

Kellermorph:
When a PLC slave is unlinked from its master, or when the cross-reference position is changed from "Text field" to another position in Project Settings, the elementInformations key "xref" was never cleared. This caused the cross-reference text to remain visible in the slave's dynamic text field even after unlinking or position change.

Root cause:

The "xref" key is written into the slave's elementInformations by DynamicElementTextItem::updateXref() / ElementTextItemGroup::updateXref() when the xref position is set to "Text field" (Qt::AlignHCenter), but was never removed on unlink or position change.
Additionally, XRefPropertiesChanged was only connected to updateXref() when m_slave_Xref_item already existed. If the initial position was "Text field", no m_slave_Xref_item was created, so the label text was never notified of position changes.
Changes:

masterelement.cpp: Remove "xref" from slave's elementInformations during unlink
linkelementcommand.cpp: Remove "xref" in all unlink code paths (undo slave, undo master, makeLink cleanup)
dynamicelementtextitem.cpp: Remove stale "xref" from elementInformations in updateXref() cleanup; connect XRefPropertiesChanged for slave elements in itemChange() so all text fields are notified of position changes
elementtextitemgroup.cpp: Remove stale "xref" in updateXref() cleanup; connect XRefPropertiesChanged in constructor


Kellermorph:
When a PLC slave is unlinked from its master, or when the cross-reference position is changed from "Text field" to another position in Project Settings, the elementInformations key "xref" was never cleared. This caused the cross-reference text to remain visible in the slave's dynamic text field even after unlinking or position change.

Root cause:

The "xref" key is written into the slave's elementInformations by DynamicElementTextItem::updateXref() / ElementTextItemGroup::updateXref() when the xref position is set to "Text field" (Qt::AlignHCenter), but was never removed on unlink or position change.
Additionally, XRefPropertiesChanged was only connected to updateXref() when m_slave_Xref_item already existed. If the initial position was "Text field", no m_slave_Xref_item was created, so the label text was never notified of position changes.
Changes:

masterelement.cpp: Remove "xref" from slave's elementInformations during unlink
linkelementcommand.cpp: Remove "xref" in all unlink code paths (undo slave, undo master, makeLink cleanup)
dynamicelementtextitem.cpp: Remove stale "xref" from elementInformations in updateXref() cleanup; connect XRefPropertiesChanged for slave elements in itemChange() so all text fields are notified of position changes
elementtextitemgroup.cpp: Remove stale "xref" in updateXref() cleanup; connect XRefPropertiesChanged in constructor

scorpio810:
Thanks — this looks like the right place to fix it: clearing the stale key at the source rather than stripping it at duplication time.

Please move the explanation you posted in the comment into the commit message. It is a good description, and a commit body reading only the title loses all of it for anyone reading the history later.

Two questions before I look closely.

This touches the undo paths in linkelementcommand.cpp. A bad undo means silent project corruption, so what did you test there? Link, unlink, undo, redo, save, reopen — and the same with the xref position changed in between.

The missing XRefPropertiesChanged connection when the initial position is "Text field" is a second, separate bug, and probably one users have hit without being able to name it. Worth putting in the title or splitting out.

And the point that matters for #856: if this clears stale xref at the source, is the unconditional xref stripping in that PR still needed?

> There is an existing preference for this: the "do not keep element labels when copying and pasting" checkbox. On my installation it is unchecked, meaning I keep my labels.
> 
> Does this PR read that setting? From the description, ELMT_LABEL is stripped from every slave unconditionally, which would override the user's choice — and if the justification for the behaviour is that it matches PasteDiagramCommand, then it should match paste on this point too: paste honours the preference, so duplication should as well.
> 
> Same question for xref.

Kellermorph:
Good catch on the undo concern — you were right to flag it. There was indeed a bug: on_m_unlink_pb_clicked() never called setGroupIndex(), so m_group_index stayed at -1. When undo called makeLink(m_linked_before), the slave was re-linked, but makeLink() only populates PLC variables when group_idx >= 0. Result: the xref text appeared (because linkedElementChanged → updateXref fires), but PLC_TYPE, PLC_ADDRESS, PLC_FUNCTION etc. were never restored — the slave looked unlinked even though linkedElements() was non-empty.

Additionally, m_update_slave_Xref_connection was only cleared inside the if(m_slave_Xref_item) block in the updateXref() cleanup path. For AlignHCenter (Text field) position, no m_slave_Xref_item is ever created, so the old connections were never cleaned up on unlink. On re-link, the stale entries prevented new connections from being established.

Both fixed in the latest commit. Tested: link slave to PLC master → unlink → undo → all PLC fields and cross-reference restored → redo → fields cleared → undo again → fields restored. Also tested changing xref position between Text field and Top/Bottom after unlink/re-link cycles.

You're right that the XRefPropertiesChanged connection fix is a second, separate bug. It affects a different scenario: when the initial xref position is "Text field" (AlignHCenter), the slave's dynamic text items never connect to XRefPropertiesChanged. So if the user later changes the position to e.g. Top or Bottom in Project Settings, the label text is never notified and the xref doesn't appear at the new position.

I'm keeping both fixes in this PR since they both address PLC slave cross-reference handling and are closely related. But I've made this explicit in the title and description.

PR #856 strips xref unconditionally from every slave during page duplication. Our fix clears xref at the source — on unlink (masterelement.cpp, linkelementcommand.cpp) and when the xref position changes (dynamicelementtextitem.cpp, elementtextitemgroup.cpp).

These address different code paths: #856 is about duplicateDiagram() copying stale xref data from a source page, while #869 is about the runtime unlink/position-change flow. So our fix does not make the stripping in #856 redundant — both are still needed. However, I agree with your point that the unconditional stripping should respect the user's "keep labels when copying" preference, and that the link/label handling deserves its own discussion.

scorpio810:
https://qelectrotech.org/forum/viewtopic.php?pid=23332#p23332

related to this post : https://qelectrotech.org/forum/viewtopic.php?pid=23331#p23331

Kellermorph:
Enjoy your vacation and take some time to relax.

ispyisail:
Merged onto master, Qt6 build clean, ctest 6/6. Branch is 28 commits behind — needs a merge. Corpus reaches 23 of 39 changed lines (59%); the unlink paths aren't among them.

Diagnosis checks out. Moving `m_update_slave_Xref_connection.clear()` out of the `if(m_slave_Xref_item)` block also clears connections left dangling after unlink in the Text-field case — worth keeping.

One request: the new block at `dynamicelementtextitem.cpp:1505` lacks the `!parentGroup()` guard its store branch has (line 1393). For an `"xref"` text inside a group it deletes what `ElementTextItemGroup::updateXref()` just stored, while still linked. From reading the code, not reproduced.

`QETInformation::ELMT_XREF` already exists — prefer it to `QStringLiteral("xref")`.


**Blocking:** none.
**Non-blocking:** the `!parentGroup()` guard — please check it, I read it but did not reproduce it; `QETInformation::ELMT_XREF`; merge master (28 behind); indentation at `plclinkwidget.cpp:358`.

Kellermorph:
Thanks for the detailed feedback. All non-blocking items addressed in the latest commit:

Added !parentGroup() guard to the stale xref removal block at dynamicelementtextitem.cpp:1505 — without it, a grouped "xref" text would have its entry deleted by DynamicElementTextItem::updateXref() right after ElementTextItemGroup::updateXref() stored it.
Replaced QStringLiteral("xref") with QETInformation::ELMT_XREF in linkelementcommand.cpp (3 sites) and masterelement.cpp (1 site).
Fixed indentation of the comment block in plclinkwidget.cpp.
Merged master (28 commits behind → now up to date).

PR #872, opened 2026-09-14
Add Linux CI: build, unit tests, and the IPC regression gate

ispyisail:
Stacked on #871 — that PR adds the test this job runs. The last commit here is the only new one.

Nothing currently builds QElectroTech on Linux in CI, and nothing runs `tests/` at all; the existing workflows build Windows and generate documentation. This adds one job: configure and build with Qt 6 Debug, run `ctest`, then run the IPC open-forwarding regression test from #871.

I ran the job's exact step sequence in a clean `ubuntu:26.04` container, both directions:

| | unit tests | IPC gate | job |
|---|---|---|---|
| with the #868 fix | 5/5 pass | passes | green |
| with the fix reverted | 5/5 **still** pass | fails, exit 139 | red |

The unit tests passing in both arms is the point — the existing suite cannot see this class of bug.

Three choices that are not cosmetic:

- **Runs in a container, not on the runner.** `ubuntu-latest` ships Qt 6.4; the gate needs 6.10.2, and whether the crash reproduces on 6.4 has never been checked. A job that cannot go red is worse than no job.
- **Debug, not Release.** The pre-fix commit survives every attempt built `-O3 -DNDEBUG`.
- **An inconclusive gate run warns rather than fails.** It means the crash path was not exercised — that proves nothing, which is not the same as a regression, and failing on it would make the job flaky rather than useful.

Adds roughly a build plus two minutes per run. If you would rather have the Linux build and unit tests without the IPC gate, say so and I will drop that step — the first two thirds stand on their own.


arummler:
Can you take into account #722 ? I still intend to bisect that large MR at stome point as you proposed. Regarding the Qt minimum version: I was compiling on native Ubuntu 22, 24, 26 taking it as a requirement. If one could agree on a higher Qt LTS version, e.g. 6.10 it would make things easier as several case switches could be removed.

ispyisail:
@arummler You're right, and I'll get out of the way. #722 is the better foundation — eight build combinations against my one, and your matrix already includes `ubuntu-26.04`, which provides Qt 6.10.2 natively and makes my container pointless. I'll drop the workflow from this PR.

One thing worth knowing before you rebase: diffed against master, #722 currently reads as removing `enable_testing()` and deleting `tst_smart_device` and `tst_qetstrings`. It isn't — all three landed on master *after* your branch point of 9 September, so a rebase picks them up and clears the merge conflict at the same time. I did check the effect is real: configuring your branch with `-DPACKAGE_TESTS=ON` gives `Total Tests: 0`, and restoring that one line brings them back.

The gap that remains after #722 lands: across all thirteen workflows, `ctest` appears zero times, every build passes `-DPACKAGE_TESTS=OFF`, and the binary is never executed. The matrix proves QElectroTech compiles and packages on eight platforms — but a change that builds clean and crashes on launch passes every job in it.

@scorpio810 — proposed order, if you agree with it:

1. #871 merges on its own. It is the test script only, no build or CI changes.
2. @arummler rebases #722, then it merges.
3. A small follow-up adds one test job to his `linux-build.yml`: `ubuntu-26.04`, Debug, `PACKAGE_TESTS=ON`, running `ctest` and the regression test. Debug is not cosmetic — the use-after-free does not fault at `-O3 -DNDEBUG`.
4. I close this PR.

Happy to write step 3 against his branch once it lands.

scorpio810:
Thanks @ispyisail, this fills a real gap: nothing built QET on Linux in CI and nothing ran `tests/`. The workflow is simple and the non-obvious choices are well justified in the comments:

- **`ubuntu:26.04` container.** It provides the Qt version the #868 crash was reproduced on, instead of whatever `ubuntu-latest` ships.
- **Debug build.** Given the measurements in the README, a Release job would never catch this regression.
- **Inconclusive runs.** They only warn and don't fail, which avoids a flaky job.
- **Validation.** The README documents runs against both the unfixed and fixed builds, and records the earlier false passes. That will save the next person a lot of time.

One thing I'd like changed before merging:

**Pass the test project explicitly.** Without `--project`, `run.sh` picks the first `examples/*.qet` in alphabetical order, while the README says `industrial.qet` is the default. If a smaller example sorts first, the gate will be inconclusive on every run and only emit a warning nobody reads. Could you pass it in the workflow?

```yaml
tests/ipc-regression/run.sh --binary build/qelectrotech --project examples/industrial.qet
```

and align the README (or make `industrial.qet` the script's default)?

Optional, not blocking:

- `permissions: contents: read` at the top of the workflow.
- A `concurrency` group with `cancel-in-progress: true` and/or `ccache`, since every push and PR triggers a full Debug build.

Also a note for merging: this PR is stacked on #871, so merging it brings #871's commit in too.

Thanks again!

scorpio810:
Thanks @ispyisail, merging this.

I read through `run.sh`. It stays inside its sandbox, only kills the processes it started, keeps `examples/` untouched by working on a copy, and documents every past false pass right where someone would be tempted to "simplify" it. Measuring the fix against both builds is exactly what gives this gate its value.

CI is green and the gate reported a real PASS (both dialogs seen on all 3 attempts, no crash).

Three small follow-ups, none blocking. Happy to take them as a separate PR:

1. **Inconclusive attempts.** If one of 3 attempts is inconclusive, the job still passes on 2 real attempts, which gives about 89% detection instead of 96%. We could require N conclusive attempts, or retry an inconclusive one.
2. **Logs on failure.** On a crash, `KEEP_LOGS=1` keeps the logs, but they're in the container's `/tmp` and disappear with the job. Copying them into the workspace and adding an `actions/upload-artifact` step with `if: failure()` would make CI crashes debuggable.
3. **Empty dialog title.** The second dialog is logged as `''`, which makes the log harder to read. Could we log the window class or geometry as a fallback? This is purely cosmetic, since the unfixed build going red already proves the right path is exercised.

Thanks again for the thorough work!

PR #874, opened 2026-09-14
Add Tab selection cycling and select-all conductors/text fields

ispyisail:
Resubmits #585, rebased onto current master. That PR was closed on 10 September to cut @scorpio810's review backlog, not on merit. It completes the last outstanding part of discussion #574; the other two parts shipped in #586 and #587.

**What this is for:** working QElectroTech without a mouse. Not as a speed-up — clicking is faster when you can see what you want — but for when you cannot click: a broken or unavailable mouse, or a user who cannot use one reliably. Today there is no way to select anything on a folio from the keyboard at all.

## Tab steps through everything on the folio, one item at a time

Press `Tab` and the selection moves to the next item on the folio; press it again for the one after. `Shift+Tab` goes backwards, and it wraps around at the end.

"Item" means everything you could otherwise click: elements, conductors and text fields, in one list. Exactly one is selected at a time, and the view scrolls to whatever it lands on, so you can walk a whole folio without touching the mouse.

The order is the folio's stacking order. A consequence worth seeing before you try it:

Tab cycling

Same folio, same build, one `Tab` press between frames. The first press lands on the element's **label**, the second on the **element itself**. Those are two separate items that sit next to each other in the stacking order — not two modes, and not Tab behaving differently the second time. Every press does the same thing: go to the next item.

## Escape steps back out

Because Tab is taken for items, it can no longer move focus off the drawing area — which would strand the very person this is for. `Escape` reverses that: the first press drops the selection, the second hands focus back to the rest of the window, and `Tab` then moves between panels again as usual.

Tab is also ignored while you are typing in a text item, so it never interrupts editing.

## Two new Edit menu entries

**Select all conductors** and **Select all text fields**, beside the existing select-all, deselect and invert. Each selects that one kind of thing on the current folio and deselects the rest — handy for restyling every conductor at once. Both are registered with the shortcuts manager with no default key, so anyone can bind one.

## Verified

- Builds clean on Qt 5 and Qt 6; QElectroTech's own tests pass 5/5 on both.
- Tab stepping and the Escape exit were both exercised in a real session under Xvfb. The screenshot is from that run. For the exit: with an item selected, Escape clears it; a second Escape changes nothing on screen; a `Tab` after that moves focus in the toolbar rather than selecting on the canvas — a view that no longer holds focus.
- **Select all conductors** was triggered from the Edit menu in a running session: every conductor on the folio is selected and the elements are left alone. Select all text fields shares the same dispatch path.

One thing to note about the two new entries and the Ajouter submenu: they are new French source strings with no translation yet, so they currently read in French in an English interface. Happy to run lupdate and add English if you would rather they land translated.


scorpio810:
Please add a little video howto.
On diagram select all conductor ok, but you can only remove all, F2 shorcut not work.

PR #876, opened 2026-09-15
Give the keyboard the folio's context menu, not a generic one

ispyisail:
Pressing the Menu key on a folio gives a bare Undo/Redo/Cut/Copy/Paste/Delete/Select All menu, with almost everything greyed out, instead of the menu a right-click gives. Anyone working without a mouse gets the wrong menu for whatever they have selected.

before and after

Same element selected, same branch, only this change differing between the two frames.

## Cause

A context menu raised from the keyboard carries no useful position — Qt does not aim the event at the selection. `DiagramView::contextMenuEvent()` passes the event to `QGraphicsView` first, which hands it to whichever item holds focus. That item answers with its own default menu and accepts the event, so the early return fires and the folio's menu is never built. Even past that, the `itemAt()` lookup below would have used an unrelated point.

## Change

A keyboard-raised menu is built directly rather than being offered to the items first, and aimed at the centre of the selection — or the middle of the view when nothing is selected. The mouse path is untouched.

## Verified

Measured on this branch with only this change applied: the menu went from 7 actions, all but one disabled, to 16, positioned on the selected element. Builds clean on Qt 5 and Qt 6; QElectroTech's own tests pass 5/5 on both.


scorpio810:
Please add a video howto, thanks.

scorpio810:
Maybe an keyboard_whithout_mouse_shorcut.md can help?

PR #880, opened 2026-09-15
Remove Qt5 

arummler:
Update README and INSTALL instructions to use CMake and Qt6. Removal of vestigal Qt5 code (is that already fine? otherwise to be split).


scorpio810:
Thanks for this cleanup, @arummler! Since we only package with Qt6 + KF6 now (Debian/PPA, Flatpak, Snap and Windows CI all pass `-DQT_VERSION_MAJOR=6 -DPACKAGE_TESTS=OFF`), this is safe on the packaging side. The only side effect is a harmless CMake warning about `QT_VERSION_MAJOR` being unused, and I'll drop that flag from the scripts later.

A few things to fix before merging:

**1. `sources/TerminalStrip/ui/terminalstripmodel.h`: the wrong branch was kept.**
The Qt5 `uint` overload survived instead of the Qt6 one. It still compiles through implicit conversion, but the `size_t` seed gets truncated on 64-bit. Please restore:
```cpp
inline size_t qHash(const QColor &key, size_t seed = 0) {
	return qHash(key.rgba(), seed);
}
```
The `QPointer<Element>` overload right below is also `uint` (pre-existing), so it would be good to align it too.

**2. `tests/*/CMakeLists.txt` still have the Qt5 fallback.**
The root CMakeLists no longer sets `QT_VERSION_MAJOR` before `add_subdirectory(tests)`. As a result, catch, qttest, googletest and googlemock fall back to `find_package(QT NAMES Qt6 Qt5 ...)`. On a machine with Qt5 and Qt6 in different prefixes, they can pick Qt5 and mix versions at link time. `PACKAGE_TESTS` defaults to ON, so a plain `cmake ..` is affected. Either keep `set(QT_VERSION_MAJOR 6)` in the root, or make the tests Qt6-only. `tests/ipc-regression/run.sh` also still greps for `libQt5Core`.

**3. `cmake/fetch_singleapplication.cmake`: `CACHE` without `FORCE`.**
The previous code forced it. An existing build dir with `QT_DEFAULT_MAJOR_VERSION=5` in its cache will keep 5. A plain `set(QT_DEFAULT_MAJOR_VERSION 6)` would be enough.

**To your question about splitting:**
In `projectprintwindow.cpp`, `setOption(QAbstractPrintDialog::PrintShowPageSize, true)` is new behaviour on Qt6, where the branch was only a `qDebug()` before. It's probably fine, but since it isn't pure removal, I'd prefer it in a separate commit.

**Pre-existing bug made visible:**
In `diagramview.cpp` (`mouseReleaseEvent`), the kept Qt6 branch calls `menu->popup(e->pos())`. `pos()` is widget-local, while `popup()` expects global coordinates (Qt5 used `globalPos()`), so the menu opens at the wrong place. The fix is `menu->popup(e->globalPosition().toPoint());`, either here or in a follow-up.

**Leftovers (optional / follow-up):** dead code in `bordertitleblock.cpp:59` (`QT_VERSION < 4.8`) and the `titleblocktemplate` caching exception you mentioned.

The rest looks good to me: the `#if/#else/#endif` blocks are balanced, and `qgimanager`, `fontToString` and `elementslocation` are fine. I only reviewed statically, though.

@ispyisail, since this PR touches a lot of files, it would be great if you could run your testing tools on this branch once the fixes above are in, as a second check before we merge.

Thanks again!

ispyisail:
Merged onto master, Qt6 build clean, ctest 6/6. Branch is 3 commits behind — needs a merge.

All 9 Qt6-*minor* guards (6.2/6.5/6.6/6.7) survive the sweep: 9 before, 9 after. The `setXml()` `QRegularExpression` port and the print-dialog `setOption` are real fixes, not just deletions.

One request: `cmake/fetch_singleapplication.cmake:38` drops the `FORCE` master had. Cache values stick — I confirmed with a minimal project that a tree already holding `QT_DEFAULT_MAJOR_VERSION=5` keeps it across reconfigures, so SingleApplication and kdeaddons would build against Qt5. Fresh build dirs are fine.

`projectprintwindow.cpp` is `Q_OS_WIN`-only; unverified on Linux.


**Blocking:** restore `FORCE` on `QT_DEFAULT_MAJOR_VERSION`.
**Non-blocking:** merge master (3 behind); four stray blank lines in `main.cpp`; the `Q_OS_WIN` branch needs someone on Windows to confirm.

arummler:
Done. And I was mistaken / read too fast: the caching is actually used by default...and was disabled only for Qt4. So it stays and I removed only the old switch.

ispyisail:
Correction to my earlier comment: I credited this with two Qt6 fixes. It has one. Master's `setXml()` `#else` branch already holds the working `QRegularExpression` code calling `addElementDefinition()` — the `qDebug() << "Help code for QT 6 or later"` only prints ahead of it. Deleting the dead Qt5 branch is cleanup, not a repair. My mistake.

The `QT_DEFAULT_MAJOR_VERSION` change is better than the `FORCE` I asked for: a plain `set()` shadows the cache outright. Verified — a tree cached at 5 reads 6.

Rebuilt at `cfed78a0`: clean, ctest 6/6, 0 behind.

Physically tested the one real fix. Print dialog in both arms, collapsed and expanded — identical. `PrintShowPageSize` looks ignored by the Qt/CUPS dialog on Linux, so its effect is unverified here and wants a Windows check.

**Blocking:** none.
**Non-blocking:** the `setXml` hunk is cleanup rather than a fix, worth not claiming otherwise in the description.

arummler:
Correction of correction: the caching should be indeed disabled for anything newer according to the old distinction. I will fix it.

scorpio810:
Thanks @arummler, all the blocking points from my review are addressed:

- `QT_DEFAULT_MAJOR_VERSION`: the plain `set()` is fine (and verified by @ispyisail on a tree cached at 5).
- `qHash(const QColor &, size_t)`: Qt6 overload restored.
- `tests/*/CMakeLists.txt` are Qt6-only now, and `run.sh` no longer greps for `libQt5Core`.
- `diagramview.cpp`: context menu now uses `e->globalPosition().toPoint()`, thanks for fixing that pre-existing bug.
- Dead code in `bordertitleblock.cpp` is gone.

The branch is 25 commits behind master but merges without conflicts.

About 261cca52 (cache removal in `TitleBlockTemplateRenderer`): it matches the effective behaviour since Qt 4.8, so no functional change is expected. However, it came after @ispyisail's test run. Could you (@arummler and @ispyisail) please check the titleblock rendering on your side (display, print, PDF export) before we merge?

Follow-ups, non-blocking, can be done in a separate PR:
- `m_rendered_template`, `m_last_known_titleblock_width`, `renderToQPicture()` and `invalidateRenderedTemplate()` are now unused leftovers of the cache (the last one is still called from `bordertitleblock.cpp`).
- `qHash(const QPointer<Element> &, uint)` in `terminalstripmodel.h` could be aligned to `size_t` too.
- `PrintShowPageSize` in `projectprintwindow.cpp` still needs a check on Windows.

Great cleanup, thanks again!

INSTALL.md
```diff
@@ -0,0 +1,349 @@
...
+the closest match.
+
+### Debian / Ubuntu
+
+```sh
+sudo apt install \
+    build-essential cmake ninja-build git \
+    qt6-base-dev qt6-base-private-dev qt6-tools-dev qt6-tools-dev-tools \
+    libsqlite3-dev \
+    libkf6coreaddons-dev libkf6widgetsaddons-dev
```
scorpio810, line 110: cmake,
               ninja-build,
               qt6-base-dev,
               qt6-svg-dev,
               qt6-l10n-tools,
               qt6-tools-dev,
               libkf6widgetsaddons-dev,
               libkf6coreaddons-dev,
               libsqlite3-dev,
               libqt6sql6-sqlite,
               qt6-base-private-dev,
               libqt6pdf6,
               qt6-pdf-dev,
               libcups2-dev,
               pkgconf,
               libpugixml-dev,
               qt6-declarative-dev

PR #881, opened 2026-09-15
Add Hide full masters checkbox to slave linking widget

Kellermorph:
Add a checkbox Masquer les éléments maîtres pleins to LinkSingleElementWidget that allows hiding masters that have reached their slave limit. When checked, full masters are hidden from the candidate list; when unchecked, all masters are shown (previous behavior).

The setting is persisted via QSettings and applies only to slave elements (not reports). The search field respects the hide state so full masters stay hidden during search.

The unused m_hidden_masters_label hint text from the .ui file is removed since it is now replaced by the checkbox.


ispyisail:
Builds clean on Qt6, ctest 6/6, current with master. One bug, and a couple of smaller things.

**The search interaction from #858 does not recur here** — both loops in `on_m_search_field_textEdited()` respect the hide state, so clearing a search no longer brings full masters back. Worth saying, since that was the defect last time.

Also: the label being removed claimed *"les éléments maîtres ayant atteint leur nombre maximal d'esclaves sont masqués"*, and `buildTree()` on master has no such filter — nothing hid them. The label was simply false. This makes it optionally true and drops the label.

**Toggling the checkbox while a search is active silently drops the search.** `on_m_hide_full_masters_cb_toggled()` calls `buildTree()`, which rebuilds every item, and `buildTree()` never re-applies the filter. The search box still shows your text but everything is listed again — the mirror image of the #858 bug. Re-applying the search at the end of the toggle handler would cover it:

```cpp
on_m_search_field_textEdited(ui->m_search_field->text());
```

**`isMasterFull()` is never called.** The two occurrences in the diff are its declaration and its definition.

**The label survives for report elements.** `delete ui->m_hidden_masters_label` is inside the `if (slave)` block, so a NextReport/PreviousReport still shows a hint that is now both false and irrelevant — a report links to reports, not masters.

That makes one line of the description inaccurate too: it says the label "from the .ui file is removed", but the .ui is not in the diff. It is deleted at runtime, for slaves only.

Minor: rows 2/3/4 are hardcoded in the .cpp to mirror the .ui, so adding a row to the .ui would silently misplace things.

I could not exercise any of this: the example corpus reaches 0 of the 23 executable changed lines, so the findings are from reading rather than running.

---

**Blocking:** none of the above. All of it is read, not run — the corpus reaches 0 of the 23 changed lines. @scorpio810's crash on open and use-after-free on toggle are the two that gate the merge.

**Non-blocking:** unused `isMasterFull()`; the label surviving for report elements; the description line about the `.ui`; hardcoded rows 2/3/4.

Correction to my search point: the text isn't kept, it's cleared by `setUpCompleter()`, so the fix I suggested would have filtered on an empty string. @scorpio810's `updateItemsVisibility()` supersedes it and fixes the green highlight and the dangling `m_pending_qtwi` in the same change.

scorpio810:
Thanks @Kellermorph, the feature makes sense, and keeping it unchecked by default preserves the current "advisory limit" behaviour (see the comment in `availableElements()`). There are two blocking issues though, plus a few smaller things.

**1. Crash when opening the widget (blocking)**
The checkbox is created, connected and `setChecked()` in the constructor *before* `setElement(elmt)` runs. If the stored setting is `true`, `toggled` fires and calls `buildTree()`, then `setUpHeaderLabels()`, which dereferences `m_element` while it is still null. So once a user has ticked the box, the next time a slave's properties are opened, QET crashes.
To fix it, restore the state before connecting (or use a `QSignalBlocker`), and don't call `buildTree()` from the toggle at all (see point 2).

**2. Use-after-free on toggle (blocking)**
`on_m_hide_full_masters_cb_toggled()` rebuilds the whole tree. In non-live-edit mode, if the user has already picked a master (green row), `m_pending_qtwi` now points to a deleted item. The next `linkTriggered()` then calls `setBackground()` on freed memory. The rebuild also clears the search field (via `setUpCompleter()`) and drops the green highlight while the link is still pending.
The checkbox only changes visibility, so there is no need to rebuild. A single helper can be used by both the search slot and the toggle slot:
```cpp
void LinkSingleElementWidget::updateItemsVisibility()
{
	const QString text = ui->m_search_field->text();
	const bool hide_full = m_hide_full_masters_cb
			&& m_hide_full_masters_cb->isChecked();

	for (auto it = m_qtwi_strl_hash.cbegin(); it != m_qtwi_strl_hash.cend(); ++it)
	{
		bool match = text.isEmpty();
		for (const QString &str : it.value()) {
			if (str.contains(text, Qt::CaseInsensitive)) {
				match = true;
				break;
			}
		}
		it.key()->setHidden(!match
			|| (hide_full && m_full_masters.contains(it.key())));
	}
}
```
The toggle slot then only saves the setting and calls this helper.

**3. `on_` slot name**
`ui->setupUi()` calls `QMetaObject::connectSlotsByName()`. Because the checkbox doesn't exist yet at that point, `on_m_hide_full_masters_cb_toggled` will print a *"No matching signal"* warning at every startup of the widget. Please rename it (e.g. `hideFullMastersToggled`), since it is connected explicitly anyway.

**4. Do the layout in the `.ui` file**
The description says the hint label was removed from the `.ui`, but the `.ui` isn't part of the commit. As a result, the label is still shown for folio report elements, with a text about hidden masters that doesn't apply to them. Shuffling grid rows and deleting a `ui->` member at runtime is also fragile. It would be cleaner to remove `m_hidden_masters_label` and add the checkbox directly in `linksingleelementwidget.ui`, then just `hide()` the checkbox for non-slave elements.

**5. Visibility when already linked**
`showButtons()`/`hideButtons()` hide and show the search field depending on the link state, but the checkbox stays visible above an empty tree when the slave is already linked. It should follow the search field.

**6. Minor**
- `isMasterFull()` is never used, so it can be removed.
- Please also clear `m_full_masters` in `clearTreeWidget()`, so it never keeps pointers to deleted items.
- The new `tr()` string will need a `.ts` update. That's not required in this PR, I can run lupdate afterwards.

Thanks again!

scorpio810:
Thanks @ispyisail for the review, I agree on the label, the unused `isMasterFull()` and the `.ui` not being part of the diff. Since the changed lines aren't exercised by the example corpus, here are the runtime issues found by reading the surrounding code. Two of them are blocking.

**1. Crash when opening the widget (blocking)**
The checkbox is created, connected and `setChecked()` in the constructor *before* `setElement(elmt)` runs, so `m_element` is still null at that point. If the stored setting is `true`, `toggled` fires and calls `buildTree()`, then `setUpHeaderLabels()`, which dereferences the null `m_element`. So once a user has ticked the box, the next time a slave's properties are opened, QET crashes.
To fix it, restore the state before connecting (or use a `QSignalBlocker`).

**2. Use-after-free on toggle (blocking)**
`on_m_hide_full_masters_cb_toggled()` rebuilds the whole tree, which deletes every item. In non-live-edit mode, if the user has already picked a master (green row), `m_pending_qtwi` now points to a deleted item. The next `linkTriggered()` then calls `setBackground()` on freed memory.

**About the search being lost:** one small correction. The search text isn't kept, it is cleared, because `buildTree()` ends with `setUpCompleter()`, which calls `ui->m_search_field->clear()`. Re-applying `m_search_field->text()` after `buildTree()` would therefore filter on an empty string.
The checkbox only changes visibility, so the toggle shouldn't rebuild anything. This fixes the lost search, the lost green highlight and the dangling `m_pending_qtwi` at once. A single helper can be used by both slots:
```cpp
void LinkSingleElementWidget::updateItemsVisibility()
{
	const QString text = ui->m_search_field->text();
	const bool hide_full = m_hide_full_masters_cb
			&& m_hide_full_masters_cb->isChecked();

	for (auto it = m_qtwi_strl_hash.cbegin(); it != m_qtwi_strl_hash.cend(); ++it)
	{
		bool match = text.isEmpty();
		for (const QString &str : it.value()) {
			if (str.contains(text, Qt::CaseInsensitive)) {
				match = true;
				break;
			}
		}
		it.key()->setHidden(!match
			|| (hide_full && m_full_masters.contains(it.key())));
	}
}
```
The toggle slot then only saves the setting and calls this helper.

**3. `on_` slot name**
`ui->setupUi()` calls `QMetaObject::connectSlotsByName()`. Because the checkbox doesn't exist yet at that point, `on_m_hide_full_masters_cb_toggled` triggers a *"No matching signal"* warning. Please rename it (e.g. `hideFullMastersToggled`), since it is connected explicitly anyway.

**4. Layout in the `.ui`**
This also covers @ispyisail's remarks on the report label and the hardcoded rows: remove `m_hidden_masters_label` and add the checkbox directly in `linksingleelementwidget.ui`, then just `hide()` the checkbox for non-slave elements.

**5. Visibility when already linked**
`showButtons()` hides the search field once the slave is linked, but the checkbox stays visible above an empty tree. It should follow the search field in `showButtons()`/`hideButtons()`.

**6. Minor**
- Clear `m_full_masters` in `clearTreeWidget()`, so it never keeps pointers to deleted items.
- The new `tr()` string will need a `.ts` update. I can run lupdate afterwards.

Thanks @Kellermorph!

Kellermorph:
Blocking fixes:

Crash on open: QSignalBlocker prevents setChecked() from firing toggled before m_element is initialized
Use-after-free: Toggle now calls updateItemsVisibility() instead of buildTree(), so tree items are never deleted mid-interaction
Other fixes:

Moved checkbox into .ui file, removed hardcoded row numbers
Renamed slot to hideFullMastersToggled to avoid QMetaObject::connectSlotsByName warning
Checkbox follows search field visibility in showButtons()/hideButtons()
clearTreeWidget() now clears m_full_masters and resets m_pending_qtwi
Removed unused isMasterFull() and orphaned member variable
updateItemsVisibility() shared helper keeps search text and green highlight intact on toggle
Thanks to @ispyisail and @scorpio810 for the thorough review — the blocking issues would have been nasty in production.

PR #883, opened 2026-09-15
Persist the folio uuid, derived deterministically for legacy folios

bhangart:
## Persist the folio uuid, derived deterministically for legacy folios

`Diagram::m_uuid` was created in the constructor and never written, so a folio got a new uuid on every load. Inside a running instance that is enough (the project database keys on it), but nothing outside it could tell which folio is which: in the file, folios were only identified by their position.

### Motivation

More and more `.qet` projects live in version control (a Git repository on GitHub or GitLab, reviewed through pull requests, sometimes edited by several people) or are synchronised through a cloud or key-value store. A `.qet` file is plain XML, so in principle it can be diffed, merged and split up, but only if the same folio can be recognised in two versions of the file. Today it cannot:

- **Diffs:** Inserting, deleting or reordering a folio shifts every following `<diagram>` element. A line-based diff, and GitHub's review view, then pair up unrelated folios and show far more change than was made.
- **Merges:** A three-way merge of two branches that both touched the project has no way to match "folio 3" on one side with "folio 3" on the other if either side reordered folios.
- **Tooling:** Any tool that wants to say "folio X changed in this commit", keep per-folio history, lock a single folio, or store folios as separate objects has nothing stable to key on. The title and the folio number are user-editable and not unique.

Element uuids are already persisted and used for cross-folio links, so the file format already relies on uuids for identity; the folio itself was the missing piece. A stable folio uuid is the prerequisite for later work towards better version control support: per-folio diffs and locks (check-out / check-in), and possibly storing a project as a directory with one file per folio.

### Change

Write the uuid as an attribute of `<diagram>` when the whole content is saved, and restore it first thing when the project is loaded, before any item is created. Older versions ignore the attribute, so files stay readable in both directions.

### Folios without a uuid: why not a random one

The obvious migration (keep the random uuid created by the constructor and save it) conflicts with #754 / #779: saving an unmodified project must give the same bytes every time. Every example project predates the attribute, so each load would invent different uuids and write them out.

Measured on the 24 example projects, each resaved 3–4× from the same original, with `QT_HASH_SEED=0` so that QDom's attribute order is stable, and an isolated `HOME` per run:

| Variant | Byte-identical |
|---|---|
| upstream master | 23/24 |
| persist, random uuid | **0/24** |
| persist, derived uuid (this PR) | **23/24** |

The remaining project, `schema_indus.qet`, differs only in element uuids, the known residual #779 leaves for elements; its folio uuid is stable.

#779 solved the same problem for conductors by not writing an invented uuid back at all. That is not an option here: legacy folios would never get a persistent uuid, which is the whole point of the change. Instead, a folio without a uuid gets a name-based (version 5) uuid, derived only from data read from the file:

```cpp
QUuid::createUuidV5(<fixed QET folio namespace>,
                    "legacy" + project title
                             + position of the folio in the file
                             + folio title)
```

- **Reproducible:** The same input file always yields the same uuids, so resaving an unmodified legacy project stays reproducible.
- **Derived once:** The uuid is derived at load time and saved from then on. After that it is read, never recomputed: renaming, reordering or editing the folio later does not change it. Renaming in the same session as the migration does not change it either, since it was derived from the title as loaded.
- **Merge-friendly:** Two people opening the same legacy file on different branches get the same uuid for each folio, even if one of them reorders or renames folios before saving. With random uuids the two branches would disagree about every folio, and a later merge could not match them.
- **No content hash:** The folio content is deliberately not part of the name. QDom keeps attributes in a hash whose iteration order changes between runs, so hashing the content would need a canonical form for no real gain.

> **Note:** Folio uuids are only guaranteed unique within their project. Two unrelated legacy projects with the same title and the same first folio title get the same uuid for that folio, so anything keying folios globally has to combine the folio uuid with a project identifier. The project uuid is not persisted yet; that is a separate change.

### Duplicated uuids

A hand-edited or merged file can contain the same uuid twice, e.g. a folio copied by duplicating its XML block. Since the uuid is used as a key, the second folio gets a derived uuid as well (`"duplicate"` + the clashing uuid + the same inputs as above), so this case is reproducible too. If a derived uuid is ever already taken, which only happens with a hand-crafted file, the name is salted with a counter until it is free.

> **Important:** The namespace uuid is fixed in the code and must never change, or every legacy folio would get a different uuid.

### Tests

Qt 6.4, offscreen, `qelectrotech --resave` / `--set-titleblock`:

- [x] 24 example projects, 3–4 resaves each from the same original: results as in the table above; all folio uuids identical across runs, no duplicates within any project.
- [x] Resaving an already migrated file is byte-identical to the first output.
- [x] Renaming a folio in a migrated file keeps its uuid.
- [x] Swapping two `<diagram>` blocks in a migrated file: each uuid moves with its folio.
- [x] Migrating and renaming in the same run (`--set-titleblock title=...` on a legacy file) gives the same uuids as a plain resave.
- [x] A file with a duplicated uuid: the second folio gets a new uuid, the same one on every run.
- [x] A migrated file opened with upstream master loads normally; the attribute is ignored and dropped on save.

Refs #754, #779


ispyisail:
Current with master, Qt6 build clean, ctest 6/6.

Physically verified the three properties this exists for, using `--resave` on real and hand-edited files (my first pass used `--resave` with one argument, which is a silent no-op -- caught it via a vacuous "0 mismatches" and redid all of this with the correct two-argument form):

- 24 example projects, two independent resaves each: 23/24 byte-identical, matching your table. The one mismatch (`schema_indus.qet`) diffs only on element uuids -- zero folio-uuid involvement, confirming it's the pre-existing #779 residual and not something new here.
- Swapped two `<diagram>` blocks by raw XML edit, reloaded, resaved: both uuids followed their content to the new position, not the old one.
- Hand-inserted a duplicate uuid (copied a `<diagram>` block wholesale): the original kept its uuid, the copy got a fresh derived one, reproducible across two independent runs.

Traced every caller of `Diagram::fromXml`/`initFromXml`: exactly two exist, the project-load loop (`consider_informations=true`) and `duplicateDiagram()` (`false`, so this code never runs there -- a duplicated folio correctly keeps its fresh random uuid rather than copying the source's). Collisions under the "legacy" reason are structurally impossible within one project, since `index` is part of the hash and unique by construction -- confirmed live on the ten folios in `industrial.qet` all titled "Nomenclature", each still distinct.

**Blocking:** none.
**Non-blocking:** `derivedUuid()` reads `root.ownerDocument().documentElement().attribute("title")` for the project title, which only works because the two current callers guarantee that context. If `consider_informations` is ever set true on a paste/duplicate path later, this would silently read the diagram's own title instead. Not reachable today; a one-line comment noting the assumption would save someone a debugging session.

PR #884, opened 2026-09-15
Persist the project uuid, derived from the file content for legacy files

bhangart:
QETProject::m_uuid was created in the constructor and never written, so a project got a new uuid every time it was opened. Inside a running instance it is only used to name the SQLite connection, but nothing outside the instance could tell which project a file belongs to.

Motivation

A .qet file is increasingly handled by tools outside QElectroTech: Git repositories on GitHub or GitLab, cloud storage, key-value stores, per-project locks. All of them need a stable key for "this project":

- The file name and path are not stable: files get renamed, moved, checked out in different places.
- The project title is user-editable and not unique.
- Folio uuids (persisted separately) are only unique within their project; keying folios globally needs a project identifier as well, e.g. projects/{projectUuid}/folios/{folioUuid}.

Change

Write the uuid as an attribute of <project> and restore it in QETProject::openFile(), right after parsing and before the project is built from the XML. Older versions ignore the attribute, so files stay readable in both directions.

The project database is not affected: it takes its connection name from the uuid created at construction (m_uuid is declared before m_data_base), before the file is read. Two open projects carrying the same persisted uuid therefore still get distinct connections.

Files without a uuid: why not a random one

Keeping the random uuid created by the constructor and saving it conflicts with #754 / #779: saving an unmodified project must give the same bytes every time. Every example project predates the attribute. Measured on the 24 example projects (resaved 3x each from the same original, QT_HASH_SEED=0 so that QDom's attribute order is stable, isolated HOME per run):

  upstream master              23/24 byte-identical
  persist, random uuid          0/24
  persist, derived uuid (this) 23/24

The remaining project, schema_indus.qet, differs only in element uuids, the known residual #779 leaves for elements; its project uuid is stable.

Instead, a project file without a uuid gets a name-based (version 5) uuid derived from the raw content of the file:

  QUuid::createUuidV5(<fixed QET project namespace>,
                      "qet-project-legacy\n" + file content without CR)

- The same file always yields the same uuid, so resaving an unmodified legacy project stays reproducible.
- Different projects practically never share a uuid, because any difference in content gives a different one. This is unlike folios, where only data such as title and position could be used; the raw file bytes are stable input for the whole project.
- Carriage returns are dropped before hashing. QFile's Text mode already strips them on Windows but not elsewhere, and git's autocrlf can change them on checkout; either way the uuid is the same on every platform.
- The uuid is derived once, at load time, and saved from then on. After that it is read, never recomputed: renaming the project, editing it or changing it in the same session as the migration does not change it.
- Two people opening the same legacy file on different branches get the same project uuid.

The namespace uuid is fixed in the code and must never change, or every legacy project would get a different uuid.

Known limitations, open for discussion

- Copies share the uuid. Two byte-identical legacy files get the same uuid (examples/cablage-eclairages_sikli-v5.qet and câblage-éclairages-sikli-v5.qet are such a pair), and so does a migrated file copied in the file manager or saved with "Save as". That is what identity means for a copy, and the same happens with Git, but a tool that treats the uuid as globally unique has to cope with it. Regenerating the uuid on "Save as" could be a follow-up, if that is the preferred behaviour.
- A legacy file that differs from another only in formatting (e.g. re-indented) gets a different uuid. The two sides of a merge only agree if they started from the same bytes, which is the normal case.

Tests (Qt 6.4, offscreen, qelectrotech --resave / --set-titleblock / --info)

- 24 example projects, 3 resaves each from the same original: results above; the project uuid is identical across runs. All 24 uuids are distinct, except the byte-identical pair mentioned above.
- Resaving an already migrated file is byte-identical to the first output.
- The same legacy file with CRLF line endings gets the same uuid as with LF.
- Changing the project title in a migrated file keeps its uuid.
- Migrating and modifying in the same run (--set-titleblock on a legacy file) gives the same uuid as a plain resave.
- Re-indenting a legacy file gives a different uuid (expected).
- A migrated file opened with upstream master loads normally; the attribute is ignored and dropped on save.
- --info on a migrated file still works.

Refs #754, #779


ispyisail:
Current with master, Qt6 build clean, ctest 6/6.

Physically verified with the correct two-argument `--resave` this time:

- 24 example projects, two independent resaves each: 23/24 byte-identical. The one mismatch (`schema_indus.qet`) diffed on the `<project uuid=...>` attribute specifically: empty diff, identical -- the instability there is pure element-uuid churn (#779), unrelated to this change.
- CRLF claim tested directly: same file resaved as LF and as hand-converted CRLF gives the identical uuid.
- Reproduced the disclosed "copies share a uuid" limitation on purpose: two byte-identical copies get the identical project uuid on independent resaves, as documented.
- Merged #883 and this PR onto master together: zero conflicts, combined build clean.

Traced the connection-safety claim rather than trusting it: `createDataBase()` -- the only caller of `addDatabase(..., "qet_project_db_" + m_project->uuid().toString())` -- runs inside `projectDataBase`'s constructor, and `m_data_base(this, this)` is in every `QETProject` constructor's initializer list, which completes before the constructor body runs `openFile()`. So the SQLite connection name is always keyed on the original per-instance random uuid, never the persisted one -- structurally guaranteed, not incidental, and it's exactly what lets two projects sharing a persisted uuid still get distinct connections.

Worth noting: folios have no resource keyed on their uuid at construction, so #883 never faced this ordering hazard. This one does, and handles it correctly.

**Blocking:** none.
**Non-blocking:** none.

PR #889, opened 2026-09-16
Add "Reload element drawings" to refresh placed elements

### LINKED ISSUE #802: refresch page after inserting a modifyed element
When i modify a element and reinsert the element, i get the popup to remove the old of keep them both.
After removing the old element, i have to close my project file and reopen it to see the changed element correct,
Is there a refresch button or mechanisme to refresch the page?

ispyisail:
Fixes #802 (option 1 from my own analysis there). A placed element is drawn once at construction: `buildFromXml()` only turns `terminal`/`input`/`dynamic_text` tags into live child objects, everything else (line, rect, ellipse, polygon, arc, text) is pre-rendered into a `QPicture` by `ElementPictureFactory` and cached forever under the element's uuid, with no invalidation path anywhere in the codebase. Edit and save a symbol's drawing and every placed instance keeps showing the old one until the project is closed and reopened.

## Change

Scoped to what's safe without ever risking a conductor or a dynamic text's per-instance state:

- `ElementPictureFactory::dropCache(location)` -- forgets the cached drawing for one location.
- `Element::reloadPicture()` -- re-fetches and repaints one instance.
- **Projet > "Recharger les dessins des éléments"** -- walks every diagram, drops each distinct location's cache once, then reloads every placed instance.

Deliberately does **not** touch terminals or dynamic texts. Terminal positions are what conductors attach to, and a definition whose terminals moved still needs the existing remove-and-reinsert workflow -- a wrong guess there would silently misconnect wires, which is worse than the bug this fixes.

## Tests

Qt6, build clean, ctest 6/6.

Triggered the new action on a real, densely-wired project (76 elements) via exact keyboard-menu navigation cross-checked against the menu's own `addAction` order in the source: ran to completion, showed the correct confirmation dialog, no crash, diagram unchanged and uncorrupted afterward.

I was not able to complete a full live "edit a symbol, watch the placed instance update without reopening" trace -- opening the element editor on a selected item via GUI automation was unreliable in my environment, and I had no file-based (`common://`) element available to mutate on disk as a shortcut (every example project embeds its elements). The mechanism is traced correct rather than assumed: `ElementsLocation::xml()` for an `embed://` location reads the project's live in-memory collection DOM on every call, so a dropped cache rebuilds from whatever was most recently saved there.


scorpio810:
Thanks @ispyisail, the approach is good and nicely scoped. A few things before merging; I've attached a patch on top of your branch covering them.

1. **Element can go blank.** `reloadPicture()` clears `m_picture` before `getPictures()`, which returns early if the location no longer exists or `build()` fails. The element then disappears from the folio instead of keeping its old drawing. The patch now builds into temporaries and only swaps on success.

2. **Geometry / terminal mismatch.** If the new definition changes size, hotspot, or terminals (added, removed or moved), the new drawing no longer matches the bounding rect or the live terminals. The user sees a terminal in one place while the conductor stays attached elsewhere. The patch compares these first and skips such elements. The action then lists them and warns that they must be removed and re-inserted, which deletes their connected conductors.

3. **`dropCache()` is a no-op in the main case.** `ElementScene::toXml()` generates a new uuid on every save, so `cacheKey()` returns the *new* uuid and the old entry is never removed. It still works because the new uuid forces a rebuild, but old entries leak until the app quits, and the doc comment describes a mechanism that isn't the one actually at work. Not blocking; the comment should be fixed.

4. **Not undoable.** Fine for a purely visual refresh, but I added it to the status tip so users know.

The three touched files build clean on Qt 6.4.

Could you run the real scenario before we merge? Place an element, edit and save its symbol, then run the action. Then do it again after moving a terminal, to check the warning path.


```
From 355750c81c151889f97ca8e2901ce451daccc197 Mon Sep 17 00:00:00 2001
From: Laurent Trinques <scorpio@qelectrotech.org>
Date: Wed, 16 Sep 2026 08:52:39 +0000
Subject: [PATCH] Reload element drawings: keep old drawing on failure, skip
 elements whose geometry changed (#802)

- Element::reloadPicture() now returns a ReloadPictureResult and never
  clears the current drawing before a successful rebuild: a missing or
  unreadable definition leaves the element as it was instead of blank.
- Elements whose size, hotspot or terminals (added, removed or moved)
  differ from the new definition are not redrawn: the new drawing would
  no longer match their bounding rect and live terminals.
- The action lists those elements and warns that they must be removed
  and re-inserted, which deletes the conductors already connected to
  them.
- Status tip states the action is not undoable.
---
 sources/qetdiagrameditor.cpp        |  71 ++++++++++++---
 sources/qetgraphicsitem/element.cpp | 135 ++++++++++++++++++++++++----
 sources/qetgraphicsitem/element.h   |   9 +-
 3 files changed, 189 insertions(+), 26 deletions(-)

diff --git a/sources/qetdiagrameditor.cpp b/sources/qetdiagrameditor.cpp
index 2804e7d7..597c6356 100644
--- a/sources/qetdiagrameditor.cpp
+++ b/sources/qetdiagrameditor.cpp
@@ -555,7 +555,7 @@ void QETDiagramEditor::setUpActions()
 	m_reload_element_drawings = new QAction(QET::Icons::ViewRefresh, tr("Recharger les dessins des éléments"), this);
 	m_reload_element_drawings->setStatusTip(
 		tr("Redessine chaque élément placé d'après sa définition actuelle,"
-		   " sans avoir à fermer et rouvrir le projet"));
+		   " sans avoir à fermer et rouvrir le projet (action non annulable)"));
 	connect(m_reload_element_drawings, &QAction::triggered, this, &QETDiagramEditor::slot_reloadElementDrawings);
 
 	#ifdef QET_EXPORT_PROJECT_DB
@@ -2990,9 +2990,10 @@ void QETDiagramEditor::slot_terminalNumbering() {
 	is closed and reopened.
 
 	Purely visual and not undoable, the way pressing a "refresh" button
-	would be: it does not touch position, rotation, links, elementInformations,
-	labels or dynamic texts, and does not detect or handle a definition whose
-	terminals moved -- those still need the usual remove-and-reinsert.
+	would be: nothing is pushed on the undo stack and the project is not
+	marked as modified. Elements whose size, hotspot or terminals changed
+	are skipped and listed: they must be removed and re-inserted, which
+	deletes the conductors already connected to them.
 */
 void QETDiagramEditor::slot_reloadElementDrawings() {
 	QETProject *project = currentProject();
@@ -3015,12 +3016,62 @@ void QETDiagramEditor::slot_reloadElementDrawings() {
 		}
 	}
 
-	for (Element *elmt : elements) {
-		elmt->reloadPicture();
+	int reloaded = 0;
+	int unavailable = 0;
+	QStringList geometry_changed;
+	for (Element *elmt : elements)
+	{
+		switch (elmt->reloadPicture())
+		{
+			case Element::ReloadPictureResult::Reloaded:
+				++reloaded;
+				break;
+			case Element::ReloadPictureResult::Unavailable:
+				++unavailable;
+				break;
+			case Element::ReloadPictureResult::GeometryChanged:
+			{
+				const Diagram *diagram = elmt->diagram();
+				const QString folio = diagram
+						? tr("folio %1").arg(project->folioIndex(diagram) + 1)
+						: QString();
+				geometry_changed << QStringLiteral("%1 (%2)").arg(elmt->name(), folio);
+				break;
+			}
+		}
 	}
 
-	QET::QetMessageBox::information(
-		this,
-		tr("Recharger les dessins des éléments"),
-		tr("%n élément(s) redessiné(s).", "", elements.size()));
+	QString message = tr("%n élément(s) redessiné(s).", "", reloaded);
+
+	if (unavailable) {
+		message += QStringLiteral("\n\n")
+				% tr("%n élément(s) dont la définition est introuvable ou illisible :"
+					  " leur dessin actuel a été conservé.", "", unavailable);
+	}
+
+	if (geometry_changed.isEmpty())
+	{
+		QET::QetMessageBox::information(
+			this, tr("Recharger les dessins des éléments"), message);
+		return;
+	}
+
+	message += QStringLiteral("\n\n")
+			% tr("%n élément(s) non redessiné(s) : leur taille, leur point de saisie"
+				  " ou leurs bornes ont changé (borne ajoutée, supprimée ou déplacée).",
+				  "", geometry_changed.size())
+			% QStringLiteral("\n\n")
+			% tr("Pour les mettre à jour, il faut les supprimer puis les réinsérer."
+				  " Attention : cette opération supprime les conducteurs déjà reliés"
+				  " à ces éléments, qu'il faudra retracer.");
+
+		//The full list goes in the expandable, scrollable details area
+		//so the dialog stays readable on large projects.
+	QMessageBox box(QMessageBox::Warning,
+					tr("Recharger les dessins des éléments"),
+					message,
+					QMessageBox::Ok,
+					this);
+	box.setDetailedText(geometry_changed.join(QLatin1Char('\n')));
+	box.exec();
 }
diff --git a/sources/qetgraphicsitem/element.cpp b/sources/qetgraphicsitem/element.cpp
index ef55e42a..1e2d729e 100644
--- a/sources/qetgraphicsitem/element.cpp
+++ b/sources/qetgraphicsitem/element.cpp
@@ -1792,24 +1792,129 @@ ElementsLocation Element::location() const
 	(buildFromXml()), and nothing afterwards ever makes it look again --
 	editing and saving the definition leaves every already-placed instance
 	showing the old drawing until the project is closed and reopened
-	(bugtracker #802). This is the per-instance half of the fix: the caller
-	is expected to have already dropped the shared ElementPictureFactory
-	cache for this location, or the fresh fetch below just returns the same
-	cached picture unchanged.
-
-	Deliberately limited to the drawing: terminals and dynamic texts are
-	live objects carrying state a reload cannot safely fabricate --
-	terminal positions are what conductors are attached to, and dynamic
-	texts carry per-instance overrides (position, visibility) a rebuild
-	would have to invent a default for. A definition whose terminals moved
-	still needs the element removed and re-inserted, same as today.
+	(bugtracker #802). This is the per-instance half of the fix.
+
+	Deliberately limited to the drawing. Terminals are what conductors are
+	attached to: if the new definition adds, removes or moves a terminal,
+	or changes the element size or hotspot, the new drawing would no longer
+	match the live terminals and bounding rect. Such an element is left
+	untouched and GeometryChanged is returned; it has to be removed and
+	re-inserted, which deletes the conductors already connected to it.
+
+	If the definition cannot be found or read, the current drawing is kept
+	and Unavailable is returned, so the element never goes blank.
+
+	Purely visual: nothing is pushed on the undo stack and the project is
+	not marked as modified.
+	@return what happened to this element
 */
-void Element::reloadPicture()
+Element::ReloadPictureResult Element::reloadPicture()
 {
-	m_picture = QPicture();
-	m_low_zoom_picture = QPicture();
-	ElementPictureFactory::instance()->getPictures(m_location, m_picture, m_low_zoom_picture);
+	if (!m_location.exist()) {
+		return ReloadPictureResult::Unavailable;
+	}
+
+	const QDomElement definition = m_location.xml();
+	if (definition.isNull()) {
+		return ReloadPictureResult::Unavailable;
+	}
+
+	if (!definitionGeometryMatches(definition)) {
+		return ReloadPictureResult::GeometryChanged;
+	}
+
+	QPicture picture;
+	QPicture low_zoom_picture;
+	ElementPictureFactory::instance()->getPictures(m_location,
+												   picture,
+												   low_zoom_picture);
+	if (picture.isNull()) {
+		return ReloadPictureResult::Unavailable;
+	}
+
+	m_picture = picture;
+	m_low_zoom_picture = low_zoom_picture;
 	update();
+	return ReloadPictureResult::Reloaded;
+}
+
+/**
+	@brief Element::definitionGeometryMatches
+	Compare the geometry described by @p definition with this live element:
+	size and hotspot (normalized the same way setSize()/setHotspot() do it)
+	and the set of terminal positions (same parsing rules as
+	TerminalData::fromXml()).
+	@param definition : the <definition> root of the element
+	@return true if the new drawing can be applied without desynchronizing
+	the bounding rect or the terminals
+*/
+bool Element::definitionGeometryMatches(const QDomElement &definition) const
+{
+	int w = 0, h = 0, hot_x = 0, hot_y = 0;
+	if (!QET::attributeIsAnInteger(definition, QStringLiteral("width"), &w)         ||
+		!QET::attributeIsAnInteger(definition, QStringLiteral("height"), &h)        ||
+		!QET::attributeIsAnInteger(definition, QStringLiteral("hotspot_x"), &hot_x) ||
+		!QET::attributeIsAnInteger(definition, QStringLiteral("hotspot_y"), &hot_y)) {
+		return false;
+	}
+
+		//Same rounding as setSize()
+	while (w % 10) ++w;
+	while (h % 10) ++h;
+	if (QSize(w, h) != dimensions) {
+		return false;
+	}
+
+		//Same clamping as setHotspot()
+	const QPoint new_hotspot = dimensions.isNull()
+			? QPoint(0, 0)
+			: QPoint(qMin(hot_x, w), qMin(hot_y, h));
+	if (new_hotspot != hotspot_coord) {
+		return false;
+	}
+
+		//Terminal positions described by the new definition
+	QList<QPointF> new_terminals;
+	for (QDomElement description = definition.firstChildElement(QStringLiteral("description")) ;
+		 !description.isNull() ;
+		 description = description.nextSiblingElement(QStringLiteral("description")))
+	{
+		for (QDomElement terminal = description.firstChildElement(QStringLiteral("terminal")) ;
+			 !terminal.isNull() ;
+			 terminal = terminal.nextSiblingElement(QStringLiteral("terminal")))
+		{
+			qreal x = 0.0, y = 0.0;
+			if (QET::attributeIsAReal(terminal, QStringLiteral("x"), &x) &&
+				QET::attributeIsAReal(terminal, QStringLiteral("y"), &y)) {
+				new_terminals << QPointF(x, y);
+			}
+		}
+	}
+
+	if (new_terminals.size() != m_terminals.size()) {
+		return false;
+	}
+
+		//Every live terminal must still exist at the same place
+	for (const Terminal *terminal : m_terminals)
+	{
+		const QPointF live_pos = mapFromScene(terminal->dockConductor());
+		bool found = false;
+		for (int i = 0 ; i < new_terminals.size() ; ++i)
+		{
+			const QPointF delta = new_terminals.at(i) - live_pos;
+			if (qAbs(delta.x()) < 0.01 && qAbs(delta.y()) < 0.01) {
+				new_terminals.removeAt(i);
+				found = true;
+				break;
+			}
+		}
+		if (!found) {
+			return false;
+		}
+	}
+
+	return true;
 }
 
 /**
diff --git a/sources/qetgraphicsitem/element.h b/sources/qetgraphicsitem/element.h
index 709645c4..1b9ee940 100644
--- a/sources/qetgraphicsitem/element.h
+++ b/sources/qetgraphicsitem/element.h
@@ -83,6 +83,7 @@ class Element : public QetGraphicsItem
 			Element::kind link_type = Element::Simple);
 		~Element() override;
 	private:
+		bool definitionGeometryMatches(const QDomElement &definition) const;
 		Element(const Element &);
 
 		// attributes
@@ -153,7 +154,13 @@ class Element : public QetGraphicsItem
 
 		QString name() const override;
 		ElementsLocation location() const;
-		void reloadPicture();
+		/// Result of Element::reloadPicture()
+		enum class ReloadPictureResult {
+			Reloaded,          ///< drawing replaced by the current definition
+			Unavailable,       ///< definition missing or unreadable, old drawing kept
+			GeometryChanged    ///< size, hotspot or terminals changed, old drawing kept
+		};
+		ReloadPictureResult reloadPicture();
 		virtual void setHighlighted(bool);
 		void displayHelpLine(bool b = true);
 		QSize size() const;
-- 
2.43.0
```

ispyisail:
Pushed your patch as-is. Checked all four points against the code before applying:

- **#1/#2**: confirmed `reloadPicture()` never touches `m_picture` on failure (builds into locals first), and `definitionGeometryMatches()`'s rounding/clamping matches `setSize()`/`setHotspot()` exactly, byte for byte.
- **#3**: verified directly — `ElementScene::toXml()` calls `QUuid::createUuid()` unconditionally on every save, so `dropCache()` is a confirmed no-op in the main case, as you said. Doc comment fix still outstanding, not done here.

Ran the scenario you asked for, in one running process (no restart):
1. Edited a placed element's drawing on disk, reran the action → clean "11 élément(s) redessiné(s)", drawing updated live.
2. Then moved one of its terminals and reran → "10 redessiné(s) / 1 non redessiné(s)... doit être supprimé puis réinséré... supprime les conducteurs" — exactly the warning path.

Qt 6.10.2, builds clean, ctest 6/6.

Issue #734, opened 2026-08-13
Conductors render a short self-retracing spur at docked terminals (0.100.1-dev r9269) — also in the shipped affuteuse example


philippeagray:
**QElectroTech's own shipped example project exhibits this defect.**
`affuteuse_modern.qet` — authored in the GUI, distributed with QET,
never modified by us — exports with **30 self-retracing conductor
paths**, including small pure out-and-back ticks (e.g. a 3-segment,
zero-width vertical at (498.8, 333.8)–(498.8, 341.2) pt). We found the
problem in files we generate, but it is not a property of generated
files; it is visible in QET's own content.

## Summary

Many conductors render with a short out-and-back excursion at one end:
the path leaves the terminal, doubles back over itself, then continues.
The excursion is 3–5 px (~1 mm at our sheet scale), perpendicular or
parallel to the run depending on the terminal pair. Electrically the
connection is correct, but it reads as an accidental tick, loop or hook
and blocks use on an issued drawing.

Example path, taken from the SVG export:

```
M637,565 L630,565 L630,560 L630,565 L615,565
                   ^^^^^^^^^^^^^^^^ out and back
```

## Reproduction

Attached via gist (GitHub's issue API doesn't accept binary file
uploads, so the repro files are hosted there instead of inline):

**https://gist.github.com/philippeagray/7ff74ae434735fd459576b9857e1fdb8**

1. `qet_bug_repro_resaved.qet` — **primary**. This is the file *after*
   QET's own `--resave`, i.e. written by QET itself.
2. `qet_bug_repro.qet` — the original, for comparison.

(An annotated screenshot of the three artifact classes will be added
as a follow-up comment/attachment.)

Export either headless:

```
qelectrotech --export-pdf qet_bug_repro_resaved.qet out.pdf
```

Three artifacts appear, one per connection class:

| PDF location (pt) | Class | Path as exported |
|---|---|---|
| 258.8, 363.8 | series device → bus stub | 5 segments, 2.25 pt lateral loop |
| 461.2, 701.2 | side tap (tee → device) | 4 segments, doubles back vertically |
| 483.8, 828.4 | series device → device | 5 segments |

Also reproducible with **no custom files at all**:

```
qelectrotech --export-pdf affuteuse_modern.qet out.pdf
```

(the shipped example) shows 30 such paths.

## The file is canonical QET output

`--resave` makes QET parse the reproduction and rewrite it in its own
[...]

PR #890, opened 2026-09-16
Fix bugtracker #734: stop routing a bridge conductors don't need

ispyisail:
Refs #734, following up on my own root-cause comment there. `generateConductorPath()`'s cas "3"/"4" branches each insert a two-point bridge along one axis, at a coordinate on the other axis computed as the midpoint of `depart`/`arrivee` and snapped to the routing grid by a loop that walks it strictly downward until it divides evenly.

When `depart` and `arrivee` already agree on the axis the bridge would run along, no bridge is needed at all -- the midpoint starts on the correct, already-shared value. But the snap can still walk it off that value, or, when it's already on-grid, the two bridge points just duplicate `depart`/`arrivee` outright. Either way: an unnecessary out-and-back excursion or small looping detour at a join that needed neither.

## Fix

Skip the bridge in exactly that degenerate case, per branch, on the axis that branch actually bridges on. `descendant` and `montant` are not mirror images of each other in which axis each `cas` guards on -- each site says so in a comment.

The four guards now compare `qRound()`ed coordinates, matching how the
bridge coordinate itself is computed -- an exact `!=` would miss a pair
already grid-equal after rounding but off by a sub-pixel remainder, and
still route a degenerate bridge for it. Made no difference on the shipped
corpus (every coordinate there already lands exactly on-grid), verified by
diffing per-file counts before/after: identical.

Added `tests/qttest/tst_conductorselfretrace.cpp` (fixture:
`qet_bug_repro_resaved.qet`, the report's own canonical reproduction):
exports it via the built binary's `--export-svg` and asserts no conductor
path is self-retracing. Confirmed it actually catches the bug, not just
passes vacuously -- reverted `conductor.cpp` to master and reran: fails,
1 self-retracing path found.

## Tests

Qt6, build clean, ctest 7/7 (was 6/6; the new regression test above is the
7th).

- **The report's own canonical reproduction** (`qet_bug_repro_resaved.qet` from the linked gist): 1 self-retracing path -> 0.
- **Full corpus, all 24 shipped example projects** (stored `<segment>`s, as shipped, not regenerated): 74 -> 65 self-retracing conductor paths. Only two files changed, both improved; zero files got worse.

**One disclosed trade-off**, found while checking for regressions rather than assumed away: in `schema_unifilaire_voltaique2.qet`, 8 terminal pairs closer together than twice the extension length (their docked stubs already cross before any bridge is even considered) go from an existing small rectangular-loop artifact to a straight out-and-back retrace over the full gap -- same count of defective paths (8 -> 8), a different shape, no connectivity change either way. Not chased further here; a proper fix for that narrower "crossed stubs" case looks like separate work.

**The shipped `affuteuse_250h.qet`'s own 12 self-retracing paths are unchanged** by this fix (verified by stripping all 185 stored `<segment>` elements to force full regeneration). They have a structurally different point-count signature -- a 4-point back-and-forth reachable from `cas "1"/"2"`, not the `cas "3"/"4"` grid-snap bridge this fixes -- consistent with what the issue thread already flagged as a separate, undiagnosed mechanism.


scorpio810:
Thanks, the change is small, well targeted and the reasoning is sound: when depart and arrivee already share the coordinate the bridge is computed on, the midpoint can only be moved off or duplicated, so skipping it is correct. The comments explaining why descendant and montant guard on different axes are much appreciated.

A few remarks:

1. In the descending cas "4", `depart.y() != arrivee.y()` is always true (descending requires `depart.y() < arrivee.y()`). Fine to keep for symmetry as you said, just noting it's dead code.
2. The guards use exact `!=` on qreal. Since the bridge coordinate is computed with `qRound()`, comparing rounded values (`qRound(depart.x()) != qRound(arrivee.x())`) would be more consistent and would also catch nearly-aligned cases that still produce a degenerate bridge.
3. The `schema_unifilaire_voltaique2.qet` trade-off: a straight out-and-back retrace overlapping itself is arguably harder to spot for the user than a small loop. I'll check it visually before merging.
4. It would be great to add a regression test based on `qet_bug_repro_resaved.qet` so this doesn't come back.
5. Since `affuteuse_250h.qet`'s 12 paths come from a different mechanism, #734 isn't fully solved yet. Could you change "Fixes #734" to "Refs #734" so the issue stays open when this is merged?

Thanks again!

ispyisail:
Addressed 2, 4, 5.

**2.** All four guards now compare `qRound()`ed coordinates, matching how the bridge coordinate itself is computed. No change on the shipped corpus (every coordinate there already lands exactly on-grid) — verified by diffing per-file counts before/after the change: identical.

**4.** Added `tests/qttest/tst_conductorselfretrace.cpp`, fixture `qet_bug_repro_resaved.qet` (your canonical repro). It exports via the built binary's `--export-svg` and asserts no self-retracing path. Confirmed it isn't a vacuous pass: reverted `conductor.cpp` to master and reran — fails, 1 self-retracing path found. ctest is now 7/7.

**5.** PR title/body changed from "Fixes #734" to "Refs #734" so the issue stays open.

**1 and 3** left as-is — 1 is dead code kept for symmetry by agreement, and 3 is your own visual check to make. Thanks for flagging both.

Issue #882, opened 2026-09-15
CLI: add a global safety net against modal dialogs in headless mode + CI timeout test


scorpio810:
@ispyisail, you already know this code well from #661, 9285d123 and #839, so I'd like to assign this one to you.

### Context
Headless verbs (`--info`, `--resave`, `--export-*`, `--set-titleblock`, `--check-elements`) rely on `QET::QetMessageBox::setNonInteractive(true)`. If a modal dialog is shown while no one can answer it, the process hangs forever in `QDialog::exec()`.

This has been fixed call site by call site:
- e3d11a49 (#661): message boxes routed through `QetMessageBox`
- 9285d123: last load-path message box routed through the wrapper
- 20a0a888 (#839): `checkInsufficientRowsCount()` had been missed

### Current state (audit of master, 2026-09-15)
I went through all direct `QMessageBox::` statics, `exec()`, `QFileDialog::get*`, `QInputDialog::get*` and `QProgressDialog` calls. **None of them is currently reachable from `CLIExport::run()`.** A few look like traps, but are GUI-only paths:
- `QETProject::importElement()` (`ImportElementDialog`): only called on element drag & drop
- `PotentialSelectorDialog` (`ConductorAutoNumerotation`, `LinkElementCommand`): only on conductor creation or interactive linking
- `WiringListExport::toCsv()` and `ConductorNumExport::toCsv()`: the CLI uses `toCsvString()` / `wiresNum()` instead

The weak point is that protection depends on every contributor remembering to use `QetMessageBox`. #839 shows that one call site is easy to miss, and a single new `QMessageBox::warning()` or `QDialog::exec()` in the load/export path will bring the hang back.

### Proposal

**1. Global safety net in the CLI branch of `main.cpp`**
Add an application-wide event filter that rejects any `QDialog` shown in headless mode and reports it on stderr:
```cpp
class HeadlessDialogKiller : public QObject
{
protected:
	bool eventFilter(QObject *obj, QEvent *ev) override
	{
		if (ev->type() == QEvent::Show) {
			if (auto *dlg = qobject_cast<QDialog *>(obj)) {
				QTextStream(stderr) << "Unexpected dialog in headless mode: "
				                    << dlg->metaObject()->className() << " \""
				                    << dlg->windowTitle() << "\" -- rejected\n";
				QMetaObject::invokeMethod(dlg, "reject", Qt::QueuedConnection);
			}
		}
[...]

PR #891, opened 2026-09-16
Add JavaScript scripting: read, export, edit geometry, undo, navigate

### LINKED ISSUE #162: Feature Request: CLI for scripts

ispyisail:
Fixes #162. A JavaScript scripting surface, built on QJSEngine rather than an embedded Python interpreter -- no new toolchain to package (QJSEngine ships in every Qt SDK QET already targets, via the Qml module, probed the same non-fatal way QtPdf already is), no GIL, no version pinning, automatic reflection of the QObject-derived core classes' own methods with no hand-written binding layer.

## Two entry points

- `qelectrotech --run script.js project.qet` -- headless/CI, matching the original use cases in the issue (export derived files after a revision, drive it from CI).
- **Projet > "Exécuter un script..."** -- an interactive macro against the currently open project, via a real file-open dialog.

## What a script can do

A single global, `qet`:

- **Read the model**: project title, file path, folio count/titles, element/conductor counts per folio.
- **Export**: pdf/png/svg/cables/wires/bom/wiring/nets/links/info, plus set-titleblock and save -- thin wrappers around `CLIExport::run()`, reusing its already-tested logic rather than duplicating it.
- **Edit geometry, through the real undo commands**: `addElement`/`setElementPosition`/`moveElement`/`deleteElement`, using the same `AddGraphicsObjectCommand`, `QPropertyUndoCommand`, and `DeleteQGraphicsItemCommand` the GUI itself uses -- Ctrl+Z undoes a script's edits exactly as it would the equivalent manual ones.
- **Undo/redo**: `undo`/`redo`/`canUndo`/`canRedo` against the project's real `QUndoStack`, the same one Ctrl+Z is wired to.
- **Navigate and message**: `selectElement`, `deselectAll`, `zoomFit`/`zoomToContent`/`zoomReset`, `showMessage`.

Deliberately **not** exposed: triggering an arbitrary menu action by name. A script that could invoke any `QAction` could just as easily open a modal `QDialog::exec()` with nobody there to dismiss it -- the same hang class investigated for #882. Every method here is either non-blocking by construction, or (for `showMessage`) safe under `QET::QetMessageBox`'s existing non-interactive mode, already active for headless runs.

## Two real bugs caught by testing, not assumed away

1. **`save()` didn't actually save geometry edits.** It inherited the same reopen-from-disk pattern as every export method -- opened a second, unmodified copy of the project and rewrote *that*. An `addElement()` call counted correctly in memory and then silently vanished from the saved file. Caught by checking the saved XML directly. Fixed: `save()` now writes `m_project->toXml()` directly, the only method touching the live instance rather than a fresh reopen.
2. **Consecutive position changes didn't behave as two independent undo steps.** Traced to `QPropertyUndoCommand::mergeWith()` (pre-existing, correct): consecutive commands on the same object+property merge when their `text()` also matches, which `setElementPosition`/`moveElement` always produce for a given element -- the same collapsing dragging an item repeatedly gets. Not a bug in the new code, a wrong assumption in the first test. Re-verified against the merge-aware prediction and it landed exactly right, confirmed against the saved file's actual XML, not just in-memory counters.

## Optional dependency, not a hard requirement

`Qt::Qml` is probed the same way `QtPdf` already is: `QUIET`, non-fatal, behind a `QET_HAS_SCRIPTING` compile definition. A build without it compiles and links identically; the CLI flag and menu action are simply absent, or compile to a clear "not available" stderr message rather than disappearing silently.

## Tests

Qt6, build clean from a fresh reconfigure, ctest 6/6.

- Headless: read-model calls, export calls (a real PDF confirmed with `file`), the full add/select/move/undo/redo/delete/save cycle -- verified against the saved XML's actual content, not just in-memory counters.
- Corpus: the read-model script re-run against all 24 shipped example projects, 0 failures.
- `zoomFit` confirmed to return `false` headless (no view to act on) rather than silently no-op unnoticed.
- GUI: "Exécuter un script..." opened a real `*.js`-filtered file dialog; running a script against the live project marked it `[modifié]` in the title bar -- the same change-tracking path a manual edit goes through, consistent with the undo command landing on the project's real stack rather than a side channel.
- Error paths: a thrown script exception reports file:line:message and exits 1; missing arguments exit 2; a missing project file is reported and doesn't hang.

One real CMake ordering bug caught and fixed along the way, documented in the second commit: the new source files were silently excluded from the build the first time, only caught by an undefined-symbol link error, not by "it compiles."


ispyisail:
https://github.com/qelectrotech/qelectrotech-source-mirror/wiki/scripting

https://github.com/qelectrotech/qelectrotech-source-mirror/wiki/cli_reference

https://github.com/qelectrotech/qelectrotech-source-mirror/wiki/api_reference

scorpio810:
Btw, another cli export tool: https://github.com/qelectrotech/qet-web

scorpio810:
Thanks @ispyisail, the scope is well thought out: no arbitrary menu actions, optional Qml dependency, and the two bugs you caught are nicely documented. It builds clean here (Qt 6.4 with Qml). Before merging, though, a few things need fixing.

1. **`addElement()` doesn't import the element into the project.** The GUI path (`DiagramEventAddElement::buildElement()`) calls `project->importElement()` first. The script creates the element straight from `common://...`, so the saved .qet references a definition that isn't embedded, and the element goes missing on another machine. Please do the same import as the GUI. Also, the editing methods should refuse when `m_project->isReadOnly()`.

2. **`save()` bypasses `QETProject::write()`.**
   - It writes with a plain `QFile`, whereas `QET::writeXmlFile()` uses `QSaveFile`, so an interruption can corrupt the .qet.
   - It ignores read-only mode.
   - It leaves the project marked modified and doesn't update `saveddate`/`savedtime`.
   - From the GUI, it also silently writes the user's unsaved edits.

   Suggestion: use `m_project->write()` when no path is given, and `QET::writeXmlFile()` otherwise.

3. **Exports from the GUI open a second `QETProject` on the same file.** `CLIExport::run()` does this, and backups are only disabled on the headless path. Could you check this doesn't leave a stray `KAutoSaveFile` or trigger a restore prompt on next launch? Also worth a clear note in the API doc: `addElement()` followed by `exportPdf()` exports without the new element unless you save first.

4. **No timeout.** `while(true){}` freezes the GUI, or hangs a CI job forever with `--run`. `QJSEngine::setInterrupted()` is callable from another thread, so a watchdog timer could stop runaway scripts.

Smaller things:
- Wrap `runOnProject()` in `undoStack()->beginMacro()`/`endMacro()` so one Ctrl+Z undoes the whole script, including a script that throws halfway.
- In the GUI, script errors only go to stderr, which is invisible on Windows. A `QetMessageBox` on failure would help.
- The new strings need adding to the .ts files.

Could you have a look at 1-4? Happy to review again after.

ispyisail:
Pushed fixes for 1-4.

1. **addElement() import.** Now goes through `QETProject::importElement()`, same as the drag-from-collection-panel path. While wiring it up: `ElementsLocation::setPath()` forces *any* path to `embed://` once a non-null project is passed, so the existing `ElementsLocation(locationPath, m_project)` silently broke every `common://`/`custom://` call before this fix ever ran — confirmed by actually calling it, got "does not resolve to an element" on a real file. Fixed by only passing `m_project` for already-`embed://` paths. Also refuses on an import-name-collision that would otherwise reach `importElement()`'s own modal dialog, and on a read-only project (all four editing methods now check `isReadOnly()`).

2. **save().** Empty path now goes through `QETProject::write()` (read-only handling, saveddate/savedtime, `QSaveFile`); an explicit path goes through `QET::writeXmlFile()` directly, per your suggestion.

3. **Second QETProject on export.** Confirmed the risk: backups are now disabled around just that temporary project's `CLIExport::run()` when called interactively, then restored — headless `--run` is untouched since it needs backups off for the whole run.

4. **Timeout.** Watchdog thread calls `setInterrupted()` after 30s. First version used `sleep_for()` and blocked *every* run — including instant ones — for the full 30s on `join()`; caught by testing a one-line script, fixed with `wait_for()` on a condition variable.

Smaller: GUI errors now also show a `QetMessageBox`. Ran `update_translations` — picked up 37 strings this branch hadn't added yet.

**Not applied:** the `beginMacro()`/`endMacro()` wrap. `QUndoStack` ignores `undo()`/`redo()` calls while a macro is open, so wrapping the whole script run would turn a script's own `qet.undo()`/`qet.redo()` into silent no-ops for the run's duration — that's an already-shipped capability, explicitly asked for earlier in this feature's design. Happy to revisit if you see a way to keep both.

Qt 6.10.2, builds clean, ctest 6/6. Ran each fix against a real project — details in the commit message.

scorpio810:
Thanks for the quick follow-up on the review, the code side of 4a835213 looks good to me: the non-fatal `Qt::Qml` probe with the always-compiled stub, the `wait_for()` watchdog, `save()` going through `QETProject::write()`, and `addElement()` importing into the embedded collection are all fine.

One blocker before merging though: the `lang/*.ts` changes.

Running `update_translations` rewrote all 35 .ts files (+27k/-23k lines). Most of it is just line-number churn, but I compared against master (2eed3acd) and lupdate marked a number of **existing, still-used translations** as `type="vanished"`:

- `sources/qetgraphicsitem/ViewItem/ui/graphicstablepropertieseditor.ui` (29 strings)
- `sources/ui/inditextpropertieswidget.ui` (12 strings)
- `sources/qetgraphicsitem/ViewItem/ui/projectdbmodelpropertieswidget.ui` (3 strings)

These three .ui files still exist and are valid XML, but they no longer have any `<location>` in the new .ts files. Vanished count per language went from 0 on master to ~38 (de, it, nl, ru, uk, zh, en, sv, pt_BR...), 43 (ca, pl), 3 (fr). Since lrelease drops vanished entries, those dialogs would lose their translations in about twenty languages. I don't know why lupdate skipped those files in your run, but the result isn't safe to merge.

Could you please drop the translation changes from this PR?

```
git checkout master -- lang/
git commit -m "Revert .ts changes, translations will be updated separately"
```

I'll run lupdate myself in a separate commit after the merge, so the 37 new strings get picked up without losing anything, and it avoids conflicts with the translation work currently in progress.

Optional, not blocking: there is currently no way to force or disable scripting from the command line, it only depends on whether Qt Qml is found. A `QET_ENABLE_SCRIPTING` option (e.g. AUTO/ON/OFF, where ON fails at configure time if Qml is missing) would make packaging builds more predictable. Also, the fixed 30s timeout applies to interactive runs too, so a legitimately long script (large project export) would be interrupted, and the error message doesn't say it was a timeout. Could be improved in a follow-up.

Thanks again for the work on this!

ispyisail:
Reverted `lang/` to master exactly (`git checkout upstream/master -- lang/`, verified `git diff HEAD~1 -- lang/` is empty). `sources/` untouched. Good catch — didn't think to diff the vanished entries, just checked that lupdate found the new strings.

Left the two optional items as follow-ups rather than folding them in here, since you flagged them as non-blocking and this PR's already had enough scope creep from my own review fixes:
- `QET_ENABLE_SCRIPTING` (AUTO/ON/OFF) for predictable packaging builds
- Timeout: make it apply less aggressively to interactive runs and say "timed out" rather than just "Interrupted"

Happy to take either in a follow-up PR once this lands, or here if you'd rather bundle it.

scorpio810:
Thanks, I checked e62d43d0: `lang/` is identical to master now, and the PR only touches the scripting code and CMake. Perfect.

Agreed on the two optional items, let's keep them for a follow-up PR once this one lands, this one is big enough already.

One more thing before I can merge: master moved on since your branch point, and the PR now conflicts in `sources/qetdiagrameditor.cpp` and `sources/qetdiagrameditor.h`, with the "Reload element drawings" changes from #802 (43d27a95 and my follow-up fix 9004db1d). Could you rebase on current master (or merge it) and resolve the conflict? It's probably just both PRs adding a menu action in the same place, but please make sure my fix from 9004db1d is kept.

Thanks!

ispyisail:
Merged master and resolved the conflict — both were just additive (each PR adding its own menu action/slot next to the other's). Kept your `slot_reloadElementDrawings()` from 9004db1d exactly as-is, byte for byte; the diff is purely `<<<<<<<`/`=======`/`>>>>>>>` markers removed, both functions kept in full.

Builds clean, ctest 7/7 (picked up `tst_qetpalette` from master too). Re-ran the scripting smoke test post-merge to confirm nothing regressed. `mergeable=MERGEABLE`/`CLEAN`.

PR #894, opened 2026-09-16
Give Fusion a palette it can draw with on macOS

jp2images:
Fixes the low-contrast controls on macOS in both light and dark mode. Refs #870, #467, #553, and the forum thread [macintosh dark mode problems](https://qelectrotech.org/forum/viewtopic.php?id=3187).

*Version française plus bas.*

### Problem

`main.cpp` has forced the Fusion style on macOS since 2019, but the palette still comes from Qt's macOS platform theme, which is built for the native style. It hands Fusion a Window, Button and Base that are the same color, a Dark lighter than Light, and in dark mode an Inactive ButtonText of black. Fusion derives its frames, gradients and indicators from those roles, so:

- fields and buttons have no edges;
- the radio buttons in the text alignment dialog vanish;
- the "Handles" combo box in the diagram editor toolbar draws its text black on a dark combo the moment the window loses focus, which is exactly when a screenshot is taken.

Light mode has the same flatness, with Window, Button and Base all white.

### Change

- New `QET::Palette` (`sources/qetpalette.{h,cpp}`) with a light and a dark palette laid out the way Fusion expects. Every text role pair reads at WCAG 4.5:1, disabled text at 3:1, and Inactive equals Active. The platform's accent color is kept when it reads at 4.5:1 against its highlighted text.
- `QETApp::initStyle()` installs it on macOS when the running style is Fusion, choosing light or dark from the platform palette's lightness. On Qt 6.5 and newer the app follows the OS light/dark switch through `QStyleHints::colorSchemeChanged`.
- Two small readability fixes that the new palette exposed: the Selection properties "texts" tab painted the color value's own text in that color, so black was unreadable on dark. It now shows a swatch. The terminal plan preview draws black ink on a white brush like the printed page, so its view gets a white background.

### What "Use system colors" means on macOS after this

On macOS the base palette is now always the QET Fusion palette, because the system one cannot be drawn by Fusion. The checkbox keeps one job: unchecked, the user's `style.css` is layered on top; checked, it is not. This is the setting #467 reported as having no visible effect. Nothing changes on Linux or Windows: the platform palette stays in effect there, and the checkbox behaves as before.

### Scope

macOS only. Fusion is also Qt's default style on Linux desktops without a platform theme, and the palette there carries the user's desktop colors, so this change does not touch it. Whether Fusion and this palette should become the default on every platform is one of the questions in #870.

### Testing

[...]

Issue #413, opened 2026-01-04
Copy-pasting conductors without labels adds a `_` label


p-himik:
QElectroTech V 0.100.0-dev

Steps to reproduce:
1. Add any two elements with terminals to the folio
2. Join some of their terminals with a conductor
3. Select and copy the elements together with the conductor
4. Paste

Expected result: the pasted content is exactly the same as the copied one.

Actual result: underscores are added to each copy-pasted conductor.

A screenshot of an example - the original is on the left, the pasted copy is on the right:

<img width="441" height="296" alt="Image" src="https://github.com/user-attachments/assets/0e0b7d58-288e-46b1-9ac2-c2cd9eecfb8c" />

Here's the copied content for the above image. Notice how all conductors have `num=""`, so no label.
If you copy the pasted content again and examine its XML, all conductors will have `num="_"`.

```xml
<diagram projectId="0">
    <elements>
        <element z="10" uuid="{35959da5-c61a-4311-8023-3b1839a514a8}" orientation="0" type="embed://import/10_electric/10_allpole/140_connectors_plugs/20_socket_outlets/pc_mono_armoire.elmt" y="1850" freezeLabel="false" x="1770" prefix="X">
            <terminals>
                <terminal id="0" orientation="0" y="-13" x="-6"/>
                <terminal id="1" orientation="0" y="-13" x="6"/>
                <terminal id="2" orientation="2" y="13" x="0"/>
            </terminals>
            <inputs/>
            <elementInformations/>
            <dynamic_texts>
                <dynamic_elmt_text rotation="0" font="Liberation Sans,9,-1,5,50,0,0,0,0,0,Regular" Valignment="AlignTop" uuid="{786bbb3f-feaf-498f-850b-58d66e79fd79}" keep_visual_rotation="false" frame="false" y="-19.95" text_width="-1" text_from="ElementInfo" x="11" Halignment="AlignLeft">
                    <text></text>
                    <info_name>label</info_name>
                </dynamic_elmt_text>
            </dynamic_texts>
            <texts_groups/>
        </element>
        <element z="10" uuid="{659fa417-c6c4-4f06-a08f-57f8258fafeb}" orientation="0" type="embed://import/10_electric/10_allpole/140_connectors_plugs/20_socket_outlets/pc_mono_armoire.elmt" y="1850" freezeLabel="false" x="1730" prefix="X">
            <terminals>
                <terminal id="3" orientation="0" y="-13" x="-6"/>
[...]

PR #895, opened 2026-09-16
Clean up stale data when duplicating diagram pages

Kellermorph:
When duplicating a diagram page via the project tree context menu, copied slave elements retain stale data from the source diagram — labels, descriptions, link references, and PLC master information (type, address, function, comment, cross-ref, timer values) all remain in the copies.

Changes:

Add clearPendingLinks() method to Element — clears tmp_uuids_link and m_group_index_map so copies don't link back to source elements via stale UUIDs.
In ElementsPanelWidget::duplicateDiagram(), clean up copied element data after fromXml():
Slaves always lose their label, formula, comment, location, and PLC master data, since their text comes from a master element which is not available on the copy. The displayed text on slave elements is cleared directly via setPlainText() because the text items are not connected to any element after copying (no linked master).
Non-slave elements respect the existing "erase-label-on-copy" preference, same as PasteDiagramCommand::redo().
Text alignment is preserved by wrapping setElementInformations() with m_block_alignment, same as Element::fromXml().
Builds on top of #894 (fix-copy-page).


scorpio810:
Thanks Kellermorph, this is a useful cleanup: stale pending links, PLC master data and labels on duplicated pages were a real problem. `clearPendingLinks()` is a good defensive move (`fromXml()` doesn't call `initLink()`, so stale uuids would otherwise stay pending), and blocking the alignment around `setElementInformations()` matches `Element::fromXml()` nicely.

A few remarks:

1. **`setPlainText(QString())` on all dynamic texts of slaves**: this also clears `UserText` items (free text typed by the user, not linked to the master). Only the display is cleared, `m_text` is still saved, so the text disappears on screen and comes back after reopening the project. Could you restrict it to non-user texts, e.g. `if (deti->textFrom() != DynamicElementTextItem::UserText)`?
2. The comment says "matching PasteDiagramCommand::redo()", but there PLC data is only removed when "erase-label-on-copy" is enabled, while here slaves are always cleaned. That makes sense for slaves, but could you update the comment so it reflects the actual difference?
3. Thanks for taking the "erase-label-on-copy" checkbox into account as I asked. To go all the way with it: `PasteDiagramCommand::redo()` also resets the conductor texts when this option is enabled (#413). Could you do the same in `duplicateDiagram()`, so the duplicated page behaves exactly like a paste regarding that setting?
4.  Minor: `QSettings` is read for every element, it could be read once before the loop.

Kellermorph:
Thanks for the thorough review! All four points have been addressed:

1. UserText items preserved — setPlainText() now checks deti->textFrom() != DynamicElementTextItem::UserText before clearing, so free text typed by the user is kept.

2. Comment updated — The comment now clearly describes the difference from PasteDiagramCommand::redo(): slaves are always cleaned regardless of the "erase-label-on-copy" setting, while non-slaves honour it.

3. Conductor text reset — duplicateDiagram() now resets conductor labels (cp.text = "") when "erase-label-on-copy" is active, matching PasteDiagramCommand::redo() and fixing issue #413.

4. QSettings read once — Moved outside the per-element loop, read once before iteration.

Bonus fix for PLC slave text positions — For PLC slaves, elementInfoChanged() fires during setElementInformations() (because elementUseForInfo() returns self). I originally blocked finishAlignment() with m_block_alignment during that call, which left wrong positions for the now-empty texts. Fixed by not blocking alignment for slaves during setElementInformations(), and instead only blocking it during the subsequent setPlainText() clearing pass.

scorpio810:
Thanks again @Kellermorph, this is a solid fix. `clearPendingLinks()` closes a real hole (`fromXml()` fills `tmp_uuids_link`, but `duplicateDiagram()` never calls `initLink()`), and giving conductors a new uuid is a great catch: a duplicated primary key means a wire silently missing from the wiring list.

A few follow-up points after re-reading the merged code:

1. **Formula on duplicated pages.** Since `erase-label-on-copy` defaults to `true`, duplicating a page now clears `formula` and `label` on every non-slave element, masters included. When duplicating a whole page, keeping the formula is usually what users want: a label like `%f-%l%c` recomputes itself for the new folio. Could you check what `PasteDiagramCommand::redo()` does with the formula (and with `location` / `comment`)? If paste keeps the formula, duplication should probably do the same and only clear the label, so it gets re-evaluated. Otherwise, someone duplicating a page to reuse it ends up with all element labels empty.

2. **CI.** No checks had run on this PR when it was merged (branched from an older base). @scorpio810, could you confirm the next master build is green on all platforms (Linux, Windows MinGW, macOS)?

3. **Direct access to `m_block_alignment`.** `ElementsPanelWidget` now writes `DynamicElementTextItem::m_block_alignment` directly. It works, but a small setter or an RAII guard (like in `Element::fromXml()`) would keep that detail inside the class.

4. **Line-number reference in a comment.** The comment "same as Element::fromXml() (line 890-896)" will go stale at the next edit of `element.cpp`. Referencing the function name alone is enough.

5. **Test project.** A small test project with a master, a simple slave, a PLC slave, a free `UserText` and a conductor with a label would make it easy to check duplication behaviour manually before releases.

Point 1 is the only one that can change user-visible behaviour; the others are cleanup. Happy to open an issue for point 1 if that's easier to track.

Kellermorph:
Thanks for the suggestions! I'll keep them in mind, but I'd like to focus on the two open bugs first before moving on to cosmetic cleanups:

There's still a bug with the normal copy-paste behaviour
The PDF export checkboxes aren't being saved
Once those are sorted I can circle back to the refactoring items (m_block_alignment setter, line-number comment, test project).

PR #897, opened 2026-09-17
Per-element IEC 81346 designation letter + revive live placeholder

ispyisail:
Adds a per-element designation-letter override and revives the live placeholder label from #668, now compatible with it.

The per-element approach here was proposed independently by @VonkeBoer in #666, who has built an equivalent field (`IECClass`) of his own — we're reconciling the two in that thread, and I'm happy to align on his naming.

Discussion #666: an element's own `designation_letter` attribute (set once in the Element Editor's Type tab) now overrides the category-derived prefix from `qet_labels.xml` for every placed instance — most specific wins, since a category can't express "this one element in a mixed folder is different." Verified: an overridden element gets `prefix="ZZ"` when placed; an unmodified element still gets its category prefix (e.g. `"F"`) unchanged.

Also revives #668 (closed, unmerged): `DynamicElementTextItem` shows `"<prefix>?"` instead of nothing when an element has no label formula or manual label yet — display-time only, never written into `elementInformations`. #668's original fix reordered `setPrefix()` before `buildFromXml()` in the constructor; that's incompatible with the new field, since `designation_letter` isn't populated until `m_data.fromXml()` runs *inside* `buildFromXml()`. Fixed instead by resolving the prefix inside `buildFromXml()`, right after `fromXml()` and before any text item is constructed — verified an overridden element renders `"ZZ?"` on first paint.

Two follow-up commits:

- An empty saved `prefix=""` no longer overwrites the freshly-resolved letter. This bites a project saved by a build that didn't know about `designation_letter`: the embedded definition carries the attribute, the placed instance carries `prefix=""`, and re-opening resolved the letter then immediately discarded it. A non-empty saved prefix still wins, so existing drawings keep their designation. Safe because `setPrefix()` has no UI path — an empty saved value always means "nothing resolved when this was saved", never "the user cleared it".
- The letter field is restricted to uppercase letters (validator + uppercase on edit). Typing `kf3!x` now yields `KFX`.

**Scope limit worth knowing before review:** an element already placed in a saved project does *not* pick up a `designation_letter` added to the shared collection afterwards. Placing an element copies its definition into the project (`embed://import/...`), and that snapshot is what `buildFromXml()` reads. So the companion element-collection PRs (qelectrotech/qelectrotech-elements#79, #80) only affect newly-placed elements. Making library updates propagate into existing projects is a separate problem, not attempted here.

In passing, fixed a self-comparison bug in `ElementData::operator==()` that made edits to the "Informations" field alone never register as a change.

Qt 6, build clean, ctest 8/8. Verified headlessly via `--run` scripting against real `.elmt`/`.qet` XML, and the editor field verified in the real dialog under Xvfb.


plc-user:
Do you think the "designation_letter" is right in definition-line of an element?
In my opinion, it's more appropriate in the “elementInformations” section next to "elementInformation label=...".
Come to think of it: 
Isn't the “label” tag in “elementInformations” already designed specifically for that purpose?

ispyisail:
Good catch on the round-trip — you're right, and I hadn't checked it. ElementScene::toXml() rebuilds <definition> from a fixed attribute list, so an older QET that opens and re-saves an element in the Element Editor silently drops designation_letter. A key in elementInformations would survive, because DiagramContext reads and writes back keys it doesn't recognise. That's a real advantage of your placement.

On label though, I don't think it can do this job. 231 shipped elements already set label in their definition, and the values are things like EXT1, -F1, -BA, PS1, HMIET6400, 16DI — complete designations with the number already in them, not class letters. There are two mechanical problems too: %prefix reads Element::getPrefix(), not elementInformations, so a K in label never reaches %prefix%sequ_1; and PasteDiagramCommand clears label on copy when erase-label-on-copy is set, which is the default.

So the real question is a <definition> attribute vs a new key in elementInformations — not label. The trade as I see it: the attribute keeps it a property of the type, alongside type and link_type; a key in elementInformations round-trips through old versions but becomes per-instance state, since the definition's block is only a default that the placed element's own block replaces.

Which would you prefer? @VonkeBoer has built the same field independently (see #666), so it would be good if the three of us settled on one.

Happy to do your suggestion.......

plc-user:
When I look at the existing entries under “label,” almost all of them are used incorrectly!

```16DI
4-AI
4-AO
4-DO
-A
A
Actuator
-AE
AS1
-BA
Compteur d'énergie
Configurator ⇅
Configurator 0
Configurator 1
Configurator 2
Configurator 3
Configurator 4
Configurator 5
Configurator 6
Configurator 7
Configurator 8
Configurator 9
Configurator AUX
Configurator CEN
Configurator GEN
Configurator GR
Configurator ⇅M
Configurator OFF
Configurator O/I
Configurator ON
Configurator PUL
Configurator SLA
Dimmer
Dimmer 2
Double Switch 2
EL8 EC
EL8 RS
EXT1
-F1
Fibaro RGBW
Gateway
HMIET6400
Input meter
Memory
MS1
P
P1
-PG1
Power
PP1
PS1
Raspberry Pi Pico W
SICK UM 30
-SL
-SM
-SU
-SV
SW1
-T
Toroid
Web server
-X1
X1
-XX
```

The English term “label” corresponds to the German “Betriebsmittelkennzeichen BMK” (equipment identification code), and this can only be set properly in the diagram editor because when creating a generic element, you have no way of knowing where it will be used!

That's why I think the existing tag "label" would be the right one to store the prefix that overrides the one from qet_labels.xml.

Before making this change, we should delete the existing incorrect entries in the element files. Shouldn’t be a major problem for automated search-and-replace!

ispyisail:
> automated search-and-replace!

Don't encourage me. I can't stop hitting the AI button.

plc-user:
The definition line contains data for the graphic side of an element: width, height, hotspots, etc.

Information on the technical aspects of the element can be found under “elementInformations.” In my opinion, this should also include the specification for the BMK / label.

plc-user:
> Don't encourage me. I can't stop hitting the AI button.

I see that several times every day!  :stuck_out_tongue_winking_eye:

plc-user:
And: 
Label-Prefix can be more than one letter.
As far as I know it can be up to three letters for a device.
There is a German wikipedia-entry for that:
[https://de.wikipedia.org/wiki/EN_IEC_81346](https://de.wikipedia.org/wiki/EN_IEC_81346#Tabelle_der_Hauptklassen_(A1),_Unterklassen_(A2)_und_Unter-Unterklassen_(A3))
(show table with "Ausklappen")

plc-user:
What I said about the tag "label" in "elementInformations" section

> (...) can only be set properly in the diagram editor because when creating a generic element, you have no way of knowing where it will be used!

... also applies, in my opinion, to the tags "location" and "plant" in that section.

ispyisail:
Closing, not on merit — plc-user's review (#666, and here) surfaced real, unresolved design questions this needs before code: whether `designation_letter` belongs on the definition line or in `elementInformations`, that the existing `label`-as-prefix data is largely wrong to begin with, and that a prefix can be up to three letters per the IEC 81346 standard rather than one. Those need settling as a concept, not patched around in this diff.

Branch stays, so this can be reopened once that's worked out. Not withdrawn on merit.

PR #899, opened 2026-09-17
Fix paste problems

### LINKED ISSUE #898: New diagramevent 'diagrameventaddpaste'
Hi,
I’ve tested the new 'diagrameventaddpaste' (#878) and have a few observations:

- Pressing Escape does not terminate the event.
- If a command is executed while the 'diagrameventaddpaste' event is active (e.g., clicking a toolbar button or using the Ctrl+V shortcut), QET crashes.
  With other events (e.g., 'diagrameventaddelement'), the event terminates and the command is executed.
- if the selection is moved with the mouse after pasting, the elements are repositioned, but the conductors remain at the original insertion point (see image).
- a single element is not aligned to the grid after pasting.

Just a side note: I’m not sure how or when the new crash report is supposed to trigger, but the report appeared only once despite there being 10 or more crashes.

Image

arummler:
fixes #898 
a) paste
- conductor does not move
- pressing escape does not abort
- pressing anything during the process crashes the program instead of aborting
 
b) similar issue with the escape not working fixed for graphics items, texxt fields (and probably others)


ispyisail:
Reviewed — non-blocking, I'd merge this. I'd independently found the same two root causes (destructor emitting `finish()` into a handler that deletes the object being destroyed; `DiagramView` swallowing Escape), and you got there first and covered more. The `updatePath()` call in `moveTo()` is the conductor fix I misread — I took "moved after pasting" to mean the committed selection and couldn't find a mechanism; during-placement is the right reading.

One thing worth changing: `hasEventInterface()` tests for non-null, but the paste constructor has three early returns before `m_running = true`:

```cpp
if (clipboard_text.isEmpty()) return;
if (!document_xml.setContent(clipboard_text)) return;
if (!m_content.count()) return;
```

`setEventInterface()` installs the object anyway, so an empty or non-diagram clipboard leaves a *non-running* interface behind. Escape then diverts to a dead tool that ignores it, and `DiagramView`'s own Escape handling (clear selection, then release focus) is lost until something replaces the interface. Repro: copy nothing, Ctrl+V, Escape. `isRunning()` is already virtual on the base, so `m_event_interface && m_event_interface->isRunning()` covers it.

Two nits:

- The destructor dropped the `m_diagram` guard, and `removeItems()` dereferences it unguarded. It's a `QPointer<Diagram>`, and "the editor closed" — the case the original comment names — is when it can go null.
- `cancel(); event->setAccepted(true);` — `cancel()` emits `finish()`, which deletes `this`, so that line runs on a destroyed object. Benign today since no member is touched, but same pattern in `mouseReleaseEvent`.

I had a duplicate branch for this; closing it in favour of yours.

PR #905, opened 2026-09-17
Harden the crash reporter: keep every dump, and say what crashed

ispyisail:
Companion to #902, independent of it. #902 makes the crash report *appear*; this makes it worth reading when it does. They touch different files and can merge in either order.

Three weaknesses, all visible in @ChuckNr11's report on #898 — *"the report appeared only once despite there being 10 or more crashes"*.

**One dump per run, instead of one per install.** `crashDumpPath()` was a single fixed `crash_dump.log`, and the handler opens it `O_TRUNC` — so each crash destroyed the previous one's evidence. Ten crashes left one dump. Dumps now go to `crashes/crash_<timestamp>_<pid>.log`, and every pending one is offered together, newest first, with a count. A crash that repeats is exactly when the earlier dumps matter, because the difference between them is the evidence.

**The dump now says which signal fired.** The header is built once at `install()`, so every dump looked identical regardless of what killed the process — and SIGSEGV and SIGABRT point at very different bugs.

**And where it died.** The ring said what the program was doing; nothing said where. The dump now carries a backtrace.

The four invariants in `crashhandler.h` are deliberately untouched — nothing added here allocates, blocks, takes a lock, or swallows the crash. The filename is built in normal context and copied into the same preallocated buffer, so the handler still writes one fixed path. `backtrace()` is warmed once in `install()` because its first call can allocate; `backtrace_symbols_fd()` is the async-signal-safe variant (`backtrace_symbols()` mallocs and must never be used there). The signal number uses a small integer formatter into a stack buffer, since `snprintf` is not on the POSIX safe list. Guarded on `__has_include(<execinfo.h>)`.

Known limitation: QET's own frames resolve as offsets rather than names, because the binary doesn't export dynamic symbols. They remain resolvable offline — the header records the exact git SHA. `-rdynamic` would give names directly, but that's a build-flag decision that deserves its own change.

**Verified:** three consecutive SIGSEGVs leave three separate dumps, each carrying `Signal: 11` and a backtrace with resolved Qt frames; launching afterwards offers all three in one dialog, newest first, and clears them once shown.

Qt 6, ctest 8/8. Branched from `adca07f26` rather than current master, which doesn't configure here (`ff3c2e091` requires ECM ≥ 6.28.0; this box has 6.24.0).


scorpio810:
Thanks Shane, good diagnosis: the single `crash_dump.log` opened with `O_TRUNC` is exactly what explains ChuckNr11's report on #898, and the approach keeps the handler invariants in mind. I'd like to merge it, but a few things first.

**Before merge**

1. **`clearPendingCrashDump()` doesn't do what its comment says.** The doc says it only removes the dumps that were just offered, but it calls `pendingCrashDumpFiles()` again at clear time, so a dump written in between would be deleted unseen. Please pass the offered list in (e.g. `clearPendingCrashDump(const QStringList &offered)`), or keep it in a member when `pendingCrashDumpContents()` builds it.

2. **Write the ring before the backtrace.** The ring is the most valuable and safest part of the dump. `backtrace()` goes through the libgcc unwinder, which calls `dl_iterate_phdr` and takes the loader lock. Warming avoids the allocation, not the lock. If we crash during a `dlopen` (Qt plugin loading), or on a corrupted heap or stack, the handler can hang or fault again and we lose the ring. Order: header → signal → ring → backtrace. (So "nothing added here takes a lock" isn't strictly true.)

3. **FreeBSD link.** `__has_include(<execinfo.h>)` is true there, but `backtrace()` lives in `libexecinfo`, so the build will fail at link time. Could you use `find_package(Backtrace)` in CMake, link `${Backtrace_LIBRARIES}`, and gate `QET_CRASH_BACKTRACE` on `Backtrace_FOUND`?

4. **Legacy `crash_dump.log`.** A dump left by a previous version in `m_log_dir/crash_dump.log` will never be offered or removed after upgrading, which is precisely the "user updates after crashing" case. Please move it into `crashes/` at startup, or include it once in `pendingCrashDumpFiles()`.

5. **CI / base.** The branch is based on `adca07f26` and no checks have run. Please rebase on master so CI (Linux, Windows MinGW, macOS) can run.

**Nice to have**

- Cap the number of kept dumps (e.g. the 10 newest), in case of a crash loop before the dialog shows.
- A small ctest for `formatInt()` (0, negative, `INT_MIN`, short buffer) and for the pending list excluding the current run's own path.
- The timestamp in `crash_<timestamp>_<pid>` is the launch time, not the crash time. That's fine, but the commit message suggests otherwise.
- `crashDumpDir()` calling `mkpath` from a const getter is a bit of a hidden side effect.
- Please check that `redact()` masks the absolute paths `backtrace_symbols_fd` writes (`/home/<user>/…`, `/tmp/.mount_xxx/…` for the AppImage).

**Follow-up, not for this PR:** without `-rdynamic` and without archived debug symbols, QET frames won't be resolvable in practice for AppImage/Flatpak/Snap/Debian builds. Let's discuss that separately.

Thanks!

ispyisail:
All five done, plus the nice-to-haves.

1. **`clearPendingCrashDump()`.** You were right that it deleted whatever was in the directory at clear time. It now takes the offered list; `checkCrashDump()` lists once and passes the same list to both calls. Worth spelling out why it bites: SingleApplication keys its socket on the binary path, so a second build running alongside is a separate process that really can crash while the dialog is open.
2. **Ring before backtrace.** Done — header, signal, ring, backtrace. The class comment claimed the handler takes no locks; it now says which one it takes and why that call is last.
3. **FreeBSD.** `find_package(Backtrace)`, `QET_CRASH_BACKTRACE` gated on `Backtrace_FOUND`, `${Backtrace_LIBRARIES}` linked. The header probe is gone.
4. **Legacy `crash_dump.log`.** Migrated into `crashes/` at startup, named from its own mtime so it sorts correctly, then offered and deleted like any other.
5. **CI.** Merged master rather than rebasing, since you've asked before not to force-push open PRs.

Nice-to-haves: capped at the 10 newest, pruned at startup; `crashDumpDir()` no longer creates anything (`ensureCrashDumpDir()` does, for the callers that write); `redact()` now masks `/tmp/.mount_XXXXXX` as well as `$HOME`, which is what `backtrace_symbols_fd` puts in every frame.

Two test executables, each checked to fail against the behaviour it replaces:

- `tst_crashhandler` — `formatInt()`: zero, negatives, `INT_MIN`, `INT_MAX`, truncation, zero-sized buffer, each against a sentinel-filled buffer so a write past the reported length fails.
- `tst_crashdumps` — ordering, empty dumps, this run's own path excluded, the cap, concatenation, and that clearing deletes only what was offered. `qetlogger.cpp` needs exactly one symbol from the application, `QETApp::dataDir()`, which the test supplies.

You were right about the timestamp: it is the launch time. The commit message was what was wrong, not the code.

Verified on the merged branch: ctest 11/11, and a real `SIGSEGV` against the built binary gives a dump ordered header → `Signal: 11` → ring → backtrace. Legacy migration and the 10-dump cap checked end-to-end against a sandboxed `$HOME`.

Happy to take `-rdynamic` and archived debug symbols to a separate discussion.

PR #908, opened 2026-09-17
Refuse to close an editor while a modal dialog is running

### LINKED ISSUE #904: Quitting while the "Créer une copie de sauvegarde ?" dialog is open aborts in BackupDialog::~BackupDialog()
**Steps.** Open an existing project so that the backup-copy question appears. While it is showing, quit the application (application menu, Quit). The process aborts:

```
___BUG_IN_CLIENT_OF_LIBMALLOC_POINTER_BEING_FREED_WAS_NOT_ALLOCATED
BackupDialog::~BackupDialog()            backupdialog.cpp:63
QObjectPrivate::deleteChildren()
QWidget::~QWidget()
QETMainWindow::~QETMainWindow()          qetmainwindow.cpp:60
QETDiagramEditor::~QETDiagramEditor()    qetdiagrameditor.cpp:164
QObject::event(QEvent*)                  (DeferredDelete)
```

macOS 27, Qt 6.11.2, commit 90798402f (master plus an icon branch that does not touch this code). QET's own crash dump was written too.

**Cause.** `QETDiagramEditor::openAndAddProject()` creates the dialog on the stack with the editor as parent, after the project has been added to the editor:

```cpp
BackupDialog backup_dialog(this);
if (backup_dialog.exec() == QDialog::Accepted)
```

`exec()` runs a nested event loop. Quitting during it runs `QETApp::closeEveryEditor()`, the editor's `closeEvent()` closes the project, sets `WA_DeleteOnClose` and accepts, and the resulting `deleteLater()` is processed inside the nested loop. `~QWidget()` then deletes its children, the stack-allocated dialog among them, and `delete` on a stack object aborts. The same nested-loop hazard is what qetapp.cpp:1666 already guards against on the SingleApplication path; this is the local-quit variant. `QET::QetMessageBox` builds its boxes the same way (stack object, parent, `exec()`), so the font warning shown a few lines earlier has the same exposure.

**Possible fixes**, keeping the dialog on the stack:

- Refuse the close while a modal dialog is running: at the top of `QETDiagramEditor::closeEvent()`, `if (QApplication::activeModalWidget()) { qce->ignore(); return; }`. The quit is denied until the question is answered, and no object is created during shutdown.
- Or give the dialog no QObject parent (`BackupDialog backup_dialog;` with `setWindowModality(Qt::ApplicationModal)`), so no parent destructor can reach the stack object.

Allocating the dialog on the heap with `deleteLater()` would also avoid the abort, but it creates an object during shutdown only to delete it. A Qt Test for either fix: open a project, post a `QTimer::singleShot` that calls `QETApp::closeEveryEditor()` while the dialog's `exec()` runs, then check that the editor is still alive after the dialog returns.

---

## Version française

**Étapes.** Ouvrir un projet existant de sorte que la question « Créer une copie de sauvegarde ? » s'affiche. Pendant qu'elle est affichée, quitter l'application. Le processus s'arrête sur `pointer being freed was not allocated` dans `BackupDialog::~BackupDialog()`, appelé depuis `QETDiagramEditor::~QETDiagramEditor()` via un `DeferredDelete`. macOS 27, Qt 6.11.2.

**Cause.** `QETDiagramEditor::openAndAddProject()` crée la boîte de dialogue sur la pile avec l'éditeur comme parent, puis appelle `exec()`. Quitter pendant cette boucle d'événements imbriquée passe par `QETApp::closeEveryEditor()` ; le `closeEvent()` de l'éditeur ferme le projet, pose `WA_DeleteOnClose` et accepte, et la suppression différée est traitée dans cette même boucle. `~QWidget()` supprime alors ses enfants, dont l'objet sur la pile. C'est le même schéma que celui contourné dans qetapp.cpp:1666 pour SingleApplication. `QET::QetMessageBox` construit ses boîtes de la même façon (objet sur la pile, parent, `exec()`).

**Pistes**, en gardant la boîte sur la pile : refuser la fermeture tant qu'une boîte modale est active (`if (QApplication::activeModalWidget()) { qce->ignore(); return; }` en tête de `closeEvent()`), ou ne pas donner de parent QObject à la boîte (`setWindowModality(Qt::ApplicationModal)`). L'allouer sur le tas avec `deleteLater()` éviterait aussi le plantage, mais crée un objet pendant la fermeture pour le supprimer aussitôt. Un test Qt Test pour l'une ou l'autre correction : ouvrir un projet, déclencher `QETApp::closeEveryEditor()` par `QTimer::singleShot` pendant l'`exec()` de la boîte, puis vérifier que l'éditeur existe encore au retour.

ispyisail:
Fixes #904.

Closing an editor while a modal dialog is running is now refused, and the dialog is raised so the refused quit isn't silent. Without this, the close's `deleteLater()` is processed by the dialog's nested loop, `~QWidget()` deletes the stack-allocated dialog, and the process aborts.

- Done in `QETMainWindow::event()`, before `closeEvent()` — which starts closing projects before deciding whether to accept. Covers the diagram and title-block editors; `QETElementEditor` (a plain `QMainWindow`) calls the same helper from its `closeEvent()`.
- **Correction to my comment on #904:** I suggested guarding `QETApp::closeEveryEditor()`. That misses the reported route — File > Quit is connected to `QETDiagramEditor::close()`, not `quitQET()`. @jp2images's original placement was the right one.
- Not unparenting the dialog: the editor's own constructor calls `openAndAddProject()` and then `slot_updateActions()` on `this`, so that turns the abort into a silent use-after-free.

**Verified** on Linux by calling `close()` from gdb inside the dialog's loop (both `quitQET()` and the editor's `close()`): unfixed aborts with `free(): invalid size` under `~QETDiagramEditor()`, matching the report; fixed, `close()` returns false and Ctrl+Q works normally once the dialog is answered. The element-editor path uses the same helper but wasn't exercised on its own.

**Regression gate:** `tests/modal-quit-regression/run.sh --binary <qet>` — gdb + offscreen, no titles or window ids, ~10 s. Exit 1 on a build without this change, 0 with it.

ctest 8/8. Branched from `adca07f26`; current master doesn't configure here (`ff3c2e091` needs ECM ≥ 6.28.0).


scorpio810:
Thanks Shane, this one looks good. The fix is minimal and in the right place: guarding in `QETMainWindow::event()` before `closeEvent()` runs is correct, since `QETDiagramEditor::closeEvent()` starts closing projects before deciding whether to accept. Thanks also for correcting the `closeEveryEditor()` suggestion from #904; File > Quit indeed goes through `close()`.

I like that the guard checks `QApplication::activeModalWidget()` globally. It may look broad, but it keeps `quitQET()` consistent: every editor refuses, so we never end up with half the windows closed. Raising the dialog so the refusal isn't silent is a nice touch.

The regression script is well done: no titles or window ids, offscreen, isolated HOME and binary copy for SingleApplication, a watchdog, and PASS/FAIL/ERROR distinguished and checked both ways.

**Before merge**

1. **Rebase / CI.** The branch is still based on `adca07f26` and no checks have run. Please rebase on master so CI (Linux, Windows MinGW, macOS) goes green.

**Small things to check or consider**

2. **`close()` called from inside a modal.** If any dialog slot calls `window()->close()` on the editor while the dialog is still the active modal widget (e.g. `accept()` followed by `close()` in the same slot), that close will now be silently refused. Could you grep for `close()` calls reachable from dialog slots, just to be sure?
3. **Hook the script into ctest.** You could add it as an optional CTest, enabled only when gdb with Python is available and returning 77 (`SKIP_RETURN_CODE`) instead of 2 when gdb is missing, so CI can run it.
4. **Symbols.** The script calls `QETApp::instance()` and `quitQET()` through gdb, so it needs a non-stripped binary. On a stripped release build it exits 2, which is correct, but a line about this in the README would help.
5. **macOS.** That's the reported route, and so far it has only been reproduced on Linux via gdb. It would be good to have someone on macOS try File > Quit with the backup dialog open, and to check the element editor path, which hasn't been tested on its own.

Once CI is green, it's good to merge from my side. Thanks!

ispyisail:
All five addressed except macOS, which still needs someone with a Mac.

1. Merged master (`3ff02e528`) rather than rebasing, since you've asked before not to force-push open PRs. It configures here again now that #900's ECM bump is reverted, so CI should run.
2. Checked, and nothing is affected. No `close()` from a dialog slot reaches a guarded window: `ProjectPrintWindow` is a plain `QMainWindow`, `RichTextEditorDialog::on_buttonBox_accepted()` and `ElementPropertiesEditorWidget` close themselves, and `qettemplateeditor.cpp:1023` / `qetelementeditor.cpp:1555` are the Quit actions, where refusing is the point.
3. Done — `modal_quit_regression`, Linux only, `SKIP_RETURN_CODE 77`.
4. Done, and it now exits 77 rather than 2 on a stripped binary, so CI skips it instead of failing.
5. Element editor is covered on its own: passing a read-only `.elmt` drives `QETElementEditor` instead of the diagram editor, because QET picks the editor from the extension. Without the fix it aborts with `double free or corruption` under `~QETElementEditor()`.

Doing (5) turned up two bugs in my own script, both of which made it report success it hadn't earned: `nm | grep -q` under `pipefail` gave nm a SIGPIPE, so the symbol check skipped the test on every build that could actually run it; and a hardcoded sandbox filename renamed the `.elmt`, so the "element editor" run was really the diagram editor again. Separately the watchdog orphaned a `sleep` that held stdout, so through a pipe every run took the full 120 s — ctest is 12 s now, was 133 s.

ctest 10/10 on the merge.

scorpio810:
Thanks for the thorough follow-up, @ispyisail — really appreciate you double-checking the other close() paths and catching those two bugs in your own test script rather than letting a false-positive slide through. Merging this now; I'll test on macOS myself when I have a time and report back if anything comes up there. @jp2images @arummler can test it also.

PR #918, opened 2026-09-18
Persist QColorDialog custom colors across application restarts

Kellermorph:
Bug
Custom colors added to QColorDialog's user-defined color palette (the 16 color slots at the bottom of the dialog) are lost when QElectroTech is restarted.

Root cause
Qt's QColorDialog loads the 16 custom color slots from QSettings via QColorDialog::customColor() on first access, but QColorDialog::setCustomColor() only updates an in-memory static array -- it never writes back to QSettings. The custom colors survive within a single session but are discarded on exit.

Fix
Explicitly serialize/deserialize the 16 custom color slots to QSettings at application boundaries:

QET::loadCustomColors() -- called in QETApp constructor after initStyle(), restores the saved colors into Qt's internal array before any QColorDialog is opened.
QET::saveCustomColors() -- called as the first action in QETApp destructor, writes the current custom colors to QSettings before the process exits.
This approach covers every QColorDialog usage in the application transparently, without modifying individual call sites.


ispyisail:
The bug is real — I checked with a standalone Qt 6.10 program: `setCustomColor(0, red)` then exit writes no settings at all, and a fresh process reads back `#ffffff`. Worth fixing.

**Blocking: the `delete color_dialog;` in `diagramview.cpp`.**

It isn't in the description, and it's the #904 hazard in a new place. `DiagramView` (`diagramview.cpp:54`) and `QETDiagramEditor` (`qetdiagrameditor.cpp:1087`) both set `WA_DeleteOnClose`. If anything closes the editor while the colour dialog's `exec()` loop is spinning, that `deleteLater()` is processed by *this* loop: `~DiagramView()` deletes its children, the parented dialog among them, and then `delete color_dialog;` frees it a second time.

`QPointer<QColorDialog>` plus `if (color_dialog) delete color_dialog;` closes it. (#908 refuses close-while-modal for `QETMainWindow` and would prevent it too, but it isn't merged.)

**Non-blocking**

1. The root cause isn't quite right: Qt never reads custom colours from QSettings — no read, no write, the array simply defaults to white. The fix is correct; only the explanation is off.
2. `loadCustomColors()` doesn't check `s.status()`. Fed a short buffer, the stream reaches `ReadPastEnd` and slots 4–15 are silently set to **black**, each reporting `isValid() == true`. Reading into a temporary array and applying only on `QDataStream::Ok` avoids replacing a good palette with black.
3. The `QDataStream` version isn't pinned, so the on-disk format follows whichever Qt wrote it.
4. `"customColors"` is a bare key; most settings here are grouped and lower-cased (`diagrameditor/Xgrid`, `elements-collections/…`).

ispyisail:
Correction to my comment above: #908 **is** merged — it went in this morning, I checked a stale value. `QETMainWindow::event()` now refuses a close while a modal is up, so the editor cannot be destroyed underneath `color_dialog->exec()` by the route I described.

That downgrades my first point from blocking to defensive: the double free needs a way to destroy the `DiagramView` during the dialog's nested loop, and on current master the guard closes the one I could name. `QPointer` is still the cheaper thing to write than an argument about whether any such route remains, but treat it as hardening, not a merge blocker. Sorry for the noise.

The four smaller points are unaffected, and the `s.status()` one is the one I would actually act on — a truncated value silently turns slots 4–15 black.

Kellermorph:
Thanks for the thorough review — all good catches. Here's what changed:

**s.status() check (loadCustomColors):**
Colors are now read into a temporary `QColor colors[16]` array. After the stream finishes, we check `s.status() != QDataStream::Ok` and bail out entirely if the buffer was truncated. This prevents slots 4–15 from silently turning black on a corrupt or short value.

**QPointer in diagramview.cpp:**
Changed `QColorDialog *` to `QPointer<QColorDialog>` with a guarded `if (color_dialog) delete`. Defensive hardening — agreed it's not a merge blocker with #908 in place, but it's cheap insurance.

**Settings key:**
Moved from bare `"customColors"` to `"color/customColors"` to match the grouped, lowercase convention used elsewhere. The loader falls back to the old key for backward compatibility with existing configs.

**QDataStream version:**
Pinned to `QDataStream::Qt_6_0` so the on-disk format is deterministic regardless of which Qt version builds it.

**Root cause description:**
You're right — I incorrectly stated Qt reads custom colors from QSettings. It doesn't; the array simply defaults to white. Updated the comment accordingly.

PR #921, opened 2026-09-18
Fix: "group rotate" 

arummler:
Conductors attached to elements were ripped off as the conductors themselves were not rotated. Unexpected behaviour.


ispyisail (commented):
Tested merged onto current master: builds clean on Qt 6, 12/12 ctest.

**It fixes the reported problem.** Selecting everything on folio 1 of `Habitat-Schemas_developpes.qet` and pressing Shift+Space: on master the wires stay behind while the symbols move; with this branch they follow.

**It also stops something quieter, which your description undersells.** On master, rotating a group and saving writes wire paths that no longer match the terminals they run to. On reopening, QET notices the mismatch and silently drops the stored path in favour of an automatic route — so a wire the user had shaped by hand loses that shape, and the reopened file looks perfectly fine. One rotate of one folio: **7 wires affected on master, 0 with this branch**.

Undo is clean: back to 159 stored segments, the same as the original file.

Ordinary saving is untouched — all 23 example projects resave byte-for-byte identical with and without the change.

**One measured cost.** Moving an element with 20 wires attached: 0.55 ms → 0.91 ms. A diagonal move costs twice that extra (1.29 ms), because the update runs once per coordinate — `xChanged` and `yChanged` both fire. That is well inside a frame, so it should not be visible. Iterating `m_terminals` instead of `childItems()` made no measurable difference, so the cost is inside `updatePath()`.

**Blocking:** none.
**Non-blocking:** the update running twice on a diagonal move.

PR #922, opened 2026-09-18
Minimal bundled kf

elevatormind:
### Minimal bundled library of KDE's KCoreAddons and KWidgetsAddons libraries.

New CMake option, `BUILD_KF_MINIMAL`; which when enabled together with `BUILD_KF` compiles and includes a minimal bundled version of KCoreAddons and KWidgetsAddons. Sources are taken from KDE's git repository, and care has been taken to include appropriate licenses and notices.

If these minimal bundled libraries ever reaches a distributed installer/package though, KDE-Frameworks.LICENSE and KDE-Frameworks.NOTICE must be included.

They're added to the CMake `install()` command in this PR, which I realize now maybe should be guarded by `BUILD_KF_MINIMAL`, they're probably not needed if the minimal build is not enabled.


ispyisail:
I built this rather than only reading it — 2650 vendored lines are the kind of thing that passes CI and breaks in the field.

**Verified here**, configuring with `-DBUILD_KF_MINIMAL=ON`: builds clean, ctest 9/9, the binary carries 476 KF symbols with **zero** KF6 shared-library dependencies, and `KColorButton` renders correctly in Configuration → Nouveau projet → Conducteur.

Three things that hold up well and are worth saying out loud:

- Licensing is right. The vendored files declare `LGPL-2.0-or-later` and the bundled `KDE-Frameworks.LICENSE` is the matching Library GPL v2 text. LGPL-2.0 vs 2.1 is easy to get wrong; this isn't.
- The `PROVENANCE` files are exemplary — upstream URL, git tag, and an explicit list of which files are KDE-derived versus added by QElectroTech.
- The subset is correct. QET uses `KAutoSaveFile`, `KColorButton` and `KColorCombo`; `KRandom` follows as `KAutoSaveFile`'s dependency. Nothing is missing.

**Three things to consider**

1. **CI never exercises this path.** The workflow configures with only `-DCMAKE_BUILD_TYPE=Debug -DQT_VERSION_MAJOR=6` and installs system KF6, and `BUILD_KF_MINIMAL` defaults to OFF — so the green check proves the default build still works, not the new code. A second job with the flag on would be the difference between tested and untested.
2. **Version skew.** The bundled sources are v6.30.0 while the FetchContent default `KF_GIT_TAG` is v6.10.0, so the two build modes ship two different KF versions. Worth aligning, or documenting why not.
3. The unconditional `install()` of the KF licences that you flagged yourself — confirmed, it installs them even with `BUILD_KF_MINIMAL=OFF`. Guarding it is right.

One argument the description does not make for itself: this sidesteps ECM entirely. #900 forced `0a7ea032d` to revert the KF 6.28 bump because it needed ECM >= 6.28, and I hit that same wall today on a box with ECM 6.24 — master would not configure. A bundled build needs no ECM at all, which is the strongest case for the change.

**Non-blocking** from me. All three points are small and the risk is contained by the option defaulting off.

scorpio810:
More reviews: @jp2images @arummler

elevatormind:
- I'm not too familiar with the CI actions, but this should probably be in there in some way if merged
- Regarding version alignment minimal <-> FetchContent path, I don't think that's necessary for now. But can do version 6.10.0 for minimal path if requested.
- KF license files installed conditionally now with commit [15c2baa](https://github.com/qelectrotech/qelectrotech-source-mirror/pull/922/commits/15c2baa8f0122a78e06b74106e6239c80b074268). Just a thought here, shouldn't they be installed with KF FetchContent path as well? Or does the FetchContent lift them in?

arummler:
> * I'm not too familiar with the CI actions, but this should probably be in there in some way if merged

The CI should indeed ideally check all the (option) combinations as one of its purposes is to check whether someone did not break one of them. That's not really difficult, more a bit tedious to implement all. I can add them to my PR but it still takes a bit of time.

elevatormind:
That makes me think as few options as possible is wanted, adding this would make four different ways of using the KF features making endless of build paths needed for testing.

Are the KF libraries git FetchContent build path actually used anywhere? `BUILD_KF` is `YES` as default, and the only place I see it's not explicitly set to `NO` is ironically in the `linux-build` workflow.

scorpio810:
Please try this Ai generated patch to consolidate nokde-kautosavefile.
[nokde-kautosavefile.patch](https://github.com/user-attachments/files/32596834/nokde-kautosavefile.patch)

elevatormind:
I'm happy as long as KF git FetchContent is removed :) I'm ready to drop this PR if 'nokde' is implemented and KF is removed completely.

The patch compiles fine with MSVC 2022 compiler / Qt Online installer.

I did an AI review of the 'nokde' folder, three points worth noting below. Tests for 'kautosavefile' are only run on Unix and should perhaps be looked upon

---
**[P2] Hidden project files can be missed during recovery** — [`kautosavefile.cpp:183–186`](sources/ui/nokde/kautosavefile.cpp). Autosave filenames retain the project’s basename. For a project whose name starts with `.`, the autosave file also starts with `.`, but the scan uses `QDir::Files` without `QDir::Hidden`. After a crash, `allStaleFiles()` can therefore omit that backup. Include hidden entries in the scan or ensure autosave filenames never start with a dot; add a dotfile recovery test.

- **[P2] `setColor()` does not notify `changed` listeners** — [`kcolorbutton.cpp:43–47`](sources/ui/nokde/kcolorbutton.cpp). The fallback updates its color and appearance without emitting `changed`. That differs from the behavior the repository documents for the KDE `KColorButton` equivalent, so listeners can behave differently depending on the build configuration. Decide whether the fallback must preserve that signal contract and test it.

- **[P2] The autosave recovery test skips non-Unix platforms** — [`kautosavefile_test.cpp:20–24`](tests/catch/src/kautosavefile_test.cpp). On Windows and other non-Unix platforms, the test reports success without exercising locking or stale-file recovery. That leaves an important production path unverified, particularly given the platform-specific path handling in this implementation.

ispyisail:
@scorpio810 I tested the patch: it applies to master, and in a standalone crash-and-recover test on Linux it behaves exactly as before. Worth applying.

Two corrections to the description:

- The Windows data loss is point 5, not point 1. Without the patch, a recovered project reopens with its drive letter gone (`C:/foo.qet` comes back as `/foo.qet`), so the next save goes to the wrong place. Point 1's "autosave never written" does not happen: the autosave file's name is never built from the project path.
- Point 2 fixes the recovery dialog showing `:/foo.qet`. The comparison it corrects only hides projects that are already open, so recovery was still offered.

One question: `urlFromLocalPath()` keeps only the path part of the URL, so a project on a network share (`\\server\share\x.qet`) would be recovered as `/share/x.qet`. Could someone on Windows check that case?

**Blocking:** none.
**Non-blocking:** network-share paths (read, not run, since I have no Windows); a Windows test that saves and reloads a path before nokde becomes the default there.

elevatormind:
Tested this, on Windows when opening ‪"\\\hefs01\Filservern\CAD\namnlös.qet" it creates a .path file with "/Filservern/CAD/namnlös.qet".

"namnlös.qet" also gets escaped to:
"namnl%C3%B6s.qet_7573cde4feda83fb_a75d4a10d1c23fbf.qetautosave.path"

don't know if this is intended or a potential issue.

PR #924, opened 2026-09-18
Add global auto-numbering rules to QElectroTech settings

Kellermorph:
Added a "Numérotation auto" tab to QElectroTech's global settings page (Settings > Nouveau projet) where users define default auto-numbering rules that are automatically transferred to every new project.

What was done:
NumerotationContext serialization (numerotationcontext.h/.cpp): Added saveToSettings()/loadFromSettings() static helpers that serialize a QHash<QString, NumerotationContext> to/from QSettings using XML via toXml()/fromXml().
"Numérotation auto" tab (configpages.h/.cpp): Added a new tab to NewDiagramPage (global settings only, not per-project) with three sub-tabs — Conducteurs, Eléments, Folios — each using SelectAutonumW widgets. Users can create, name, edit, and delete rules. Changes are persisted to QSettings immediately on every Apply or Remove action via persistAutonumSettings().
New project integration (qetproject.cpp): The QETProject constructor loads the saved global auto-numbering rules from QSettings and transfers them to every new empty project via addConductorAutoNum()/addElementAutoNum()/addFolioAutoNum().
Key design decisions:
Immediate persistence: Every save/remove action writes to QSettings right away (not only on dialog OK), preventing data loss if the dialog is closed via Cancel or X.
Reuses existing SelectAutonumW widget — same UI users already know from project properties.
5 files changed, 334 insertions.


ispyisail:
Built it and drove the feature end to end rather than only reading the diff. Build clean, ctest 9/9.

**It works.** I defined a conductor rule in Configuration → Nouveau projet → Numérotation auto, made a new project and saved it, and the file contains exactly what I defined:

```xml
<conductor_autonum formula="%sequ_1" title="RuleTwo">
    <part increase="1" type="unit" value="5"/>
</conductor_autonum>
```

The immediate-persistence design holds too: I applied a rule and then dismissed the dialog with **Cancel**, and it survived. I also checked the thing most likely to go wrong — the new code is in the `QETProject(QObject*)` empty-project constructor, and neither file-opening constructor delegates to it, so opening an existing project is genuinely unaffected.

**Blocking: the UI placeholder leaks into every new project file.**

```xml
<element_autonums current_autonum="Nom de la nouvelle numérotation" freeze_new_elements="false"/>
```

`persistAutonumSettings()` writes `element/current` from the combo's text, which for a tab the user never touched is the placeholder. The constructor then runs `if (!elementData.second.isEmpty()) setCurrrentElementAutonum(...)`, so every new project gets a `current_autonum` naming a rule that does not exist — and since it is a translated string, the junk differs per locale. Applying the same guard the save methods already use, or only writing `current` when the hash contains that name, fixes it.

**Non-blocking**

1. Deleted rules are never removed from the config. After deleting my only rule:

   ```ini
   conductor\rules\size=0          ← claims zero
   conductor\rules\1\name=RuleTwo  ← still present
   conductor\rules\1\xml="..."
   ```

   `beginWriteArray()` does not clear entries beyond the new size. Harmless today because `beginReadArray()` returns 0 and skips them, but the file grows without bound. `settings.remove(prefix + "/rules")` before the write clears it.
2. The three save/remove/combo-activated paths are ~200 lines of near-identical code; one helper parameterised by (widget, hash, prefix) would carry all three.
3. `name == tr("Nom de la nouvelle numérotation")` is locale-dependent — the persisted value is in whatever language was active when it was written, so changing language defeats the guard.
4. Pressing Enter in the rule-name combo activates the dialog's default button and closes the settings dialog.

One note in case anyone retests: Apply stays disabled until every `NumPartEditorW` is valid, and `isValid()` is false while the value field is empty. My first two attempts only typed a rule *name* and I very nearly reported "Apply does nothing" — it was my test, not the code. Put a digit in the value field and it enables.

Kellermorph:
Thanks for the thorough review — all issues addressed in the amended commit:

Placeholder leaking into new projects (blocking): persistAutonumSettings() now validates that the combo's current text actually exists in the contexts hash before writing it as current. Untouched tabs write an empty string, so no bogus current_autonum appears in project files.
Stale QSettings entries after delete: saveToSettings() now calls settings.remove(prefix + "/rules") before beginWriteArray() to clear old array entries beyond the new size, preventing unbounded config file growth.
Refactoring: Replaced the three near-identical save/remove method triples (~200 lines) with an AutoNumTab struct and two parameterized helpers (saveAutoNumContext, removeAutoNumContext). The struct bundles the widget, contexts hash, and settings prefix together.
Locale-dependent placeholder check: Replaced name == tr("Nom de la nouvelle numérotation") with isPlaceholder(combo, name) which compares against combo->itemText(0) — the actual first item from the .ui file — so it works regardless of locale.
Enter key closes dialog: Added an eventFilter() that intercepts Return/Enter keypresses in the combo box line edits and eats them, preventing the dialog's default button from firing.

ispyisail:
Re-checked by running it, not by reading the diff.

**The blocking one is fixed.** Built `e6f030c88`, launched with an isolated config, opened Configuration and pressed OK with every tab untouched:

```ini
[autonum]
conductor\current=      element\current=      folio\current=
```

All three empty, and `Nom de la nouvelle numérotation` appears nowhere in the config file. A new project now saves `<element_autonums current_autonum="" …>` instead of a rule name that does not exist.

**The stale entries are cleared too.** I seeded exactly the leftovers from my first review — `conductor\rules\1\name=StaleRule` sitting past `size=0` — relaunched and pressed OK. Gone.

The refactor, `isPlaceholder()` and the `eventFilter()` all read correctly. `isPlaceholder()` is sound because both the loader and the save path *append*, so the placeholder stays at index 0. ctest 9/9.

Two small things I only read, so worth your own check rather than my word:

- `removeAutoNumContext()` reads `currentText()` untrimmed (line 416) where `saveAutoNumContext()` trims (line 399). A name typed with surrounding spaces is stored trimmed, so Remove would take the combo entry away while the rule survives in settings and returns on restart.
- `removeItem(currentIndex())` — typing a new name does not change `currentIndex()`, so it can still be 0 while the text is the typed name. That passes the placeholder guard and removes item 0. After that `itemText(0)` is a real rule and `isPlaceholder()` would reject it for good.

**Blocking:** none.
**Non-blocking:** the two above.

Kellermorph:
Good catches, both fixed:

Trimmed name in removeAutoNumContext(): Now uses .trimmed() to match the save path. A name typed as " test " is stored as "test" and removed as "test".
Wrong item removed by currentIndex(): Replaced removeItem(currentIndex()) with findText(name) so the correct item is always removed regardless of what the combo's current index happens to point to.
Build clean, ctest 8/9 (the menu-bar keyboard test is a pre-existing headless-environment issue).

PR #926, opened 2026-09-18
Add optional SQLite amalgamation sources

elevatormind:
Adds a new option, QET_USE_BUNDLED_SQLITE3, to use SQLite amalgamation sources instead of a system package. Defaults to OFF.


ispyisail:
Built this with the option on rather than only reading it. It configures and builds cleanly and produces a working binary plus a 2.9 MB `libsqlite3.a`, and the CMake is well structured — opt-in, and the system path keeps the `SQLite3::SQLite3` alias shim intact.

**What is this intended to solve?** I ask because of what the build actually produces.

With the default options, the bundled library is dropped entirely by the linker:

```
sqlite3 symbols in the built binary: 0
```

The only place QElectroTech calls the SQLite C API directly is `sqlite3_backup_*` in `sources/dataBase/projectdatabase.cpp:1314`, and that sits behind `#ifdef QET_EXPORT_PROJECT_DB`, which `cmake/start_options.cmake:32` defaults to OFF. Nothing references SQLite, so 270k lines of C compile and are then discarded.

The option therefore only does something when `QET_EXPORT_PROJECT_DB` is also on — and that is the combination I would be careful about:

```cpp
QVariant v = db->driver()->handle();               // handle created by Qt's SQLite
handle = *static_cast<sqlite3 **>(v.data());
sqlite3_backup_init(file_db_handle, "main", ...);  // called into the bundled SQLite
```

Qt's `libqsqlite.so` links `libsqlite3.so.0` here (checked with ldd), so with both options on a handle created by the system SQLite is passed into the statically linked 3.53.4. Different library, different internal state — it links fine and is undefined at runtime.

Both options default to OFF, so nothing shipped hits this; it is a latent trap rather than a live bug. But as written the option is a no-op in the safe configuration and unsafe in the one where it takes effect, which seems worth settling before merge. If there is a platform where this pays off that my Linux build does not capture — a static Qt, or MinGW where Qt bundles its own SQLite — say so, because that would change the picture.

**Smaller points**

1. No provenance file. It is 3.53.4 (source id dated 2026-07-24), which I had to read out of `sqlite3.h`. Your `PROVENANCE` files in #922 are a good model, and a vendored SQLite especially wants a recorded version so it can be tracked against CVEs.
2. CI does not exercise the option — it defaults OFF and the workflow passes no extra flags, so the new path is never built there. Same gap as #922.
3. `cmake_minimum_required(VERSION 3.5)` in `thirdparty/sqlite3/CMakeLists.txt:17` raises a CMake deprecation warning.
4. No `SQLITE_*` compile options are set for the amalgamation.

elevatormind:
I have dug deeper into how Qt handles SQLite, and linking QET against an internal compiled version will unfortunately not work. Exporting the project DB crashes, as Qt will use one version of SQLite, and QET linked against another. I will close this PR, and try another solution.

Issue #461, opened 2026-05-16
[Feature] Color preset


psolyca:
Could we have an way to create preset for conductor properties ?

scorpio810:
Select conductor and hit F2 for color picker

psolyca:
Thanks; it's usefull ^^

I was thinking of a more complete feature, with preset of type of conductor and color.

scorpio810:
Hello @psolyca,

This is closely related to #476.

QET already has a mechanism that covers this use case: **conductor types** defined at the project level (project conductor list). A conductor type acts exactly as a preset — it bundles color, style, thickness, and numbering rules, and all conductors assigned to that type inherit its appearance automatically.

Changing the type definition propagates to all conductors of that type at once, which also preserves equipotentiality consistency (conductors sharing the same potential stay visually coherent).

Could you clarify what is missing compared to the current conductor type system? For example:
- Presets shared across projects (global/user-level presets, not just per-project)?
- A quicker way to assign a type to a conductor during drawing?
- Something else entirely?

Understanding the gap will help determine whether this is a UI/workflow improvement on top of the existing type system, or a genuinely new feature.

Best regards

scorpio810:
Is the expected behavior that the last used conductor properties are remembered for the next drawn conductor (session-level "sticky" default), rather than a named preset system? If so, should this override the conductor type, or should it select/create a matching type automatically?

psolyca:
> Hello [@psolyca](https://github.com/psolyca),
> 
> This is closely related to [#476](https://github.com/qelectrotech/qelectrotech-source-mirror/issues/476).
> 
> QET already has a mechanism that covers this use case: **conductor types** defined at the project level (project conductor list). A conductor type acts exactly as a preset — it bundles color, style, thickness, and numbering rules, and all conductors assigned to that type inherit its appearance automatically.
> 
> Changing the type definition propagates to all conductors of that type at once, which also preserves equipotentiality consistency (conductors sharing the same potential stay visually coherent).
> 
> Could you clarify what is missing compared to the current conductor type system? For example:
> 
> * Presets shared across projects (global/user-level presets, not just per-project)?
> * A quicker way to assign a type to a conductor during drawing?
> * Something else entirely?
> 
> Understanding the gap will help determine whether this is a UI/workflow improvement on top of the existing type system, or a genuinely new feature.
> 
> Best regards

Could be this but I do not find any "project conductor list" on my project properties !
[...]

PR #929, opened 2026-09-18
Add a one-click conductor colour to the toolbar (#461)

ispyisail:
Colouring wires costs a dialog per wire today. F2 opens the colour picker and only acts when exactly one conductor is selected, so an electrician drawing 400 V and 24 V circuits on the same folio changes colour one conductor at a time. Asked for on the forum (`viewtopic.php?pid=23296`), and the quick-access half of #461.

This puts a colour swatch in the **Schéma** toolbar. Picking a colour recolours every selected conductor in a single undo step, and becomes the colour of the next conductor drawn — the mechanism #888 added for exactly that. With nothing selected it only sets the pen for what comes next.

The menu lists the colours the trade names, then any custom colour picked during the session, then the full colour dialog.

**Deliberately left out:** nothing is written to the project or to QSettings, so there is no file-format change. Named presets stored per project — what #461 actually asks for — are a larger feature that still needs your decision; that question has been open since June.

**Tested** on a virtual display, reading colours back from the saved project rather than the screen:

| gesture | result |
|---|---|
| select all on a folio, pick Rouge | 21 conductors in five colours → all `#ff0000` |
| one Ctrl+Z | back to the original mix, exactly |
| pick Marron with nothing selected, then draw a wire | the new conductor is `#7b3f00` |

ctest 12/12, Qt 6.


scorpio810:
What about equipotentialy conflicts?

ispyisail:
Good question — I measured it rather than guessed.

Colour uniformity across a potential is **not** an invariant. Of 75 multi-conductor potentials in the 23 example projects, 8 already carry more than one colour, all in `weneedpolonez-Polonez_MR89_wiring_diagram.qet` — a yellow and a cyan wire meeting at one connector pin, which is ordinary in a car loom.

So the button recolouring only the selection does not break a rule. It matches the properties dialog with "apply to all" unticked.

Three paths do disagree, which is worth knowing: editing a conductor's text propagates to the whole potential unconditionally (`conductor.cpp:1743`), the properties dialog propagates by default but is opt-out (`m_apply_all_cb` ships checked), and this button never propagates. Forcing propagation here would overwrite the Polonez case, so I would rather leave it and say so in the description — but if you want the button to follow the dialog default instead, say and I will change it.

One thing I checked and it is unaffected: the potential chooser fires on potential *size*, not on properties differing (`conductorcreator.cpp:110`), so this cannot make it appear more often.

scorpio810:
OK, thanks.

PR #930, opened 2026-09-18
List auto conductor creation in the Shortcuts page

ispyisail:
An electrician asked on the forum for a keyboard shortcut for automatic conductor creation: *"je n'utilise pas par défaut, mais ponctuellement c'est très pratique"* (`viewtopic.php?pid=23296`).

The action already exists — a checkable entry in the **Schéma** toolbar and the Project menu. It was never handed to `ShortcutManager`, so it did not appear in **Configuration → Raccourcis** and could not be reached from the keyboard at all. This registers it. One line.

**No default key is set.** He asked for it to be *in* the shortcuts list so he can bind it himself, and that avoids spending one of the few free keys on a setting many people never touch. The page already treats "no shortcut" as a normal state — it shows an empty field and can filter for actions with and without a binding.

**Tested** on a virtual display, with the binding set to Ctrl+Alt+A:

| check | result |
|---|---|
| Configuration → Raccourcis, filtered on "conducteur" | lists "Création automatique de conducteur(s)" under "Éditeur de schémas" with that key |
| pressing it, pointer away from the toolbar | the toolbar button changes |
| pressing it again | returns to its starting state — it toggles, exactly like clicking |

ctest 12/12, Qt 6.


scorpio810:
@ispyisail retry to create new account on forum.
Edit: you see a new Mantis send email to change your password, I'm tried with new account and it's work, only captcha is dificult, but is the game.

ispyisail:
im in

PR #529, opened 2026-06-22
feat(wirecatalogue): wire/cable catalogue + colour pickers + conducto…

Gokulakannan750:
…r integration

Custom feature set under sources/custom/wirecatalogue/ (isolated from upstream):

- Wire/Cable Catalogue: SQLite-backed CRUD (WireCatalogueDb), WireSpec data model, dockable panel (WireCatalogueWidget) with search/filter and an add/edit dialog, registered via a Display-menu toggle in qetdiagrameditor.
- Phase 3: IEC 60757 colour utilities (iec60757), per-core colour pickers and a multi-core cable builder (CoreColorEditor/WireColorComboBox). Colour swatches in the catalogue table; gradient dialog header; tinted toolbar.
- Conductor integration: "From catalogue" picker in the conductor properties Appearance tab sets the conductor colour and puts cross-section + colour on the line (shared catalogue via WireCatalogueDb::defaultPath()).

Build: files registered in both CMake (qet_compilation_vars.cmake) and qmake (qelectrotech.pro); add QET_UPDATE_SUBMODULES option to skip the slow recursive submodule fetch; BUILD_WINDOWS.md documents the MSYS2/Qt5 build; ignore /build/.

PR #936, opened 2026-09-19
Wire/cable catalogue (#529 restored): French UI, search and green/yellow fixes

ispyisail:
**This is @Gokulakannan750's work, restored.** #529 was closed on 16 September when his fork was deleted by accident, not withdrawn — his five commits are here unchanged, with his authorship. @scorpio810 said at the time he wanted to test and merge it, so rather than let it disappear I have brought it up to date and cleared the blockers. @Gokulakannan750 — if you would rather take this back and push it yourself, say so and I will close this.

## What it adds

A wire/cable catalogue: a SQLite-backed dock with search and CRUD, IEC 60757 colours with per-core pickers for multi-core cables, CSV import/export, and an **Affecter un fil** tab in the conductor properties that sets a conductor's colour, section and label from a catalogue entry.

## What I changed

**Brought it to master** — it was 887 commits behind. Two conflicts, both mechanical: an include added on each side, and `qelectrotech.pro`, which upstream deleted when the project went CMake-only.

**Translated the interface.** All 71 `tr()` strings were English, which is what @scorpio810 asked about on 13 September — the Display menu read "Wire / cable catalogue" between French entries. They are French now, using QET's own vocabulary. "Core" became **brin** rather than *conducteur*, since *conducteur* already means the wire on the folio and both words appear in the same dialog.

**Three defects, each verified by running it:**

| | Before | After |
|---|---|---|
| Searching `2.5` | matched only wires whose *name* contained "2.5" — a wire recorded as 2.5 mm² under any other name was invisible | the query casts the numeric columns, so it matches the section and the core count |
| A green/yellow wire | drawn plain green: `colorForName("Green-Yellow")` returned the same RGB as `Green`, and the shield path hardcodes that name | drawn as a bicolour conductor, green + yellow, using QET's existing support — `bicolor="true"`, `color2="#ffd700"` |
| Build | two deprecated `invalidateFilter()` warnings | no warnings from the feature |

I also corrected a comment claiming the cable field makes "the terminal strip / BOM treat the bundle as one cable". Nothing reads that field today — the terminal strip's Cable column is a stub returning an empty string — which is what discussion #934 is about.

**Left alone deliberately:** the `Copyright 2026 Trovo Tech Solutions` headers, which @scorpio810 explicitly accepted; the `sources/custom/` location; and the catalogue database being opened when the dock is constructed rather than on first use.

**Tested:** builds clean on Qt 6, ctest 12/12, and driven in a real session — added a wire, searched for it, assigned it to a conductor and read the result back out of the saved project.


sources/ui/conductorpropertieswidget.cpp
```diff
@@ -478,6 +481,17 @@ void ConductorPropertiesWidget::applyWireAppearance(const QString &section,
 	if (c.isValid())
 		ui->m_color_kpb->setColor(c);
 
+		//Green/yellow is the one colour with a regulatory meaning, and it is
+		//two colours. QET already draws bicolour conductors, so use that
+		//rather than approximating the pair with a single swatch: the
+		//protective conductor then looks like a protective conductor.
+	const bool green_yellow = (colour == QLatin1String("Green-Yellow"));
```
arummler, line 488: QLatin1StringView is more modern afaik

ispyisail:
Follow-up: the first translation commit was incomplete, and running the UI in English found it.

Moving the `tr()` sources to French without touching `qet_en.ts` meant **English users would have seen French** — a regression I introduced while fixing the translation chain. `qet_en.ts` now carries 80 translations: the 78 catalogue and conductor-tab strings, plus two in `qetdiagrameditor.cpp` my first pass missed entirely, which is why the Display menu still read "Wire / cable catalogue" inside a French interface.

Verified by running it: with `lang=en`, the menu bar reads File / Edit / Project / Display / Settings and the catalogue entry reads "Wire / cable catalogue", translated from the French source. French is unaffected — with no translation loaded the sources show, and those are now French.

**What is and is not covered**, so a reviewer knows where to look:

| Exercised in a running session | Not exercised |
|---|---|
| dock opens, catalogue lists and persists across restarts | Edit and Delete in the catalogue (only Add) |
| Add a wire, then find it by cross-section | CSV import / export |
| assign a wire to a conductor; result read back from the saved file | multi-core cable editing and per-core assignment |
| green/yellow assignment → bicolour conductor | the shield path specifically (same code, not driven) |
| English and French interfaces | undo/redo of an assignment; Windows and macOS |

No automated test covers the catalogue: `ctest` is 12/12 but none of those twelve touch this feature.

cmake/git_update_submodules.cmake
```diff
@@ -16,9 +16,16 @@
...
+# The build only needs pugixml + SingleApplication; the large `elements`
+# symbol library is runtime content and not required to compile or launch.
+# Configure with -DQET_UPDATE_SUBMODULES=OFF to skip (ensure the two
+# build-critical submodules are already checked out).
+option(QET_UPDATE_SUBMODULES "Auto update git submodules during configure" ON)
+
 find_package(Git QUIET)
 
-if(GIT_FOUND AND EXISTS "${PROJECT_SOURCE_DIR}/.git")
+if(QET_UPDATE_SUBMODULES AND GIT_FOUND AND EXISTS "${PROJECT_SOURCE_DIR}/.git")
```
arummler, line 28: Should this be part of the wire catalogue PR? It seems unrelated.

lang/qet_en.ts
arummler, line 1: I believe usually the TS is addressed in a separate PR as it needs to be created against the most current HEAD of master. Any changes inbetween leed to info loss.

sources/custom/wirecatalogue/wirecatalogueui.cpp
```diff
@@ -0,0 +1,239 @@
...
+		return;
+
+	WireSpecDialog dlg(this);
+	dlg.setWireSpec(current, true);
+	if (dlg.exec() != QDialog::Accepted)
+		return;
+
+	if (!m_db->updateWire(dlg.wireSpec())) {
+		QMessageBox::warning(this, tr("Catalogue de fils"),
+			tr("Impossible de mettre à jour le fil :\n%1").arg(m_db->lastError()));
```
arummler, line 179: I guess the error has to stay untranslated? It is coming directly from the db?

arummler: Okay, it seems that they are already translated where they get generated which makes sense.

BUILD_WINDOWS.md
arummler, line 1: Why is a Windows build instruction part of this MR? It replicates the already existing one partially and refers to outdated (Qt5) infos.

.gitignore
arummler, line 1: Useful although in-source build asre discouraged. But should it be part of this MR?

sources/ui/conductorpropertieswidget.cpp
```diff
@@ -248,6 +266,241 @@ void ConductorPropertiesWidget::initWidget()
...
+		const QString colour = QStringLiteral("Green-Yellow");
+		if (ui->m_line_style_cb->count() > 1)
+			ui->m_line_style_cb->setCurrentIndex(1); // dashed
+		applyWireAppearance(section, colour, w.wireId,
+							QStringLiteral("%1:SH  Shield").arg(w.wireId));
+	} else if (w.isCable()) {
+		const QStringList core = w.coreColors.value(sel);
+		const QString colour = core.value(0);
+		// Per-core section (falls back to the cable's nominal section).
+		const QString core_section = QStringLiteral("%1mm²")
```
arummler, line 452: Did not ccheck where it is used but does that needs a translation?

sources/custom/wirecatalogue/wirecataloguedb.cpp
```diff
@@ -275,6 +275,16 @@ bool WireCatalogueDb::removeWire(const QString &wireId)
 		setError(QStringLiteral("removeWire"), q);
 		return false;
 	}
+	// A DELETE that matched nothing is a successful statement, so q.exec()
+	// alone cannot tell "removed" from "was never there". Report the
+	// difference: addWire() refuses an id that already exists, so
+	// removeWire() owes the caller the same answer about one that does not,
+	// and the catalogueChanged() signal must not fire for a no-op.
+	if (q.numRowsAffected() == 0) {
+		m_last_error = QStringLiteral("removeWire: no wire with id \"%1\"")
```
arummler, line 284: No translation? French source as default?

tests/qttest/CMakeLists.txt
```diff
@@ -138,6 +138,20 @@ add_test(NAME tst_qetstrings COMMAND tst_qetstrings)
...
 
+# The wire/cable catalogue's storage layer: CRUD, search, CSV round trip,
+# multi-core cables and the IEC 60757 colour table. Headless -- it opens
+# databases in a QTemporaryDir, never the user's real catalogue. Two of its
+# cases are regressions: searching by cross-section or core count, and
+# Green-Yellow resolving to a different colour than plain Green.
+add_executable(
+  tst_wirecatalogue
+  tst_wirecatalogue.cpp
+  ${QET_DIR}/sources/custom/wirecatalogue/wirecataloguedb.cpp
```
arummler, line 149: absolute path? not a CMake variable to be used?

ispyisail:
Reviewing my own branch turned up four defects, all now fixed with tests. The one that mattered: assigning a wire to a conductor destroyed its wire number. Wire numbering stores the number in the conductor's text formula, and the assignment wrote its description — `2.5mm² Brown` — into that same field, so a numbered conductor lost its number the moment a wire was assigned to it. Assignment now fills that field only when the conductor has no label of its own, and sets the section, colour and cable either way.

The other three:

- Assigning a cable shield set the line dashed and nothing set it back, so every core assigned afterwards stayed dashed.
- A note containing a line break exported to a valid catalogue file the importer could not read back: it read line by line, tearing the record in half and shifting every field after it.
- Removing a wire that was not in the catalogue reported success.

**Evidence.** New `tst_wirecatalogue`, 19 cases, registered in ctest — storage, search, import/export round-trip, multi-core cables, the IEC 60757 colours. Five fail on the code before these fixes and pass after; the other fourteen pass either way. All 13 ctest suites pass. All 23 example projects resave byte-identical, so the feature is inert for anyone who never opens it. Not tested: the catalogue's own edit, delete and import/export buttons — only the storage layer beneath them.

Two things that are a maintainer's call rather than mine:

- The code carries a vendor's name throughout: a `sources/custom/` directory, twenty files headed *"Copyright 2026 Trovo Tech Solutions — part of a custom feature set built on QElectroTech"*, and `(Trovo Tech)` comments in three existing files. It is GPLv2+ so it can be merged, but it reads as a fork's code rather than a contribution.
- Three unrelated build changes are bundled in: `BUILD_WINDOWS.md`, a `/build/` gitignore entry, and a new `QET_UPDATE_SUBMODULES` option. Happy to split these out.

**Blocking:** none outstanding.
**Non-blocking:** the vendor naming; the three build changes; the catalogue dialogs are exercised only by hand.

tests/qttest/tst_wirecatalogue.cpp
arummler, line 1: Not sure that tests need to be translated (developers common language English) but just to mention.

cmake/git_update_submodules.cmake
ispyisail: Agreed — removed.

BUILD_WINDOWS.md
ispyisail: Agreed — removed. You're right on both counts: it overlapped the existing instructions and described a Qt5 build the project no longer does.

.gitignore
ispyisail: Agreed — removed.

tests/qttest/CMakeLists.txt
ispyisail: That is `${QET_DIR}`, the same variable the tests above it use — it just wraps awkwardly in the diff view.

tests/qttest/tst_wirecatalogue.cpp
ispyisail: Agreed, and it is — the file is English throughout. Only the shipped interface strings follow the French-source convention.

sources/ui/conductorpropertieswidget.cpp
ispyisail: You're right that it's the modern spelling. The tree has 205 uses of `QLatin1String` and none of `QLatin1StringView`, so I've left this one matching its neighbours rather than making it the odd one out. Happy to do the sweep as its own change if you'd like it.

lang/qet_en.ts
ispyisail: These entries are hand-written, not regenerated — 354 lines added, 0 removed — so nothing existing is dropped and a later `lupdate` against master loses nothing.

Without them the catalogue ships French-only to English users, since the sources are French by convention. That regression is what the previous commit was fixing, so dropping the file here would put it back. Happy to split it out if you'd still rather have it separate.

sources/ui/conductorpropertieswidget.cpp
ispyisail: Good question, and the answer is that it must *not* be translated: that string is saved, not just displayed. It lands in the conductor's `conductor_section` attribute in the `.qet` file (`conductorproperties.cpp:291`), so translating it would make a saved project depend on the language of whoever drew it.

sources/custom/wirecatalogue/wirecataloguedb.cpp
ispyisail: These are diagnostics rather than interface text. They get appended to an already-translated sentence — `tr("Impossible de supprimer le fil :\n%1")` — so the user reads a translated message with the raw cause on the end. Wrapping them in `tr()` would put SQL wording into the translators' file.

ispyisail:
Thanks — that was a useful pass. Pushed in `2fcf10198`:

- `BUILD_WINDOWS.md`, the `/build/` ignore entry and the `QET_UPDATE_SUBMODULES` option are gone. All three were unrelated to a wire catalogue.
- Chasing your styling question turned up something none of us had flagged: this branch carried **the only hard-coded foreground and background colours in QET** — a light gradient with dark blue labels on the dock toolbar, a blue/green gradient with white text on the dialog heading, and a `#666` hint. Under a dark palette the labels were unreadable. The toolbar now takes the application palette like every other dock, the heading is bold through its font, and the hint uses `PlaceholderText`.
- A comment claiming the project targets Qt5 is corrected.

The rest I've answered in the threads. Short version: `QLatin1String` matches 205 existing uses and no `QLatin1StringView`; the `mm²` string is file data, not interface text; the database errors are diagnostics inside an already-translated sentence; and the `.ts` change adds 354 lines and removes none, so nothing is lost.

13/13 ctest suites pass and all 23 example projects still resave byte-identical.

scorpio810:
Asked to @Gokulakannan750 in https://github.com/qelectrotech/qelectrotech-source-mirror/pull/529
Could someone make an explanatory video about this feature for QET fans so I can post it on the QET YouTube channel?

ispyisail:
I created manuals and this feaure is not what i expected

https://github.com/qelectrotech/qelectrotech-source-mirror/wiki/cable_catalogue
https://github.com/qelectrotech/qelectrotech-source-mirror/wiki/cable_catalogue-FR
https://github.com/qelectrotech/qelectrotech-source-mirror/wiki/cable_catalogue-DE

still more work to do

scorpio810:
> I created manuals and this feaure is not what i expected
> 
> https://github.com/qelectrotech/qelectrotech-source-mirror/wiki/cable_catalogue https://github.com/qelectrotech/qelectrotech-source-mirror/wiki/cable_catalogue-FR https://github.com/qelectrotech/qelectrotech-source-mirror/wiki/cable_catalogue-DE
> 
> still more work to do

Thanks @ispyisail

ispyisail:
I was underwhelmed by this feature at first, but after doing some more research, I can see that it’s a stepping stone towards a much bigger picture, and I think it should be pushed.

scorpio810:
I'll side with arummler here: the `.ts` file shouldn't be part of this PR, regardless of how clean the diff is.

It's not about whether these 354 lines get lost on a later `lupdate` — it's about keeping translation work on its own track, separate from feature PRs. Hand-editing a `.ts` inside a feature branch bypasses the normal translation workflow (and the contributors who handle it), and it sets a precedent where every feature PR carries its own ad-hoc translation diff instead of going through one place.

The French-only regression for English users is real, but the fix for that isn't to hand-write `qet_en.ts` entries here — it's to merge the feature with French sources (which is the project's convention) and let the translation pass happen afterward, the normal way, against master HEAD.

Could you drop `qet_en.ts` from this PR? Once the feature itself is merged, the missing English strings can be filled in through the usual translation process.

```
git fetch https://github.com/ispyisail/qelectrotech-source-mirror.git revive/529-wire-catalogue
git show FETCH_HEAD:lang/qet_en.ts > /tmp/qet_en_wirecatalogue.ts
lconvert -i /tmp/qet_en_wirecatalogue.ts -o /tmp/wirecatalogue_en.qph

```

arummler:
> I was underwhelmed by this feature at first, but after doing some more research, I can see that it’s a stepping stone towards a much bigger picture, and I think it should be pushed.

I tried it and I agree that it has issues and could be much more. But that requires a strategic discussion before implementing. This MR is pretty self-contained but if it is merged now then I would propose with the understanding that no one builds their workflow on top of it and that it can be completely changed / broken in the future. I will write down some ideas.

arummler:
Some thoughts:
1. Up to now there is a mix between logical and physical objects in general (does not refer only to this MR). I know the analogy goes only that far but from a PCB design perspective one has logical "nets" in the schematic  and physical "traces" on the board. The same applies to anything placed on the board like a transistor: logical symbol, physical foot print, nowadays physical 3d model and mostly export with separate models for things like thermal simulation or EMI FEM. 
2. In the moment the Wire catalogue write directly into the conductor properties. E.g. if there is logical/physical colour mismatch there is a fight for that field's content. 
3. In QElectrotech we have the conflict at several places: if there is a connection used (dot) it is clearly a logical connection. A physical connection connection requires to e.g. go to the end point (connector on a device or terminal and plug-in a second wire with a double sleeve or install a seaparete terminal block if it does not fit / is not rated by the manufacturer).
4. The "real, physical" wire should have in the end end-marking / ferrule text (which nowadays is folded into the label off the logical connections), cut vs routed length (in case of a cabinet builder 3d module).
5. Proposal, don't apply the catalogue spec directly to a conductor. Add at least one link object in between e.g. PhysicalWire which is an instance of a wire catalogue wire (as it can be used multiple times). That one says that conductor UUID points at core X of cable instance XYZ but not at the spec/template itself. The catalogue stays but the "applying" part gets replaced. 
6. The wire catalogue property model can only describe very simple cables. One hasShield flag can't describe a e.g. a CAT7 pair (each pair shielded and an overall shield) or triax (or double shielded foil shield inside a braided shield). We could use a nested tree (cable → pairs/shields/fibers → conductors) instead of flat fields.
7. Regarding the descriptions: fixed fields make it difficult as there a bunch of other entries commonly (and very uncommonly) used: RoHS, fulfillment of fire safety e.g. EN 50200, uncommonly radiation hardness... Probably some well defined common fields but option to add additionally ones? That database one has to create anyhow oneself depending on company internal catalogue or what you have available.
8. Another capability is connectors and full harnesses (probably out of scope, Zuken has an own CAD for that). But would foresee hooks in case...

arummler:
If one wants to think a lot in advance probably also some import/export/replacement by an inventory tool, PLM, etc.

arummler:
Having defaults for the "logical conductors" to apply to the conductor properties (or better to link to a conductor property so that it can be mass changed) would be still an interesting separate version. One just has to be careful with the vocabulary: conductor, PhysicalConductor, PhysicalCable, CableType, etc. Better to get it right before the confusion starts.

ispyisail:
Proposed Fix
https://github.com/qelectrotech/qelectrotech-source-mirror/discussions/956

ispyisail:
Closing, not on merit — the review here (arummler, scorpio810) turned into a real architecture question this needs settled before more code: logical vs. physical objects, whether a "conductor" should link to a shared property rather than copy values, and where this sits relative to nets/PLM import. That's exactly what discussion #956 (and #951 before it) now scopes, replacing the design this PR was building toward.

Branch stays, so this can be reopened once that concept work lands somewhere buildable. Not withdrawn on merit.

PR #938, opened 2026-09-19
Fix various singular/plural cosntructions

arummler:
To be avoided in translations: manual construction of sentences which all assume the same sentence order logic as in French. Better to have separate sentences completely translated. For enumerated lists there is a special Qt linguist function.


ispyisail:
This is the right fix, and the reason is worth stating plainly: building a sentence out of separately translated fragments assumes every language puts the words in the French order. German puts the verb at the end, so no translator could ever repair those strings, however well each fragment was translated. Whole sentences plus `createSeparatedList()` is what lets them do the job.

I built it and checked both mechanisms. A whole-sentence `%n` string does let the translator move the count and the strip name independently — `Ajouter %n borne(s) au groupe de bornes %1` comes out as `5 Klemmen zur Klemmenleiste X1 hinzufügen`. And the list separator is correct per language: **und**, **and**, **et**. Current with master, merges cleanly, Qt6 build clean, 11/11 tests.

**It also fixes a bug you haven't claimed.** The undo entry for moving element texts counted texts and groups together — `.arg(m_items_hash.size())`, where that hash holds both (`elementtextsmover.cpp:64,71`). Two texts and one group read *"3 textes d'élément et un groupe de texte"*. Using `m_text_count` corrects it.

**One line reaches further than the title suggests.** `QLocale::setDefault()` is QET's first, and it governs more than sentences: 129 `QDoubleSpinBox` widgets take their number format from the default locale. Choosing Arabic then renders every numeric field as `١٢٣٤٫٥٠`, and anyone whose interface language differs from their system locale gets different number entry than before.

It is avoidable — `QLocale(desired_language).createSeparatedList(parts)` gives the identical result without touching the app-wide default. Both checked.

**Smaller:** "a enumerative" should be "an enumerative" (8 places). And the disambiguation is not part of the runtime lookup — a mismatched comment still resolves — so rewording those comments buys nothing and only churns the `.ts` files.

**Worth telling the translators:** about 580 finished entries across the 34 language files stop matching, because the fragments they translate are gone. Unavoidable for this kind of fix, but the `.ts` files aren't regenerated here, so those strings show French until someone runs lupdate.

**Blocking:** the `QLocale::setDefault()` side effect — scope it to the two list calls, or keep it and say so in the description.
**Non-blocking:** "a enumerative"; the reworded disambiguations; the ~580 entries needing lupdate.

sources/TerminalStrip/UndoCommand/addterminaltostripcommand.cpp
```diff
@@ -166,19 +165,30 @@ MoveTerminalCommand::MoveTerminalCommand(QSharedPointer<PhysicalTerminal> termin
...
-	auto str_2 = strip_name.isEmpty() ? QObject::tr(" d'un groupe de bornes") :
-										QObject::tr(" du groupe de bornes %1").arg(strip_name);
-
-	auto str_3 = new_strip_name.isEmpty() ? QObject::tr("vers un groupe de bornes") :
-											QObject::tr("vers le groupe de bornes %1").arg(new_strip_name);
-	setText(str_1 % " " % str_2 % " " % str_3);
+	QString text;
+	if (t_label.isEmpty()) {
+		if (strip_name.isEmpty() && new_strip_name.isEmpty())
+			text = QObject::tr("Déplacer une borne d'un groupe de bornes vers un groupe de bornes");
```
scorpio810, line 174: text = QObject::tr("Déplacer une borne d'un groupe de bornes vers un autre groupe de bornes");

sources/TerminalStrip/UndoCommand/addterminaltostripcommand.cpp
```diff
@@ -166,19 +165,30 @@ MoveTerminalCommand::MoveTerminalCommand(QSharedPointer<PhysicalTerminal> termin
...
+		else if (strip_name.isEmpty())
+			text = QObject::tr("Déplacer une borne d'un groupe de bornes vers le groupe de bornes %1").arg(new_strip_name);
+		else if (new_strip_name.isEmpty())
+			text = QObject::tr("Déplacer une borne du groupe de bornes %1 vers un groupe de bornes").arg(strip_name);
```
scorpio810, line 178: text = QObject::tr("Déplacer une borne du groupe de bornes %1 vers un autre groupe de bornes").arg(strip_name);

sources/TerminalStrip/UndoCommand/addterminaltostripcommand.cpp
```diff
@@ -166,19 +165,30 @@ MoveTerminalCommand::MoveTerminalCommand(QSharedPointer<PhysicalTerminal> termin
...
+		else
+			text = QObject::tr("Déplacer une borne du groupe de bornes %1 vers le groupe de bornes %2").arg(strip_name, new_strip_name);
+	} else {
+		if (strip_name.isEmpty() && new_strip_name.isEmpty())
+			text = QObject::tr("Déplacer la borne %1 d'un groupe de bornes vers un groupe de bornes").arg(t_label);
```
scorpio810, line 183: text = QObject::tr("Déplacer la borne %1 d'un groupe de bornes vers un autre groupe de bornes").arg(t_label);

sources/TerminalStrip/UndoCommand/addterminaltostripcommand.cpp
```diff
@@ -166,19 +165,30 @@ MoveTerminalCommand::MoveTerminalCommand(QSharedPointer<PhysicalTerminal> termin
...
+		else if (strip_name.isEmpty())
+			text = QObject::tr("Déplacer la borne %1 d'un groupe de bornes vers le groupe de bornes %2").arg(t_label, new_strip_name);
+		else if (new_strip_name.isEmpty())
+			text = QObject::tr("Déplacer la borne %1 du groupe de bornes %2 vers un groupe de bornes").arg(t_label, strip_name);
```
scorpio810, line 187: text = QObject::tr("Déplacer la borne %1 du groupe de bornes %2 vers un autre groupe de bornes").arg(t_label, strip_name);

sources/TerminalStrip/UndoCommand/addterminaltostripcommand.cpp
```diff
@@ -191,17 +201,18 @@ MoveTerminalCommand::MoveTerminalCommand(QVector<QSharedPointer<PhysicalTerminal
...
-
-	const auto str_3 = new_strip_name.isEmpty() ? QObject::tr("vers un groupe de bornes") :
-												  QObject::tr("vers le groupe de bornes %1").arg(new_strip_name);
-
-	setText(str_1 % " " % str_2 % " " % str_3);
+	const auto count = m_terminal.size();
+
+	QString text;
+	if (strip_name.isEmpty() && new_strip_name.isEmpty())
+		text = QObject::tr("Déplacer %n borne(s) d'un groupe de bornes vers un groupe de bornes", "", count);
```
scorpio810, line 208: text = QObject::tr("Déplacer %n borne(s) d'un groupe de bornes vers un autre groupe de bornes", "", count);
￼

sources/TerminalStrip/UndoCommand/addterminaltostripcommand.cpp
```diff
@@ -191,17 +201,18 @@ MoveTerminalCommand::MoveTerminalCommand(QVector<QSharedPointer<PhysicalTerminal
...
+	else if (strip_name.isEmpty())
+		text = QObject::tr("Déplacer %n borne(s) d'un groupe de bornes vers le groupe de bornes %1", "", count).arg(new_strip_name);
+	else if (new_strip_name.isEmpty())
+		text = QObject::tr("Déplacer %n borne(s) du groupe de bornes %1 vers un groupe de bornes", "", count).arg(strip_name);
```
scorpio810, line 212: text = QObject::tr("Déplacer %n borne(s) du groupe de bornes %1 vers un autre groupe de bornes", "", count).arg(strip_name);

sources/qet.cpp
```diff
@@ -271,84 +271,88 @@ QString QET::ElementsAndConductorsSentence(
...
+	QStringList parts;
 	if (elements_count) {
-		text += QObject::tr(
-			"%n élément(s)",
-			"part of a sentence listing the content of a diagram",
-			elements_count
+		parts.append(
+			QObject::tr(
+				"%n élément(s)",
+				"part of a enumerative partial sentence listing the content of a diagram",
```
scorpio810, line 279: English here?

ispyisail: English is right here, and it is not an exception to the French-source rule: the disambiguation argument is translator metadata, never shown to a user. `sources/` currently has 765 of them in English and none in French.

Only the *rewording* is worth undoing — it changes nothing at runtime but marks those messages for re-review in every `.ts`. And "a enumerative" should be "an enumerative".

sources/TerminalStrip/UndoCommand/addterminaltostripcommand.cpp
ispyisail: Agreed, and the set is complete: these six are exactly the branches where the destination strip is unnamed. The six where it is named keep `vers le groupe de bornes %1`, which is right.

They are reachable, just rarely: the creation dialog substitutes `X` for a blank name (`terminalstripcreatordialog.cpp:79`), but the strip editor has no such guard (`terminalstripeditor.cpp:184`), so a name can be cleared there.

sources/TerminalStrip/UndoCommand/addterminaltostripcommand.cpp
ispyisail: Agreed — source named, destination unnamed. `du groupe de bornes %1 vers un autre groupe de bornes` reads correctly.

sources/TerminalStrip/UndoCommand/addterminaltostripcommand.cpp
ispyisail: Agreed — labelled terminal, both strips unnamed.

sources/TerminalStrip/UndoCommand/addterminaltostripcommand.cpp
ispyisail: Agreed — labelled terminal, source named. The three placeholders stay in order with `.arg(t_label, strip_name)`.

sources/TerminalStrip/UndoCommand/addterminaltostripcommand.cpp
ispyisail: Agreed. I checked the plural form still resolves with this wording at n=1 and n=5, so the `%n` substitution is unaffected by the longer sentence.

sources/TerminalStrip/UndoCommand/addterminaltostripcommand.cpp
ispyisail: Agreed. `%n` plus `.arg(strip_name)` works in this order — the count is substituted during translation, `%1` afterwards; verified both forms render correctly.

sources/qet.cpp
scorpio810: It's my bad, read so quickly..

PR #939, opened 2026-09-19
Fix issue #937

### LINKED ISSUE #937: Compile problems when Backtrace are not found
When Backtrace is not found and tests are configured in CMake with `-DPACKAGE_TESTS=ON`, CMake configure and compile fails.

`${Backtrace_LIBRARIES}` are not guarded when linked to targets `tst_crashhandler` and `tst_crashdumps`. Removing them makes configure work, but compile fails with this then:
```
tst_crashhandler.cpp:103:18: error: use of undeclared identifier 'SIGBUS'
tst_crashhandler.cpp:103:46: error: use of undeclared identifier 'SIGBUS'
```

This is when compiling without msys2, instead using `clang` or `gcc` toolchains from the Qt Online installer on Windows. libbacktrace doesn't seems to be shipped with compilers in the Qt Online installer, but msys2 seems to have a package for libbacktrace though, so it's perhaps not a real issue. The downside with msys2 is that `pacman` doesn't keep a version history as far as I know, you can't target a specific version of Qt for example. Though using the compiler toolchain from msys2 together with Qt from the Qt online installer would solve that.

I'm curious about `SIGBUS`, doesn't seem to be explicitly in the c standard signal.h? [signal - cppreference.com](https://en.cppreference.com/c/program/signal)

elevatormind:
Fixes [Issue #937](https://github.com/qelectrotech/qelectrotech-source-mirror/issues/937)

Note, completely disables backtrace detection on Windows.


CMakeLists.txt
```diff
@@ -107,13 +107,19 @@ endif()
...
-    list(APPEND QET_PRIVATE_LIBRARIES ${Backtrace_LIBRARIES})
-    include_directories(${Backtrace_INCLUDE_DIRS})
-    add_compile_definitions(QET_CRASH_BACKTRACE)
-else()
-    message(STATUS "backtrace() not available: crash dumps will carry the log ring without a backtrace")
+# Windows uses a separate native exception path and must not probe for the
+# POSIX execinfo/backtrace API.
+if(NOT WIN32)
+    find_package(Backtrace QUIET)
+    if(Backtrace_FOUND)
```
arummler: There are some more native ways how to handle backtrace in CMake described in the docu:
https://cmake.org/cmake/help/latest/module/FindBacktrace.html

elevatormind: Thanks for making try a little harder on this - seems to work in msys2 with libbacktrace installed after all!

`Backtrace_INCLUDE_DIR` and `Backtrace_LIBRARY` must be set  for CMake to work correctly:
`Backtrace_INCLUDE_DIR` must point to the folder where backtrace.h resides, `Backtrace_LIBRARY` must point to backtrace.a. This wasn't exactly clear by the errors CMake threw me.

`SIGBUS` must be guarded anyway, it's not defined when using msys2 clang.
I will address the PR for this, `Backtrace_INCLUDE_DIR` and `Backtrace_LIBRARY` should be mentioned in the build docs also.

elevatormind: Adding -DBacktrace_INCLUDE_DIR=/c/msys64/clang64/include -DBacktrace_LIBRARY="/c/msys64/clang64/lib/libbacktrace.a" did the trick for me, otherwise CMake threw a lot of errors regarding Backtrace.

PR #952, opened 2026-09-19
Fix/pin fetchcontent dependencies

bhangart:
Pin fetched dependencies to commit hashes instead of git tags
CMake fetches pugixml, SingleApplication and the three KDE Frameworks
modules by git tag. A tag is a mutable pointer that its owner can move,
so two builds of the same QElectroTech commit can silently get different
third-party sources, and a compromised upstream account can change what
every builder downloads without anything changing in this repository.
Pinning each dependency to the commit its tag currently points at closes
that, while keeping the tag name in a trailing comment so the intended
version stays readable.

No versions change. Every pinned commit is the one its tag resolves to,
checked with git ls-remote and confirmed by fetching each one and
verifying that git describe reports exactly the tag. The three KDE
modules live in separate repositories and therefore need separate
commits, so the single KF_GIT_TAG variable becomes three per-module
variables; passing -DKF_GIT_TAG=<ref> still selects one ref for all
three, unpinned, exactly as before, and KF_GIT_TAG stays defined so the
build summary in define_definitions.cmake is unaffected.


scorpio810:
Thanks for this, pinning by commit hash instead of tag is the right call and the write-up on the attack mechanism is clear and well documented.

I checked each pinned hash against what its tag currently resolves to:

- `pugixml` → `ee86beb30e4973f5feffe3ce63bfa4fbadf72f38` matches `v1.15` exactly (`git ls-remote https://github.com/zeux/pugixml.git v1.15`)
- `SingleApplication` → `aede311d28d20179216c5419b581087be2a8409f` matches `v3.2.0` exactly (`git ls-remote https://github.com/itay-grudev/SingleApplication.git v3.2.0`)

But the three KDE Frameworks pins don't match what `v6.10.0` actually resolves to. Verified directly against invent.kde.org:

```
$ git ls-remote https://invent.kde.org/frameworks/extra-cmake-modules.git v6.10.0
9e3a27e6d48b8df373dc3fd64f959f3436782c88        refs/tags/v6.10.0

$ git ls-remote https://invent.kde.org/frameworks/kcoreaddons.git v6.10.0
f7939b05312ef72be41297dbd93bea29f1d6b26d        refs/tags/v6.10.0

$ git ls-remote https://invent.kde.org/frameworks/kwidgetsaddons.git v6.10.0
bd99eaa21bd277fe0f47fbdf8191e20137526a5f        refs/tags/v6.10.0
```

| Module | Pinned in this PR | What `v6.10.0` actually resolves to |
|---|---|---|
| extra-cmake-modules | `7dd28cc56c339c3f8fb356f7c53c0e8f61433d81` | `9e3a27e6d48b8df373dc3fd64f959f3436782c88` |
| kcoreaddons | `c569f974dab24b4784ad186a3db4b76b2fa36612` | `f7939b05312ef72be41297dbd93bea29f1d6b26d` |
| kwidgetsaddons | `1abbed8a280d6626c59fb197f2c4667d2b1e7445` | `bd99eaa21bd277fe0f47fbdf8191e20137526a5f` |

None of the three match, and this is confirmed straight from the canonical KDE GitLab, not a mirror. Given pugixml and SingleApplication matched exactly, and all three (and only the three) KDE modules are off, this looks like a mix-up when generating these three pins — maybe the wrong ref was resolved, or a different branch/tag got picked up.

Could you regenerate these three commits and push an update before this merges? A pin that doesn't actually match what it claims to pin would undercut the whole point of the PR.

Happy to re-check once updated.

arummler:
It is a bit hypothetical. A lot of software uses git tags to request certain versions. If one of our dependencies goes "rogue" I think we have another problem. But admittedly it does not hurt, only issue is that one does not see immediately which version is installed as SemVer is expressed in tags and not hashes... If you do it at least put a comment above with the git tag. When I see a hash in fetch requests I otherwise assume that it is random master branch point.

bhangart:
thanks, checked it, added more verbose comments incl. adding version in the comments.  KDE uses annotated tags (vs. lightweight) and need to resolve the hash pointing to the commit, otherwise the hash of the tag itself is referenced. So the resolved hashes should actually be correct. 


image

image

image

